PR #9, opened 2026-01-17
Write backslash in man pages as \e (not \\)

RolandRosenfeld:
According to groff(7):
 Printable backslashes must be denoted as \e.

See also https://bugs.debian.org/966803.
Forwarded: https://lists.xymon.com/archive/2024-February/048288.html


canepan (approved):
I am not familiar with groff, but this sounds reasonable

PR #12, opened 2026-01-19
Docs: add Git setup and baseline alignment procedures

bonomani:
This PR adds a new Git documentation file under docs/git/.

- Documents setup and maintenance procedures
- Clarifies baseline verification
- Makes baseline alignment non-destructive by default
- No code changes

Documentation only.

### DIFF
--- a/.github/workflows/workflow-release.yaml
+++ b/.github/workflows/workflow-release.yaml
@@ -1,6 +1,6 @@
 name: Release tarball on tag
 concurrency: 
-  group: workflow-java
+  group: workflow-tarball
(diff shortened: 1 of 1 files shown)


RolandRosenfeld:
I'd suggest renaming 4.x-master branch before writing that branch name to the docs.
- "master" isn't politically correct (at least if we name the stable branch "main").
- "4.x" includes the stable 4.3 version numbers and is stuck at versions starting with "4.".

I'd suggest renaming 4.x-master to "devel", "unstable", or "next" (descending preference for me).

PR #14, opened 2026-01-20
Create CODE_OF_CONDUCT.md

bonomani:


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,32 @@
...
+
+Diversity is one of our huge strengths, but it can also lead to communication issues and unhappiness. To that end, we have a few ground rules that we ask people to adhere to. This code applies equally to founders, mentors and those seeking help and guidance.
+
+This isn’t an exhaustive list of things that you can’t do. Rather, take it in the spirit in which it’s intended - a guide to make it easier to enrich all of us and the technical communities in which we participate.
+
+This code of conduct applies to all spaces managed by the Xymon project or xymon-monitoring. This includes IRC, the mailing lists, the issue tracker, DSF events, and any other forums created by the project team which the community uses for communication. In addition, violations of this code outside these spaces may affect a person's ability to participate within them.
+
+If you believe someone is violating the code of conduct, we ask that you report it by emailing [xymon@xymon.com](mailto:xymon@xymon.com). For more details please see our
```
RolandRosenfeld: isn't there something missing after "see our"?

bonomani: Yes, updated thanks

RolandRosenfeld:
Maybe it would be good to merge the two commits and force-push only one commit (I don't think that the first version, which is nearly completely rewritten is very usefull for the project history).

The test refers PROJECT-ORGANIZATION.md, which seems not to exist.  So the referral can be removed.

I don't know what "Speak Up! project" is nor where to find it or its CoC. So I'm not fully sure, whether it makes sense referring it here (especially without a link). Since I wasn't able to find it using search strings from the above CoC, IMHO the referral to them can be removed.

bonomani:
Thanks for the review, Roland. All three points addressed:

- Squashed the two commits into a single one and force-pushed.
- Removed the reference to `PROJECT-ORGANIZATION.md` (it doesn't exist), along with the whole *Questions* section — the contact email is already in *Scope*.
- Removed the "Speak Up! project" attribution.

Let me know if anything else needs changing.

RolandRosenfeld (approved):
Now I'm fine with it.

PR #36, opened 2026-01-21
ci: build on FreeBSD, OpenBSD, NetBSD and in Debian and RPM containers

montjoie:
Add CI build for all recent Debian and all recent BSD.
Note that all added CI failed currently for reasons.
This PR need at least PCRE2 support to be added.

I have added also 2 build fix for some BSD

### DIFF
--- a/docker/debian/Dockerfile
+++ b/docker/debian/Dockerfile
@@ -1,0 +1,13 @@
+ARG OS
+FROM $OS
+
+COPY . /xymon
+RUN ls -l
+RUN ls -l xymon
+
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y install build-essential libpcre2-dev
+
+COPY docker/debian/start.sh start.sh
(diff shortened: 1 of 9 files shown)


docker/debian/Dockerfile
```diff
@@ -0,0 +1,13 @@
+ARG OS
+FROM $OS
+
+COPY . /xymon
+RUN ls -l
+RUN ls -l xymon
+
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y install build-essential libpcre2-dev
```
RolandRosenfeld, line 8: <s>Seems that there are some packages missing for building server on Debian:
libc-ares-dev libldap-dev librrd-dev libssl-dev libtirpc-dev</s> (just noticed, that this is done in debian/start.sh)

If you use libpcre2-dev, this requires PCRE2 (#5) beeing installed.  Otherwise you need libpcre3-dev, which is no longer available in Debian trixie.

PR #37, opened 2026-01-21
Add ALPN support for xymonnet tests

feld:
This makes it possible to add a custom protocol like this:

[imaps443]
    options ssl,alpn=imap
    port 443

And now it will be able to monitor IMAP being reverse proxied on port 443 multiplexed with other protocols.

### DIFF
--- a/docs/manpages/man5/protocols.cfg.5.html
+++ b/docs/manpages/man5/protocols.cfg.5.html
@@ -56,6 +56,21 @@ to the service. It will then expect a response beginning with &quot;220&quot;.
 Any data returned by the service (a so-called &quot;banner&quot;) will be recorded 
 and included in the status message.
 <P>
+For services that require ALPN protocol negotiation (such as IMAPS multiplexed 
+by Nginx on the same port as HTTPS), you can define:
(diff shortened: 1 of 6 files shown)


bonomani (approved):

bonomani:
Thanks @feld — added a regression test guarding this on my fork, in the `feat/tests-bootstrap` branch (#101):

**`tests/network/alpn-support.sh`** — https://github.com/bonomani/xymon/blob/feat/tests-bootstrap/tests/network/alpn-support.sh

A static guard (a behavioural run would need xymonnet built plus a TLS server advertising ALPN, and the build CI already compiles these files): it asserts the ALPN wiring stays wired across its layers — the `TCP_ALPN` flag + `alpns` field (`lib/netservices.h`), the `alpn=` option parser (`lib/netservices.c`), and `SSL_CTX_set_alpn_protos()` in the SSL setup (`xymonnet/contest.c`), plus the `protocols.cfg(5)` docs. Auto-discovered by `tests/testsuite` and run in CI via `tests.yml`; skips cleanly on branches predating this PR.

PR #38, opened 2026-01-21
Fix critical buffer overflow in trends page graph generation

### LINKED ISSUE #30: Trends page has a buffer overflow
If you have too many graphs of a specific type, it just quits rendering graphs once the overflow happens. I have a patch for this already and will open a PR soon.

feld:
This fixes a critical bug that caused graphs to stop appearing mid-page on the trends status page, particularly for hosts with many filesystems (disk/inode graphs). The issue was most noticeable when servers had 20+ filesystems, where only some disk graphs would appear and all subsequent graphs would be missing.

The problem had two components:

1. snprintf return value mishandling (line 308): The code used the snprintf return value directly to advance the buffer pointer without checking for truncation. Per POSIX/C99, snprintf returns the number of characters that WOULD have been written if the buffer was large enough, even if output was truncated. This caused the pointer to advance past the buffer end, resulting in:
   - Subsequent writes to invalid memory locations
   - Silent failures (no error messages)
   - Incomplete trends pages

2. Inadequate buffer size checking (line 304): The buffer reallocation check used '>=' instead of '>' and didn't account for the null terminator, allowing edge cases where the buffer would be exactly at capacity but still attempt to write.

The fix:

- Pre-calculate exact space needed including null terminator
- Grow buffer sufficiently in a loop (not just +4096) to accommodate actual content size
- Properly validate snprintf return value:
  * Check for errors (< 0)
  * Detect truncation (>= available space)
  * Only advance pointer by bytes actually written
- Add error logging so buffer issues are no longer silent

This explains why the problem would move when reordering GRAPHS env var: whichever graph caused the overflow would fail, and all graphs after it in the GRAPHS order would be lost.

Symptoms that this fixes:
- Trends page showing only partial graphs
- Missing graphs that should appear after disk/inode graphs
- Problem varying based on GRAPHS configuration order
- Silent failures with no error messages

Fixes #30

### DIFF
--- a/web/svcstatus-trends.c
+++ b/web/svcstatus-trends.c
@@ -298,14 +298,43 @@ char *generate_trends(char *hostname, time_t starttime, time_t endtime)
 		if (rwalk) {
 			int buflen;
 			char *onelink;
+			size_t onelinklen;
+			size_t available;
+			int written;
 
 			buflen = (allrrdlinksend - allrrdlinks);
 			onelink = rrdlink_text(myhost, rwalk, 0, starttime, endtime);
-			if ((buflen + strlen(onelink)) >= allrrdlinks_buflen) {
-				SBUF_REALLOC(allrrdlinks, allrrdlinks_buflen+4096);
+			onelinklen = strlen(onelink);
+
+			/* Ensure we have enough space for the string plus null terminator */
+			if ((buflen + onelinklen + 1) > allrrdlinks_buflen) {
+				size_t newsize = allrrdlinks_buflen;
+				/* Grow buffer to accommodate the new content plus extra headroom */
+				while ((buflen + onelinklen + 1) > newsize) {
+					newsize += 4096;
+				}
+				SBUF_REALLOC(allrrdlinks, newsize);
 				allrrdlinksend = allrrdlinks + buflen;
 			}
-			allrrdlinksend += snprintf(allrrdlinksend, (allrrdlinks_buflen - (allrrdlinksend - allrrdlinks)), "%s", onelink);
+
+			available = allrrdlinks_buflen - buflen;
+			written = snprintf(allrrdlinksend, available, "%s", onelink);
+
+			/* Check if snprintf truncated the output */
+			if (written < 0) {
+				errprintf("Error writing graph link for %s\n", graph->xymonrrdname);
+			}
+			else if ((size_t)written >= available) {
+				/* Output was truncated - this should not happen due to our check above */
+				errprintf("Buffer overflow prevented for graph %s (needed %d, had %zu)\n",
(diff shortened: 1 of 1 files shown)


bonomani:
Code looks correct — the `snprintf()` truncation fix is sound: it sizes the buffer to fit (incl. the null terminator) and recomputes `allrrdlinksend` after `SBUF_REALLOC()`. No functional concerns; graph-rendering logic is untouched. Verified locally that it merges and builds clean against current `main`.

Non-blocking nit: `buflen` could be `size_t`.

bonomani:
**Verified locally (2026-05-27):** merges clean with current `main` and **builds clean** (server, 0 errors, `svcstatus.cgi` produced) — the stale green is fine against today's `main`; it just needs a branch update + re-CI + undraft. Independent of #46 (disjoint files). Recommend landing this (and #46) before #44, since #44 rewrites the same `generate_trends()` buffer block and must keep this PR's loop-grow.

bonomani (commented):
Reviewed — looks correct. Well-scoped fix for a real heap overflow: on the old truncation path the write pointer advances past the buffer, so the next `snprintf` gets a wrapped-around size. The loop-grow plus validated `snprintf` return are sound (`SBUF_REALLOC` allocates `LEN+1`, so the sizing is safe), the normal path is unchanged, and it builds clean on `main`. Locally, the old code truncates the trends output and drops trailing graphs on a large single link, while this code renders them in full.

Non-blocking: worth adding that large-link case as a regression test in CI so this can't be silently reverted later — notably by #44, which rewrites this same `generate_trends()` buffer block.

bonomani (approved):
LGTM

feld:
> worth adding that large-link case as a regression test in CI so this can't be silently reverted later

I'm not up to speed on the current CI situation, so how would you suggest we approach this?

bonomani:
Thanks @feld — I added a regression test guarding this fix on my fork, in the `feat/tests-bootstrap` branch:

**`tests/web/svcstatus-trends-overflow.sh`** — https://github.com/bonomani/xymon/blob/feat/tests-bootstrap/tests/web/svcstatus-trends-overflow.sh

It is a static guard (a behavioural repro would need the full web CGI plus an RRD environment, and the build CI already compiles the file): it asserts the dangerous advance-by-`snprintf()` pattern stays gone, and that the size check (`len+1` vs buffer) and the truncation/error handling introduced by this PR stay present in `web/svcstatus-trends.c`. It skips cleanly if the file is absent (e.g. after a web/CMake rework).

It is auto-discovered by the `tests/testsuite` runner and run in CI via `tests.yml`. Verified both ways: it passes with this PR applied, and fails on the current pre-fix code — so it is a real guard, not a no-op.

PR #40, opened 2026-01-21
FreeBSD XymonClient: Filter out autofs for df and inode

feld:
autofs shows up in df output, but it's not a real filesystem and also cannot be parsed the same way as it includes an extra column in the output.

Example of df output (trimmed for clarity):

Filesystem                    Size    Used   Avail Capacity  Mounted on
zflash/ROOT/default           156G    9.5G    146G     6%    /
map auto_plex                   0B      0B      0B   100%


bonomani:
Seem good to me.

bonomani (approved):

PR #41, opened 2026-01-21
FreeBSD XymonClient: collect pkg info

feld:
Debian ships a similar change for dpkg in their `09_hobbitclient-debian.patch`


RolandRosenfeld:
The Debian patch is now in https://github.com/xymon-monitoring/xymon/pull/48

bonomani (commented):
Small nit/bug: `echo "[pkg"]` — the quote is in the wrong place, should be `echo "[pkg]"`. It happens to still print the right thing because bash concatenates `"[pkg"` and the bare `]`, but it's inconsistent with the surrounding `[ports]` / `[ps]` headers and could break if reformatted.

Otherwise looks good — useful addition, matches the Debian counterpart in #48.

feld:
> Small nit/bug: `echo "[pkg"]` — the quote is in the wrong place

really funny that I overlooked that, good catch

Issue #29, opened 2026-01-21
JC's Terabithia RPMs


feld:
Tracking issue for auditing the downstream changes shipped in J.C. Cleaver's
Terabithia RPM packages and upstreaming the generally useful fixes.

The RPM sources and supporting documentation are publicly available:

- repository index: https://repo.terabithia.org/rpms/xymon/
- RPM spec: https://repo.terabithia.org/rpms/xymon/xymon.spec
- Terabithia README: https://repo.terabithia.org/rpms/xymon/xymon.README.terabithia
- Terabithia changes: https://repo.terabithia.org/rpms/xymon/xymon.CHANGES.terabithia
- known issues: https://repo.terabithia.org/rpms/xymon/xymon.KNOWNISSUES.terabithia
- source RPMs, for example EL8: https://repo.terabithia.org/rpms/xymon/el8/SRPMS/

J.C. explained that the packages use the upstream 4.3.30 tarball plus a large
number of patches and build/install layout changes. The spec / SRPM is the
authoritative source: https://lists.xymon.com/archive/2023-August/048088.html

## Scope

Separate the downstream contents into: RPM/packaging policy (keep downstream),
fixes already in `main`, changes covered by an open issue/PR, changes on `devel`
tracked for backport, and remaining fixes/features worth focused upstream PRs.

Related tracking: Debian patches #28 · FreeBSD patches #103 · `devel` feature
backports #106 · repository ownership/continuity
xymon-monitoring/xymon-problems#5.

## Audit summary

The Terabithia stack is upstream 4.3.30 plus **210 active patches** (el8 SRPM).
A forced-direction comparison (both ways, `--fuzz=0`) against `upstream/main`
`7ed6bc72` and `upstream/devel` `3686fbee` (reproduced 2026-06-09) splits them:
**151** apply cleanly forward, **59** diverge (bucketed below), **0** are strict
reverse-only matches on `main`. Exactly one diverged patch strictly matches an
open PR (`proxytune` → #46).

## Inventory (2026-06-08)

[...]

PR #44, opened 2026-01-21
Fix custom graphs

### LINKED ISSUE #31: Custom RRD graphs do not really work as documented
The [howtograph.html](https://xymon.sourceforge.io/xymon/help/howtograph.html) FAQ explains how to make custom graphs. The documentation is mostly correct, except there are some things that are completely wrong:

> If you want the graph included with the other graphs on the trends column, you must add it to the GRAPHS setting in the ~xymon/server/etc/xymonserver.cfg file.

No, if you want your graph to work at all it must be in the `GRAPHS` setting. And then if you want it rendered on a specific status page, you need to also declare `GRAPHS_statusname=graphname,graphname...`

But even that will not make it work. A custom graph cannot be rendered on a status page unless that status page already has a graph. This is where the confusion really sets in because the docs seem to show you creating a custom graph on a custom status page.

Fixing this requires a pretty big rework of the graphs/RRD code. I have a patch for it and we can review it when I open the PR. It probably shouldn't go in the 4.3.31 release anyway because it's a pretty big change.

feld:
Custom graphs are supposed to work and are documented as being possible through NCV/SPLIT_NCV formatting included in a status message or through a custom submission using the data channel reporting to the "trends" test which then will let you create/update RRD files.

This part works. Actually using the graphs does not.

I spent too much time fighting with configuration thinking I was doing something wrong and then I had an idea of trying my S.M.A.R.T. graphs by applying them to a different test that already has graphs, like cpu:

in xymonserver.cfg:
GRAPHS_cpu="smart-temp"
GRAPHS="...,smart-temp"

And this worked! So graphs work but not on my custom smart status page.

This led to debugging of Xymon server itself where I discovered there was a requirement of if (rrd && graph) guarding the parsing of GRAPHS_%s variables.

This required some refactoring to move that out, and it worked!

But then I noticed it wasn't showing all the graphs, just one graph. How do pages like the disk staus page show graphs for all the filesystems?

It turns out this is a real mess -- it uses linecount to guess how many graphs there are supposed to be before it does its loop to generate the HTML for embedding the graphs in the page. And the graphs with multiple datasets are declared in the code or in the xymon CGI options config as multigraphs. Putting my test in this list did nothing, likely because the graph name doesn't match the status page name.

But what is this magical linecount value anyway?

It's literally the amount of lines in the status that submitted the data. But how does that work for disks, where df has a header before the list of filesystems? It turns out this is also hacked into the code to subtract one from the total count. There are other tests with their own linecount offsets as well...

Ok, is there another way to deal with this? Yes, there are some things you can embed into statuses like , but that won't work for how I'm publishing the data (trends, data channel). It's also a big mess and I can't believe we have these random rules to work around this problem.

So the correct thing to do for this is to actually count the number of RRD files. But this was never done, likely because it was considered to expensive in disk IO years ago...

Well it's being done now. Instead of relying on linecount sorcery we actually read graphs.cfg, find the FNPATTERN (and EXFNPATTERN), and use that to count the number of matches of that PCRE pattern in the rrd directory.

It works. It hasn't been tested with graphs.cfg includes of another file/directory yet, but I intend for that to work.

Anyway, now we can have custom graphs on custom status pages and all is well.

Summary:

- Remove all line counting logic as we can just count the RRDs

- Move the RRD graph counting using better logic with FNPATTERN awareness into xymonrrd, and make htmllog and svcstatus-trends use it for consistency

- Now custom graphs will show up on the trends page

- Do not bother generating graph HTML when status updates are received; only when actually rendering graphs. Much less expensive as we skip some unnecessary IO looking for the count of RRD files.

Fixes #31


feld:
This has not at all been tested with PCRE2, and it does have some new pcre code, so this needs to be rebased after the PCRE2 change is in

bonomani:
**Status / coordination.** This PR collides with two others — `web/svcstatus-trends.c` (↔ #38) and `web/svcstatus.c` (↔ #46), both confirmed by real 3-way merges. #38 and #46 are independent of each other and both verified to merge + build clean against current `main`, so the clean path is: **land #38 and #46 first, then rebase this PR** onto them — keeping #38's loop-grow and combining #46's `skiploc`. The rebase only resolves the collisions, though; this PR also has its own blockers (PCRE1→PCRE2 won't compile on current `main`, plus a few correctness findings). Full per-task verification checklist: #107.

bonomani:
Heads-up on a small overlap with **#200** (TBT 237, tracked in #29) — a daemon-side `xymond_rrd` cleanup. The two are conceptually independent, but commit `261e46435040` ("Fix custom graphs") touches the same two spots in `lib/xymonrrd.{c,h}` that #200 does, so a trivial conflict will surface for whichever lands second.

**Suggested order: land #200 first** (it's small/self-contained/ready). This PR is already `CONFLICTING` against `main` and must rebase regardless, so the resolution below folds into that rebase — no extra work.

**What conflicts in `261e46435040`, and how to resolve once #200 is in `main`:**

1. `lib/xymonrrd.c` — the `rrd_setup()` signature.
   - This commit changes `static void rrd_setup(void)` → `void rrd_setup(void)` (to call it externally), keeping the old body (`setup_done` + the 5-min timer).
   - #200 rewrites that body: it drops `static int setup_done` and the 5-minute guard and builds the tables once via `if (xymonrrdtree != NULL) return;` — but leaves the function `static`.
   - **Resolution:** keep #200's build-once body verbatim and just drop the `static` keyword → `void rrd_setup(void)`. Discard this commit's now-stale context lines referencing `setup_done`. Net result: `void rrd_setup(void)` **+ build-once**.
   - This is functionally safe: build-once fully serves this PR's external `rrd_setup()` call. The 5-min reload only ever rebuilt identical env-derived tables (`TEST2RRD`/`GRAPHS`), so nothing is lost.

2. `lib/xymonrrd.h` — the extern block.
   - This commit adds `extern void rrd_setup(void);` and `extern int count_rrd_files_for_graph(...);`.
   - #200 adds `extern void rrd_destroy(void);` in the same block.
   - **Resolution:** keep all three.

Nothing else overlaps — `count_rrd_files_for_graph()`, `lib/htmllog.c`, `web/showgraph.c`, and `xymond/xymond_filestore.c` are untouched by #200 (#200's daemon edit is in `xymond/xymond_rrd.c`).

bonomani:
Superseded by **#206**, which keeps this PR's diagnosis of the `(rrd && graph)` guard (credited — the fix commit carries Mark Felder as co-author) but fixes #31 additively, without the RRD-counting machinery: a deep review of a corrected, rebased revision of this PR's approach ([findings summarized in #107](https://github.com/xymon-monitoring/xymon/issues/107#issuecomment-4866693882)) showed the count-gate regressed locator/distributed setups, filestore static pages, TEST2RRD-mapped columns and `--multigraphs` paging. Leaving this PR open for reference until #206 lands.

PR #46, opened 2026-01-21
4.4 branch backports (TBT 70/116/562/973/987)

feld:
Here are a bunch of commits that are probably safe to backport from the 4.4 branch.

We'll need to move the Changes entries before merging.


bonomani:
## Status (updated 2026-07-03)

**Rebased onto `main`, `web/svcstatus.c` dropped, mergeable.** The branch (`f28484791`) carries **24 backports** replayed onto current `main`. The one conflict was resolved **inside** `3366cfc4` (`RRDADDUPDATED`) — the guard wraps only the modern escaped-colon `strftime`, matching `main`'s removal of the pre-1.2 `RRDTOOL12` ifdefs (`5d9e77da8`); default behavior unchanged (`RRDADDUPDATED=TRUE` ships as an `xgetenv` built-in and in `xymonserver.cfg.DIST`). The `skiploc` change (`a0822c27`, `web/svcstatus.c`) was dropped from this bundle per the documented plan — **#151 is its sole carrier**. All 24 commits remain unique to this branch (`git cherry` vs `main`). Remaining to land: move the `Changes` entries (per the PR description), undraft.

**Collisions with other PRs** — none. The #44 rework landed as **#206** (touches only `lib/htmllog.c` plus docs and tests — no file overlap), and #139 test-merges cleanly alongside.

**Terabithia cross-reference (#29)** — Terabithia Patch 121, `xymon.proxytune.patch`, reverse-applies cleanly to this PR's head; the implementation is carried by commit b0e04497 (proxy retry counts and interval, send retries, initial/growth buffer sizes, socket limit, combo-size limit, initial buffer size for connections). The only strict patch-level match among the 59 diverged Terabithia patches; assigned here, not resolved until the PR merges.

bonomani:
### Packaging dependency

The 14 files under `tools/` and `rpm/` in this PR are the complete packaging support layer for Xymon, and are currently vendored downstream in [xymon-monitoring/xymon-rpm](https://github.com/xymon-monitoring/xymon-rpm) so RPMs can be built before this merges.

Last in the packaging merge chain: the unit's `ExecReload` needs #172, and `xymon-tmpfiles.conf`'s `/run/xymon` needs #219, so both should land first or those two files ship inert.

**One caveat worth recording.** `xymonlaunch.service`, `xymon.server-init` and `xymon.client-init` all invoke `xymoncmd --no-env`, and **`--no-env` does not exist on `main`** — it comes from devel `ec161e994` (`lib/stdopt.c`), tracked in #106's stdopt group, and `main` has no `lib/stdopt.c` at all. I checked the 4.3.30 variants of those three files in Terabithia's SRPM: they are identical apart from that flag. As it stands the unit would fail to start on a build from `main`, so either the stdopt work lands first, or these three drop `--no-env` until it does. (xymon-rpm currently ships the 4.3.30 variants for exactly this reason.)

**Merge order for the packaging set:** #163 (independent, no prerequisites) · #219 → #172 → #46

bonomani:
Flagging a blocker found while checking this against the RPM packaging: as it stands, this PR ships a client that cannot start.

`client/Makefile` gains

```make
cat xymonclient.cfg.DIST | sed -e 's!@XYMONTOPDIR@!$(XYMONTOPDIR)!g' ...
```

and `client/xymonclient.cfg.DIST` gains `XYMONCLIENTHOME="@XYMONTOPDIR@"`.

But `XYMONTOPDIR` never reaches that sub-make. `build/Makefile.rules:78` invokes it as

```make
CC="$(CC)" CFLAGS="$(CFLAGS)" XYMONHOME="$(XYMONCLIENTHOME)" XYMONHOSTIP="$(XYMONHOSTIP)" ... $(MAKE) -C client all
```

— `XYMONHOME` is passed, `XYMONTOPDIR` is not, and it is not exported anywhere either. This PR does not touch `build/Makefile.rules`, so nothing compensates.

### Reproduced

`./configure --client` with `XYMONTOPDIR=/opt/xymon`, then `make && make install`, on this branch:

```
$ grep -n '^XYMONCLIENTHOME\|^XYMONHOME' /opt/xymon/etc/xymonclient.cfg
12:XYMONCLIENTHOME="" # Directory for the Xymon client files
13:XYMONHOME="$XYMONCLIENTHOME"       # Directory for the Xymon client files
```

Both are empty. Every client task loads this file through `ENVFILE` in `clientlaunch.cfg`, so the whole client breaks at runtime, not at build time — which is why CI would not catch it.

### Fix

`@XYMONHOME@` instead of `@XYMONTOPDIR@`, since that *is* passed and already carries the client home (`XYMONCLIENTHOME` on the server build, `XYMONTOPDIR` on a client-only build):

```diff
-XYMONCLIENTHOME="@XYMONTOPDIR@"
+XYMONCLIENTHOME="@XYMONHOME@"
```

The `sed` already substitutes `@XYMONHOME@`, so the added `-e 's!@XYMONTOPDIR@!...!g'` can go too.

Worth having: with that one token changed, this line becomes genuinely useful beyond packaging. `XYMONCLIENTHOME` is currently derived only by `client/runclient.sh` from its own path, which is why a service manager or any other caller has to re-derive it — see #408. Baking it into the installed config removes that from the wrapper's job.

Marking as draft so it does not merge in this state; happy to be corrected if I have misread how `XYMONTOPDIR` reaches `client/Makefile`.

bonomani:
Three call sites here pass `--no-env` to `xymoncmd`, which `main`'s `xymoncmd` does not implement:

- `rpm/xymon.server-init.redhat` — `daemon --user=$user /usr/bin/xymoncmd --no-env $exec $XYMONLAUNCHOPTS`
- `rpm/xymon.client-init.redhat` — same line
- `tools/xymonlaunch.service` — `ExecStart=/usr/bin/xymoncmd --no-env /usr/sbin/xymonlaunch --no-daemon $XYMONLAUNCHOPTS`

`devel` implements it in `lib/stdopt.c`, which is not part of this backport and does not exist on `main`. `main`'s `xymoncmd` parses its options inline (`common/xymoncmd.c:118-129`) and knows only `--debug`, `--env=`, `--area=` and `--version`. Anything else falls to the `else` at `common/xymoncmd.c:133`, which treats the first unrecognised argument as **the command to run** — so `--no-env` becomes the program, and the real command becomes its argument.

Measured against `main`'s `xymoncmd`:

```
$ xymoncmd --env=<cfg> --no-env /bin/echo STARTED
execvp() failed: No such file or directory

$ xymoncmd --env=<cfg> /bin/echo STARTED
STARTED
```

So as written, the two init scripts and the unit start nothing at all.

Dropping `--no-env` is also the behaviour you want on `main`: letting `xymoncmd` load the environment probes `xymonserver.cfg` before `xymonclient.cfg`, so the role resolves at runtime — the same way `xymonlaunch` already picks `tasks.cfg` before `clientlaunch.cfg`. The alternative is backporting `lib/stdopt.c` too, which is a much larger piece of `devel`.

Found while bringing the same unit across in #415, which hit this and removed the flag for that reason.

bonomani:
Closing this — not because the work is unwanted, but because the bundle is the wrong unit for it, and every commit in it is already tracked individually.

**Nothing here is lost.** All 24 non-merge commits are itemised on the two trackers, each with its `devel` commit id, a verdict and a dated measurement: #29 for the Terabithia-derived ones, #106 for the rest. GitHub keeps `refs/pull/46/head` indefinitely, so the branch stays fetchable — including the two commits that exist nowhere else, the July merge resolution (`cfccccdfd`) and the `Fix compile` repair of the hosts.cfg-cache pair.

**Why the bundle rather than the content.** 25 commits across 44 files and 12 areas — packaging, init scripts, SELinux, firewalld, C buffer sizing, xymond caching, web icons — is not reviewable as one unit, and after 7.5 months it has drawn no review. Meanwhile the trackers' own porting instructions bypass it: most of these lines say to take the `devel` commit rather than this branch's replay of it.

**Three blockers found while checking it, now recorded on the trackers** so they survive this closing:

- `6aff7d411` (`XYMONCLIENTHOME`) breaks the client as written. It substitutes `@XYMONTOPDIR@`, but `build/Makefile.rules`'s `client:` target passes `XYMONHOME` and not `XYMONTOPDIR`, and nothing exports it — the installed `xymonclient.cfg` gets `XYMONCLIENTHOME=""`, and every task loading it through `ENVFILE` fails at runtime. Port it as `@XYMONHOME@`. (#29, TBT `36`)
- `bec94d76b` (`--no-env` in the RH init scripts) needs `ec161e994` first: `main` has no `lib/stdopt.c`, so the flag does not exist and the unit fails to start. (#106)
- `f509be0b1` needs #172 for its `ExecReload`, and `5a0b5b79d`'s `xymon-tmpfiles.conf` needs #219 for `/run/xymon`; both ship inert otherwise. (#106)

**How this lands instead.** Per subject, in reviewable PRs, ported from `devel` — the packaging/init set (9 commits, no C) is independent and can go first; the two correctness fixes (garbled-banner guard, missing `</td>`) are small enough to land on their own. The trackers hold the running list, so nothing needs re-deriving.

Thanks @feld — the inventory this assembled is what made the per-item tracking possible.

PR #50, opened 2026-01-23
Add "IFEXIST" patch to hobbit-clients.cfg/files test

RolandRosenfeld:
Author: Bernd Zeimetz <bzed@debian.org>
Acked-By: Christoph Berg <myon@debian.org>
Forwarded: https://lists.xymon.com/archive/2017-January/044216.html
Last-Update: 2015-10-30

### DIFF
--- a/client/localclient.cfg
+++ b/client/localclient.cfg
@@ -206,6 +206,8 @@
 #             - "NOEXIST" triggers a warning if the file exists. By default,
 #               a warning is triggered for files that have a FILE entry, but
 #               which do not exist.
+#             - "IFEXIST" only checks the file if it exists. If the file is
+#                reported as missing by the client, it is ignored.
 #             - "TYPE=type" where "type" is one of "file", "dir", "char", "block", 
 #               "fifo", or "socket". Triggers warning if the file is not of the
 #               specified type.
--- a/xymond/analysis.cfg.5
+++ b/xymond/analysis.cfg.5
@@ -294,6 +294,9 @@ For files, you can check one or more of the following:
 triggers a warning if the file exists. By default,
 a warning is triggered for files that have a FILE entry, but
 which do not exist.
+.IP "ifexist"
+only checks the file if it exists. If the file is
+reported as missing by the client, it is ignored.
 .IP "type=TYPE"
 where TYPE is one of "file", "dir", "char", "block",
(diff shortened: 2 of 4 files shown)


bonomani:
I would not merge this patch as written.

The downstream `IFEXIST` name is useful for compatibility with existing Debian configs, but upstream already has the same behavior with `OPTIONAL`: skip the file check when the file does not exist.

Recommendation: keep `IFEXIST`, but implement it only as an alias for `OPTIONAL`.

The code change should be in `xymond/client_config.c`, in the `FILE` rule parser, next to the existing `optional` handling:

```c
else if (strcasecmp(tok, "optional") == 0 ||
         strcasecmp(tok, "ifexist") == 0)
    currule->chkflags |= CHK_OPTIONAL;
```

The documentation change should be in `xymond/analysis.cfg.5`, near the existing `OPTIONAL` paragraph, saying that `IFEXIST` is an alias for `OPTIONAL`.

This keeps downstream compatibility without adding a new flag or a second runtime path.

The current patch's `FCHK_IFEXIST (1 << 32)` is unsafe because the file-check flags are stored in an `int` bitmask. Shifting a 32-bit `int` by 32 is undefined/invalid, and it also runs past the existing flag space. Reusing `CHK_OPTIONAL` avoids that problem completely.

bonomani:
Following up on my review above: I've opened #161, which implements `IFEXIST` as a one-line alias for the existing `OPTIONAL` keyword (`CHK_OPTIONAL`) instead of a dedicated flag. That reuses the dump and `check_file()` paths already in place, and sidesteps the `FCHK_IFEXIST (1 << 32)` issue entirely — that shift is undefined on the `uint32_t` flags field and folds to `0` on common toolchains, which makes the feature a silent no-op as written here.

#161 keeps your `IFEXIST` spelling and the `analysis.cfg.5` documentation, so it preserves the compatibility goal of this PR. Proposal: close this one in favor of #161 — or, if you'd rather keep authorship here, I'm happy to instead push the alias change onto your branch and close #161. Your call; thanks for getting the ball rolling on this one.

RolandRosenfeld:
I confirm, that IFEXIST is just an alias of OPTIONAL, so this PR does not make sense and also is broken because auf number of bits.

PR #51, opened 2026-01-23
Ignore duplicate submissions for the "/" partition

RolandRosenfeld:
https://salsa.debian.org/debian/xymon/-/blob/master/debian/patches/69_disk-no-duplicate-root.patch

Author: Christoph Berg <myon@debian.org>
Forwarded: https://lists.xymon.com/archive/2015-September/042223.html
Last-Update: 2015-09-10
Reviewed-By: Axel Beckert <abe@debian.org>

On some systems, including the Debian Wheezy default setup, df shows two entries for / (one for "rootfs", one for the real device). Skip the second one or else the rrd files produced contain ugly gaps. (A complete fix would do this for all filesystems, but this case should be rare.)

### DIFF
--- a/xymond/rrd/do_disk.c
+++ b/xymond/rrd/do_disk.c
@@ -20,6 +20,7 @@ int do_disk_rrd(char *hostname, char *testname, char *classname, char *pagepaths
 	static int ptnsetup = 0;
 	static pcre *inclpattern = NULL;
 	static pcre *exclpattern = NULL;
+	int seen_root_fs = 0;
 
 	if (strstr(msg, "netapp.pl")) return do_netapp_disk_rrd(hostname, testname, classname, pagepaths, msg, tstamp);
 	if (strstr(msg, "dbcheck.pl")) return do_dbcheck_tablespace_rrd(hostname, testname, classname, pagepaths, msg, tstamp);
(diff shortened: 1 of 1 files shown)


bonomani:
Quick context before merging: this patch defends against a duplicate-`/`
line in `df` output that **GNU coreutils itself fixed around 2012-2013**.
The current `df` manual states it elides the early-boot `rootfs`
pseudo-filesystem whenever the real root device is mounted
([reference](https://www.gnu.org/software/coreutils/manual/html_node/df-invocation.html)).

The Debian patch (2015) was needed on Wheezy, which had older coreutils.
On any Linux distribution Xymon supports today (Jessie+, Ubuntu 14.04+,
RHEL 7+), `df` no longer emits the duplicate line — so this code is
effectively dead.

A few remaining concerns if it stays:

1. **Silent skip** — no `dbgprintf` when the second `/` is dropped, so
   admins can't tell when the workaround triggers.
2. **Wrong layer** — the real source of the duplicate is `df`'s output,
   which is parsed client-side (`xymonclient.sh`). Filtering there
   (e.g. drop the line whose source is literally `rootfs`) is precise
   and doesn't risk hiding anything else.
3. **First-occurrence wins** is brittle — on a system with two
   legitimate `/` entries (bind-mounts, namespaces) one is silently
   dropped, and which one depends on order.

### Conclusion

Best path forward: **close the PR**. The underlying coreutils bug has
been gone for over a decade. If a current reproducer exists, please
share — otherwise this is a maintenance liability with no remaining
benefit.

If kept for the rare legacy system, at minimum **gate it behind a
config/compile flag** rather than enabling it by default, so modern
hosts cannot silently lose an edge-case `/` entry.

RolandRosenfeld:
You are right, this was only needed in Debian 7 (wheezy) and is gone in Debian 8 (jessie).

PR #52, opened 2026-01-23
Make the date header of html man pages reproducible

RolandRosenfeld:
While re-generating the HTML man pages I noticed the following issues:

a) Generating the man pages places the current day in the header, so
   the files cannot be generated reproducible (see
   https://en.wikipedia.org/wiki/Reproducible_builds).  Using
   SOURCE_DATE_EPOCH (if it is set) solves this problem.

b) Without "-H localhost" as parameter of man2html the generated HTML
   files differ from what is packaged in the tarball.  So I added this
   option.

https://lists.xymon.com/archive/2024-February/048293.html https://salsa.debian.org/debian/xymon/-/blob/master/debian/patches/97_makehtml.patch

### DIFF
--- a/build/makehtml.sh
+++ b/build/makehtml.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 export LANG=C
-DATE=`date +"%e %b %Y"`
+DATE=`date +"%e %b %Y" --date="@${SOURCE_DATE_EPOCH:-$(date +%s)}"`
 VERSION="$1"
 if [ "$VERSION" = "" ]
 then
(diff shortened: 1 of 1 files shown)


feld:
We haven't done many merges yet, but things that are more than just like fixing typos should probably have Changes file entries. Can you add that? Then I will hit the review approval for this :)

RolandRosenfeld:
@feld 
> We haven't done many merges yet, but things that are more than just like fixing typos should probably have Changes file entries. Can you add that? Then I will hit the review approval for this :)

I saw this in one of your merge requests and thought about doing so myself, too.
But I decided not to do so, since this would complicate the pull request work flow a lot.
At the moment there are no Changes entries for the already merged commits.  If I start with an empty 4.3.31 chapter now, this would be wrong, since there are some entries missing.  And if every pull request contains a Changes line, this would imply that after merging one pull request, all other pull requests have to be rebased to move their Changes line below the already merged line. 
So I fear, that merge requests containing a Changes line will hinder merging.

My idea was to keep Changes file in mind and add a Changes file commit every now and then (not sure whether to do this via a merge request or directly in the main branch).

Another question is, whether we want to continue with the old format or maybe switch to some markdown format, which may be recycled for the release text (which ideally links to github issues and pull requests).

feld:
that's fair. I guess we can look at the list of commits between branches and write a changelog just before release until we come up with something better.

feld (approved):

feld:
I get this "Merging is blocked \\ Cannot updated this protected ref." that doesn't make sense. I haven't seen that before. Perhaps it's because you're submitting this from your fork instead of the branch being directly on this repo? It also really wants me to squash by default which seems wrong. Strange. I'll just do the override.

PR #59, opened 2026-01-31
lib: include stdlib for abort

### LINKED ISSUE #58: CI clang/CMake build fails: missing <stdlib.h> for abort() in lib/suid.c
**Title:** CI clang/CMake build fails: missing stdlib include for abort()

**Description:**

CI fails when building with **clang** using the **CMake packaging preset**.

**Context:**

* Job: linux_cmake / PRESET=packaging / VARIANT=client / COMPILER=clang

**Error:**
lib/suid.c:45:3: error: call to undeclared library function 'abort'
note: include the header <stdlib.h>

**Cause:**
`abort()` is used in lib/suid.c without including the required standard header.
Clang (C99+) does not allow implicit function declarations.

**Fix:**
Add `#include <stdlib.h>` to lib/suid.c.

**Notes:**
This is exposed by CMake-based CI builds with clang; GCC may not flag it.

bonomani:
**Description:**

Fixes #58

This pull request fixes a CI build failure observed with clang and CMake-based builds.

`abort()` is used in `lib/suid.c` without including the required standard C header.
With modern C standards (C99 and later) and stricter compilers such as clang, implicit function declarations are not allowed, which causes the build to fail.

**Change:**

* Add the missing standard header:

  ```c
  #include <stdlib.h>
  ```

  in `lib/suid.c`.

**Context:**

* Triggered by CI builds using clang (CMake / packaging preset)
* Improves portability and standards compliance

**Impact:**

* Fixes clang CI failure
* No functional change at runtime

### DIFF
--- a/lib/suid.c
+++ b/lib/suid.c
@@ -15,6 +15,7 @@ static char rcsid[] = "$Id$";
 
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <stdlib.h>
 #include <unistd.h>
 
 #include "libxymon.h"
@@ -71,4 +72,3 @@ void drop_root_and_removesuid(char *fn)
 
 	drop_root();
 }
-


RolandRosenfeld (approved):
Since I run into the same issue with the same fix, I approve this merge request.
I'm personally not really happy with commits generated by "actions-user" instead of seeing real people in the history, but for this minimal fix this sounds okay.

bonomani:
I understand and I agree, and honestly I had not even noticed it myself 🙂

Thanks a lot for pointing this out, Roland. I have updated my setup accordingly and also updated the contribution documentation to reflect this.

Much appreciated.
Bruno

Issue #43, opened 2026-01-21
GCC Warnings fixes backport needed


feld:
We need to cherry pick this but there will be conflicts as even just applying the patch normally causes some rejects. Some are due to missing files as we don't various changes from the 4.4 branch and others might need manual adjustment: https://github.com/xymon-monitoring/xymon/commit/54870eda23a8b12a25cf3fdfd9b7ce17217155e0

RolandRosenfeld:
The above commit id seems is now reported as "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.".  Seems that this was changed a little, the current commit is https://github.com/xymon-monitoring/xymon/commit/3686fbeef5ad7e1668878fab4eaf6c1e498699f7

bonomani:
Status update: the backport requested here is now carried by #113 (the safe, no-behavior-change subset) and #114 (the remaining root-cause fixes, no pragma suppression), which together supersede #60. Once both land, the r8123 warnings fix (Tom Schmidt, with Ralph Mitchell) is fully on main — except #60's two unrelated side-changes (XYMONHELPDIR, the sig.c crash-message/getenv change), which were deliberately not carried over.

PR #60, opened 2026-01-31
Fix a large number of GCC build warnings and issues (Tom Schmidt, with Ralph Mitchell)

RolandRosenfeld:
I tried to merge/backport https://github.com/xymon-monitoring/xymon/commit/3686fbeef5ad7e1668878fab4eaf6c1e498699f7 from 4.4-alpha2 to 4.3.30.

The original Changes entry be jccleaver was:
* A large number of build warnings and possible buffer issues have been resolved
  (Thanks, Tom Schmidt, with Ralph Mitchell)

This reduced the number of warnings in the server build log from 84 to 24.

This was requested in https://github.com/xymon-monitoring/xymon/issues/43

### DIFF
--- a/lib/links.c
+++ b/lib/links.c
@@ -144,9 +144,13 @@ void load_all_links(void)
 		load_links(dirname, notesskin);
 	}
 
-	/* Change xxx/xxx/xxx/notes into xxx/xxx/xxx/help */
-	strncpy(dirname, xgetenv("XYMONNOTESDIR"), sizeof(dirname));
-	p = strrchr(dirname, '/'); *p = '\0'; strncat(dirname, "/help", (sizeof(dirname) - strlen(dirname)));
+	/* If no XYMONHELPDIR, change xxx/xxx/xxx/notes into xxx/xxx/xxx/help */
+	if (xgetenv("XYMONHELPDIR")) strcpy(dirname, xgetenv("XYMONHELPDIR"));
+	else {
+		strncpy(dirname, xgetenv("XYMONNOTESDIR"), sizeof(dirname));
+		dirname[sizeof(dirname)-1] = '\0'; /* Make sure it is null terminated */
+		p = strrchr(dirname, '/'); *p = '\0'; strncat(dirname, "/help", (sizeof(dirname) - strlen(dirname)));
+	}
 	load_links(dirname, helpskin);
 
 	linksloaded = 1;
--- a/client/logfetch.c
+++ b/client/logfetch.c
@@ -326,8 +326,16 @@ char *logdata(char *filename, logdef_t *logdef)
 
 			dbgprintf(" - Last position was %u, found curpos location at %u in current buffer.\n", logdef->lastpos[1], bytesin);
 			t = strdup(fillpos);	/* need shuffle about to insert before this line */
+#if __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6)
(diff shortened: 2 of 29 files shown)


feld (approved):

feld:
approving this if you have no other changes to add.

lib/links.c
```diff
@@ -144,9 +144,13 @@ void load_all_links(void)
 		load_links(dirname, notesskin);
 	}
 
-	/* Change xxx/xxx/xxx/notes into xxx/xxx/xxx/help */
-	strncpy(dirname, xgetenv("XYMONNOTESDIR"), sizeof(dirname));
-	p = strrchr(dirname, '/'); *p = '\0'; strncat(dirname, "/help", (sizeof(dirname) - strlen(dirname)));
+	/* If no XYMONHELPDIR, change xxx/xxx/xxx/notes into xxx/xxx/xxx/help */
+	if (xgetenv("XYMONHELPDIR")) strcpy(dirname, xgetenv("XYMONHELPDIR"));
```
RolandRosenfeld, line 148: The xgetenv("XYMONHELPDIR") triggers a errprintf(), if XYMONHELPDIR is not defined, which results in a yellow xymongen column.

RolandRosenfeld:
Since I applied this patch, I observe the message
```
xgetenv: Cannot find value for variable XYMONHELPDIR
```
every minute in xymongen.log (resulting in a yellow xymongen column for the xymon server).

So at least https://github.com/xymon-monitoring/xymon/pull/60/changes#diff-33e8c01fe04407628710ef46d67675587529331179081743141f880e71bc3935 does more than just fixing warnings, but changes behavior.

Except this, I'm not fully sure, whether temporarily quieting warnings (-Wformat-overflow, -Wformat-truncation) via these #pragma calls is the right way.  Wouldn't it be better to fix the root cause instead (by increasing buffers or limiting number of copied bytes or the like).  My C skills are quite limited, so this may be easier to fix for someone else.

So from my point of view, I tend not to apply this patch, maybe cherry pick only some \0 terminations?

bonomani:
Thanks for carrying this warning-cleanup work forward.

I split the useful parts of this PR into two smaller PRs that fix the underlying warnings without carrying over the problematic pieces:

- #113: safe/no-behavior-change subset
- #114: remaining warning fixes without pragma suppression

Together these cover the real warning/root-cause fixes from this PR: unterminated strncpy buffers, strncpy(dst, src, strlen(src)) misuse, fixed-buffer sprintf risks, bad strncat bounds, misleading indentation, the runtime_s off-by-one, and RRD path truncation handling.

The split intentionally avoids the parts discussed above as problematic: #pragma-based warning suppression, the unrelated XYMONHELPDIR behavior change, and other non-warning behavior changes. Once #113 and #114 are merged, I think this PR can be closed as superseded while keeping this PR as the attribution/reference for the original warning-cleanup work.

bonomani:
Closing as superseded by #113 and #114, which carry this same r8123 warnings fix (Tom Schmidt, with Ralph Mitchell) as a reviewable pair: #113 is the safe, no-behavior-change subset; #114 is the remaining root-cause fixes without pragma suppression. Between them they cover everything here except the two unrelated behaviour changes (XYMONHELPDIR, the sig.c crash-message/getenv change), which were left out deliberately.

Thanks @RolandRosenfeld for getting this backport started — both PRs were extracted from it. Tracking continues in #43 and #106.

PR #67, opened 2026-02-10
improve explanation of escaping hex chars for cont tests in man page

### LINKED ISSUE #66: hosts.cfg man page has error for escaping hex characters
The section of the  hosts.cfg man page documenting the usage of escaped strings for "content" tests is wrong: it documents using "\XX" but "\xXX" is required. See also lib/encoding.c ; using just "\XX" leads to 

```
Unknown escape sequence ....
```

in xymonnet.log

alt36:
- make clear that only specific escape sequences are allowed
- fix explanation for escpaing a hex char
- use NN rather than XX for consistency

Fixes #66

### DIFF
--- a/common/hosts.cfg.5
+++ b/common/hosts.cfg.5
@@ -1104,9 +1104,9 @@ as a regular expression (except that <space> is not allowed)
 or as a message digest (typically using an MD5 sum or 
 SHA-1 hash).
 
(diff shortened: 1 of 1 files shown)


RolandRosenfeld (approved):
Sounds reasonable to me.

PR #69, opened 2026-02-12
Fix handling of $CARESLIB in configure.server.

### LINKED ISSUE #68: Building with CARESLIB set to something results in empty "-L"
During tests to build xymon on FreeBSD I noted, that linking is tried with 
```
cc [...] httpcookies.o -L -lcares -L/usr/local/lib -lldap [...]
```
Note the `-L` without a path before the `-lcares` option.

The root cause is in the following code in configure.server:
```
    if test "$CARESLIB" != ""; then
	echo "CARESLIBS = -L$LIB -lcares" >>Makefile
	echo "RPATHVAL += ${LIB}"        >>Makefile
    else
	echo "CARESLIBS = -lcares"       >>Makefile
    fi
```
If `$CARESLIB` is set, it uses `$LIB` (instead of `$CARESLIB`). Since `$LIB` is empty, this results in `-L` without path parameter.

RolandRosenfeld:
If $CARESLIB is set, it should be used and not $LIB (which is not defined). 

This fixes https://github.com/xymon-monitoring/xymon/issues/68.

### DIFF
--- a/configure.server
+++ b/configure.server
@@ -453,8 +453,8 @@ then
 	echo "CARESINCDIR = -I$CARESINC" >>Makefile
     fi
     if test "$CARESLIB" != ""; then
-	echo "CARESLIBS = -L$LIB -lcares" >>Makefile
-	echo "RPATHVAL += ${LIB}"        >>Makefile
+	echo "CARESLIBS = -L$CARESLIB -lcares" >>Makefile
+	echo "RPATHVAL += ${CARESLIB}"        >>Makefile
     else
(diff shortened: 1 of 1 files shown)


bonomani (approved):
The issue is correctly identified: configure.server tested $CARESLIB but used $LIB, which could generate an invalid -L flag.

The proposed change properly uses $CARESLIB for both CARESLIBS and RPATHVAL, restoring deterministic and correct linker behavior.

Build impact (minor binary size changes) is expected due to corrected link flags and does not indicate functional regression.

Approved

PR #71, opened 2026-02-14
Use cc (instead of gcc) in OpenBSD

RolandRosenfeld:
clang is the default compiler in OpenBSD, not gcc.

Since the compiler is hard coded in build/Makefile.OpenBSD, this should be changed to their standard compiler.


bonomani:
Hi Roland,

I had a more permissive version that still allows using GCC via environment variables — see PR #73.

Bruno

RolandRosenfeld:
#73 looks better than mine...

PR #72, opened 2026-02-14
RRDtool: compatibility for < 1.9 and >= 1.9

bonomani:
# RRDtool: ensure compatibility with < 1.9 and >= 1.9 (const argv ABI)

## Summary

RRDtool 1.9.x changed several APIs from `char **argv` to `const char **argv`.  
This PR replaces version-based checks with an ABI probe at build time and introduces a unified compatibility layer (`rrd_compat.h`).

All RRDtool calls are now routed through `xymon_rrd_*` wrappers, removing scattered `RRDTOOL19` conditionals and centralizing compatibility handling.

## What Changed

- ABI detection added to `build/rrd.sh`
- New `include/rrd_compat.h` with wrapper functions
- Refactored:
  - `xymond/do_rrd.c`
  - `web/perfdata.c`
  - `web/showgraph.c`
  - `build/test-rrd.c`
- Removed redundant `rrdfn == NULL` check

No functional changes intended.

## Checklist

- [x] Build without RRDtool (graphs disabled)
- [x] Build with RRDtool < 1.9
- [x] Build with RRDtool >= 1.9
- [x] No incompatible pointer warnings
- [ ] Basic runtime: create / update / fetch / graph

### DIFF
--- a/web/showgraph.c
+++ b/web/showgraph.c
@@ -31,6 +31,7 @@ static char rcsid[] = "$Id$";
 #include <rrd.h>
 
 #include "libxymon.h"
+#include "rrd_compat.h"
 
 #define HOUR_GRAPH  "e-48h"
 #define DAY_GRAPH   "e-12d"
@@ -1214,24 +1215,18 @@ void generate_graph(char *gdeffn, char *rrddir, char *graphfn)
 	/* All set - generate the graph */
 	rrd_clear_error();
 
-#ifdef RRDTOOL12
-    #ifdef RRDTOOL19
-	result = rrd_graph(rrdargcount, (const char **)rrdargs, &calcpr, &xsize, &ysize, NULL, &ymin, &ymax);
-    #else
-	result = rrd_graph(rrdargcount, rrdargs, &calcpr, &xsize, &ysize, NULL, &ymin, &ymax);
-    #endif
+	result = xymon_rrd_graph(rrdargcount, rrdargs, &calcpr, &xsize, &ysize, &ymin, &ymax);
 
 	/*
 	 * If we have neither the upper- nor lower-limits of the graph, AND we allow vertical 
 	 * zooming of this graph, then save the upper/lower limit values and flag that we have 
 	 * them. The values are then used for the zoom URL we construct later on.
 	 */
+#ifdef RRDTOOL12
 	if (!haveupperlimit && !havelowerlimit) {
 		upperlimit = ymax; haveupperlimit = 1;
 		lowerlimit = ymin; havelowerlimit = 1;
 	}
-#else
-	result = rrd_graph(rrdargcount, rrdargs, &calcpr, &xsize, &ysize);
 #endif
 
 	/* Was it OK ? */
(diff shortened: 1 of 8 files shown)


web/showgraph.c
```diff
@@ -31,18 +31,14 @@ static char rcsid[] = "$Id$";
...
 #define DAY_GRAPH   "e-12d"
 #define WEEK_GRAPH  "e-48d"
 #define MONTH_GRAPH "e-576d"
 
 /* RRDtool 1.0.x handles graphs with no DS definitions just fine. 1.2.x does not. */
-#ifdef RRDTOOL12
-#ifndef HIDE_EMPTYGRAPH
-#define HIDE_EMPTYGRAPH 1
-#endif
-#endif
```
RolandRosenfeld, line 45: This change implies, that HIDE_EMPTYGRAPH is no longer defined.
But before your change it was defined for rrdtool >= 1.2.
Later in this file we still have `#ifdef HIDE_EMPTGRAPH` parts, that are now disabled by precompiler.
This sounds wrong to me.
Seems that `#ifdef HIDE_EMPTGRAPH` below has to be removed (with the `blankimg` content enabled always).

web/showgraph.c
```diff
@@ -517,6 +513,7 @@ char *expand_tokens(char *tpl)
 {
 	static strbuffer_t *result = NULL;
 	char *inp, *p;
+	char numstr[10];
```
RolandRosenfeld: Why is this added here? numstr is already declared in line 590 and 619 locally.
Declaring it here again makes code less readable.

web/showgraph.c
bonomani: Fixed in d454cabc. I restored explicit empty-graph fallback behavior by default-enabling `HIDE_EMPTYGRAPH` in `web/showgraph.c` and keeping the guarded blank-image path active (`#if HIDE_EMPTYGRAPH` around `blankimg` and the `rrddbcount == 0` return block). This addresses the dead-guard/behavior-regression concern.

web/showgraph.c
bonomani: Fixed in a19fb7c2. Removed the redundant outer `numstr` declaration in `expand_tokens()` and kept the two local declarations where they are actually used. This matches upstream `main` style and removes the unused-variable warning.

bonomani:
Follow-up updates pushed on branch fix/rrdtool-const-argv-core:

- 433777fc: fail compile probe correctly in build/rrd.sh (returns failure instead of false compile-ok report).
- 433777fc: keep a minimal RRDTOOL12 -> RRD_CONST_ARGS=0 shim in rrd_compat.h for stale generated Makefiles.
- adb5414d: make build/test-rrd.c -Werror clean (time_t start/end, remove unused result).
- 8313b22c: refresh obsolete STACKIT compatibility comment in showgraph.c (comment-only).
Pre-RRDTOOL12 support remains obsolete; the RRDTOOL12 shim is transitional build compatibility only.

bonomani:
Simplifiy it with new PR

PR #73, opened 2026-02-14
openbsd: default to cc and gate GCC-only warning flag

bonomani:


RolandRosenfeld (approved):
Looks good to me. Especially the option to set CC=gcc to overrule the default.

PR #74, opened 2026-02-15
PCRE2 Migration - Exploration Branch

bonomani:
# PCRE2 Exploration Branch

---

## Branch Purpose

This branch explores a possible migration path from PCRE1 to PCRE2.

It is **not** intended for immediate merge.  
Its goal is to evaluate technical feasibility, scope, risks, and impact before preparing a smaller and reviewable proposal.

---

## Context

Initial PCRE compatibility work was introduced by @canepan.

This branch expands on that work to:

- Assess what is required to build cleanly with PCRE2 on modern systems.
- Evaluate whether a compatibility layer is justified or excessive.
- Identify minimal changes needed for 4.3.x.
- Measure behavioral impact.

It also incorporates feedback and compatibility concerns raised by @RolandRosenfeld and others regarding distribution-level PCRE2 adoption.

---

## Exploration Scope

This branch currently includes:

- A `pcre_compat.[ch]` boundary to test abstraction feasibility.
- PCRE2-first implementation experiments.
- Replacement of direct `pcre_*` calls in key modules.
- CI validation across:
  - PCRE1
  - PCRE2
  - Strict PCRE2 builds

Some hardening and cleanup work was also introduced during experimentation.

---

## Important Notes

- The current diff is intentionally larger than what would be acceptable for 4.3.x.
- A full compatibility layer may prove unnecessary or too complex.
- Behavior preservation remains a priority.
- Any compatibility impact must be explicitly documented if adopted.

This branch should be treated as a **technical sandbox**, not a ready-to-merge solution.

---

## Evaluation Goals

The purpose of this exploration is to determine:

- Can we move to PCRE2 with minimal behavioral change?
- Is a broad compatibility layer justified?
- What is the smallest viable delta for 4.3.x?
- Are there low-risk code improvements that can be extracted independently?

---

## Validation Areas

Testing focus areas include:

- Clean build under PCRE1 and PCRE2
- Strict PCRE2 builds (no legacy shims)
- Regex behavior parity
- RRD filtering paths
- Mail ACK processing
- CGI filtering paths
- Memory safety (valgrind)

---

## Next Steps

Depending on findings, the outcome may be:

- A reduced, reviewable proposal targeting 4.3.x
- Extraction of isolated improvements only
- Or abandonment of the abstraction layer approach

---

## Objective

Evaluate a sustainable and minimal path toward PCRE2 support, while preserving Xymon’s stability and behavior on the 4.3.x production line.


.github/workflows/build1.yml
```diff
@@ -0,0 +1,80 @@
...
+
+          - name: client CT=server
+            variant: client
+            conftype: server
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install build dependencies (PCRE1 only, strict)
```
canepan: (nit) This should mention PCRE3

.github/workflows/build1.yml
```diff
@@ -0,0 +1,80 @@
...
+          MANROOT=/usr/share/man \
+          INSTALLBINDIR=/usr/lib/xymon/server/bin \
+          INSTALLETCDIR=/etc/xymon \
+          INSTALLWEBDIR=/etc/xymon/web \
+          INSTALLEXTDIR=/usr/lib/xymon/server/ext \
+          INSTALLTMPDIR=/var/lib/xymon/tmp \
+          INSTALLWWWDIR=/var/lib/xymon/www \
+          ./configure --${{ matrix.variant }}
+
+      - name: Build (PCRE1 strict)
```
canepan: (nit) This should mention PCRE3

.github/workflows/build3.yml
```diff
@@ -0,0 +1,82 @@
+name: build3 (pcre2-strict-no-shims)
```
canepan: Does this obsolete build1?

build/pcre.sh
```diff
@@ -1,73 +1,147 @@
...
-	cd ..
-
-	if test "$PCREOK" = "NO"; then
-		echo "Missing PCRE include- or library-files. These are REQUIRED for xymond"
-		echo "PCRE can be found at http://www.pcre.org/"
-		echo "If you have PCRE installed, use the \"--pcreinclude DIR\" and \"--pcrelib DIR\""
-		echo "options to configure to specify where they are."
-		exit 1
-	fi
+echo "Checking for PCRE ..."
```
canepan, line 1: It's probably worth passing this through https://www.shellcheck.net/
i.e.: use `(`..`)` instead of backticks, etc

build/pcre.sh
```diff
@@ -1,73 +1,147 @@
...
+        PCRETESTSRC="test-pcre2.c"
+        PCRETESTOBJ="test-pcre2.o"
+        PCRETESTBIN="test-pcre2"
+        PCRELINKLIB="-lpcre2-8"
+        PCREDEF="-DPCRE2"
+fi
+
+echo "PCRE probe mode: PCRE$PCRE_MAJOR"
+
+cd build
```
canepan: ^------^ SC2164 (warning): Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
This is from shellcheck, but a better alternative would be to add `set -eu` at the start of the script

build/pcre.sh
```diff
@@ -1,73 +1,147 @@
...
+if test "$PCRELIB" != ""; then LIBOPT="-L$PCRELIB"; fi
+
+OS=`uname -s | sed -e's@/@_@g'` $MAKE -f Makefile.test-pcre clean
+
+OS=`uname -s | sed -e's@/@_@g'` \
+        PCREINC="$INCOPT" \
+        PCRETESTSRC="$PCRETESTSRC" \
+        PCRETESTOBJ="$PCRETESTOBJ" \
+        PCRETESTBIN="$PCRETESTBIN" \
+        PCRELINKLIB="$PCRELINKLIB" \
```
canepan, line 99: Are these needed?

build/pcre.sh
```diff
@@ -1,73 +1,147 @@
...
+        PCRETESTBIN="test-pcre"
+        PCRELINKLIB="-lpcre"
+
+        OS=`uname -s | sed -e's@/@_@g'` $MAKE -f Makefile.test-pcre clean
+        OS=`uname -s | sed -e's@/@_@g'` \
+                PCREINC="$INCOPT" \
+                PCRETESTSRC="$PCRETESTSRC" \
+                PCRETESTOBJ="$PCRETESTOBJ" \
+                PCRETESTBIN="$PCRETESTBIN" \
+                PCRELINKLIB="$PCRELINKLIB" \
```
canepan, line 127: Are these needed?

build/pcre.sh
```diff
@@ -1,73 +1,147 @@
...
+        echo "ERROR: Cannot compile using PCRE library."
+        PCREOK="NO"
+fi
+
+OS=`uname -s | sed -e's@/@_@g'` \
+        PCRELIB="$LIBOPT" \
+        PCRETESTSRC="$PCRETESTSRC" \
+        PCRETESTOBJ="$PCRETESTOBJ" \
+        PCRETESTBIN="$PCRETESTBIN" \
+        PCRELINKLIB="$PCRELINKLIB" \
```
canepan, line 143: Are these needed?

build/pcre.sh
```diff
@@ -1,73 +1,147 @@
...
+fi
+
+echo "PCRE probe mode: PCRE$PCRE_MAJOR"
+
+cd build
+
+if test "$PCREINC" != ""; then INCOPT="-I$PCREINC"; fi
+if test "$PCRELIB" != ""; then LIBOPT="-L$PCRELIB"; fi
+
+OS=`uname -s | sed -e's@/@_@g'` $MAKE -f Makefile.test-pcre clean
```
canepan: (nit) Either remove `-e` or add a space in between: `sed -e 's@/@_@g'`
Same for below occurrences

build/pcre.sh
```diff
@@ -1,73 +1,147 @@
...
+
+if test $? -eq 0; then
+        echo "Linking with PCRE library works OK"
+else
+        echo "ERROR: Cannot link with PCRE library."
+        PCREOK="NO"
+fi
+
+OS=`uname -s | sed -e's@/@_@g'` $MAKE -f Makefile.test-pcre clean
+cd ..
```
canepan: I am not sure this is needed

build/pcre.sh
```diff
@@ -1,73 +1,147 @@
...
+# ------------------------------------------------------------
+if test -z "$PCRE_MAJOR"; then
+        echo "Auto-detecting PCRE version..."
+
+        if command -v pkg-config >/dev/null 2>&1; then
+                if pkg-config --exists libpcre2-8 2>/dev/null; then
+                        PCRE_MAJOR=2
+                        echo "PCRE2 detected via pkg-config"
+                elif pkg-config --exists libpcre 2>/dev/null; then
+                        PCRE_MAJOR=1
```
canepan, line 14: Should we be more explicit and use `legacy` (or `3`)?

canepan (commented):
No blockers, but I have no way to verify if it works on RedHat/CentOS

RolandRosenfeld:
Please don't feel offended, but is this really the way we want to go?
- I still don't think that we need backwards compatibility for legacy pcre, since all current systems support pcre2.  If you run some legacy hardware/OS, you usually won't run the xymon server on this, but only a client (with server configuration), where no pcre is used at all.
- I understand that switching from legacy pcre to pcre2 is a big step in a point release, so I also understand, that it may be a good idea to have a way back to the legacy pcre code, which is proven to work for ages.
- But this pull requests is not only a simple compatibility layer (as the original compat patch by @canepan was), but it seems to be a complete rewrite of pcre code including several logical changes (removal of null pointer checks and the like).  This pull request consists of 57 commits and the diff is over 4000 lines (excluding the .github directory) and still over 3000 lines when ignoring whitespace changes.
- My C and PCRE skills are too limited to review all these massive changes.
- Usually I would request to clean up the commit history and split out whitespace changes to a separate pull request, but the longer I think about pcre2 the less I think that we need legacy-pcre or such a complex compatibility layer. Especially if the main purpose of the compatibility layer is to avoid big code changes in a point release.

I'm really sorry for such a negative (maybe destructive) comment after all the work that has been invested into this pull request, but I think that this is the wrong way...

bonomani:
Hi Roland,

Thanks for the frank feedback - I am not offended. You are right: this PR grew far beyond a thin compat layer, and the scope/diff (plus formatting changes) makes it hard to review and too risky for a point release. I likely overreached.

Quick update: I have applied canepan’s line-level fixes (CI wording + shell-script safety/nits) on this branch, and I am applying the same safe fixes on your branch.

I agree it should not be merged as-is. I will compare both branches and see whether anything useful can be extracted from mine for yours, or kept for a next round.

On the breaking-change concern: it is valid and should not be hidden. With devel already at 4.4 (alpha) and production on 4.3.x, the practical path may be a clearly documented compatibility adjustment in a 4.3.x minor release, with explicit release and migration notes.

Target: 4.3.x only (4.3.30 production). Approach: PCRE2-first, minimal wrapper only where strictly needed, minimal behavior change, and explicit documentation of any compatibility impact.

Bruno

build/pcre.sh
bonomani: Yes, they are needed. `build/pcre.sh` passes `PCRETESTSRC/PCRETESTOBJ/PCRETESTBIN/PCRELINKLIB` so `build/Makefile.test-pcre` can switch between PCRE1 and PCRE2 test targets. These are consumed in `build/Makefile.test-pcre` at the variable definitions (lines 3-6) and compile/link rules (lines 9 and 12). Without them, the Makefile falls back to defaults (`test-pcre.c`, `-lpcre`) and we lose the PCRE2/fallback behavior.

build/pcre.sh
bonomani: Also needed here. This block is the fallback compile path after switching mode, and `PCRETESTSRC/PCRETESTOBJ/PCRETESTBIN/PCRELINKLIB` are the mechanism used by `build/Makefile.test-pcre` (lines 3-6, 9, 12) to select the active test tuple. Dropping them would revert to defaults and make the fallback behavior unreliable.

build/pcre.sh
bonomani: Needed as well. This is the `test-link` invocation, and `build/Makefile.test-pcre` uses these variables directly in the link rule (line 12) to link the selected test object with the selected PCRE library. Keeping the tuple explicit avoids silently falling back to defaults.

build/pcre.sh
bonomani: Good question. In the current structure this is needed because `build/pcre.sh` is sourced from `configure.server` / `configure.client` (not executed in a subshell). We `cd build` to run probe targets, so we must return with `cd ..` before control goes back to configure; otherwise the next sourced scripts (e.g. `build/c-ares.sh`) would be resolved from the wrong directory. I added guards around these `cd` calls in this branch for safety.

build/pcre.sh
bonomani: Valid point on naming clarity. In this branch `PCRE_MAJOR` is intentionally numeric with a strict domain (`1` or `2`) because configure/CI already pass and consume these values directly. Current mapping is: `1` = legacy PCRE API (Debian package `libpcre3-dev`, link `-lpcre`), `2` = PCRE2 (`libpcre2-dev`, link `-lpcre2-8`). I kept that contract for compatibility with existing configure paths and CI knobs.

bonomani:
Thanks canepan.

Understood on the RedHat/CentOS coverage - I cannot verify that yet either. At the moment, the CI signal is limited to the environments we already exercise on this PR branch.

For context on the scope reset and release targeting, see my note to Roland:
[PR comment (Roland note)](https://github.com/xymon-monitoring/xymon/pull/74#issuecomment-3910386468)

Current situation: this branch became a useful deep exploration, but the resulting diff is too large and mixed (logic + formatting) to be safely merged as-is. The exploration was still valuable because it clarified what looks low-risk to carry now vs what should be split out.

Progression I propose from here:
- Treat this as an exploration branch, not a merge-ready change set.
- Compare the parallel variants and extract only the smallest, reviewable, low-risk subset.
- Target the production line (4.3.x) with a PCRE2-first approach and minimal wrapper only where strictly needed, with explicit release/migration notes for any compatibility impact.

Separately, I also have ongoing workflow work on `cmake/bootstrap`, mainly around the reference-validation suite. The workflows on my branch are visible individually, for example:

[ref-validate-all.yml](https://github.com/bonomani/xymon/actions/workflows/ref-validate-all.yml)
[ref-valid-alpine.yml](https://github.com/bonomani/xymon/actions/workflows/ref-valid-alpine.yml)
[ref-valid-debian.yml](https://github.com/bonomani/xymon/actions/workflows/ref-valid-debian.yml)
[ref-valid-freebsd.yml](https://github.com/bonomani/xymon/actions/workflows/ref-valid-freebsd.yml)
[ref-valid-macos.yml](https://github.com/bonomani/xymon/actions/workflows/ref-valid-macos.yml)
[ref-valid-netbsd.yml](https://github.com/bonomani/xymon/actions/workflows/ref-valid-netbsd.yml)
[ref-valid-openbsd.yml](https://github.com/bonomani/xymon/actions/workflows/ref-valid-openbsd.yml)
[ref-valid-opensuse.yml](https://github.com/bonomani/xymon/actions/workflows/ref-valid-opensuse.yml)
[ref-valid-ubuntu.yml](https://github.com/bonomani/xymon/actions/workflows/ref-valid-ubuntu.yml)
[ref-generate-all.yml](https://github.com/bonomani/xymon/actions/workflows/ref-generate-all.yml)
[ref-make-linux.yml](https://github.com/bonomani/xymon/actions/workflows/ref-make-linux.yml)
[ref-make-freebsd.yml](https://github.com/bonomani/xymon/actions/workflows/ref-make-freebsd.yml)
[ref-make-openbsd.yml](https://github.com/bonomani/xymon/actions/workflows/ref-make-openbsd.yml)
[ref-make-netbsd.yml](https://github.com/bonomani/xymon/actions/workflows/ref-make-netbsd.yml)
[ref-sync.yml](https://github.com/bonomani/xymon/actions/workflows/ref-sync.yml)

That `cmake/bootstrap` work is still WIP and not ready for merge yet; it needs cleanup and stabilization before I can propose it.

bonomani:
PCRE2 - Compatibility layer (PCRE1 + PCRE2) vs full migration

Build validation
- Built the current codebase with PCRE2-only on all platforms listed below.
- Oldest tested: CentOS 7 (build OK).
- Note: CentOS 7 packages are no longer on standard RHEL repos; an archive mirror is required.

Decision
- Close this PR.
- Extract the useful parts.
- Proceed with the full PCRE2 migration based on Roland's work.

PR #76, opened 2026-02-20
make: fix parallel legacy builds (make -j4)

### LINKED ISSUE #3: Building in parallel (make -j4) does not work
When you try to build xymon (server as well as client) with 
```
./configure --server    # or --client
make -j4
```
this results in random compiler errors like
```
gcc -g -O2 -Wall -Wno-unused -I/usr/include/tirpc -Wno-pointer-sign -D_REENTRANT -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -DLINUX -I/home/runner/work/xymon/xymon/include -DCLIENTONLY=1 -I../include    -c -o osdefs.o osdefs.c
In file included from osdefs.h:15,
                 from osdefs.c:16:
/home/runner/work/xymon/xymon/include/config.h:2: error: unterminated #ifndef
    2 | #ifndef __CONFIG_H__
      | 
make[1]: *** [<builtin>: osdefs.o] Error 1
```
As a workaround using `make -j1` does not trigger these problems.

We should fix Makefile dependencies to make sure, that building happens in right order with `make -j4`, too.

bonomani:
## Summary
This fixes race conditions in legacy `make` builds when running in parallel (`-j4`, `-j8`, ...).

### What changes
- `build/Makefile.rules`
  - Make `build-build` depend on `include/config.h` and the correct lib target for each mode:
    - `CLIENTONLY=yes` -> `lib-client`
    - otherwise -> `lib-build`
  - Ensure `lib-client` dependencies are explicit and mode-correct.

- `lib/Makefile`
  - Add missing prerequisite on `test-endianness` for targets that execute it:
    - `sha1.o`, `rmd160c.o`, `md5`, `sha1`, `rmd160`
  - Add missing library prerequisites for `locator`:
    - `$(XYMONCOMMLIB)` and `$(XYMONLIB)`

### Why
Without these edges, parallel builds can race and fail with errors like:
- `./test-endianness: not found` / `Endianness is UNDEFINED`
- `cannot find libxymoncomm.a`

## Validation
Tested locally with parallel builds:
1. `./configure --server && make -j8`
2. `CONFTYPE=client ./configure --client && make -j8`
3. `CONFTYPE=server ./configure --client && make -j8`

All three complete successfully.

Fixes #3


RolandRosenfeld:
Could you please cross-check this with a very big concurrency level?
When I try to run this (to build server) with `make -j 1000`, it still fails on my system with
```
/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/14/../../../x86_64-linux-gnu/Scrt1.o: in function `_start':
(.text+0x17): undefined reference to `main'
collect2: error: ld returned 1 exit status
```
Sadly I cannot see what command triggered this error because of the concurrency...

Seems that the problem is triggered on my system randomly with `-j 22` to `-j 30`, while `-j 1000`  always seems to fail.

Since the error message reports missing `main` this seems to be building an executable that failed.
Several tests show that always one or more of client/xymon client/xymoncfg client/xymondigest client/xymongrep client/xymonlaunch common/xymon common/xymoncfg common/xymondigest common/xymongrep common/xymonlaunch are missing (started from common/Makefile).

But I'm not able to find the root cause. All I found was a missing xymondigest dependency on $(XYMONCOMMLIB), but adding this doesn't solve the issue here.
```diff
diff --git a/common/Makefile b/common/Makefile
-xymondigest: $(DIGESTOBJS) $(XYMONLIB)
+xymondigest: $(DIGESTOBJS) $(XYMONLIB) $(XYMONCOMMLIB)
    $(CC) $(CFLAGS) -o $@ $(DIGESTOBJS) $(XYMONCOMMLIBS) $(XYMONLIBS)
```

I played myself with build dependencies a month ago, too, but gave up a the end...

bonomani:
I cannot reproduce your problem: it passes for me even for 1000 (but maybe I am lucky)
but we have maybe a still a missing entry in the build/Makefile:

```
setup-newfiles: setup-newfiles.c
-	$(CC) -o $@ $(CFLAGS) $< ../lib/libxymonclient.a
+	$(CC) -I../include -o $@ $(CFLAGS) $< ../lib/libxymonclient.a

This resolve that:
bc@pc31:~/repos/github/bonomani/xymon [fix/make-j-gt1]$ make -C build clean
make: Entering directory '/home/bc/repos/github/bonomani/xymon/build'
rm -f merge-lines merge-sects setup-newfiles renamevars renametasks testfile *.o *~
make: Leaving directory '/home/bc/repos/github/bonomani/xymon/build'

bc@pc31:~/repos/github/bonomani/xymon [fix/make-j-gt1]$ make -C build -j1000 --output-sync=recurse V=1
make: Entering directory '/home/bc/repos/github/bonomani/xymon/build'
cc -o setup-newfiles  setup-newfiles.c ../lib/libxymonclient.a
setup-newfiles.c:11:10: fatal error: libxymon.h: No such file or directory
   11 | #include "libxymon.h"
      |          ^~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:14: setup-newfiles] Error 1
make: *** Waiting for unfinished jobs....
cc -o renamevars  renamevars.c
cc -o merge-sects  merge-sects.c
cc -o renametasks  renametasks.c
cc -o merge-lines  merge-lines.c
make: Leaving directory '/home/bc/repos/github/bonomani/xymon/build'
bc@pc31:~/repos/github/bonomani/xymon
```

bonomani:
"Forget my previous post. I’ll try to provide a proper analysis of the problem. It’s interesting that you can reproduce the issue on your machine; that’s the foundation we need to fix it."

bonomani:
By setting `-j=1000` (or another very high value), you are effectively testing whether the build can still complete under extreme parallelism; in that case, some processes could block and the compilation order can change significantly. This goes beyond the scope of the issue title, which only mentions `-j=4`.

In my tests, the proposed fix does not fully address the `-j` problem on GitHub-hosted runners (which already build relatively fast). My investigation suggests that if we want to reliably support `-j=1000`, we would need additional guards to enforce a deterministic build order and prevent deadlocks. That does not appear mandatory for now.

For the time being, this fix allows us to remove the `-j=1` constraint on GitHub runners, i.e., not force any specific `-j` value, while keeping local builds conservative (e.g., `-j=1`). This already speeds up the build workflow.

We can track the `-j=1000` extreme-parallelism case in a separate issue, to be addressed either against the current Make-based build or later in the CMake build once that work is ready.

Bruno

feld:
If this reduces the chance of running into the issue it's probably worth merging

bonomani:
It build much faster (which is very nice), but could fail (but i never saw it failing with j4). If it fails just retry to be sure if it is really a bug or revert to j1

RolandRosenfeld:
@bonomani: could you please get rid of the merge commit https://github.com/xymon-monitoring/xymon/pull/76/changes/a6a408168d08527c4d24840fe1e0447f4d035a26 from your merge request, which doesn't seem to add anything usefull.

In addition I'd suggest to add some dependencies on include/config.h (see https://github.com/RolandRosenfeld/xymon/commit/16752279cf5898d287047bbfedc0a0aca0b9c67a).

Last but not least, I'd suggest replacing the -j1 in github workflows (see https://github.com/RolandRosenfeld/xymon/commit/0864222f01bc9cb5460d3f7776049534822f8331).

Feel free to cherry-pick them from my https://github.com/RolandRosenfeld/xymon/tree/fix/make-j-gt1 branch.

bonomani:
Hi Roland,
  
  I did 2 of your 3 suggestions. I haven't added the include/config.h commit for now — we can always add it later. My
  tests (about 500 commits) never failed, so I'm not sure including the include/config.h commit would add much value.
  Also, adding a dependency (if it doesn't really bring value) doesn't seem like a good idea, since moving to another
  build process is already painful with the dependencies we currently have.

RolandRosenfeld (approved):
It's at least much better than the current behavior.
Maybe we have to add some other dependencies later.

PR #79, opened 2026-02-22
RRDtool: compatibility for < 1.9 and >= 1.9

bonomani:
# RRDtool: ensure compatibility with < 1.9 and >= 1.9 (const argv ABI)

## Summary

RRDtool 1.9.x changed several APIs from `char **argv` to `const char **argv`.

This PR replaces version-based checks with ABI probing at build time and introduces a unified compatibility layer in `include/rrd_compat.h`.

All RRDtool calls now go through `xymon_rrd_*` wrappers, removing scattered `RRDTOOL19` conditionals and centralizing argv-ABI handling.

## What changed

- Added ABI detection in `build/rrd.sh`:
  - compile-time probe for mutable vs const argv signatures
  - `pkg-config` fallback when probing is inconclusive
  - emits `-DRRD_CONST_ARGS=0|1`
- Added `include/rrd_compat.h`:
  - `xymon_rrd_argv_item_t`
  - `xymon_rrd_update/create/fetch/graph` wrappers
  - manual-build guard/error if `RRD_CONST_ARGS` is not defined
- Refactored call sites to use the compatibility layer:
  - `xymond/do_rrd.c`
  - `web/perfdata.c`
  - `web/showgraph.c`
  - `build/test-rrd.c`
- Updated build probe harness:
  - `build/Makefile.test-rrd` now includes `-I../include`
- Updated docs:
  - `docs/configure.txt` now documents manual `RRD_CONST_ARGS` definition for non-configure builds
- Removed obsolete/legacy branches tied to old macros (`RRDTOOL12`/`RRDTOOL19`) in touched paths
- Removed redundant `rrdfn == NULL` check in `xymond/do_rrd.c` (`rrdfn` is a static array)

## Notes

No functional change is intended for supported RRDtool versions (`>= 1.2`, including `< 1.9` and `>= 1.9`).

## Checklist

- [x] Build without RRDtool (graphs disabled)
- [x] Build with RRDtool < 1.9
- [x] Build with RRDtool >= 1.9
- [x] No incompatible-pointer warnings
- [ ] Basic runtime validation: create / update / fetch / graph

### DIFF
--- a/build/rrd.sh
+++ b/build/rrd.sh
@@ -73,7 +73,7 @@
 
 	# Probe + compile/link validation
 	RRDOK="YES"
-	OS=`uname -s | sed -e's@/@_@g'`
+	OS=$(uname -s | sed -e's@/@_@g')
 	if test "$RRDINC" != ""; then INCOPT="-I$RRDINC"; fi
 	if test "$RRDLIB" != ""; then LIBOPT="-L$RRDLIB"; fi
 
@@ -83,19 +83,19 @@
 		f=
 		n=0
 		if command -v mktemp >/dev/null 2>&1; then
-			f=`mktemp "${TMPDIR:-/tmp}/${prefix}.XXXXXX" 2>/dev/null` && { echo "$f"; return 0; }
+			f=$(mktemp "${TMPDIR:-/tmp}/${prefix}.XXXXXX" 2>/dev/null) && { echo "$f"; return 0; }
 		fi
 		while test $n -lt 50; do
 			f="${TMPDIR:-/tmp}/${prefix}.$$.$n"
 			(umask 077; : > "$f") 2>/dev/null && { echo "$f"; return 0; }
-			n=`expr $n + 1`
+			n=$(expr "$n" + 1)
 		done
 		return 1
 	}
 
 	# Probe whether rrd_update() expects const char ** (newer) or char ** (legacy).
 	detect_rrd_const_args() {
-		TESTOBJ=`mktemp_xymon "xymon-rrd-abi-obj"` || return 2
+		TESTOBJ=$(mktemp_xymon "xymon-rrd-abi-obj") || return 2
 		CONSTOK=0
 		MUTABLEOK=0
 		RRD_PROBE_CFLAGS=""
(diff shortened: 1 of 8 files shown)


build/rrd.sh
```diff
@@ -70,51 +70,93 @@
...
+			RRDOK="NO"
+		fi
+		;;
+	esac
+
+	# --- Compile / Link ---
+	cd build || exit 1
+	OS=$OS $MAKE -f Makefile.test-rrd clean
+
+	if OS=$OS RRDDEF="$RRDDEF" RRDINC="$INCOPT" \
```
canepan, line 134: I think `OS=$OS` can be omitted

build/rrd.sh
```diff
@@ -70,51 +70,93 @@
...
+		$MAKE -f Makefile.test-rrd test-compile >/dev/null 2>&1
+	then
 		echo "Compiling with RRDtool works OK"
 	else
 		echo "ERROR: Cannot compile with RRDtool."
 		RRDOK="NO"
 	fi
 
-	OS=`uname -s | sed -e's@/@_@g'` RRDLIB="$LIBOPT" PNGLIB="$PNGLIB" $MAKE -f Makefile.test-rrd test-link 2>/dev/null
+	OS=$OS RRDLIB="$LIBOPT" PNGLIB="$PNGLIB" $MAKE -f Makefile.test-rrd test-link >/dev/null 2>&1
```
canepan, line 143: I think `OS=$OS` and `PNGLIB="$PNGLIB"` can be omitted

build/rrd.sh
```diff
@@ -70,51 +70,93 @@
...
 	if test $? -ne 0; then
 		# Could be that we need -lz for RRD
 		PNGLIB="$PNGLIB $ZLIB"
 	fi
-	OS=`uname -s | sed -e's@/@_@g'` RRDLIB="$LIBOPT" PNGLIB="$PNGLIB" $MAKE -f Makefile.test-rrd test-link 2>/dev/null
+	OS=$OS RRDLIB="$LIBOPT" PNGLIB="$PNGLIB" $MAKE -f Makefile.test-rrd test-link >/dev/null 2>&1
```
canepan, line 148: I think `OS=$OS` and `PNGLIB="$PNGLIB"` can be omitted

bonomani:
Thank you very much, Nicola.

I checked all those fields, and it appears that RRD needs to be built with different flags depending on the operating system, and that libpng also needs to be properly exported. I will therefore keep those parts as they are.

I will close this PR, as Roland has already addressed the compatibility work for 1.9 in the current main branch. However, I intend to propose a more robust and cleaner implementation. I will also remove the deadline; this can still be included in a future release if we agree with Roland. Even if the implementation appears correct, I do not expect it to be perfect, so I would prefer that Roland have a final validation step. I am not entirely sure whether I can formally assign that, but I would like him to review it before any merge.

PR #80, opened 2026-02-28
rrd: detect argv const-API (1.9+/1.8 backports) and drop RRDtool <1.2 gates

### LINKED ISSUE #81: Detect RRDtool const-argv API by signature, not version (Oracle Linux 10 / RRDtool 1.8.0)
On Oracle Linux 10.1, the packaged RRDtool reports version 1.8.0, but the installed headers expose the newer `const char **` API.

This means version-based detection is not sufficient: `1.8.0` alone does not predict whether the API expects `char **` or `const char **`.

Facts

1. The system is Oracle Linux 10.1.

```text
/etc/oracle-release: Oracle Linux Server release 10.1
```

2. The installed RRDtool version is 1.8.0 (`rrdtool`, `pkg-config`, and `rrdtool-devel`).

```text
rrdtool --version: RRDtool 1.8.0
pkg-config --modversion librrd: 1.8.0
```

3. The headers expose `const char **`.

From `/usr/include/rrd.h`:

```c
int       rrd_graph(
int,
const char **,
...
);

int       rrd_update(
int,
const char **);

int       rrd_fetch(
int,
const char **,
...
);
```

These are the lines that show the relevant information: despite the package version being 1.8.0, the API signature is already `const char **`.

For comparison, the “normal” non-const form would be:

```c
int       rrd_update(
int,
char **);

int       rrd_graph(
int,
char **,
...
);

int       rrd_fetch(
int,
char **,
...
);
```

Impact

Code that passes `char **` will hit warnings or errors such as `expected 'const char **'` on this platform.

A feature test like `RRD_CONST_ARGS=1` looks like the right pivot, because this is an API property and not reliably derivable from the RRDtool version string.

Additional note

There is a possible include shadowing risk because the compiler searches `/usr/local/include` before `/usr/include`, and `/usr/local/include/rrd.h` exists on this machine.

However, `pkg-config --cflags librrd` is empty, and the system `/usr/include/rrd.h` already shows the `const` signatures above, so this Oracle Linux 10 package appears to expose the const API even without a local shadow header being involved.

One command note from the investigation: using `cc -H -E` on `/dev/null` does not prove which `rrd.h` is actually included, because it preprocesses an empty source. A minimal check is:

```sh
printf '#include <rrd.h>\n' | cc -H -E -xc - 2>&1 | head -n 50
```

Conclusion

Oracle Linux 10 ships RRDtool 1.8.0 with `const char **` function signatures in `rrd.h`. Xymon should detect the signature exposed by the headers instead of assuming it from the version number.

bonomani:
## Summary

This PR updates Xymon's RRDtool integration to key off the ABI exposed by
`rrd.h`, instead of guessing from the RRDtool version number.

It does three things together:

- detect whether the installed RRDtool headers use `char **argv` or `const char **argv`
- remove the old `RRDTOOL19` argument-type gate by routing callers through a shared ABI shim
- remove the legacy `RRDTOOL12` graph path and keep the modern `rrd_graph()` path

## Why

Version-based detection is not reliable enough here.

A concrete example is Oracle Linux 10, which ships RRDtool 1.8.0 headers with a
`const char **` ABI. In that environment, checking for "1.9+" is not sufficient;
the build has to detect the API that the headers actually expose.

## Changes

- detect the RRDtool argv ABI in `build/rrd.sh`
- export `RRD_CONST_ARGS`, with an explicit manual override when detection is ambiguous
- update the configure help and documentation for that override
- add a private compatibility shim in `lib/rrd_abi_compat.h`
- migrate RRD callers in:
  - `xymond/do_rrd.c`
  - `web/perfdata.c`
  - `web/showgraph.c`
  - `build/test-rrd.c`
- remove obsolete `RRDTOOL19` branches from the migrated code
- remove obsolete `RRDTOOL12` graph branches from the migrated code

## Notes

- this branch intentionally keeps the ABI detection and the runtime caller migration together, because they are part of the same compatibility change
- tested successfully on Oracle Linux 10 with RRDtool 1.8.0

## Supersedes

Supersedes #57.

## Tracking

Fixes #81

### DIFF
--- a/build/test-rrd.c
+++ b/build/test-rrd.c
@@ -1,14 +1,11 @@
 #include <stdio.h>
 
 #include <rrd.h>
+#include "../lib/rrd_api_compat.h"
 
-int main(int argc, char *argv[])
+int main(void)
 {
-#ifdef RRDTOOL19
-	const char *rrdargs[] = {
-#else
-	char *rrdargs[] = {
-#endif
+	xymon_rrd_argv_item_t rrdargs[] = {
 		"rrdgraph",
 		"xymongen.png",
 		"-s", "e - 48d",
@@ -28,7 +25,8 @@ int main(int argc, char *argv[])
 
 	for (pcount = 0; (rrdargs[pcount]); pcount++);
 	rrd_clear_error();
-	result = rrd_graph(pcount, rrdargs, &calcpr, &xsize, &ysize, NULL, &ymin, &ymax);
+	result = xymon_rrd_graph(pcount, rrdargs, &calcpr, &xsize, &ysize, NULL, &ymin, &ymax);
+	(void)result;
 
 	return 0;
 }
--- a/build/test-rrd.c
+++ b/build/test-rrd.c
@@ -28,11 +28,7 @@ int main(int argc, char *argv[])
 
 	for (pcount = 0; (rrdargs[pcount]); pcount++);
 	rrd_clear_error();
-#ifdef RRDTOOL12
 	result = rrd_graph(pcount, rrdargs, &calcpr, &xsize, &ysize, NULL, &ymin, &ymax);
-#else
-	result = rrd_graph(pcount, rrdargs, &calcpr, &xsize, &ysize);
-#endif
 
 	return 0;
 }
--- a/build/rrd.sh
+++ b/build/rrd.sh
@@ -75,11 +75,54 @@
 	RRDOK="YES"
 	if test "$RRDINC" != ""; then INCOPT="-I$RRDINC"; fi
 	if test "$RRDLIB" != ""; then LIBOPT="-L$RRDLIB"; fi
-	if pkg-config --atleast-version=1.9.0 librrd > /dev/null 2>&1; then
-		echo "Found RRDtool >= 1.9.0 via pkg-config"
-		RRDDEF="-DRRDTOOL19"
+
+	# Detect the RRDtool argv API by probing rrd.h directly, rather than
+	# inferring it from the RRDtool version (which is unreliable: e.g. some
+	# 1.8 backports already use the "const char **" prototype). We try to
+	# redeclare rrd_update() both ways and keep the one that matches.
+	probe_rrd_const_args() {
+		${CC:-cc} ${INCOPT} -x c -c -o /dev/null - >/dev/null 2>&1 <<'PROBE'
+#include <rrd.h>
+int rrd_update(int, const char **);
+int main(void) { return 0; }
+PROBE
+	}
+	probe_rrd_mutable_args() {
+		${CC:-cc} ${INCOPT} -x c -c -o /dev/null - >/dev/null 2>&1 <<'PROBE'
+#include <rrd.h>
+int rrd_update(int, char **);
+int main(void) { return 0; }
+PROBE
+	}
+
+	if test "$RRD_CONST_ARGS" != ""; then
+		case "$RRD_CONST_ARGS" in
+		0|1)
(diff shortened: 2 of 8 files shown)


bonomani:
Tested on a fresh Oracle Linux 10 installation with RRDtool 1.8.0.

With this branch, I was able to run `configure`, `make`, and `make install` successfully.

- `configure` completed without warnings
- Xymon startup was normal
- graph generation worked as expected
- graph zooming also worked correctly

This looks good on Oracle Linux 10 and matches the issue reported in #81.

If additional tests would be useful on this platform, let me know.

bonomani:
From Mailing list:
>Hi Bruno,
>sorry for the delay on my end.
>I did the steps you’ve asked me to do using the updated branch, pls find the results in the attached files.
>Further, again I’ve performed the debugging steps as you asked me in one of your earlier emails, the output is included also.
>The newly installed Xymon is working, rrd graphs show up fine, zooming works as expected.
>Hope you’ll find the information you’re looking for – for me I couldn’t see any difference compared to my previous install (but I don’t have your developer’s eyes).
>Regards
>Christian

Real testing completed successfully.
Ready for review - I have converted this PR from draft to ready for review.

RolandRosenfeld:
I try to review this pull request, but I'm a little unhappy with it mixing three things:

1. Auto-probing RRD API (with/without const) (incl. option to force one or the other)
2. Adding an RRD API abstraction layer (instead of #ifdef RRDTOOL19)
3. Removal of support for RRDTool <1.2

I like all three parts, but it hurts me a little to review them in one commit.

I'd prefer to first remove RRDtool <1.2 stuff to make code easier to read (including removal of HIDE_EMPTYGRAPH, the `#ifdef HIDE_EMPTYGRAPH`, which is always true on 1.2 and later. There's also a left over comment about rrdtool 1.0, that can be removed).

After this you could add the RRD API abstraction layer (based on `#ifdef RRDTOOL19`). I'd name this rrd_api, not rrd_abi, since not the binary interfaces changes but the programming interface.

Last but not least, replacing RRDTOOL19 option based on rrdtool version can be replaced by the API-variant-autoprober with (env-var override).

Doing this in 3 steps makes it easier to check, that every api call is catched and we avoid to accidentally used the rrdtool1.0 stuff somewhere or the like.

bonomani:
Should be corrected as asked

RolandRosenfeld:
I can approve commit [5d9e77d](https://github.com/xymon-monitoring/xymon/commit/5d9e77da8b266cce75349a8d161bd8b5e6e9eafa) (drop RRDtool <1.2 support).

build/test-rrd.c
```diff
@@ -26,9 +23,11 @@ int main(int argc, char *argv[])
 	int pcount, result, xsize, ysize;
 	double ymin, ymax;
 
+	(void)argc; (void)argv;
```
RolandRosenfeld: Why do we have these here? Maybe you could add an comment, that you set these parameters "used", otherwise this line is surprising (at least in this commit).

RolandRosenfeld:
72741588df1417d1ca9bd22e7eb244c09677414d also looks good.

RolandRosenfeld:
24f97c2a23f8e8548265fabbf3b3454d8592de68 also looks good.

build/test-rrd.c
bonomani: Good point. Since the probe never reads the command line, I dropped the parameters entirely (`int main(void)`) instead of commenting the casts — that removes the cause of the `-Wunused-parameter` warning, so there's nothing left to suppress. Compiles clean under `-Wall -Wextra`. Fixed in the latest push.

RolandRosenfeld (approved):
Now I'm happy with the code.

PR #82, opened 2026-03-04
rrd: apply Linux utimes workaround consistently

bonomani:
## Summary

This updates the Linux `utimes()` workaround in `xymond/do_rrd.c::flush_cached_updates()`.

The workaround is now applied to all Linux RRDtool builds, not only builds with
`RRDTOOL12` defined. This keeps the mtime refresh behavior consistent for modern
RRDtool builds using mmap-backed updates.

The `utimes()` call is also gated on `result == 0`, so a failed `rrd_update()`
does not refresh the file mtime and hide the failure from stale/nostale checks.

## Why

This is a follow-up to #80. Once the old RRDtool ABI gate is removed, the Linux
mtime workaround should no longer depend on `RRDTOOL12`.

Gating the timestamp refresh on success avoids expanding an existing edge case:
previously, a failed `rrd_update()` could still make the RRD file look fresh.

## Scope

- touches only `xymond/do_rrd.c`
- does not change RRDtool ABI detection
- does not change the `showgraph` cleanup from #80

### DIFF
--- a/xymond/do_rrd.c
+++ b/xymond/do_rrd.c
@@ -245,11 +245,12 @@ static int flush_cached_updates(updcacheitem_t *cacheitem, char *newdata)
 
 #if defined(LINUX)
 	/*
-	 * RRDtool 1.2+ uses mmap'ed I/O, but the Linux kernel does not update timestamps when
-	 * doing file I/O on mmap'ed files. This breaks our check for stale/nostale RRD's.
-	 * So do an explicit timestamp update on the file here.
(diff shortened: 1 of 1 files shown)


bonomani:
Superseded by #349 and closing.

This PR's whole change is `utimes()` only on a successful update — `do_rrd.c:289` there, plus the rest of what that timestamp needed: the stamp no longer compiled for Linux only, the value put back when RRDtool stamped the file before rejecting the reading (FreeBSD), and both failure paths reported. Nothing here is lost.

PR #89, opened 2026-05-17
PCRE2: fix leaks in headfoot/mailack, harden pickdata sizing

bonomani:
Three small focused PCRE2 hardening fixes:

- **headfoot**: free match data on class-filter early return
  (per-call leak in `wanted_host()`).
- **matching**: pattern-sized match data + NULL-check in `pickdata`.
- **xymon-mailack**: free PCRE2 resources on all exit paths.

A broader uniform hardening pass across the remaining PCRE2 call sites
is staged on `fix/pcre2-bulk-hardening` and will be opened as a follow-up
PR once this one merges — it touches `lib/headfoot.c`, `lib/matching.c`
and `xymond/xymon-mailack.c` on top of the fixes here, so it must land
after.

Note: other PCRE2 call sites still use the legacy
`pcre2_match_data_create(N, NULL)` pattern without NULL-checks. These
are addressed uniformly by the bulk follow-up, not patched piecemeal
here, to avoid churn when the same lines are rewritten to use
`pcre2_match_data_create_from_pattern()`.

### DIFF
--- a/lib/headfoot.c
+++ b/lib/headfoot.c
@@ -409,7 +409,10 @@ static void *wanted_host(char *hostname)
 
 	if (classpattern && hinfo) {
 		char *hostclass = xmh_item(hinfo, XMH_CLASS);
-		if (!hostclass) return NULL;
+		if (!hostclass) {
+			pcre2_match_data_free(ovector);
+			return NULL;
+		}
 
 		result = pcre2_match(classpattern, hostclass, strlen(hostclass), 0, 0,
 				ovector, NULL);
--- a/lib/matching.c
+++ b/lib/matching.c
@@ -168,9 +168,11 @@ int pickdata(char *buf, pcre2_code *expr, int dupok, ...)
 
 	if (!expr) return 0;
 
-	ovector = pcre2_match_data_create(30, NULL);
+	ovector = pcre2_match_data_create_from_pattern(expr, NULL);
+	if (!ovector) return 0;
+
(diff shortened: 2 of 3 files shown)


lib/headfoot.c
```diff
@@ -409,7 +409,10 @@ static void *wanted_host(char *hostname)
 
 	if (classpattern && hinfo) {
 		char *hostclass = xmh_item(hinfo, XMH_CLASS);
-		if (!hostclass) return NULL;
+		if (!hostclass) {
+			pcre2_match_data_free(ovector);
```
RolandRosenfeld, line 413: That looks correct to me.

lib/matching.c
```diff
@@ -168,9 +168,11 @@ int pickdata(char *buf, pcre2_code *expr, int dupok, ...)
 
 	if (!expr) return 0;
 
-	ovector = pcre2_match_data_create(30, NULL);
+	ovector = pcre2_match_data_create_from_pattern(expr, NULL);
```
RolandRosenfeld, line 171: This change looks good for this situation. After the change res <= 0 is also correct.

xymond/xymon-mailack.c
RolandRosenfeld, line 1: These all look good to me.

RolandRosenfeld (approved):

RolandRosenfeld (approved):
Accepted.

PR #94, opened 2026-05-21
xymongrep: add --no-dialup to filter dialup hosts (#24)

### LINKED ISSUE #24: [Xymon-developer] xymongrep & dialup hosts
```
Hi all,

The '--no-down' option to xymongrep omits hosts which are down according 
to https://xymon.sourceforge.io/xymon/help/manpages/man1/xymongrep.1.html 
. In common/xymongrep.c this is implemented with the test "if ((color == 
COL_RED) || (color == COL_BLUE))". When hosts are offline but marked as 
dialup, color == COL_CLEAR, so offline dialup hosts are probed despite 
'--no-down'.

The manpage doesn't address whether offline dialup hosts should or 
shouldn't be omitted, so it's hard to claim this behaviour is a bug. I 
expected dialup hosts to be omitted by default, but others may differ and 
it feels unwise to change well-established behaviour.

The attached patch adds a '--no-down-dialup' flag to explicitly omit these 
hosts. I'd appreciate feedback, comments, criticism etc.

Yours,

Frank
```

[xymongrep.patch](https://github.com/user-attachments/files/24756053/xymongrep.patch)

https://sourceforge.net/p/xymon/mailman/xymon-developer/?viewmonth=202203

bonomani:
Fixes #24.

`xymongrep --no-down` filters hosts whose `conn` status is RED or BLUE. Hosts declared `dialup` in hosts.cfg are typically shown as CLEAR when offline, so they slip through `--no-down` even though the user already knows they cannot be reached.

`--no-dialup` excludes them on the `dialup` **attribute** rather than on the current colour, so it cannot be confused with hosts that are CLEAR for another reason — scheduled downtime, manual suppression.

## Implementation

- new helper `dialupok()` using `xmh_item(hinfo, XMH_FLAG_DIALUP)`
- one added condition in the existing host filter: `... && dialupok(hwalk, nodialup) && knownhost(...)`
- argv parsing for `--no-dialup`, plus the man page entry under OPTIONS and the flag in SYNOPSIS

Without `--no-dialup`, behaviour is unchanged; existing options, output reconstruction and `--noextras` are left alone.

## Compared to the patch proposed in #24

The issue proposed `--no-down-dialup` driven by `color == COL_CLEAR`, which would also match non-dialup hosts that are CLEAR for other reasons. Checking the static host attribute directly makes the option do exactly what its name says.

## Testing

Manual fixture only — a `hosts.cfg` with 5 hosts tagged `ssh`, 2 of them also tagged `dialup`:

| Command | Result |
| --- | --- |
| `xymongrep ssh` | 5 hosts, including the 2 dialup |
| `xymongrep --no-dialup ssh` | the 3 non-dialup only |
| `xymongrep --no-down --no-dialup ssh` | both filters apply |
| `xymongrep --noextras --no-dialup ssh` | works with compact output |

No regression test is added under `tests/`, so the option is exercised by neither the suite nor CI — worth adding before this lands if the suite is expected to guard it.

### DIFF
--- a/common/xymongrep.1
+++ b/common/xymongrep.1
@@ -6,7 +6,7 @@ xymongrep \- pick out lines in hosts.cfg
 .br
 .B "xymongrep \-\-version"
 .br
-.B "xymongrep [\-\-noextras] [\-\-test\-untagged] [\-\-web] [\-\-net] [\-\-loadhostsfromxymond] TAG [TAG...]"
+.B "xymongrep [\-\-noextras] [\-\-test\-untagged] [\-\-no\-dialup] [\-\-web] [\-\-net] [\-\-loadhostsfromxymond] TAG [TAG...]"
 
 .SH DESCRIPTION
 .I xymongrep(1)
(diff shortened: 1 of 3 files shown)


n1tehawk (approved):
The core change relies on an existing `xmh_item()` test and looks sane. Surrounding C code is mostly trivial and non-critical. Two other files committed document the new command line flag properly.

PR #95, opened 2026-05-21
docs: use /var/log/syslog in analysis.cfg(5) LOG example

bonomani:
The example used `/var/adm/syslog`, which is the SVR4-era path still
found on some legacy Unix variants (Solaris, HP-UX). On Linux — by
far the dominant Xymon platform — the canonical path under FHS is
`/var/log/syslog`. Switch the example accordingly so users can copy
the LOG line verbatim instead of having to translate it.

The companion example on the next paragraph already uses
`/var/log/daemon.log`, so this also makes the two examples consistent
within the same manpage section.

Picked up from Debian's `21_FHS-instead-FSSTND-in-example-in-man-page.patch`.

### DIFF
--- a/docs/manpages/man5/analysis.cfg.5.html
+++ b/docs/manpages/man5/analysis.cfg.5.html
@@ -309,10 +309,10 @@ filter out any unwanted strings that happen to match &quot;pattern&quot;.
 The <B>OPTIONAL</B> keyword causes the check to be skipped if the logfile
 does not exist.
 <P>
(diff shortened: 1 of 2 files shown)


RolandRosenfeld:
We should not forget to rebuild the HTML versions of the man pages before releasing (currently docs/manpages/man5/analysis.cfg.5.html still has /var/adm/syslog).

I first thought about adding this to .github/workflow-release.yaml, but I fear this is too late, since the updated HTML files should be in the tag and workflow-release.yaml only works after tagging.

RolandRosenfeld (approved):
Looks good.

PR #96, opened 2026-05-21
client/linux: env-var-controlled df/inode filesystem filter (#49)

### LINKED ISSUE #49: Add an option to define, which filesystem types should be included/excluded from df/inode
client/xymonclient-linux.sh currently excludes several filesystems (especially nfs, nfs4, ceph) from df/inode output in the clientlog.

This is either implemented via explicitly excluding the filesystem types listed as nodev in /proc/filesystems or by using `df` option `-l` (limit listing to local file systems).

Debian patch https://salsa.debian.org/debian/xymon/-/blob/master/debian/patches/24_hobbitclient-tmpfs.patch modifies this a little (to not exclude tmpfs and squashfs and add support for Linux with FreeBSD kernel), but in the result the excluded/included filesystem types are still hardcoded in a file below /usr/lib/xymon.

I'm looking for some mechanism to disable the `-l` option of `df` and to whitelist other filesystem types (like nfs, nfs4, ceph) via an option in /etc/xymon, which isn't overridden on xymon update.

bonomani:
Closes #49.

## Summary

This PR makes Linux client filesystem filtering configurable, and updates the
default filesystem-selection policy so the common "real but `nodev`" and
"always-100%-full" cases are handled out of the box.

Previously `client/xymonclient-linux.sh` decided internally which filesystems
appeared in the `df` and `inode` sections: it excluded every `nodev` type from
`/proc/filesystems` plus a hardcoded `iso9660`, keeping `rootfs`. That policy
was too rigid for hosts using `tmpfs`, `zfs`, `virtiofs`, `nfs`, `ceph`,
`overlay`, `fuse`, snaps, etc. Downstreams already patch it locally, e.g.
Debian's `24_hobbitclient-tmpfs.patch`.

## Defaults

With none of the new variables set, the client now:

- excludes `nodev` filesystems listed by `/proc/filesystems`, **except** the
  real local types `zfs virtiofs tmpfs`, which are surfaced
  (`XYMONCLIENT_FS_INCLUDE_TYPES` default);
- additionally excludes `iso9660 squashfs fuse.snapfuse`
  (`XYMONCLIENT_FS_EXCLUDE_TYPES` default) -- read-only images reported 100% full
  by design (snaps mount as `squashfs`, or as `fuse.snapfuse` where snapd falls
  back to FUSE); none is a `nodev` type, so each must be named explicitly;
- keeps `rootfs`;
- passes `-P -l` to `df` (local filesystems only);
- in the **inode** report only, drops filesystems with no inode limit (df prints
  `-` in the `IUse%` column for them, e.g. btrfs/zfs/9p/many fuse) -- they can
  never run out of inodes, so the row is noise and may carry bogus counts.

This is a deliberate change of default behavior from the old "hide all nodev +
iso9660" policy: `zfs`/`virtiofs`/`tmpfs` now appear, and `squashfs`/snaps no
longer do. Remote filesystems (`nfs`, `ceph`, ...) are still hidden by default.

The earlier draft of this PR added a per-invocation `df` timeout; that has been
**dropped** in favor of the non-blocking df sentinel, so there is no longer any
`XYMONCLIENT_FS_DF_TIMEOUT` variable or timeout-related behavior.

When `/proc/filesystems` is **not** readable, the client disables the dynamic
nodev derivation and warns; the explicit `XYMONCLIENT_FS_EXCLUDE_TYPES`
exclusions and the local-only `df -l` behavior remain active. This replaces the
empty `df`/`inode` sections the previous code produced in that case (see Safety).

## Configuration

The following variables are documented in `xymonclient.cfg` and
`xymonclient.cfg(5)`:

| Variable | Default | Purpose |
| --- | --- | --- |
| `XYMONCLIENT_FS_INCLUDE_TYPES` | `zfs virtiofs tmpfs` | `nodev` types to surface anyway. Set to `""` to restore the historical "drop every nodev type" behavior. |
| `XYMONCLIENT_FS_EXCLUDE_TYPES` | `iso9660 squashfs fuse.snapfuse` | Non-nodev types to also hide. Set to `""` to monitor these too. |
| `XYMONCLIENT_FS_DF_LOCAL_ONLY` | `yes` | Set to `no` to drop `df -l` and report remote filesystems. |

Matching is on the exact `df` type token (literal, never glob/regex). If a type
appears in both include and exclude lists, exclusion wins. All three variables
reach both the `df` and the `inode` report.

To include remote storage, both parts of the default filtering policy must be
changed:

```sh
XYMONCLIENT_FS_INCLUDE_TYPES="zfs virtiofs tmpfs nfs nfs4 ceph"
XYMONCLIENT_FS_DF_LOCAL_ONLY="no"
```

### Relationship to Debian's `24_hobbitclient-tmpfs.patch`

The new defaults already reproduce that patch's filesystem-**type** policy
(`tmpfs`/`zfs` shown). It is still **not** a drop-in replacement for the whole
patch, which also:

- derives exclusions from `mount` when `/proc/filesystems` is unavailable -- this
  PR instead **disables** the dynamic nodev filter in that case (a behavioral
  difference);
- filters `/sys`;
- carries related `analysis.cfg` ignores (see Upgrade note below).

The mechanism does not silently adopt a downstream distribution's defaults
upstream; it just makes the common cases the shipped default.

## Safety

- Invalid `XYMONCLIENT_FS_DF_LOCAL_ONLY` values warn and fall back to `yes`.
- A `df` that exits non-zero with **no output** is reported as an explicit
  failure marker (yellow) rather than a silent empty section. An empty `inode`
  section would otherwise be read by the server as a false green ("No
  filesystems reporting inode data"); a `df` exit that still prints the healthy
  mounts (e.g. one unreadable mount, exit 1) flows through unchanged.
- If `/proc/filesystems` is unreadable, the client warns and disables the
  dynamic nodev filter.
- Filesystem type settings are handled as literal tokens, without pathname
  expansion.

## Upgrade note

Because the default now surfaces `tmpfs`, fresh installs ship `analysis.cfg`
rules that ignore the noisy `/run*` tmpfs mounts (`/run`, `/run/lock`,
`/run/user/N`) server-side. **Existing servers keep their preserved
`analysis.cfg`, which will not receive these new ignores automatically** -- after
upgrading clients, expect extra `/run*` rows on the disk report until those
ignore rules are added to the existing `analysis.cfg` manually (or it is
regenerated from the shipped default).

The client settings themselves are read from the normal client environment file.
Native installs preserve active local values through `merge-lines`; Debian
installs expose that file as `/etc/xymon/xymonclient.cfg`.

## Tests

`tests/client/fs-filter.sh` covers:

- the new default filtering policy (nodev excluded, `zfs`/`virtiofs`/`tmpfs`
  kept, `iso9660`/`squashfs`/`fuse.snapfuse` excluded, `rootfs` kept)
- include and exclude lists and precedence
- literal token handling
- local-only and remote filesystem modes, including invalid-value fallback
- a non-zero `df` with empty output emitting a non-empty, header-free failure
  marker rather than the empty section that would false-green the inode report
- consistent `df` and inode filtering (settings reach both reports)
- missing `/proc/filesystems`

All build and shell regression workflows pass.

### DIFF
--- a/client/xymonclient-linux.sh
+++ b/client/xymonclient-linux.sh
@@ -46,15 +46,14 @@ echo "[who]"
 who
 echo "[df]"
 # Default: exclude every nodev (pseudo) filesystem in df/inode output, except
-# rootfs, plus the historical explicit iso9660 exclusion (iso9660 is not a
-# nodev type, so it is added unconditionally at run_df below). This is the
-# historical upstream behavior.
+# rootfs. The always-100%-full read-only images (iso9660, squashfs) are not
+# nodev types and are excluded via XYMONCLIENT_FS_EXCLUDE_TYPES below.
 if [ -r /proc/filesystems ]; then
 	EXCLUDES=$(awk '$1 == "nodev" && $2 != "rootfs" { printf "%s%s", sep, $2; sep=" " }' /proc/filesystems)
 else
-	# Only the dynamic nodev exclusions are disabled here; the explicit iso9660
-	# exclusion and the local-only df -l behavior still apply.
-	echo "xymonclient-linux: /proc/filesystems not readable, dynamic nodev exclusions disabled (iso9660 exclusion and df -l still apply)" >&2
+	# Only the dynamic nodev exclusions are disabled here; the EXCLUDE_TYPES
+	# defaults (iso9660/squashfs) and the local-only df -l behavior still apply.
+	echo "xymonclient-linux: /proc/filesystems not readable, dynamic nodev exclusions disabled (EXCLUDE_TYPES and df -l still apply)" >&2
 	EXCLUDES=""
 fi
 # Filesystem types are literal tokens, so pathname expansion must not turn a
@@ -63,11 +62,15 @@ case $- in
 	*f*) FSRESTOREGLOB=no ;;
 	*) FSRESTOREGLOB=yes; set -f ;;
 esac
-# XYMONCLIENT_FS_INCLUDE_TYPES: whitespace-separated FS types that should
-# appear in the output even though they would otherwise be excluded
-# (e.g. "tmpfs zfs" to surface tmpfs/zfs mounts, or "nfs nfs4 ceph" to
-# include remote filesystems -- the latter also requires
-# XYMONCLIENT_FS_DF_LOCAL_ONLY=no since df -l hides remote mounts).
+# XYMONCLIENT_FS_INCLUDE_TYPES: whitespace-separated FS types to surface even
+# though they are flagged nodev and would otherwise be dropped. Defaults to the
+# real local filesystems that merely happen to be nodev: zfs (pools have no
+# single block device), virtiofs (VM-shared storage) and tmpfs (RAM-backed --
+# the noisy /run* tmpfs mounts are filtered server-side in analysis.cfg). Set
+# to "" to restore the historical "exclude every nodev type" behaviour, or add
+# remote types e.g. "nfs nfs4 ceph" (which also needs
+# XYMONCLIENT_FS_DF_LOCAL_ONLY=no, since df -l hides remote mounts).
+: "${XYMONCLIENT_FS_INCLUDE_TYPES=zfs virtiofs tmpfs}"
 if [ -n "$XYMONCLIENT_FS_INCLUDE_TYPES" ]; then
 	# Exact token comparison -- a type is matched literally, never as a
 	# regex/glob, so e.g. "fuse.sshfs" cannot collide with another token.
@@ -81,11 +84,15 @@ if [ -n "$XYMONCLIENT_FS_INCLUDE_TYPES" ]; then
 	done
 	EXCLUDES="$keep"
 fi
-# XYMONCLIENT_FS_EXCLUDE_TYPES: whitespace-separated FS types to ALSO
-# exclude, on top of the nodev default. Matching is on the exact df type
-# token; nodev types (overlay, fuse, ...) are already excluded, so an
-# effective entry names a non-nodev type, e.g. "fuse.sshfs vfat" to drop a
-# specific FUSE subtype and a device-backed mount.
+# XYMONCLIENT_FS_EXCLUDE_TYPES: whitespace-separated FS types to ALSO exclude,
+# on top of the nodev default. Defaults to "iso9660 squashfs" -- both are
+# read-only images reported 100% full by design (snap mounts one squashfs per
+# revision), and neither is a nodev type, so they must be named here. Matching
+# is on the exact df type token; nodev types (overlay, fuse, ...) are already
+# excluded, so other effective entries name a non-nodev type, e.g. adding
+# "fuse.sshfs vfat" to drop a specific FUSE subtype and a device-backed mount.
+# Set to "" to monitor iso9660/squashfs too.
+: "${XYMONCLIENT_FS_EXCLUDE_TYPES=iso9660 squashfs}"
 if [ -n "$XYMONCLIENT_FS_EXCLUDE_TYPES" ]; then
 	for t in $XYMONCLIENT_FS_EXCLUDE_TYPES; do
 		case " $EXCLUDES " in
@@ -126,17 +133,10 @@ case "$DFTIMEOUT" in
 		DFTIMEOUT=30
 		;;
 esac
-# Strip leading zeros so a value like 00030 is treated as decimal 30, not as
-# an over-length string (the length guard below keys on character count, not
-# magnitude, so 00001 would otherwise clamp to 3600) nor as octal in the
-# numeric comparison. The case above guarantees DFTIMEOUT is all digits with at
-# least one non-zero, so this loop always terminates with a non-empty value.
-while :; do
-	case "$DFTIMEOUT" in
-		0?*) DFTIMEOUT="${DFTIMEOUT#0}" ;;
-		*) break ;;
-	esac
-done
+# Normalise to decimal (expr is decimal by spec) so a zero-padded value like
+# 00030 becomes 30: the length guard below keys on character count, and the
+# numeric comparison would otherwise read a leading zero as octal.
+DFTIMEOUT=`expr "$DFTIMEOUT" + 0`
 # Cap at 3600s. BusyBox timeout(1) rejects an out-of-range duration with a
 # nonzero exit before df runs, which would silently empty both the df and
 # inode sections; an hour is already far past any sane df wait. The length
@@ -152,7 +152,6 @@ run_df()
 	set -- -P
 	[ "$DFLOCALONLY" = yes ] && set -- "$@" -l
 	[ "$DFINODES" = yes ] && set -- "$@" -i
-	set -- "$@" -x iso9660
 
 	case $- in
 		*f*) DFRESTOREGLOB=no ;;
@@ -170,24 +169,12 @@ run_df()
 	fi
 }
 # emit_df INODEFLAG LABEL
-# Run df (optionally in inode mode) and reproduce the historical sed join, but
-# guard the hung-filesystem case. timeout(1) reports 124 (timed out) or 137
-# (128+9, our `-s KILL` case) when it has to kill df; whatever partial output it
-# managed to print first is untrustworthy -- it can list the healthy mounts and
-# omit the very one that hung -- so we discard it and emit an explicit failure
-# marker instead of a silent (empty) section. The server reads an *empty*
-# [inode] section as green ("No filesystems reporting inode data"), so a silent
-# hang would otherwise be a false green; a non-empty marker that carries no
-# recognisable df header drives the server's yellow "expected strings not found"
-# path instead, and the same marker yellows the disk section. The same false
-# green can come from any nonzero exit that prints nothing at all -- e.g.
-# timeout(1) itself failing to launch df (status 125/126/127) or a BusyBox
-# timeout rejecting its arguments before df runs -- so an empty section is
-# emitted as a failure marker whenever the exit status was nonzero, not only for
-# the kill codes. df's own nonzero exits that still print the healthy mounts
-# (e.g. a single unreadable mount, exit 1) keep their partial output and flow
-# through unchanged, preserving prior behaviour; an empty section with a clean
-# exit 0 (the Solaris all-ZFS inode case) is also left untouched.
+# Run df (optionally in inode mode) and reproduce the historical sed join.
+# The server reads an empty section as green, so a hung/failed df must not pass
+# silently: on a timeout kill (124/137) or any nonzero exit with no output, emit
+# a failure marker (no recognisable df header) to drive the server yellow. A
+# nonzero exit that still prints mounts (e.g. one unreadable mount) and a clean
+# empty exit 0 (Solaris all-ZFS inodes) keep their output unchanged.
 emit_df()
 {
 	DFOUT=`run_df "$1"`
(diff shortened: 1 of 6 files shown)


client/xymonclient-linux.sh
```diff
@@ -45,14 +45,65 @@ uptime
...
+EXCLUDES=`echo $EXCLUDES | sed -e 's! ! -x !g'`
+# XYMONCLIENT_FS_DF_LOCAL_ONLY: defaults to "yes" (current upstream behavior,
+# passes -l to df). Set to "no" to drop -l so that remote filesystems
+# (nfs, ceph, ...) appear in the output.
+DFOPTS="-P"
+[ "${XYMONCLIENT_FS_DF_LOCAL_ONLY:-yes}" = "yes" ] && DFOPTS="$DFOPTS -l"
+# XYMONCLIENT_FS_DF_TIMEOUT: seconds before df is killed (default 30).
+# Set to empty string ("") to disable. df can hang on stale NFS/CIFS mounts
+# -- particularly relevant when XYMONCLIENT_FS_DF_LOCAL_ONLY=no.
+DFTIMEOUT="${XYMONCLIENT_FS_DF_TIMEOUT-30}"
```
RolandRosenfeld: shouldn't this be ${XYMONCLIENT_FS_DF_TIMEOUT:-30} (the ":" is missing)

RolandRosenfeld: shouldn't this be ${XYMONCLIENT_FS_DF_TIMEOUT:-30} (the ":" is missing)

RolandRosenfeld: ooops, the line number changed...

RolandRosenfeld: fixed now.

client/xymonclient-linux.sh
```diff
@@ -45,17 +45,175 @@ uptime
...
+done
+# Cap at 3600s. BusyBox timeout(1) rejects an out-of-range duration with a
+# nonzero exit before df runs, which would silently empty both the df and
+# inode sections; an hour is already far past any sane df wait. The length
+# guard short-circuits the numeric comparison so it never overflows the
+# shell's integer type on an absurdly long digit string.
+if [ "${#DFTIMEOUT}" -gt 4 ] || [ "$DFTIMEOUT" -gt 3600 ]; then
+	echo "xymonclient-linux: XYMONCLIENT_FS_DF_TIMEOUT '$DFTIMEOUT' exceeds 3600, using 3600" >&2
+	DFTIMEOUT=3600
+fi
```
RolandRosenfeld: Is such a complex mechanism (34 lines of code) for DFTIMEOUT sanitizing really necessary?
Can't we just say: If someone writes bullshit in this variable, it's his personal problem?
I don't like this script to be blown up by this.

client/xymonclient-linux.sh
```diff
@@ -45,17 +45,175 @@ uptime
...
+# recognisable df header drives the server's yellow "expected strings not found"
+# path instead, and the same marker yellows the disk section. The same false
+# green can come from any nonzero exit that prints nothing at all -- e.g.
+# timeout(1) itself failing to launch df (status 125/126/127) or a BusyBox
+# timeout rejecting its arguments before df runs -- so an empty section is
+# emitted as a failure marker whenever the exit status was nonzero, not only for
+# the kill codes. df's own nonzero exits that still print the healthy mounts
+# (e.g. a single unreadable mount, exit 1) keep their partial output and flow
+# through unchanged, preserving prior behaviour; an empty section with a clean
+# exit 0 (the Solaris all-ZFS inode case) is also left untouched.
```
RolandRosenfeld: Really much comment for a shell script. Can't we shorten this?
And after reading all this, I still don't understand, whether/how a timed out df results in a non-green disk/inodes column. Maybe I need to try this out myself.

EDIT: tried this out now and disk/inodes become yellow, because `Expected strings (Capacity and Mounted) not found in df output`.
So the mechanism works as expected, only the explanation is too long and too hard to understand.

client/xymonclient.cfg.DIST
```diff
@@ -18,6 +18,57 @@ XYMONCLIENTLOGS="$XYMONHOME/logs"     # Where we store the client logfiles
...
+#    to "30"; zero or another invalid value produces a warning and also falls
+#    back to "30". The timeout cannot be disabled through this setting and may
+#    be set up to a maximum of 3600 seconds -- df must never hang the client
+#    cycle. A process
+#    stuck in an uninterruptible kernel wait cannot be terminated until the
+#    kernel call returns. If timeout(1) has to kill df, the affected df and
+#    inode report is marked failed (yellow) instead of being left empty, so a
+#    hung filesystem is not silently reported green.
+#
+# XYMONCLIENT_FS_INCLUDE_TYPES=""
```
RolandRosenfeld: I'd suggest to set `XYMONCLIENT_FS_INCLUDE_TYPES="tmpfs zfs"`, since both are "normal" filesystems, that should be checked, independently of the question, whether they have "nodev" set in /proc/filesystems.
Silently ignoring them is a bug. Keeping this bug because auf backward compatibility is the wrong way.

client/xymonclient.cfg.DIST
```diff
@@ -18,6 +18,57 @@ XYMONCLIENTLOGS="$XYMONHOME/logs"     # Where we store the client logfiles
...
+# XYMONCLIENT_FS_EXCLUDE_TYPES=""
```
RolandRosenfeld: I'd suggest setting `XYMONCLIENT_FS_EXCLUDE_TYPES="squashfs"`, since this is a read only filesystem, which is always filled 100% (same as iso9660).
Debian 24_hobbitclient-tmpfs.patch is buggy in this, since squashfs is not nodev (at least not on my OpenWRT router), so `cat /proc/filesystems | grep nodev | egrep -v 'squashfs'` is just useless, as far as I can see.
Maybe we should set `XYMONCLIENT_FS_EXCLUDE_TYPES="iso9660 squashfs"` and remove the (historic) hard coded exclusion of iso9660 (so it can be removed from all documentation).

EDIT: Or alternatively add `-x squashfs` next to `-x iso9660` in run_df().

RolandRosenfeld (changes requested):
See my inline comments.

We should think about adding those options to other xymonclient-*.sh scripts, too, at least to the (still used ones: darwin, freebsd, netbsd, openbsd).
But this may be work for a separate PR.

BTW: What happens with this test case on non-linux systems?  Do I get this right, that they still use xymonclient-linux.sh for the test (which may be okay because of the /proc/filesystems and df mockups, but doesn't say anything about the really used xymonclient-<OS>.sh on this system).

bonomani:
Thanks for the thorough review, @RolandRosenfeld. Addressed in two new commits on top (history left intact so your inline comments stay anchored):

**Simplifications**
- `DFTIMEOUT`: dropped the leading-zero loop — now a single `expr "$DFTIMEOUT" + 0` (expr is decimal by spec). I kept the non-numeric/zero rejection and the 3600 cap, because a bad value otherwise makes BusyBox `timeout(1)` reject the duration and silently empty both sections (a false green — the thing this set out to prevent). So not quite "user's problem", but the loop was indeed excess.
- Shortened the long `emit_df` explanation and the `cfg.DIST` df-timeout comment. (Glad you confirmed the timeout→yellow path works.)

**squashfs / iso9660 (your point)**
Went with one mechanism, as you suggested: removed the hardcoded `-x iso9660` and defaulted `XYMONCLIENT_FS_EXCLUDE_TYPES="iso9660 squashfs"`. All type exclusions now flow through one transparent, documented, overridable list. squashfs needed adding precisely because it isn't `nodev`, which is also why the Debian `grep -v squashfs` hack never worked. `XYMONCLIENT_FS_EXCLUDE_TYPES=""` monitors them again.

**tmpfs / zfs default (your point)**
Agreed the silent drop is a bug — fixed in two layers, since the right discriminator differs:
- Client now reports real local filesystems that merely happen to be `nodev`: `XYMONCLIENT_FS_INCLUDE_TYPES` defaults to `"zfs virtiofs tmpfs"` (a zfs-root host previously reported *no* disk data at all).
- The noisy tmpfs mounts are suppressed by *mountpoint* server-side, where Xymon already does it: extended the `CLASS=linux` block in `analysis.cfg` to `IGNORE` `/run` and `/run/lock` (`/run/user/N` was already there). `/dev/shm` is left monitored on purpose. `XYMONCLIENT_FS_INCLUDE_TYPES=""` restores the old behaviour.

Note for the changelog: `analysis.cfg` is a preserved conffile, so existing servers won't auto-get the new `/run` IGNOREs on upgrade — they'd see a few extra (green, near-empty) `/run*` rows until the lines are added. New installs are clean.

**Other points**
- Porting these options to the other `xymonclient-*.sh` (darwin/bsd): agreed, better as a separate PR.
- You're right the test only exercises `xymonclient-linux.sh` regardless of host — inherent to the linux-only scope; best resolved together with the multi-OS port.

**Separately:** while digging into the `DF_LOCAL_ONLY=no` hang risk you hinted at, I built a non-blocking df path (NFS/CIFS that wedge in D-state can't stall the client) — keeping it as a focused follow-up PR rather than loading it in here.

RolandRosenfeld (dismissed):
I can't believe that I'll ever say that: I'd prefer shorter comments. None of them is wrong, but they look AI generated and are somewhat verbose and don't get to the point.
But I can live with this.
I still don't like the 34 lines of DFTIMEOUT sanitizing (I'm not sure, whether this timeout/kill mechanism works at all if we run in a NFS timeout. At least I didn't miss this feature in 25 years of using BigBrother/Hobbit/Xymon. Usually a broken NFS mount should result in xymonclient becoming purple, which is a stronger warning than the yellow, that's generated by df not creating any output).
But except these personal dislikes, I think I can approve the PR now.

bonomani:
Thanks again for the thorough review, @RolandRosenfeld — it materially improved this PR (the tmpfs/zfs default, the squashfs/iso9660 unification, the `:-` fix).

**On your remaining thread (the DFTIMEOUT sanitizing):** I removed the whole `XYMONCLIENT_FS_DF_TIMEOUT` feature rather than trimming it — so that thread is resolved by deletion. The reasoning that pushed me there:

- The timeout never actually solved the case it advertised. A `df` wedged on a **hard** NFS/CIFS mount sits in uninterruptible **D-state**, where SIGKILL is ignored; `timeout -s KILL df` then blocks in `waitpid()` on the unkillable child, so the client hangs anyway and the host goes **purple** (all metrics, not just disk) — the same outcome as no timeout.
- And it *added* a failure that wasn't there before: a slow-but-healthy `df` (many mounts / busy IO) gets killed at 30s → false yellow, and a `df` killed mid-output can silently drop the very mount that's the problem.

So the lines you flagged were guarding a stopgap that didn't earn them. **Kept:** the false-green safety (a `df` that *exits* non-zero with no output is reported yellow, not a silent empty section) and all the filtering work. The docs now state the hard-block → purple outcome plainly.

The *real* fix — a non-blocking background-`df` sentinel that survives D-state (dead mount → one red row, host stays green, auto-recovers) — is a separate follow-up branch; I'll open it as its own PR so it gets focused review.

**On merging:** the history here got iterative (the timeout was added, reworked, then removed within the PR), so it's noisy. I'd suggest a **squash merge** so `main` gets the single net change rather than the round-trip. Heads-up that my df-hang follow-up branch is stacked on this one, so I'll rebase it onto `main` after this lands.

client/xymonclient-linux.sh
bonomani: The df timeout feature this referred to was dropped entirely (commit 54e6dd450, deferring to the non-blocking df sentinel), so the DFTIMEOUT sanitizing block no longer exists. Resolving as obsolete.

RolandRosenfeld (approved):
The comments are still very verbose, but now I'm fine with the PR.

Issue #97, opened 2026-05-21
RFC: introduce a minimal test directory for regression coverage


bonomani:
## Context

The project currently has no automated test infrastructure. Each PR is
validated by hand (build + smoke testing in CI). This has worked for
20+ years but is starting to show wear in the recent burst of
contributor activity:

- PR #76 (parallel make race) was found via a CI failure on PR #90,
  not by any test; the race was timing-dependent and easy to miss.
- Several recent PRs (#92, #94, #96, …) include hand-written "test
  plan" sections that are inherently un-reproducible after the
  contributor has run them locally.
- Distro packagers (Debian, FreeBSD) carry behavior-changing patches
  with no way for upstream to validate that future changes don't
  regress what those patches expect.
- Manual review burden falls on the maintainers for every behavior
  change, no matter how small.

## Proposal

Introduce a minimal `tests/` directory at the repo root with these
properties:

- One file per area (e.g. `tests/client/fs-filter.sh`,
  `tests/parallel-make.sh`).
- Each test is a standalone executable that exits 0 on pass,
  non-zero on fail. No mandatory framework.
- Tests mock external dependencies (e.g. `/proc/filesystems`,
  `/etc/xymon/hosts.cfg`) so they run on any host.
- No required Makefile integration on day 1 — devs run them
  directly during review.
- Optional CI integration: a GitHub Actions workflow that runs
  every `tests/**/*.sh` as part of the build matrix.

## What this is NOT

- Not a proposal to adopt bats, shunit2, pytest, or any specific
  framework.
- Not a proposal to require 100% coverage or block PRs without
  tests.
- Not a proposal to retrofit tests for 20 years of accumulated
  behavior.
- Not a refactor of the existing build system.

The minimum bar is: when a PR introduces a behavior change, there is
a documented way to write a reproducible regression test for it, and
a place in the repo where that test lives.

## Concrete starting point

A starter PR could:

1. Add a `tests/` directory with a `README.md` describing the
   convention (executable, exits 0 = pass, mocks external deps).
2. Include 1-2 tests carried over from recent PRs as worked
   examples — e.g. the filesystem filter scenarios from #96
   already exist as a runnable harness.
3. Add an optional `.github/workflows/tests.yml` that runs every
   `tests/**/*.sh` file.
4. Add a short paragraph to `CONTRIBUTING` (or the equivalent) on
   how to add a test.

## Questions for maintainers

Would such an addition be welcome upstream? If yes:

[...]

PR #101, opened 2026-05-23
tests: establish test infrastructure and initial coverage (RFC #97)

bonomani:
Bootstraps `tests/` — a directory where future PRs can drop reproducible regression tests without relitigating framework choices each time. Implements RFC #97's minimum bar. Refs #97.

## How you run it

@RolandRosenfeld — this section is what shipped for your two comments (no git requirement; standalone local runner):

- **`./tests/testsuite`** — one standalone script, no arguments, works from any cwd (it resolves the repo root from its own location).
- **`make test`** — same runner, as a Makefile target.
- **No git needed, runs from a downloaded tarball.** Nothing under `tests/` calls git, and `tests/` ships in `git archive` tarballs. Verified: extracted a tarball of this branch (no `.git` directory at all) and ran `./tests/testsuite` — 9 passed, 2 skipped, 0 failed (the skips: the staged #96 fs-filter guard, and the xymongrep binary test in a no-build tree).
- **Output adapts automatically**: GitHub log annotations under CI (`$GITHUB_ACTIONS` is detected), plain readable text on a terminal. No flag to remember.
- **CI runs the exact same code**: `.github/workflows/tests.yml` is now just `run: ./tests/testsuite`, so a developer and the workflow can never diverge.

## The contract (tests/README.md)

- Each test is a bash script; exit `0` = pass, `77` = skip (the autotools / autopkgtest convention), anything else = fail.
- Tests find binaries and paths via env vars with sensible defaults, so the same test runs under in-tree make, CMake out-of-source builds, or Debian autopkgtest — no in-test `apt install`, skip-don't-fail when a precondition is absent.
- Discovery is the executable bit on `tests/**/*.sh`; helpers in `tests/lib/` (not `+x`) provide `fail`/`skip`/`pass`, `assert_*`, temp-dir cleanup, and repo-root lookup.

## What's covered — 11 tests shipped

| PR guarded | Test |
|---|---|
| #76/#92 parallel-make race | `buildsystem/parallel-make.sh` |
| #84 configure fails without RRDtool | `buildsystem/configure-no-rrd.sh` |
| #80 RRDtool argv const-ABI | `buildsystem/rrd-api-compat.sh` |
| #90 apache-doc-paths templating (#16) | `buildsystem/docs-templating.sh` |
| #38 trends graph buffer overflow | `web/svcstatus-trends-overflow.sh` |
| #8 md5hash() buffer-size fix | `server/digest-md5hash.sh` |
| #37 xymonnet ALPN support | `network/alpn-support.sh` |
| #33 xymonnet TLSv1.3 option | `network/tls13-support.sh` |
| #48 client dpkg report | `client/dpkg-report.sh` |
| #96 client fs filter (#49) — open PR | `client/fs-filter.sh` (skips until #96 merges) |
| xymongrep tag-selection contract (staging for #94) | `server/xymongrep-filter.sh` — **drives the built binary** |

They cover the five test shapes future contributors will copy from: static source guards, behavioural tests against stubbed commands, compile probes, compile-and-run checks, and **behavioural tests that drive a built binary** (`require_bin`: runs in `build.yml` after the build and under autopkgtest against the installed package; skips in the no-build `tests.yml`).

Follow-ups (test to come, they need this PR's runner first): #94 xymongrep `--no-dialup` (extends `xymongrep-filter.sh`), #98 xymoncgimsg `$XYMSRV`. Out of scope for shell tests, with reasons documented so nobody burns cycles retrofitting: PCRE2 hardening (needs valgrind), RRD utimes / SNMP MD5 (kernel/build deps), OpenBSD/NetBSD fixes (platform-specific), `$CARESLIB` wiring (build-coupled), trivial/docs/superseded PRs.

## Notes

- **CMake-ready**: tests don't compile anything of the project and don't hook into `configure`; a future `tests/CMakeLists.txt` can register the same files via `add_test()` + `SKIP_RETURN_CODE 77` with zero test edits.
- **autopkgtest** (@RolandRosenfeld): the suite now has both halves your use case needs. `server/xymongrep-filter.sh` executes the **installed binary** (`XYMONGREP=/usr/lib/xymon/client/bin/xymongrep`), so a library transition (libpcre2, OpenSSL, …) has something that can actually break at runtime; `client/{fs-filter,dpkg-report}.sh` accept `XYMONCLIENT_LINUX=/usr/lib/xymon/client/bin/xymonclient-linux.sh` and then exercise the **installed, Debian-patched client script** instead of the source copy. `tests/README.md` § *Downstream consumers* now carries a concrete `debian/tests/control` sketch (`Restrictions: skippable` maps the runner's all-skip 77; `gcc`/`make` in `Depends` keep the compile probes alive). Source-reading tests still run under autopkgtest — against the unpacked patched source — but are supplementary signal there by design.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,19 +87,13 @@ jobs:
       # there. Run it here too, where the binaries this variant produced are on
       # disk, so those tests actually execute against the build.
       #
-      # SCAFFOLDING -- as of today NO test calls require_bin: every test reads
-      # committed source or compiles its own probe, so none consumes a build
-      # artefact. That makes this step currently REDUNDANT with tests.yml (same
-      # checks, same results, after an unnecessary rebuild). It is kept on
-      # purpose: the moment a behavioural test drives a compiled tool, this is
-      # the only job where that test runs instead of skipping. Drop this step
-      # together with require_bin if the binary-test direction is abandoned.
+      # First such test: tests/server/xymongrep-filter.sh, which drives the
+      # common/xymongrep this leg just built. This step is the only place in
+      # CI where it executes instead of skipping.
       #
-      # Gated to the server leg only: until a require_bin test exists, running
-      # the identical suite in all three matrix legs triples the redundancy for
-      # nothing. The server build produces the broadest binary set, so it stays
-      # the scaffolding leg; widen the gate (or drop it) when the first test
-      # needs a binary this leg doesn't build.
+      # Gated to the server leg only: it produces the broadest binary set
+      # (including common/xymongrep). Widen the gate when a test needs a
+      # binary this leg doesn't build.
       - name: Run regression tests against the build
         if: matrix.variant == 'server'
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,10 +4,19 @@ on:
   push:
   pull_request:
 
+# Both workflows check out and run pull-request code; the token needs
+# nothing beyond reading the repo. Same grant as tests.yml.
+permissions:
+  contents: read
+
 jobs:
   build:
     name: build (${{ matrix.name }})
     runs-on: ubuntu-latest
+    # A regression test that blocks (e.g. a configure probe growing an
+    # interactive prompt) would otherwise hang the job for the runner's
+    # 6-hour ceiling. Build + suite take ~1 minute; 15 is generous.
+    timeout-minutes: 15
     strategy:
       fail-fast: false
       matrix:
@@ -72,3 +81,39 @@ jobs:
         run: |
           echo "running make with $(nproc) parallel processes"
           make -j$(nproc)
+
+      # tests.yml runs the same suite without building, so any test that needs
+      # a compiled binary (require_bin in tests/lib/assert.sh) silently skips
+      # there. Run it here too, where the binaries this variant produced are on
+      # disk, so those tests actually execute against the build.
+      #
+      # SCAFFOLDING -- as of today NO test calls require_bin: every test reads
+      # committed source or compiles its own probe, so none consumes a build
+      # artefact. That makes this step currently REDUNDANT with tests.yml (same
+      # checks, same results, after an unnecessary rebuild). It is kept on
+      # purpose: the moment a behavioural test drives a compiled tool, this is
+      # the only job where that test runs instead of skipping. Drop this step
+      # together with require_bin if the binary-test direction is abandoned.
+      #
+      # Gated to the server leg only: until a require_bin test exists, running
+      # the identical suite in all three matrix legs triples the redundancy for
+      # nothing. The server build produces the broadest binary set, so it stays
+      # the scaffolding leg; widen the gate (or drop it) when the first test
+      # needs a binary this leg doesn't build.
+      - name: Run regression tests against the build
+        if: matrix.variant == 'server'
+        run: |
+          # Mirror tests.yml: this step checks out and builds full source, so
+          # every test finds the code it guards. The runner's all-skip exit 77
+          # ("nothing verified") can therefore only mean broken discovery, not a
+          # legitimate skip -- map it to a hard failure. Actions has no neutral
+          # state, so 77 would otherwise surface as a bare nonzero failure with
+          # no explanation. `|| rc=$?` exempts the runner from bash -e so the
+          # mapping is reachable.
+          rc=0
+          ./tests/testsuite || rc=$?
+          if [ "$rc" = 77 ]; then
+            echo "::error::testsuite skipped everything (rc=77) -- with full source built this means broken discovery, not a legitimate skip"
+            exit 1
+          fi
+          exit "$rc"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,44 @@
+name: tests
+
+on:
+  push:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  shell:
+    name: shell regression tests
+    runs-on: ubuntu-latest
+    # A regression test that blocks (e.g. a configure probe growing an
+    # interactive prompt) would otherwise hang the job for the runner's
+    # 6-hour ceiling. The suite takes seconds; 10 is generous.
+    timeout-minutes: 10
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      # The runner lives in tests/testsuite so developers can run the exact
+      # same thing locally (./tests/testsuite or make test). It auto-detects
+      # GitHub Actions and emits log annotations here, plain output on a CLI.
+      #
+      # testsuite exits 77 ("everything skipped -- nothing verified"). Actions
+      # has no neutral state: any nonzero exit is a failure, so we must map 77
+      # explicitly rather than let it surface as a bare "exit code 77". This job
+      # checks out the full source, so every test finds the code it guards and
+      # runs -- an all-skip here is never the legitimate "tests shipped without
+      # source" case, it means discovery or the tree itself is broken. Treat it
+      # as a hard failure with a clear message.
+      - name: Run testsuite
+        run: |
+          # GitHub runs steps with `bash -e`, which would abort on testsuite's
+          # nonzero exit before we can read it. `|| rc=$?` exempts the command
+          # from -e so the 77 mapping below is actually reachable.
+          rc=0
+          ./tests/testsuite || rc=$?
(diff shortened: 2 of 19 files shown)


RolandRosenfeld:
Just a short comment (I'm in a hurry): I don't like the requirement of git and the git tree here.
Wouldn't it be better to be able to (re-)run the tests after downloading the tarball, too?

bonomani:
Good catch, thanks. Fixed in 9d187aa0.

Audit of where git was actually depended on:

- `find_root` (in `assert.sh`) — already uses `BASH_SOURCE`, not `git rev-parse`. Tarball-safe.
- `parallel-make.sh` — reads Makefiles directly. Tarball-safe.
- `fs-filter.sh` — reads `xymonclient-linux.sh`, runs an extracted snippet with `df`/`readlink` stubbed. Tarball-safe.
- `configure-no-rrd.sh` — was using `git worktree add --detach` to isolate. **This is the one that broke from a tarball.**

The fix replaces the worktree isolation with a selective `cp -r` of the paths `configure --server` actually reads (`configure`, `configure.server`, `build/`, plus `configure.client` for the `--help` dispatch). Stub `build/rrd.sh` overwrites it in the copy, `./configure --server` runs from there, exit code is asserted. The `command -v git || skip` guard is gone — the test now runs in any environment that has bash + cp + a C toolchain. Verified passing with `git` symlinked to `/bin/false` to simulate a tarball-only environment.

Also added a `No git assumption` bullet to `tests/README.md` so the constraint is explicit and the next contributor doesn't reintroduce it (no `git worktree`, no `git stash`, etc.).

Let me know if you want me to also draft the autopkgtest control file (`debian/tests/control` + a `debian/tests/upstream-shell` runner) that drives `tests/**/*.sh` in the format autopkgtest expects — happy to do it once you confirm what the Debian packaging side prefers.

RolandRosenfeld:
I'd like to be able to run the testsuite locally, so it would be a good idea to split out the bash script from .github/workflows/tests.yml into tests/testsuite (or the like), so it can be run by the developer.
A little switch, that modifies output for github workflows vs. direct cli usage would be great.
This tool also could be integrated into Makefile as target "test" (or the like).

bonomani:
Done in 36ff1b3 — the runner is now standalone:

- **`tests/testsuite`** — discovers the executable `tests/**/*.sh`, runs each (0=pass, 77=skip, else=fail), prints a summary. It auto-detects `GITHUB_ACTIONS`: log annotations under CI, plain output on a terminal — no flag to set. No `.sh` extension, so it never discovers itself; it resolves the repo root from its own location, so it runs from any cwd.
- **`.github/workflows/tests.yml`** is now just `run: ./tests/testsuite`, so CI and a developer run the exact same code path.
- **`make test`** invokes the same runner.

So locally it's `./tests/testsuite` (no build needed) or `make test`.

RolandRosenfeld:
I like the test framework very much, but at the moment I'm not sure, whether I can use the existing tests for Debian autopkgtest testing.  

The problem is, that autopkgtest is used to find regressions with changes in the build system or with external libraries or the like.  For this we need tests, that check, whether the built xymon server and client from the Debian package work as expected.
If I read current test scripts correct, most of them only check the source code (or compile some stub on test time) but none of them tests the behavior of the build binaries (that's what autopkgtest is used for).

So at the moment this test suite is only usable for finding problems in the source, but not for finding regressions in the built packages.

bonomani:
> [!NOTE]
> Posted while the PR sat in draft so the direction could be sanity-checked. The direction was confirmed and the PR is now **ready for review** — see the follow-up comment below for the override-contract fix that closed the remaining hole.

You're right, and the distinction matters: everything in the suite so far was a *source-tree guard* (useful for upstream CI, useless for testing what a package ships). The framework was built so installed-artifact tests slot in without changes — but until now that was a claim with zero examples. Two commits address that:

**The suite now drives a built binary.** `tests/server/xymongrep-filter.sh` executes the real `xymongrep` (via `--hosts=` against a fixture `hosts.cfg`, so it's hermetic — no server, no network) and asserts the documented tag-selection contract: exact match, `*` prefix for URL tags, only matched tags echoed, the `dialup` flag, `--noextras`. Under autopkgtest you'd export `XYMONGREP=/usr/lib/xymon/client/bin/xymongrep` and it tests the **installed package** — which is what gives a library transition (libpcre2, OpenSSL, …) something that can actually break at runtime. In upstream CI it runs in `build.yml` right after the build; in the no-build `tests.yml` it skips with 77.

**The client tests now accept the installed script.** `fs-filter.sh` and `dpkg-report.sh` read `$XYMONCLIENT_LINUX`; pointed at `/usr/lib/xymon/client/bin/xymonclient-linux.sh` they exercise the script Debian actually ships — *including your patches* (e.g. `09_hobbitclient-debian`, `24_hobbitclient-tmpfs`). If a Debian patch diverges from an invariant upstream guards, the test says so — which is the "validate packagers' behavior-changing patches" gap from #97, working in both directions.

`tests/README.md` § *Downstream consumers* now has the concrete mapping, roughly:

```
Test-Command: XYMONGREP=/usr/lib/xymon/client/bin/xymongrep \
              XYMONCLIENT_LINUX=/usr/lib/xymon/client/bin/xymonclient-linux.sh \
              ./tests/testsuite
Depends: xymon, xymon-client, gcc, make
Restrictions: skippable
```

(`skippable` maps the runner's all-skip exit 77 to SKIP; `gcc`/`make` keep the compile-probe tests alive since autopkgtest doesn't provide build-deps.)

One more angle worth naming: your patch series and this suite already overlap almost 1:1 — `100_md5_bufferoverflow` ↔ #8 ↔ `digest-md5hash.sh`, `101_rrdtool1.9` ↔ #80 ↔ `rrd-api-compat.sh`, `03_doc-paths` ↔ #90 ↔ `docs-templating.sh`, `95_fix_tls13_banner` ↔ #33 ↔ `tls13-support.sh`, and `24_hobbitclient-tmpfs` is what #96's env-var filter is designed to replace. So when you package the next release and start dropping patches, these tests are the regression harness for exactly that migration.

Fully agreed the suite still needs binary tests for the server side proper (an `xymond` message round-trip is the obvious next one) — that's follow-up material per test shape now that the first `require_bin` example exists to copy from. ~~I've moved the PR to draft while this settles; if the `xymongrep` test plus the env-var overrides match what autopkgtest needs, I'll mark it ready.~~ *(Done — marked ready for review after closing the dangling-override hole, see below.)*

bonomani:
Follow-up: re-checking the override contract against the autopkgtest use case turned up one reliability hole, now closed in cabe77c8.

`require_bin` and the `$XYMONCLIENT_LINUX` guards treated *every* missing path as a skip — including paths the caller explicitly exported. Under autopkgtest that false-greens the exact scenario the suite exists to catch: `debian/tests/control` exports the installed locations, the package stops shipping a file there, the three override-consuming tests exit 77, the source guards still pass, and the runner exits 0. Reproduced before the fix: **8 passed / 3 skipped / exit 0** with both overrides pointing at nonexistent paths.

The contract is now: an explicit override is an *existence assertion*. A dangling exported path **fails** (a broken build or package layout is precisely what the exporting caller runs the suite to catch); only the unset in-tree default still skips (binary not built in this configuration, plus the #96 staging skip in fs-filter.sh). Verified both ways: dangling overrides → 3 fail / exit 1; no overrides → unchanged 9 pass / 2 skip / exit 0; a valid `$XYMONCLIENT_LINUX` still passes. `tests/README.md` § *Path discovery* documents the rule, so the control-file sketch above can be used as-is without that footgun.

With that closed I'm marking the PR ready for review. Scope recap: the built-`xymongrep` test plus the installed-script overrides are the autopkgtest entry points; the `xymond` message round-trip test stays acknowledged follow-up material.


---

**Update (13 Jun):**

- The `debian/tests/control` sketch above now has a concrete landing spot: **#160** refreshes `debian/` from the current Debian package, and it does not (yet) include `debian/tests/` — wiring this suite in there is the natural follow-up once both PRs land.
- Re-verified today at `cabe77c8`: clean run **9 passed / 2 skipped / exit 0**; with both overrides dangling, **3 failed / exit 1** (exactly the three override-consuming tests: `xymongrep-filter.sh`, `fs-filter.sh`, `dpkg-report.sh`). The branch still auto-merges clean against current `main` (now including the #144 OpenSSL fix).

RolandRosenfeld (approved):
I tend to approve this PR, since this is a very good starting point and its existence heavily improves code quality by providing a framework for testing.
There may be edges, that can be improved or optimized, but that can be done later in future PRs.
So I didn't review every single test case included in detail, but the base test infrastructure.

PR #102, opened 2026-05-23
Pre-tag release prep workflow (refs #99)

### LINKED ISSUE #99: Release workflow: pre-tag mechanism needed
We need some mechanism (maybe via workflow-dispatch), that is run **before** the tagging of a release version (so it cannot be part of .github/workflow-release.yaml).

It seems, that build/dorelease.sh is intended for this (it updates build/md5.dat, updates the xymon version in some man pages (build/updmanver) and generates HTML versions of the man pages (build/makehtml.sh), so we can use this a start.

bonomani:
Closes #99 — implements option A from the [release-prep discussion](https://github.com/xymon-monitoring/xymon/issues/99#issuecomment-4524892824): release files are generated on a branch and reviewed as a PR, instead of being committed straight to `main`.

## How a release works now

```
 maintainer                     maintainer                      automatic
 runs "Pre-tag release prep"    merges PR, pushes tag           on tag push
 with version=X.Y.Z             rel-X.Y.Z
        │                              │                              │
        ▼                              ▼                              ▼
 release/X.Y.Z branch  ──PR──►  main with prep commit  ──tag──►  draft GitHub release
```

1. **Prep (automated).** Run the *Pre-tag release prep* workflow with `version=X.Y.Z`. It recreates `release/X.Y.Z` from current `main`, regenerates the release files (`md5.dat`, version/date stamps in manpages and their HTML), opens a PR, and dispatches build checks onto it. The run summary prints the next-step commands.
2. **Review + tag (human).** Review the generated diff, merge, push the `rel-X.Y.Z` tag. The tag is the deliberate approval gate — nothing is released without it.
3. **Publish (automated).** The tag triggers the tarball workflow: `git archive` piped through `gzip -n -9`, uploaded with a `.sha256` checksum to a **draft** release.

Because every timestamp comes from the tag (file mtimes) and `SOURCE_DATE_EPOCH` (manpage dates), the same tag always reproduces a byte-identical tarball — anyone can verify the published checksum.

Full maintainer documentation (steps, invariants, troubleshooting) lives in [`.github/RELEASING.md`](https://github.com/bonomani/xymon/blob/feat/prep-release-workflow/.github/RELEASING.md), and `build/dryrelease.sh` gives a one-command local dry-run of the whole chain in a throwaway worktree.

## What's in the diff

**New workflow.** `.github/workflows/workflow-prep-release.yaml`, plus `workflow_dispatch` on `build.yml` so checks can run on the generated branch.

**Prep scripts made deterministic and fail-fast.** `build/dorelease.sh`, `generate-md5.sh`, `updmanver`, and `makehtml.sh` now stop on errors, honor `SOURCE_DATE_EPOCH`, pin `LC_ALL=C` (so sort order and dates don't depend on the caller's locale), and use GNU/BSD-portable date handling. This matters because the whole chain used to mask failures: `dorelease.sh` ended in an unconditional `exit 0`, `makehtml.sh` reported the pipeline tail's status rather than `man2html`'s, `updmanver` could break mid-loop and still exit 0, and a failed digest wrote a malformed `md5.dat` line — so the prep workflow could have committed and opened a release PR from a partial tree. One deliberate non-change: `generate-md5.sh` still folds in the old `md5.dat` — `setup-newfiles` needs the hashes of every previously shipped version to recognize unmodified files during upgrades, so that accumulation is load-bearing, not cruft.

**Aligned with #302.** `dorelease.sh` passes the version to `makehtml.sh` as `--version` — under #302's interface a bare argument is always a page path — and the prep job installs `python3` alongside mandoc, because `makehtml.sh` pipes every page through `build/manpage-html.py` and the release path should not depend on the hosted image happening to ship an interpreter.

**Reproducible tarballs.** `gzip -n -9` (no embedded filename/mtime in the gzip header) and a published `.sha256` alongside the tarball.

**Docs.** `.github/RELEASING.md` (kept out of release tarballs by the existing `.git*` export-ignore) and `build/dryrelease.sh`.

**Guardrails.** An existing `rel-X.Y.Z` tag is rejected; a prep run that produces no diff fails loudly; reruns reuse the *open* prep PR (a closed-unmerged one gets a fresh PR); the run summary's tag commands pin the PR's merge commit, not `origin/main`, so later commits can't be scooped up. The prep job is pinned to `ubuntu-24.04` because `man2html` output is version-sensitive.

## Validation

- PR build checks and `actionlint` are clean; `git diff --check` is clean.
- `generate-md5.sh` output verified byte-identical to the old script; the digest-failure abort exercised with an unreadable file.
- `dorelease.sh 9.9.9` runs clean end-to-end in a scratch worktree (139 files stamped) and is idempotent — a second run produces an identical tree.
- Tarball reproducibility verified locally: repeated runs are byte-identical, and archiving the same commit via SHA, lightweight tag, or annotated tag gives the same bytes.
- The workflow itself can only be end-to-end tested after merge (`workflow_dispatch` needs the workflow on the default branch): the first real prep run is the live test, fixed forward if needed.

## Scope

Targets `main` only for now; maintenance branches can be added later with a branch input.


SoundGoof:
I checked the deployed man pages on xymon.com.
The index links to 69 HTML man pages.
All 69 pages return HTTP 200 with an empty body.

How are the HTML man pages deployed to xymon.com, and which source
version is used?

I could not find that deployment step in this repository.

bonomani:
The HTML manpages are deployed through the normal Xymon installation, not by a separate CI deployment step. `make install` copies `docs/manpages` to `$INSTALLWWWDIR/help/manpages/`. Since xymon.com reports that it runs Xymon 4.3.30, it should normally be serving the HTML files included with that installed release.

The normal build CI only configures, builds, and runs regression tests. It does not run `make install`, deploy files to xymon.com, or check that the generated HTML files contain content.

The new pre-tag release-preparation workflow regenerates all HTML manpages. It runs `build/dorelease.sh`, which calls `build/makehtml.sh`, and commits the generated files to a release PR for review. It then runs the normal build workflow on the release branch. This prepares the release tree and runs the build checks, but it does not deploy anything to xymon.com.

This is useful because it catches HTML manpages that were not regenerated after their source manpages were changed. However, the current generation process has some problems. Debian’s `man2html` includes patches that are not present in the original upstream version, so different installations may produce different HTML. The output may also change between package versions, and `man2html` itself is obsolete.

Pinning the release-preparation workflow to one environment gives us consistent output for now, but this should be considered a temporary solution. In the longer term, we should replace `man2html` with a maintained converter and define one reproducible way to generate the HTML manpages.

I usually regenerate and commit the HTML when changing a manpage in a PR, although I may occasionally miss one. Release preparation regenerates all HTML manpages before a release. However, regeneration alone may not detect a `man2html` command that exits successfully while producing empty output. We should therefore also add an explicit CI check that every generated HTML manpage is non-empty.

To regenerate all HTML manpages manually on Debian or Ubuntu:

```sh
sudo apt-get install man2html-base
./build/makehtml.sh VERSION
```

For the complete release preparation:

```sh
./build/dorelease.sh VERSION
```

The 69 empty pages currently served by xymon.com therefore appear to be an installation or server-side deployment problem, or a problem with the HTML files included in the installed 4.3.30 release. The repository CI does not currently deploy to or inspect the live xymon.com installation.

SoundGoof:
PR #297 updates the existing workflows to actions/checkout@v7. This PR adds workflow-prep-release.yaml with actions/checkout@v6, so please update it to v7 as well to avoid reintroducing the older version when this is merged.

bonomani:
@SoundGoof already done — both `build.yml` and `workflow-prep-release.yaml` use `actions/checkout@v7` now (pushed after your comment), and I've just fixed a stale `checkout@v6` mention in a comment for consistency. Thanks for catching it.

PR #108, opened 2026-05-27
web: default HTMLCONTENTTYPE to UTF-8 charset (fixes #34)

### LINKED ISSUE #34: CGI or templates may need changes for UTF-8
If you have any UTF-8 characters in your statuses it seems to render incorrectly because we aren't setting the HTML meta or content-type to include UTF-8. Has anyone else noticed this?

bonomani:
Resolves #34.

## Problem
The CGIs emit the HTTP `Content-Type` header directly from `HTMLCONTENTTYPE`, which defaulted to `text/html` — **no charset**. Non-ASCII status content (accented filenames, log lines, banners) then renders as mojibake unless the web server adds a default charset itself: Debian/Ubuntu's Apache does (`AddDefaultCharset utf-8`), FreeBSD's does not — which is why the breakage looked FreeBSD-specific when it's actually universal.

The bytes are already fine — clients run `LANG=C` (so tool framing is ASCII) and `htmlquoted()` passes high bytes through raw, so the output is valid UTF-8 on modern hosts. It was just never *labelled* UTF-8.

## Fix
Default `HTMLCONTENTTYPE` to `text/html; charset=UTF-8` (`lib/environ.c` + `xymonserver.cfg.DIST`). Xymon now declares the charset in the CGI HTTP header (and the templates' `<META … &HTMLCONTENTTYPE>`), correct regardless of web-server config. This is a label, **not** a transcode — nothing is converted.

`HTMLCONTENTTYPE` stays configurable, so sites whose hosts emit a different encoding (iso-8859-1, euc-jp, …) can override it.

## Compatibility
Declaring an explicit charset removes the browser's auto-detection. A site whose hosts emit non-UTF-8 bytes and relied on browser guessing should now set `HTMLCONTENTTYPE` explicitly — noted in RELEASENOTES.

### DIFF
--- a/RELEASENOTES
+++ b/RELEASENOTES
@@ -19,6 +19,13 @@ including problems with hostnames with dashes in them.
 Combostatus tests propagated up from other combostatus tests should now
 display properly.
 
+The default web charset is now UTF-8: HTMLCONTENTTYPE defaults to
+"text/html; charset=UTF-8" so non-ASCII status content renders correctly
+without relying on the web server to supply a default charset (which Debian's
+Apache does but FreeBSD's does not). Sites whose hosts emit a different encoding
+(e.g. iso-8859-1, euc-jp) and relied on browser auto-detection should set
+HTMLCONTENTTYPE explicitly in xymonserver.cfg.
+
 
 Changes for 4.3.29
 ==================
--- a/lib/environ.c
+++ b/lib/environ.c
@@ -148,7 +148,7 @@ const static struct {
 	{ "XYMONPAGEHTACCESS", "" },
 	{ "XYMONSUBPAGEHTACCESS", "" },
 	{ "XYMONNETSVCS", "smtp telnet ftp pop pop3 pop-3 ssh imap ssh1 ssh2 imap2 imap3 imap4 pop2 pop-2 nntp" },
-	{ "HTMLCONTENTTYPE", "text/html" },
+	{ "HTMLCONTENTTYPE", "text/html; charset=UTF-8" },
 	{ "HOLIDAYFORMAT", "%d/%m" },
 	{ "WEEKSTART", "1" },
 	{ "XYMONBODYCSS", "$XYMONSKIN/xymonbody.css" },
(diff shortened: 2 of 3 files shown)


RolandRosenfeld:
I don't notice problem with or without the charset header on my personal server because I usually work with html-entities (&auml; or the like), to avoid issues with an utf-8 capable editor and apache/xymon using ISO-8859-1 by default.

Someone who currently uses iso-8859-1 chars in his hosts.cfg or in some test output may run into issues (and has to either modify HTMLCONTENTTYPE in xymonserver.cfg or convert his setup).

I vote for changing this to UTF-8 (as in https://github.com/xymon-monitoring/xymon/pull/108), but I'm not sure, who is crying after the change. Mentioning it in the release notes is very important here...

Maybe we should look for more votes pro/con the change.

bonomani:
Thanks @RolandRosenfeld, I agree with your analysis.

UTF-8 is the right default to adopt: it matches the current web standard and is where Xymon should go.

**Scope of this change.** It only touches the `HTMLCONTENTTYPE` default (`lib/environ.c`) — the encoding the web UI actually renders. Today the default is `text/html` with no charset, which leaves the browser to guess; this makes it explicitly UTF-8. The XML feeds (`appfeed.cgi`, the RSS feed, `perfdata.cgi?FORMAT=XML`, and the `xymondxboard`/`xymondxlog` protocol) still hardcode `encoding='ISO-8859-1'`, but those are opt-in/legacy paths that aren't enabled by default, so they're orthogonal here and left untouched — they can be revisited separately if anyone still relies on them.

**Behavior change.** This is visible for installations that still emit raw ISO-8859-1/Windows-1252/euc-jp bytes, so I moved it to the 4.4 milestone rather than treating it as a 4.3 maintenance fix. As @RolandRosenfeld notes, mentioning it in the release notes is essential. An affected site can revert to the current non-explicit behavior (no charset, browser guesses):

```sh
HTMLCONTENTTYPE="text/html"
```

or declare its encoding properly, e.g.:

```sh
HTMLCONTENTTYPE="text/html; charset=ISO-8859-1"
```

For 4.3 we can document the UTF-8 workaround for anyone who wants it early:

```sh
HTMLCONTENTTYPE="text/html; charset=UTF-8"
```

@feld, since you opened #34, does targeting this for 4.4 sound like the right release boundary to you? And as Roland suggested, it'd be good to gather a few more pro/con votes on the change.

StefCoene (approved):

Issue #103, opened 2026-05-23
Investigate all FreeBSD patches


bonomani:
Mirror of #28, but for the FreeBSD ports tree. Patches:
- server: https://github.com/freebsd/freebsd-ports/tree/main/net-mgmt/xymon-server/files
- client: https://github.com/freebsd/freebsd-ports/tree/main/net-mgmt/xymon-client/files

Related FreeBSD source: **#22** tracks the FreeBSD 4.4-alpha patches reported on the
mailing list — a different provenance than this ports-tree list, with partial
overlap. Its items should be reconciled into the checklist below so the FreeBSD
work has a single index.

Let's try to track the status of the FreeBSD port patches here:

Server (net-mgmt/xymon-server):

- [x] patch-md5_buffer_overflow -> applied in #8 — **drop**
- [x] patch-nousemultipart -> applied in #6 — **drop**
- [x] patch-xymond_rrd_do__vmstat.c -> applied in #39 — **drop**
- [x] patch-fix_tls13_banner -> applied in #7 — **drop**
- [x] patch-xymonnet_Makefile -> see #42 — **keep** for now (fping is the best current pinger)
- [x] patch-bbcombotest-fix -> see #109 — **keep** until it merges
- [x] patch-upgrade430.sh -> dead 4.2.x -> 4.3.0 migration script, not wired into build — **drop**
- [x] patch-xymond-Makefile -> FreeBSD packaging policy — **keep**

Client (net-mgmt/xymon-client):

- [x] patch-client_xymonclient-freebsd.sh -> see #41 — **keep** until it merges
- [x] patch-client_Makefile -> FreeBSD packaging policy — **keep**

Issue #28, opened 2026-01-21
Investigate all Debian patches


feld:
They are all available here: https://salsa.debian.org/debian/xymon/-/tree/master/debian/patches?ref_type=heads

A couple are already covered by other issues/PRs, but we should try to bring everything we can upstream

RolandRosenfeld:
Let's try to track the status of the Debian patches here:

_Action tag at the end of each line = recommended disposition: **drop** (in upstream or obsolete → remove from Debian series) · **keep** (Debian-specific → stays downstream) · **upstream** (candidate to send upstream). The `[ ]/[x]` checkbox is the maintainer's — tick once actioned on the Debian side._

- [x] 03_doc-paths.patch -> replaced by https://github.com/xymon-monitoring/xymon/pull/90  — **drop**
- [x] 09_hobbitclient-debian.patch -> applied in https://github.com/xymon-monitoring/xymon/pull/48  — **drop**
- [ ] 12_hobbitvars.patch  — **keep** (Debian variable naming)
  - This is an old patch, containing multiple changes, which should be split into multiple patches. After cleanup for 4.3.31 the following changes are still needed (and should be split into separated patches):
    - [ ] Setting XYMONSERVERHOSTNAME, XYMONSERVERIP via /etc/default/xymon-client
    - [ ] Set XYMONTMP to /var/lib/xymon/tmp
    - [ ] Place PIDFILE in /var/run/xymon/xymon.pid  (may be replaced by #219)
    - [ ] include xymonclient.d in xymonserver.cfg (because in Debian server depends on client, but runs client scripts)
    - [ ] Using gifs/static instead of gifs (non blinking theme)
    - [ ] apache.conf: `Require local` instead of `Require all granted`.
- [x] 21_FHS-instead-FSSTND-in-example-in-man-page.patch -> applied in https://github.com/xymon-monitoring/xymon/pull/95  — **drop**
- [x] 24_hobbitclient-tmpfs.patch -> applied in https://github.com/xymon-monitoring/xymon/pull/96 (fixes https://github.com/xymon-monitoring/xymon/issues/49)  — **drop**
- [x] 27_hobbit_files_ifexist.patch -> see https://github.com/xymon-monitoring/xymon/pull/161  — **upstream** (in progress)
- [ ] 30_prefer-packaged-temp-plugin-over-unpackaged-devmon.patch  — **keep** (Debian packaging policy)
[...]

PR #109, opened 2026-05-27
combostatus: no spurious error on an empty xymond board (proper bbcombotest-fix)

bonomani:
## Problem
`getxymondvalue()` in `xymond/combostatus.c` reports **"Could not access xymond board, error 0"** whenever the board comes back `NULL` even though the send **succeeded** (`error 0` = `XYMONSEND_OK`) — e.g. when the board is legitimately empty. The error guard

```c
if ((xymondresult != XYMONSEND_OK) || (board == NULL)) { ... }
```

conflates two different things: a real send failure, and an empty/NULL board.

## Why not just delete `|| (board == NULL)`
That's what Debian's long-standing `42_bbcombotest-fix.patch` (Christoph Berg, 2015, `Forwarded: no`) does — and the FreeBSD port carries the same patch. But that branch also did an early `return COL_CLEAR`, so simply deleting the check means a `NULL` board with `result == OK` now falls through to the later `strncmp(board, …)` and **dereferences NULL**. It fixes the cosmetic message at the cost of a latent crash.

## Fix
Split the two concerns:

```c
if (xymondresult != XYMONSEND_OK) {              /* real failure */
    board = ""; *errptr += sprintf(...); return COL_CLEAR;
}
if (board == NULL) board = "";                   /* empty board: no error, no deref */
```

- No spurious "error 0" when the send succeeded but the board is empty.
- Still NULL-safe — `board` is never dereferenced as NULL by the `strncmp()` below.

Strictly better than both the original (noisy) and the Debian patch (deref risk). Builds clean (server variant, 0 errors).

Refs #28 (Debian patch audit) and #103 (FreeBSD carries the same patch) — this is the proper upstream version of `bbcombotest-fix`.

### DIFF
--- a/xymond/combostatus.c
+++ b/xymond/combostatus.c
@@ -207,11 +207,12 @@ static int getxymondvalue(char *hostname, char *testname, char **errptr)
 		xymondresult = sendmessage("xymondboard fields=hostname,testname,color", NULL, XYMON_TIMEOUT, sres);
 		board = getsendreturnstr(sres, 1);
 
-		if ((xymondresult != XYMONSEND_OK) || (board == NULL)) {
+		if (xymondresult != XYMONSEND_OK) {
 			board = "";
 			*errptr += sprintf(*errptr, "Could not access xymond board, error %d\n", xymondresult);
 			return COL_CLEAR;
 		}
+		if (board == NULL) board = "";	/* Empty board is not a failure; treat as no match and avoid the NULL-deref below. */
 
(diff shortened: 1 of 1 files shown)


RolandRosenfeld (commented):
Sounds reasonable to me, but since I never used combostatus myself, I'm not a good reviewer here...

feld (approved):
I have seen this error before and we know the Debian patch fixed it. The explanation for the improved fix seems sound, so let's proceed.

PR #111, opened 2026-05-28
build: drop the make-era platform targets nobody builds (#85)

### LINKED ISSUE #85: Drop obsolete legacy make-era platform targets
The repository still carries legacy support for a few obsolete platform targets that no longer look realistically maintainable as active make-era targets:

- IRIX
- OSF1 / Tru64
- SCO_SV
- GNU_kFreeBSD
- the OSX alias makefile

I would like to propose removing these targets from upstream support scope and cleaning up the remaining legacy files that still reference them.

Files to delete
- `build/Makefile.IRIX`
- `build/Makefile.SCO_SV`
- `build/Makefile.GNU_kFreeBSD`
- `build/Makefile.OSX`
- `client/xymonclient-irix.sh`
- `client/xymonclient-osf1.sh`
- `client/xymonclient-sco_sv.sh`
- `client/xymonclient-unixware.sh`
- `xymond/client/irix.c`
- `xymond/client/sco_sv.c`

Files to update
- `docs/TODO`
- `docs/known-issues.html`
- `configure.server`
- `configure.client`
- `lib/misc.h`
- `lib/misc.c`
- `client/logfetch.c`
- `xymond/rrd/do_vmstat.c`
- `xymond/rrd/do_ifstat.c`
- `xymond/rrd/do_netstat.c`
- `xymond/rrd/do_disk.c`

The intent is not to change the status of still-recognizable legacy Unix targets such as AIX, Solaris, or HP-UX in this issue.

The goal here is only to remove clearly obsolete platform branches and aliases that are now mostly dead compatibility baggage rather than supportable targets.

bonomani:
Issue #85: the tree still carried build targets and client code for platforms nobody builds — IRIX, OSF1, SCO, GNU/kFreeBSD, OS X's old `Makefile.OSX`. They are removed, along with the per-platform files that only those targets used.

One of them was not obsolete. `do_disk.c` routed a df listing **with a filesystem-type column** through a branch called `DT_IRIX`, and the platform cleanup read the name as a platform. The shape is not IRIX's: `df -PT` produces it, and the branch is chosen from the message text, not from the host's declared OS. Without it the report goes to the plain layout, which reads the capacity column as the mount point.

| | |
|---|---|
| `df -PT`, ext4 | `disk5%.rrd` — one RRD per distinct percentage, pct 9728 stored as NaN against `DS:pct:GAUGE:600:0:100` |
| after | `disk,root.rrd`, pct 5 |

The branch is selected from the header df prints (`Filesystem Type …`), not from a list of type names. The list it had — xfs, efs, cxfs — came from the IRIX df it was written for, so a Linux `df -PT` fell straight through: the arm existed but never fired for the case that justified keeping it.

`tests/rrd/disk-typecolumn-layout.sh` drives the real worker and asserts both halves — the file name *and* the stored percentage, since either alone passes on a build that got the other right. It uses ext4, because the old selector passed the xfs one. The percentage is read through librrd, not the rrdtool command line: the CLI is a separate package, and gating on it does not fail where it is absent — it quietly checks less. `test-suite-portability.sh` has the matching rule now, so the next test to reach for it is told where to look. `xymonserver.cfg.5` also listed `hpux` twice after the removal; the HTML is regenerated with `build/makehtml.sh`, not edited.

Suite on a built tree: 0 failed, 0 skipped.

**Overlaps in `tests/buildsystem/test-suite-portability.sh`.** #485 edits the same script, and both sides rewrite line 206, so this is the one place where a resolver has to pick rather than keep both.

Neither is superseded: this drops the make-era platform targets, #485 stops a test calling the runner's own helpers. Whichever lands first, the other rebases once.


bonomani:
A verification pass before merge found two real things here.

**The type-column branch was selected from a list of type names, and the list was wrong for the case I justified it with.** `xfs`, `efs`, `cxfs` — from the IRIX df it was written for. A Linux `df -PT` reporting `ext4` fell straight through to the plain layout and produced exactly the defect the branch exists to prevent:

```
df -PT, ext4 before: disk5%.rrd after: disk,root.rrd
df -PT, xfs before: disk,root.rrd after: disk,root.rrd
df -P (no Type column) unaffected
```

It is selected from the header now — `Filesystem Type …`, which is what df prints when it emits that column. The test uses `ext4`, since the old selector passed the `xfs` one.

`xymonserver.cfg.5` also listed `hpux` twice after the platform removal. HTML regenerated with `build/makehtml.sh`.

**One finding was wrong**, worth recording: the review reported five checked-in ELF binaries under `build/`. Those are untracked build artifacts in the working tree — only the `.c` sources are in the repository, and the diff adds no binary.

I also extracted an `rrdtool` binary locally so the stored-percentage half of the test actually runs; its summary line now says which half it checked. Suite on a built tree: 74 passed, 0 skipped, 0 failed.

bonomani:
One more, from a question rather than a review: the stored-percentage check gated on the `rrdtool` command line, and I had satisfied that gate by extracting the binary locally rather than by noticing the gate was the problem.

The CLI is a separate package from the library the tree links against. Where it is absent the test did not fail — it printed a note and passed, having checked only the filename. That is the wrong half to keep: the filename comes out right for two different reasons, the percentage for one.

It reads through librrd now, reusing `tests/rrd/inode-unix-columns-harness.c` (from #325, which moved the inode test the same way for the same reason). Verified with the tool absent from `PATH`: the whole suite is green and this case checks the value.

`test-suite-portability.sh` gained the rule, so the next test to reach for the CLI is told where to look. It bites on both spellings — a `command -v rrdtool` gate and a bare `rrdtool lastupdate`.

bonomani:
Conflicts with #485 in `tests/buildsystem/test-suite-portability.sh`.

Both apply cleanly to `main` on their own, so neither is broken. Whichever merges second needs a rebase before it can go in, and a rebase can cost it a fresh review — so the pair is cheapest merged back to back.

PR #113, opened 2026-05-29
Resolve GCC build warnings (safe, no-behavior-change subset)

bonomani:
The safe subset extracted from #60: warning fixes with no behavior change.

## What it does

- **`-Wstringop-truncation`** — bounded string copies use `snprintf(dst, sizeof(dst), "%s", src)`, which truncates and NUL-terminates in one self-documenting call, rather than `strncpy` + a separate `dst[sizeof(dst)-1] = '\0'`. This is byte-for-byte identical to a terminated `strncpy` for every value anything reads (in the truncation case both drop the same final byte), and matches the `snprintf` already used on the same buffers in `stackio.c`. This also absorbs the `strncpy` + manual terminator that #186 added on the xymonnet rpcinfo `cmdpath` — same semantics, one call.
- **`xymond/client/zvse.c`** keeps its explicit `pid[2] = '\0'` — a deliberate 2-char partial copy, not a whole-string copy — corrected from #60's `pid[sizeof(pid)-1]`, which left `pid[2]` uninitialised.
- **`-Wmisleading-indentation`** — split compound one-line `if` statements and de-indent one misleadingly-indented unconditional statement; control flow unchanged.
- Enlarge `xymonproxy` `runtime_s[30]->[31]` to fit the worst-case `sprintf`.

## Relation to #60

Same fixes as #60 minus the warning-suppression pragmas and behaviour-sensitive changes (those go to #114). The `zvse.c` `pid` correction above is the one deviation from #60.

## Note

Reviewing this PR surfaced a pre-existing, unrelated stack overflow in the same `zvse_getvis_report()` function (`strcpy(pid, jinfo)` / `strcpy(jobname, q)` / width-less `sscanf` on untrusted `[getvis]` data) — tracked separately as #216, out of scope here.

## Validation

* Each touched file compiles clean with `-Wstringop-truncation` (0 warnings).
* `git diff --check` clean; rebased on current `main`.

### DIFF
--- a/lib/acknowledgementslog.c
+++ b/lib/acknowledgementslog.c
@@ -224,12 +224,12 @@ void do_acknowledgementslog(FILE *output,
 		    p = strrchr(host, '.');
 		    if (p) {
                         *p = '\0';
-			strncpy(svc,p+1,  sizeof(svc));
+			snprintf(svc, sizeof(svc), "%s", p+1);
                     }
 		    /* Xymon uses \n in the ack message, for the "acked by" data. Cut it off. */
 		    p = strstr(message, "\\nAcked by:");
 		    if (p) {
-			strncpy(recipient,p+12, sizeof(recipient));
+			snprintf(recipient, sizeof(recipient), "%s", p+12);
                         *(p-1) = '\0';
 		    }
 		    else {
--- a/lib/loadhosts_file.c
+++ b/lib/loadhosts_file.c
@@ -110,7 +110,7 @@ static int prepare_fromnet(void)
 
 	if (contentbuffer) freestrbuffer(contentbuffer);
 	contentbuffer = convertstrbuffer(fdata, 0);
-	strncpy(contentmd5, fhash, sizeof(contentmd5));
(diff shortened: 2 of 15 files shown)


bonomani:
Deep review of this PR: the warning fixes themselves are correct and no-behavior-change (verified — incl. the `pid[2]='\0'` correction over #60, which is in-bounds and right). One thing surfaced adjacent to the zvse hunk: a pre-existing, untrusted-input stack overflow in the same `zvse_getvis_report()` function (`strcpy(pid, jinfo)` / `strcpy(jobname, q)` / width-less `sscanf`). Filed separately as #216 — out of scope for this no-behavior-change PR, noted here for traceability. This PR is good to merge on its own.

SoundGoof (approved):
## Review

No findings. The warning fixes are narrowly scoped and preserve existing behavior.

Validated:
- Rebase onto current `main` is conflict-free.
- `range-diff` confirms the patch is unchanged.
- Full build succeeds.
- Test suite: 36 passed, 0 skipped, 0 failed.
- Complete server starts, processes client data, generates web/RRD/history output, and shuts down cleanly.

PR #114, opened 2026-05-29
Resolve GCC build warnings properly (no pragma suppression)

bonomani:
The remaining root-cause warning-fix subset from #60: fix the code rather than hide warnings. No `#pragma GCC diagnostic ignored`, and none of #60's unrelated behaviour changes (`XYMONHELPDIR`, `sig.c` crash-message/`getenv`).

## What it does

- **Bounded string copies** use `snprintf(dst, sizeof dst, "%s", src)` (the idiom shared with #113), or `memcpy` + manual NUL where a fixed byte count is copied. No `strncpy` is left with a hand-written terminator, so the two PRs speak one idiom.
- **`lib/loadalerts.c` — real behaviour fix.** The `grouplist` copy used `strncpy(dst, src, strlen(src))`, which writes **no** terminator; `strtok_r` then read past the buffer and the **last alert group matched against heap garbage**. `snprintf` fixes both the include- and exclude-group copies. This makes group-scoped alert routing (`GROUP=` / `EXGROUP=` in alerts.cfg) **deterministic where it was previously non-deterministic** — worth noting for anyone whose alert rules key off the last group. Also replaces the fragile hand-tracked `strncat` bounds with remaining-space bounds.
- **`xymond/do_rrd.c`.** The RRD path is now built through one helper that checks for truncation; on truncation it logs and **skips the update** (`return -1` / no datasets) instead of creating/updating a truncated, possibly-colliding path. The host-directory path is bounded the same way. *(Behaviour note: pre-existing code detected-and-continued or silently truncated; this drops the update instead — the safe choice for a path that can't be represented.)*
- **`xymongen/util.c`, `pagegen.c`, `xymond/trimhistory.c`, `xymond/xymond_hostdata.c`.** Fixed-buffer `sprintf` → `snprintf`, including the remaining page-link writes so the whole function is bounded (not just one call).
- **`lib/sig.c`, `lib/links.c`.** Only NUL-terminate the affected buffers (via `snprintf`); `xgetenv()` and crash-message behaviour unchanged, per the PR's charter.
- **`web/svcstatus.c`.** Keep the copy safe and correct the CLIENT-uri `snprintf` remaining-space expression (the operands were reversed — harmless in practice due to allocation slack, but now correct).

## Relation to #60 / #113

Same real fixes as #60, minus the pragma suppression and the behaviour changes. Together with #113 (the mechanical, no-behaviour-change subset) this supersedes #60.

## Validation

* Each touched file compiles clean of `-Wstringop-truncation`; the cleanup introduces no new warnings.
* `git diff --check` clean; rebased on current `main`.

### DIFF
--- a/client/logfetch.c
+++ b/client/logfetch.c
@@ -326,8 +326,8 @@ char *logdata(char *filename, logdef_t *logdef)
 
 			dbgprintf(" - Last position was %u, found curpos location at %u in current buffer.\n", logdef->lastpos[1], bytesin);
 			t = strdup(fillpos);	/* need shuffle about to insert before this line */
-			strncpy(fillpos, curpostxt, strlen(curpostxt));		/* add in the CURRENT + \n */
-			strncpy(fillpos+strlen(curpostxt), t, strlen(t));	/* add in whatever this line originally was */
+			memcpy(fillpos, curpostxt, strlen(curpostxt));		/* add in the CURRENT + \n */
+			memcpy(fillpos+strlen(curpostxt), t, strlen(t));	/* add in whatever this line originally was */
 			*(fillpos+strlen(curpostxt)+strlen(t)) = '\0';		/* and terminate it */
 			xfree(t);						/* free temp */
 			curpos = fillpos;					/* leave curpos to the beginning of the CURRENT flag */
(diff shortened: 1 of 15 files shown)


bonomani:
Heads-up on `xymond/do_rrd.c`: this PR's truncation-check on the path-assembly `snprintf(filedir, … "%s/%s/%s", rrddir, hostname, rrdfn)` sits **inside the same hunk** that #154 reworks (`@@ -287,16 @@`, the host-dir-lazy change). So **#114 and #154 will conflict** in that region — worth coordinating merge order (e.g. land one, rebase the other).

For the record, the *remaining* unbounded `sprintf()` writes in the same function — `sprintf(filedir, "%s%s", rrddir, cacheitem->key)` plus `rrakey`/`stepsetting`/`rrdfn` — are tracked separately in #171 and are **disjoint** from this PR's lines; #171 will follow the same truncation-detecting `snprintf` style this PR introduces.

SoundGoof (approved):
No actionable findings.

- All four CI checks pass.
- Synthetic merged build succeeded.
- Full suite: 48 passed, 0 skipped, 0 failed.
- Live client/status transfer, processing, and hostdata persistence succeeded without runtime errors.

Issue #130, opened 2026-05-31
Superseded: enable SNMP by autodetect / opt-out like LDAP


bonomani:
## Problem
`configure.server` enables LDAP and SNMP asymmetrically:

- **LDAP** runs `build/ldap.sh` unconditionally, auto-detects the client library, and defaults to enabled (an empty answer → `y`). So it self-enables whenever `libldap` is present — **opt-out**.
- **SNMP** is gated behind `if test "$SNMP" = "1"`, with no autodetect and no default; anything else → `DOSNMP=no`. So it stays off unless `SNMP=1` is passed explicitly — **opt-in**.

This is inherited upstream behavior (the `SNMP=1` gate dates to 2008, `6fed27cd`; the initial revision had no SNMP at all), not a regression — but it produces a surprising result.

## Consequence
On the standard `default` build (CI and typical source builds), where neither LDAP nor SNMP is explicitly requested:

- LDAP → `DOLDAP=yes` (auto-enabled)
- SNMP → `DOSNMP=no` (never built)

So `xymon-snmpcollect` is **not compiled by default**, which means:

- The #77 build failure (snmpcollect vs an MD5-disabled Net-SNMP) has **no CI coverage** on the make/`default` lanes. It only ever surfaced on a cmake lane (which defaults `ENABLE_SNMP=ON`) — e.g. an early openSUSE Tumbleweed cmake run — and that coverage was lost once the lanes ran make/`default`.
- Users/packagers expecting parity with LDAP silently get no SNMP.

This is the concrete form of #78's note: *"make enabling this feature more similar to other features."*

## Proposal
Make SNMP **opt-out / autodetect**, mirroring LDAP:

- In `configure.server`, run `build/snmp.sh` unconditionally and set `DOSNMP=yes` when `net-snmp-config` is present.
- Keep an explicit opt-out (e.g. `SNMP=0`) for builds that don't want it.
- Result: the `default` lanes (including openSUSE Tumbleweed) build `xymon-snmpcollect` automatically, restoring #77 regression coverage. The cmake build already defaults `ENABLE_SNMP=ON`, so the two build systems converge.

### Decision embedded here
This flips SNMP to **on-by-default when Net-SNMP is available** — a behavior change for source/package builds, matching LDAP. That is the intent expressed in #78; calling it out explicitly so it is a conscious choice.

## References
- Parent / umbrella: #78
- Restores CI coverage for: #77 (fix in PR #75)

bonomani:
[...]

Issue #78, opened 2026-02-21
Xymon SNMP support


RolandRosenfeld:
Today I noticed (via https://github.com/xymon-monitoring/xymon/issues/77) that Xymon already contains SNMP support, which is well hidden and widely undocumented.
To use it, you have to set `SNMP=1` during configuration, which may result into xymon-snmpcollect being build.

This ticket tries to collect some information around SNMP support.  We should add these to some official documentation and try to make enabling this feature more similar to other features.

There are also some patches available on the list archive that should be reviewed/applied.

Let's start with some documentation:
Brian Scott's "SNMP collective adventure": https://lists.xymon.com/xymon/2021-July/047646.html pointing to https://www.dropbox.com/s/ftn695y3fv4ykur/Misc%2022%20Xymon%20SNMP.docx?dl=0 (you heed to strip `/pipermail` from the list archive links, since the archive moved in the meantime).
This documents, how Brian got SNMP to work and lists several tricks as well as patches to fix this and that.
In https://lists.xymon.com/archive/2021-July/047669.html Brian ships his patches in a more machine readable way.

<s>Except the patches in the above document I found another patch by Jeremy Laidman in https://lists.xymon.com/xymon/2022-July/047903.html (patch in https://lists.xymon.com/xymon/attachments/20220728/465e37bd/attachment.obj) fixing some off-by-one errors in the SNMPv3 code.</s> (this was already merged in 4.3.29: eda84777bbe3a862cab490465b249ed888b5b229)

I think about enabling SNMP in the Debian package, but it seems to be necessary to apply the above patches before.

bonomani:
## SNMP support — status tracker

Umbrella issue for SNMP support (build integration, docs, patches). Tracking the moving parts here:

**Build / configure**
- [x] SNMPv3 `snmphosts.cfg` parser off-by-one — merged in 4.3.29 (`eda84777`); see #23 *(closed)*
- [ ] Integrate SNMP into the `configure.server` framework, wired like the other optional features but **off by default** (SNMP is experimental) — #135
- [ ] `xymon-snmpcollect` build fails when Net-SNMP lacks the MD5 symbol (`usmHMACMD5AuthProtocol`) — fall back to SHA1 — #77 (fix in PR #75)

**Docs**
- [ ] Fold Brian Scott's "SNMP collective adventure" guide + tricks (list-archive links in the description) into official docs
- [ ] Review / apply Brian Scott's machine-readable patches from the list archive

**Related bugs**
[...]

Issue #77, opened 2026-02-21
xymon-snmpcollect build fails when Net-SNMP MD5 auth symbol is unavailable


bonomani:
`xymon-snmpcollect` fails to compile on openSUSE Tumbleweed because `usmHMACMD5AuthProtocol` is missing.

Error:
`xymonnet/xymon-snmpcollect.c:429:55: error: 'usmHMACMD5AuthProtocol' undeclared`

Expected: if MD5 is not available in Net-SNMP, build should still pass (fallback to SHA1).

PR #131, opened 2026-05-31
configure: enable SNMP by autodetect (opt-out), like LDAP

bonomani:
## Summary
SNMP was opt-in (`if test "$SNMP" = "1"`), while LDAP/SSL auto-enable when their library is present. So default builds got `DOSNMP=no` and never exercised the `xymon-snmpcollect` path.

This makes SNMP behave like LDAP: run `build/snmp.sh` (which already probes `net-snmp-config`) by default, skip only on explicit `SNMP=0`/`SNMP=n`. One-line change, branch order kept (`then` = enable).

```diff
-if test "$SNMP" = "1"
+if test "$SNMP" != "0" -a "$SNMP" != "n"
```

- **default** → enabled if Net-SNMP present, gracefully disabled if not
- **`SNMP=0` / `SNMP=n`** → opt-out
- **`SNMP=1`** → unchanged

Second commit silences the `net-snmp-config: not found` stderr noise now that the probe runs by default (see comment below).

Refs #130, #78, #77

### DIFF
--- a/build/snmp.sh
+++ b/build/snmp.sh
@@ -3,7 +3,7 @@
 	SNMPINC=""
 	SNMPLIB=""
 
-	VERSION=`net-snmp-config --version`
+	VERSION=`net-snmp-config --version 2>/dev/null`
 	if test $? -eq 0
 	then
 		echo "Found Net-SNMP version $VERSION"
--- a/configure.server
+++ b/configure.server
@@ -181,7 +181,7 @@ echo ""; echo "";
 . build/clock-gettime-librt.sh
 echo ""; echo ""
 
-if test "$SNMP" = "1"
+if test "$SNMP" != "0" -a "$SNMP" != "n"
 then
 	. build/snmp.sh
 	echo ""; echo ""


bonomani:
Follow-up commit `0cf75dc4c`: silence the `net-snmp-config: not found` stderr noise in `build/snmp.sh`.

Now that `snmp.sh` runs on every default build (the point of this PR), its Net-SNMP probe surfaces a cosmetic issue: when the library is absent, the shell itself leaks a raw `net-snmp-config: not found` line to stderr *in addition to* the script's intended graceful message. The probe was already correct (it keys off the exit code), it just didn't suppress the diagnostic.

```diff
-	VERSION=`net-snmp-config --version`
+	VERSION=`net-snmp-config --version 2>/dev/null`
```

One line. Exit-code handling (`if test $? -eq 0`) is unchanged, so behaviour is identical — only the noise is gone. Without the library the build now shows just:

```
Checking for Net-SNMP ...
Could not find Net-SNMP (net-snmp-config command fails)
Continuing with SNMP support disabled.
```

This bug pre-existed but was invisible, since `snmp.sh` was never sourced on default builds before this PR.

RolandRosenfeld:
I'm not sure, whether we should just enable SNMP support. I miss at least the following:
- Documentation (there's no hint in the package, how this could work)
- Some mechanism to set include and library path for net-snmp.
  - I miss --snmpinclude and --snmplib options in configure.server
  - I just noted that there is `net-snmp-config --cflags` and `net-snmp-config --libs` in xymonnet/Makefile, but this isn't consistent to other libraries and not intuitive.
- Also the enable-mechanism isn't consistent.  All other "features" are enabled/diabled by setting ENABLE<FEATURE>=y|n (ENABLERRD, ENABLESSL, ENABLELDAP), why not use the same mechanism for SNMP?

I'm not sure that SNMP code is production ready, so it may be a good idea that this currently disabled/hidden by default.

bonomani:
Closing in favour of #135.

You're right that SNMP shouldn't be enabled by default — the code is experimental and hasn't had functional changes in years. So instead of this PR's auto-enable, #135 makes the *mechanism* consistent while keeping SNMP **off by default**: `ENABLESNMP=y|n` like ENABLERRD/SSL/LDAP, `--snmpinclude`/`--snmplib`, configure-time resolution instead of `net-snmp-config` in `xymonnet/Makefile`, and documentation — addressing the points you raised here. `SNMP=1` stays as a backward-compatible alias.

Thanks for the review.

PR #133, opened 2026-05-31
build/rrd.sh: find RRDtool under MacPorts/Homebrew prefixes on macOS (fixes #132)

### LINKED ISSUE #132: RRDtool argv-API autodetect aborts configure.server on macOS (MacPorts/Homebrew prefixes not searched)
### Summary

On macOS, `configure.server` aborts at the RRDtool detection step even when RRDtool is installed via MacPorts or Homebrew:

```
Checking for RRDtool ...
ERROR: Unable to determine the RRDtool argv API from the installed headers.
Re-run configure with an explicit override:
  RRD_CONST_ARGS=1 ./configure.server   # rrd.h uses const char **
  RRD_CONST_ARGS=0 ./configure.server   # rrd.h uses char **
```

The build cannot proceed without manually passing `RRD_CONST_ARGS`.

### Environment

- macOS (reproduced on macOS 14, 15 and 26 in CI)
- RRDtool installed via **MacPorts** (`/opt/local`) — also affects **Homebrew on Apple Silicon** (`/opt/homebrew`)

### Root cause

The RRDtool argv-API autodetection in `build/rrd.sh` (introduced in commit `cbe373ec0`, "rrd: autodetect argv API instead of guessing from version") probes the API by **compiling** a translation unit that includes `<rrd.h>`. That probe needs the header to be on the include path (`RRDINC` / `-I`).

However, the prefix-search loop in `build/rrd.sh` only scans:

```
/opt/rrdtool* /usr/local/rrdtool* /usr/local /usr/pkg /opt/csw /opt/sfw /usr/sfw
```

It does **not** include the macOS package-manager prefixes `/opt/local` (MacPorts) or `/opt/homebrew` (Homebrew on Apple Silicon). So `RRDINC` stays empty, the probe compiles `#include <rrd.h>` without `-I/opt/local/include`, the header is not found, **both** candidate prototypes fail to compile, and the script aborts with "Unable to determine ... argv API".

The previous detection used `pkg-config --atleast-version=1.9.0 librrd`, which locates `librrd.pc` (shipped by MacPorts/Homebrew) **without needing the header path**, and never aborted. That is why this only started failing after the rework — the missing macOS prefix was latent before.

### Suggested fix

Consult `/opt/local` and `/opt/homebrew` as a fallback when the standard prefix search finds no headers, and recognise the `.dylib` shared-library extension. Keeping it as a fallback (only when nothing else matched) leaves detection on all other platforms unchanged. PR to follow.

bonomani:
Fixes #132.

## Problem

On macOS, `configure.server` aborts at the RRDtool step with "Unable to determine the RRDtool argv API from the installed headers", even when RRDtool is installed via MacPorts (`/opt/local`) or Homebrew on Apple Silicon (`/opt/homebrew`).

The argv-API autodetection added in `cbe373ec0` probes the API by compiling `#include <rrd.h>`, so it needs the header on the include path. The prefix-search loop in `build/rrd.sh` never listed the macOS package-manager prefixes, so the header was not found, both candidate prototypes failed to compile, and configure aborted. The old `pkg-config librrd` check did not need the header path, which is why this only surfaced after the rework. See #132 for the full analysis.

## Fix

Add a fallback scan of `/opt/local` and `/opt/homebrew` that runs **only when the standard search found no headers**, recognising the `.dylib` extension for `librrd`/`libpng`/`libz`.

Because it is gated on an empty `RRDINC` and placed after the existing loop, the original search loop is left **byte-for-byte unchanged** — detection on every non-macOS platform is unaffected. An explicit `--rrdinclude`/`--rrdlib` still takes precedence (applied afterwards).

`/opt/local` and `/opt/homebrew` are the default prefixes for MacPorts and Homebrew, so both macOS cases are covered; the compile probe still determines the argv API (`const char **` vs `char **`) exactly as before.

## Testing

CI (server lanes) on **macOS 14, 15 and 26**, RRDtool from MacPorts:

```
Checking for RRDtool ...
Detected RRDtool argv API: const char **
Compiling with RRDtool works OK
Linking with RRDtool works OK
```

All three macOS lanes pass; non-macOS lanes are unaffected (the standard search path is unchanged).

### DIFF
--- a/build/rrd.sh
+++ b/build/rrd.sh
@@ -64,6 +64,40 @@
 		fi
 	done
 
+	# MacPorts (/opt/local) and Homebrew (/opt/homebrew) install RRDtool outside the
+	# prefixes scanned above. Consult them only when the standard search found nothing,
+	# so this stays a pure fallback that leaves detection on other systems unchanged.
+	# macOS shared libraries use the .dylib extension.
+	if test "$RRDINC" = ""
+	then
+		for DIR in /opt/local /opt/homebrew
+		do
+			if test -f $DIR/include/rrd.h
+			then
+				RRDINC=$DIR/include
+			fi
+
+			if test -f $DIR/lib/librrd.dylib
+			then
+				RRDLIB=$DIR/lib
+			fi
+			if test -f $DIR/lib/librrd.a
+			then
+				RRDLIB=$DIR/lib
+			fi
+
+			if test -f $DIR/lib/libpng.dylib
+			then
+				PNGLIB="-L$DIR/lib -lpng"
+			fi
+
+			if test -f $DIR/lib/libz.dylib
+			then
+				ZLIB="-L$DIR/lib -lz"
+			fi
+		done
+	fi
+
 	if test "$USERRRDINC" != ""; then
 		RRDINC="$USERRRDINC"
 	fi


RolandRosenfeld:
I'm not sure, whether it is good to add pkg-config only for the rrdtool now.
Wouldn't it be better to rewrite _all_ lib/include paths (rrdtoo, pcre, ssl, ldap, cares, snmp (missing in configure options)) in the same way?
But I wouldn't change this in 4.3.31 but delay it to a later version.
As written in #132 this can easily worked around for MacPorts/Homebrew using configure options.

bonomani:
To separate what's deferred from what's kept:

- **Keep this PR** — it's a regression fix. macOS built out of the box before the argv-API rework (`cbe373ec0`); the new compile-probe is what turned a missing prefix into a hard `exit 1`. The diff just restores that: a fallback scan of `/opt/local`/`/opt/homebrew`, gated on the standard search finding nothing, original loop unchanged, **no pkg-config involved**. Needing `--rrdinclude` is itself the regression.
- **Defer the broader pkg-config rewrite** you suggest (unifying rrdtool/pcre/ssl/ldap/c-ares/SNMP detection) — agreed it's the cleaner end state and not for 4.3.31.

So: this minimal regression fix now, the uniform rewrite later. Work for you?

RolandRosenfeld:
I do not understand why you don't extend the current DIR list in line 8 and maybe add tests for .dylib after the ones for .a and .so.  This would be more strait forward.

On the other hand, I do not understand, why this is only needed for rrd.sh and not for pcre.sh, ssl.sh, ldap.sh and c-ares.sh, which have a similar mechanism and do not check for /opt/local, /opt/homebrew and .dylib.  Shouldn't it be added there, too, if it is necessary?

bonomani:
Closing in favour of a feature issue: #134.

Digging further: the RRDtool `exit 1` is a real behaviour change from cbe373ec0 (it turned RRDtool's previous graceful "disable if absent" into a hard abort) — but it is **not a user-visible regression** on a stock upstream build. It's effectively silent: macOS/MacPorts isn't buildable upstream today because the other probes don't look under `/opt/local` either, and `pcre.sh` (mandatory) aborts before configure ever reaches RRDtool. So an rrd-only change wouldn't make macOS build, and a "regression fix" framing doesn't fit.

The real gap — the make build has no MacPorts/Homebrew awareness at all — is now tracked as a feature in #134 (a minimal prefix injection in `Makefile.Darwin` plus routing the RRDtool probe through the Makefile like the others). Thanks @RolandRosenfeld — your questions are what surfaced this.

PR #136, opened 2026-06-01
build: detect macOS package-manager dependency paths in the make probes

### LINKED ISSUE #134: macOS: make build doesn't find MacPorts/Homebrew prefixes (add support)
The make build doesn't know the macOS package-manager prefixes — `/opt/local` (MacPorts) and `/opt/homebrew` (Homebrew on Apple Silicon) — so none of the dependency probes (`pcre`, `ssl`, `ldap`, `c-ares`, `rrd`, `fping`) search there.

On a stock macOS install this aborts configure (mandatory `pcre.sh`, and the `rrd.sh` argv-API probe, `exit 1`) or silently disables `ssl` / `ldap` / `c-ares`. (Intel Homebrew under `/usr/local` already works, since the loops list it.)

Fix in #136. Supersedes #132 and #133.

bonomani:
## Problem

The `make` build had no awareness of the macOS package-manager prefixes
(`/opt/local` for MacPorts, `/opt/homebrew` for Homebrew on Apple Silicon),
so on a stock macOS install the dependency probes don't find their headers
and libraries:

- `build/pcre.sh` (mandatory) aborts the build,
- `ssl`/`ldap`/`c-ares` silently disable their features,
- `build/rrd.sh`'s argv-API probe aborts with *"Unable to determine the
  RRDtool argv API"*.

## Change

Teach each `build/*.sh` dependency probe (pcre, ssl, ldap, c-ares, fping,
rrd) to look under the macOS package-manager prefixes, reusing the
per-script prefix lists already used for the other platforms, and add the
`.dylib` checks alongside the existing `.so`/`.a` ones.

- The macOS paths are **appended at the end** of each existing search list.
  Because the probes are last-match-wins, this keeps the established
  non-macOS ordering (e.g. NetBSD `/usr/pkg`) intact and only changes the
  result on macOS.
- Non-keg-only formulae (`pcre2`, `c-ares`, `rrdtool`, `fping`) resolve from
  the prefix itself: `/opt/homebrew` (Apple Silicon) and `/opt/local`
  (MacPorts); Intel Homebrew `/usr/local` is already in the lists.
- Keg-only formulae get their version-stable opt paths:
  `…/opt/openssl@3` and `…/opt/openldap` for both `/usr/local/opt` (Intel)
  and `/opt/homebrew/opt` (Apple Silicon).

Detection stays in the individual probes — where the rest of the
per-library discovery lives — rather than bolting a single global prefix
onto `Makefile.Darwin`.

## Testing

Built the make **server** variant on the project's macOS. On a MacPorts runner all dependencies (PCRE, c-ares,
RRDtool, OpenSSL, LDAP) resolved under `/opt/local` and both compiled and
linked; RRDtool's argv API was autodetected correctly.

Closes #134.

### DIFF
--- a/build/c-ares.sh
+++ b/build/c-ares.sh
@@ -2,7 +2,7 @@
 
 	CARESINC=""
 	CARESLIB=""
-	for DIR in /opt/c*ares* /usr/local/c*ares* /usr/local /usr/pkg /opt/csw /opt/sfw
+	for DIR in /opt/c*ares* /usr/local/c*ares* /usr/local /usr/pkg /opt/csw /opt/sfw /opt/homebrew /opt/local
 	do
 		if test -f $DIR/include/ares.h
 		then
@@ -14,6 +14,10 @@
 		fi
 
 		if test -f $DIR/lib/libcares.so
+		then
+			CARESLIB=$DIR/lib
+		fi
+		if test -f $DIR/lib/libcares.dylib
 		then
 			CARESLIB=$DIR/lib
 		fi
--- a/build/fping.sh
+++ b/build/fping.sh
@@ -2,7 +2,7 @@
 
 	echo "Checking for fping ..."
 
-	for DIR in / /usr /usr/local /opt /usr/pkg /opt/csw
+	for DIR in / /usr /usr/local /opt /usr/pkg /opt/csw /opt/homebrew /opt/local
 	do
 		if test "$DIR" = "/"; then DIR=""; fi
 
--- a/build/ldap.sh
+++ b/build/ldap.sh
@@ -2,7 +2,7 @@
 
 	LDAPINC=""
 	LDAPLIB=""
-	for DIR in /opt/openldap* /opt/ldap* /usr/local/openldap* /usr/local/ldap* /usr/local /usr/pkg /opt/csw /opt/sfw
+	for DIR in /opt/openldap* /opt/ldap* /usr/local/openldap* /usr/local/ldap* /usr/local /usr/pkg /opt/csw /opt/sfw /usr/local/opt/openldap /opt/homebrew/opt/openldap /opt/local
 	do
 		if test -f $DIR/include/ldap.h
 		then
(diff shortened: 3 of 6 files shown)


spiderr:
Hi @bonomani, I'm a mac guy, but develop all in linux. I have done a bunch of XCode in the past, so I was going to try and validate this change. I looked at docs/install.html and see it references Mac OS X 10.6, 10.7, and 10.8 👴🏻

Do you have a brief write up how to compile on macOS ? I will follow it and build on my desktop. I have XCode and brew.sh installed (seen in your /opt/homebrew changes). Configure fails on pcre2.h - I can figure it out, bit if you have a build steps, I'll follow those and that will help other macOS users.

```
Checking for PCRE ...
test-pcre.c:2:10: fatal error: 'pcre2.h' file not found
   2 | #include <pcre2.h>
     |          ^~~~~~~~~
```

bonomani:
I am not an expert, but from what I know, macOS has two well-known non-native package managers, in addition to the native installation methods: MacPorts and Homebrew.

I generally prefer MacPorts for software that does not change frequently, while Homebrew is also fine for testing and development. For Xymon, I think it makes sense to start with MacPorts because most users probably do not want to update their monitoring system every week. MacPorts seems better suited for long-term stability, whereas Homebrew is often more focused on newer versions and faster-moving software.

I will try to make Xymon available through both if I can.

What you are doing is slightly different, though: you are building directly from source. At the moment, that is the only option, and that is what developers typically need. Once packages are available, they should automatically install the required dependencies.

```bash
# 1. Dependencies + code
brew install pcre2 c-ares rrdtool fping openssl@3 openldap

git fetch https://github.com/xymon-monitoring/xymon.git pull/136/head:pr-136
git checkout pr-136

# 2. Create an install directory owned by your user
mkdir -p ~/xymon

# 3. Configure
./configure --server
```

Suggested answers during configure:

```text
What userid will be running Xymon [xymon]
your_current_user

Where do you want the Xymon installation
/Users/your_current_user/xymon

Logfiles
/Users/your_current_user/xymon/logs

Everything else
Use the defaults
```

Then build and install:

```bash
make
make install
```

spiderr (approved):
LGTM — macOS build dependency path detection.

PR #139, opened 2026-06-01
showgraph: anchor single-service RRD match to the FNPATTERN component (#20)

### LINKED ISSUE #20: SF Reported Bug - RRD filename matching
```
I wish I had the C programming chops to be able to provide a patch. Alas I
can only report the bug.

If I have a custom TCP test with the name "proxyconn", then this creates an
RRD file called tcp.proxyconn.rrd. This is defined in protocols.cfg:

[proxyconn]
        port 8080

The code in showgraph that displays the ping times graph, for the conn test
page, creates a URL containing "service=tcp:conn" which shows tcp.conn.rrd
on the graph.

However, it seems that the string following the colon, "conn", is matched
against RRD filenames using a substring match, which means any RRD files
that start with "tcp." where the text that follows this contains the word
"conn" will be added to the conn test page graph. So I end up with both
ping times and proxy times on the graph shown on the conn status page.

Being an unanchored substring match, this matches "tcp.conn<anything>.rrd"
and "tcp.<anything>conn.rrd".

This would also be the case for other test pages that come from xymonnet
probes, such as the "dns" status page showing graphs of any RRD file
containing the word "dns", not just tcp.dns.rrd. And ssh.

With a default installation I don't think this would ever happen, because
none of the xymonnet tests have names that match as substrings of any other
xymonnet tests, so this bug is not going to manifest without creating a
custom TCP test in protocols.cfg.

Cheers
Jeremy
```

https://sourceforge.net/p/xymon/mailman/xymon-developer/?viewmonth=202501

bonomani:
## Problem

When a bundled service is viewed as a single graph (e.g. `tcp:conn`), `showgraph` filtered the RRD directory with an unanchored `strstr(d->d_name, service)`. So service `conn` also matched `tcp.proxyconn.rrd` and `tcp.connfoo.rrd`, plotting unrelated series on the conn page; same for `dns`, `ssh`, etc. It only triggers with custom `protocols.cfg` tests whose names are substrings of (or contain) another test name.

Closes #20.

## Fix

A request can resolve to a "single" graph two different ways, which need different matching. A new `rrdparamisservice` flag distinguishes them:

1. **The bundle graph is used as a fall-back** (`tcp:conn` → `[tcp]`, or a bare test mapped via `TEST2RRD` such as `dns`). A bundle is named `<bundle>.<service>.rrd`, so the `FNPATTERN` group-1 capture is exactly the service — require an exact match: `conn` matches `tcp.conn.rrd` but no longer `tcp.proxyconn.rrd`, `tcp.connfoo.rrd` or `tcp.conn.extra.rrd`.
2. **The request resolves to its own graph definition** (`tcp.http` → `[http]`, `ncv:slab` → `[slab]`). There group 1 is a subitem (a URL, a slab name), not the service, so the unanchored `strstr` match is kept. A fall-back whose `FNPATTERN` has no capture group also keeps the substring match — there is no capture to compare, and keeping everything would *widen* selection.

Two guard rails around the new invariant:

- If the exact match rejects every candidate file, showgraph logs `no RRD file matched service 'X' - check that FNPATTERN group 1 captures the service component` instead of leaving only a silently blank image.
- An empty service name (`tcp:`) is rejected up front. Behavior change: such a URL previously rendered the entire bundle (`strstr` with an empty needle matches everything); no in-tree producer emits that form.

The fix relies on **`FNPATTERN` capture group 1 being exactly the service component** for a bundle; this holds for everything shipped. A custom definition that violates it loses series — possibly all of them — but never gains contaminated ones, and the errprintf above names the cause. The invariant is documented in `graphs.cfg` and `graphs.cfg.5`.

## Regression coverage

`tests/web/showgraph-single-service.sh` drives the real showgraph CGI with `--debug` against a fake RRD directory and asserts on the rrd_graph argument dump (using the stock `assert_contains`/`assert_not_contains` helpers): exact selection for `tcp:conn` and bare `dns`, substring match preserved for `tcp.http`, capture-less fall-back keeps `nocap.x.rrd` and excludes `nocap.y.rrd`, `tcp:` rejected. The test fails on current main and passes with this fix; full suite: 15 passed, 0 failed.

### DIFF
--- a/tests/web/showgraph-single-service.sh
+++ b/tests/web/showgraph-single-service.sh
@@ -1,0 +1,111 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/web/showgraph-single-service.sh
+#
+# Regression guard for issue #20: viewing one service out of a bundle
+# (service=tcp:conn) must select only that service's RRD file, not every
+# file whose name contains the service as a substring. Drives the real
+# showgraph CGI with --debug against a fake RRD directory and asserts on
+# the rrd_graph argument dump.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+ROOT=$(find_root)
+
+CC=${CC:-cc}
+command -v "$CC" >/dev/null 2>&1 || skip "no C compiler available (CC=$CC)"
+command -v make >/dev/null 2>&1 || skip "make not available"
+
+# Needs a configured/built tree (bare-tree CI skips; the post-build suite
+# runs it for real) and one built WITH RRD support.
+[ -f "$ROOT/include/config.h" ] && [ -f "$ROOT/lib/libxymoncomm.a" ] \
+	|| skip "tree not built (run make first; the post-build CI suite covers this)"
+[ -f "$ROOT/web/showgraph.cgi" ] || skip "tree built without RRD support (no showgraph.cgi)"
+
+# RRD and SSL build flags as configure detected them (SSLLIBS is empty
+# on a tree built without SSL - don't force -lssl on such a link)
+rrddef=$(sed -n 's/^RRDDEF *= *//p' "$ROOT/Makefile")
+rrdlibs=$(sed -n 's/^RRDLIBS *= *//p' "$ROOT/Makefile")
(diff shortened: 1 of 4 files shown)


feld (approved):
Before:

Image


After:


Image


Working exactly as expected. Approved

PR #140, opened 2026-06-04
build/clock-gettime-librt.sh: always call clean at the end.

RolandRosenfeld:
Otherwise build/test-clockgettime-librt persists after distclean.


bonomani (approved):
- Correct and minimal. make clean is idempotent, and there's already a clean at the start of
  the script, so cleaning at the end is symmetric and safe.

PR #142, opened 2026-06-05
apache: drop Apache 1.3 / pre-2.4 access config (#141)

### LINKED ISSUE #141: Remove Apache 1.3 configuration (in favour of 2.x)
We have some apache configs:
- docs/xymon-apacheconf.txt(.DIST)
- xymond/etcfiles/xymon-apache-open.DIST
- xymond/etcfiles/xymon-apache-secure.DIST
which currently support Apache 1.3 as well as 2.x.

Apache 1.3 is dead for ages now.  
According to https://bugs.debian.org/418266 Debian support for apache 1.x ended in 2011 (15 years ago).
According to https://repology.org/project/apache/versions there seems to be nearly no OS that still uses apache 1.

So I propose removal of the 1.3 configuration options from all apache configs.

bonomani:
Closes #141.

The Xymon apache configs carried two access-control variants behind `<IfModule>`
guards: the modern Apache 2.4 syntax (`Require all granted`, under
`mod_authz_core.c`) and the legacy pre-2.4 syntax (`Order`/`Allow`, under
`!mod_authz_core.c`). This drops the legacy blocks and keeps only the 2.4+ form.

## Changed

- `docs/xymon-apacheconf.txt.DIST`
- `xymond/etcfiles/xymon-apache-open.DIST`
- `xymond/etcfiles/xymon-apache-secure.DIST`
- `common/hosts.cfg.5` — modernise the `server-status` example
  (`Order`/`Deny`/`allow` → `Require ip 127.0.0.1`). Not in the issue's file
  list, but the same legacy class; included for consistency.

Headers updated to "Apache 2.4 and later". The auth blocks (`Require valid-user`,
`RequireAll`) are untouched. Net: +12 / −69.

## Scope note (compatibility)

There is no Apache-1.3-only block to remove in isolation: the legacy
`Order`/`Allow` block is the **same** syntax used by Apache 1.3, 2.0 **and**
2.2. So this necessarily drops pre-2.4 as well, not just 1.3. That is safe in
practice:

- Apache **2.2** has been EOL upstream since **July 2017** (2.2.34 was the last
  release) — ~9 years with no security updates; **2.0** is older still.
- Apache **2.4** shipped in **Feb 2012** and has been the distro default since
  ~2014–2015 (Fedora 18 2013, Ubuntu 14.04 / RHEL 7 2014, Debian jessie 2015).

So the only compatibility dropped is for releases that have been unsupported and
absent from current distros for a decade. If keeping 2.2 is desired, the
alternative is to keep the dual `<IfModule>` structure — but then the change is
purely cosmetic (just the comment), which would not actually remove any config.

### DIFF
--- a/common/hosts.cfg.5
+++ b/common/hosts.cfg.5
@@ -1379,11 +1379,7 @@ Apache configuration. The following configuration is needed:
 .br
         SetHandler server\-status
 .br
-        Order deny,allow
-.br
-        Deny from all
-.br
-        allow from 127.0.0.1
+        Require ip 127.0.0.1
 .br
     </Location>
 .br
--- a/docs/manpages/man5/hosts.cfg.5.html
+++ b/docs/manpages/man5/hosts.cfg.5.html
@@ -97,7 +97,7 @@ files (devices, sockets, pipes etc).
 <P>
 <DT>optional include/directory<DD>
 Both &quot;include&quot; and &quot;directory&quot; can be prefixed with the tag
-&quot;optional&quot;, which will preven an error message being logged if
+&quot;optional&quot;, which will prevent an error message being logged if
 the file or directory is not present on a system.
 <P>
 <P>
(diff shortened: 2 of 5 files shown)


StefCoene (approved):
Reviewed the full repo for remaining legacy Apache syntax — all instances of `Order`/`Allow from`/`Deny from`/`mod_authz_core` are in exactly the five files this PR touches. Nothing left behind.

The security note on the `SECURECGIDIR` block is a nice catch: the original `<IfModule mod_authz_core.c>` block already had a latent `<RequireAny>` bypass; the comment makes the invariant explicit for future editors.

LGTM.

PR #145, opened 2026-06-09
svcstatus: fix path traversal issue

feld:
Fixes a security issue with the svcstatus CGI - a classic path traversal issue where you can manipulate one of the parameters to read arbitrary files on the webserver with the privileges of the webserver process.

This patch came directly from Henrik

### DIFF
--- a/web/svcstatus.c
+++ b/web/svcstatus.c
@@ -89,7 +89,7 @@ static int parse_query(void)
 		else if (strcasecmp(cwalk->name, "CLIENT") == 0) {
 			char *p;
 
-			hostname = strdup(cwalk->value);
+			hostname = strdup(basename(cwalk->value));
 			p = hostname; while ((p = strchr(p, ',')) != NULL) *p = '.';
 			service = strdup("");
 			outform = FRM_CLIENT;


bonomani (approved):
Reviewed the fix and the data flow.

This closes the one remaining gap in svcstatus's existing `basename()` defense: `HOST`, `SERVICE`, the `HOSTSVC` host-part and `TIMEBUF` were already sanitized on main, and `CLIENT` was the last param still doing a raw `strdup(cwalk->value)`. The vector is real — `CLIENT` flows into `hostdatadir = "$CLIENTLOGS/$hostname"` and then `fopen("$hostdatadir/$tstamp")`, so an unsanitized value escapes the client-logs tree and reads arbitrary files with the webserver's privileges. Wrapping it in `basename()` reduces the value to a single path component and neutralizes the traversal.

The change is idiomatically consistent with the sibling lines (`<libgen.h>` is included, so POSIX `basename()`; the value is used once and the result immediately `strdup`'d, so the static-storage/arg-mutation caveats don't bite). CI is green across all three build legs.

LGTM. I'll follow up separately on hardening the remaining `service`-bearing params as a defense-in-depth pass.

PR #147, opened 2026-06-10
web: add safe_basename() and route CGI path components through it (refs #146)

bonomani:
Closes an unauthenticated path traversal in the CGI `basename()` handling that #145/#146 left open, and hardens every request-derived path/protocol/shell value in `svcstatus`, `history` and `reportlog`. Refs #145, #146.

## The bug
`basename()` normalizes instead of validating (`../realhost` → `realhost`), so `svcstatus.cgi --historical` with `CLIENT=..` built a path above the client-log directory and disclosed a file — unauthenticated in a default install.

## The fix
Confinement helpers in `lib/cgi.c`, applied per parameter:
- **`safe_basename()`** — validates, never normalizes: `/`, `.`, `..`, `""` all reject.
- **`cgi_pathcomponent()`** — STRICT (`HOST`/`CLIENT`/`TIMEBUF`, both halves of `HOSTSVC`/`HISTFILE`): a single `CGI_NAMECHARS` (`[A-Za-z0-9:,._-]`) component or refused **whole**, never truncated. Both `/` and `\` excluded; the split service stays strict because `history.cgi` shells it to `tail`.
- **`cgi_component()`** — FREE-FORM for svcstatus `SERVICE` (a PCRE pattern) and section leaves: refused only on a control byte or traversal.
- **`cgi_hasctrl()`** — one definition of the injection-dangerous byte, shared by the helpers and the `SECTION` resolver.

This also closes a control-byte protocol-injection gap and a `tail` shell-injection, and **refuses rather than truncates** malformed input. Non-ASCII canonical hostnames are refused for now; the matching `hosts.cfg` parser enforcement + docs are tracked in #309.

## Tests
A C harness pins the confinement helpers directly; the svcstatus/history/reportlog tests drive the real CGIs against canary files and a fake xymond, asserting traversal / protocol+shell injection / malformed inputs are refused *by refusal status* (not just canary absence) while legitimate spellings (dotted columns, comma-IP hosts, section filters, aggregate `SERVICE` regexes) still serve.

### DIFF
--- a/lib/cgi.c
+++ b/lib/cgi.c
@@ -20,10 +20,34 @@ static char rcsid[] = "$Id$";
 #include <unistd.h>
 #include <limits.h>
 #include <errno.h>
+#include <libgen.h>
 
 #include "libxymon.h"
 
 #define MAX_REQ_SIZE (1024*1024)
+
+/*
+ * basename() that also neutralizes "." and "..".
+ *
+ * POSIX basename() strips directory components, so "../../etc/passwd"
+ * becomes "passwd" -- but it returns ".." for "..", and "." for "." or an
+ * empty/"/" input. A lone ".." pasted into a filesystem path is still a
+ * single-level traversal, so basename() alone is not enough to keep
+ * attacker-controlled CGI parameters from escaping their intended directory.
+ *
+ * This returns "" for those cases, leaving every other result untouched.
+ * Callers that strdup() the result get a safe single path component (or an
+ * empty string, which their existing validation rejects or which collapses
+ * harmlessly in the constructed path). Like basename(3), it may modify the
+ * string pointed to by path, so callers must not rely on path afterwards.
+ */
+char *safe_basename(char *path)
+{
+	char *b = basename(path);
+
+	if (!b || (strcmp(b, ".") == 0) || (strcmp(b, "..") == 0)) return "";
+	return b;
+}
 
 enum cgi_method_t cgi_method = CGI_OTHER;
 
--- a/lib/cgi.c
+++ b/lib/cgi.c
@@ -25,6 +25,125 @@ static char rcsid[] = "$Id$";
 
 #define MAX_REQ_SIZE (1024*1024)
 
+/*
+ * Validate a CGI value as a single path component. A validator, not a
+ * normalizer: '/' (even in "../realhost", which basename() would reduce to
+ * a legitimate name), ".", ".." and "" all yield "", everything else is
+ * returned unchanged. Callers MUST reject the empty result -- appended to a
+ * path it collapses onto the parent dir and still serves a file. path is
+ * never modified and the result is read-only; copy it (callers strdup()).
+ */
+char *safe_basename(char *path)
+{
+	if (!path) return "";
+	if (!*path || strchr(path, '/') ||
+	    (strcmp(path, ".") == 0) || (strcmp(path, "..") == 0)) return path + strlen(path);
+
+	return path;
+}
+
+/*
+ * The one byte set legal in a confined CGI hostname or column name: ASCII
+ * letters and digits plus the punctuation real xymon names use -- ':'
+ * (IPv6), ',' (the URL spelling of '.' in IP hosts, and literal-comma
+ * names), '.' (FQDNs and the host.service separator), and '_' '-'. Both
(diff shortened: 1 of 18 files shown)


SoundGoof:
## Findings

### Validate separators before calling `basename()`

[`safe_basename()`](https://github.com/xymon-monitoring/xymon/blob/e660c1a6f1c34e8c2ec6ab06fbc82bad729825aa/lib/cgi.c#L57-L66) checks `strchr(b, '/')` after `basename(path)` has removed preceding path components. Consequently, inputs documented as invalid are silently normalized:

- `CLIENT=../realhost` serves `realhost`
- `CLIENT=/realhost` serves `realhost`
- `TIMEBUF=dir/20260101` serves `20260101`

I reproduced each against the built CGI; they return exit code 0 and the legitimate file contents. Check `strchr(path, '/')` before calling `basename()`, then retain the checks for `"."`, `".."`, and empty values.

The regression test should also require malformed requests to return the refusal status or `"Invalid request"`. Its current canary-only assertion does not detect this normalization.

bonomani:
Rebased and squashed to a single commit; PR description rewritten to match the final state.

Incorporates the review feedback along the way: @SoundGoof's point that a slash-containing value must be *refused*, not normalized after `basename()` (so `safe_basename()` validates the raw value up front); keeping `SECTION`'s slash-bearing section filter working while confining the `service` copy that feeds access control and the histlog path; and confining `service` where it becomes a path component. Cleanups: shared `cgi_split_hostsvc()` helper, `safe_hostname()` reuses `uncommafy()`, tightened `SERVICE`/`HOSTSVC` whitelists.

`tests/testsuite` is green (19/19).

SoundGoof:
## Findings

### Document and enforce the hostname restriction

`safe_hostname()` introduces an ASCII-only hostname contract.
The current `hosts.cfg` parser is less strict. It accepts configured names containing characters such as `ä`, `ö`, or `å`.
These hosts will still load. The affected CGI programs will reject them.

I am from Sweden, so these characters are a realistic case for me. Still, restricting canonical hostnames to a documented ASCII character set is reasonable.
Unicode can remain available through the `NAME:` display-name tag.
Document the accepted character set in `hosts.cfg(5)`.
Mention the compatibility change in `RELEASENOTES`.

Apply the same validation in the `hosts.cfg` parser. Invalid canonical hostnames should fail during configuration loading, rather than later in the CGI programs.

I could not find another issue or pull request covering this policy, documentation, or parser validation.

bonomani:
Thanks — addressed both points.

**Refuse, don't normalize:** the CGIs now validate the raw value and refuse the whole thing on anything that isn't a single legal component (`../realhost` is refused, not served as `realhost`); the tests assert the refusal status, not just canary absence.

**ASCII hostnames:** host parameters are now confined to `[A-Za-z0-9:,._-]` and the rest refused, matching your direction (Unicode still via `NAME:`). The compatibility half — enforcing the same set in the `hosts.cfg` parser and documenting it in `hosts.cfg(5)`/`RELEASENOTES` — is tracked separately in #309.

SoundGoof (commented):
this looks good now, are we pushing this or waiting for #309 ?

bonomani:
@SoundGoof let's push this now — the traversal/injection fix stands on its own and doesn't need to wait. The #309 half (the canonical-hostname character set, enforced in the `hosts.cfg` parser + documented) is now its own PR #314, stacked on this one, so this merges first and #314 follows. It keeps your "same set on both sides" point: `XYMON_HOSTNAME_CHARS` is shared by the CGIs, the loader-side check and the ghost-name guard, and a bad canonical name is warned about (not dropped) once per config load, with `NAME:` for the display spelling.

---

Release-notes text to stage on the `Changes` branch at merge (kept out of this code-only branch):

**RELEASENOTES (4.3.31):**
> The svcstatus, history and reportlog CGIs now confine the host, service and section values from a request to a single safe path component, closing a path traversal and xymond-protocol injection reachable in a default install. A host whose canonical name (field 2 of hosts.cfg) contains characters outside letters, digits and ":,._-" is no longer reachable through these pages; give it an ASCII canonical name and carry the original spelling in the NAME: display tag.

**Changes:**
> * Confine CGI host/service/section request parameters to a safe path component (#146, #147)

SoundGoof (approved):
lgtm

Issue #106, opened 2026-05-27
Tracking: backport 4.4/devel features to main (per-commit checklist)


bonomani:
Backport tracker: 4.4/`devel` (tip `3686fbee`) → `main`. Vehicles: **#113/#114** (GCC warnings, supersede #60), **#1**. Part A's bundle closed unmerged; its items re-port from `devel`, so every hash here is a `devel` hash. Last full recheck 2026-07-03.

**Scope exclusion — IPv6/TLS:** anything IPv6/TLS-dependent is not backported as-is; to be re-analysed in depth before any integration. Excluded devel commits: `a48ca9ed5` (r7747 merge), `4fbbf2a57` (tcplib), `63dad1682` (v6 hosts.cfg parsing), `129d26fb2` (client v6/TLS + cert ACL), `69dfcf443` (xymonnet v6 test fix), `4ab01c439` (http/https send, tcplib-based — the feature is noted in the http-transport section), `370dbef5a` (docs).

### Rule

Identical on #29 and #106.

**Every mark has exactly one defining clause.** box → *Checkbox* · icon → *Status icon* · verdict → *Verdict* · `needs <id>` → *Sequencing* · `_(cond)_`, `PARKED`, `no longer applies cleanly` → *Marks that sit beside a verdict* · ⚠ → *⚠ is emphasis*. **A mark with no clause is not a mark:** define it, or stop using it.

**Scope.** #29 tracks **Terabithia patches**; #106 tracks **`devel` commits**. **One line = one artifact, always** — one patch, or one commit. **A bundle is a group, not a line**: give it a heading, put each member on its own line beneath it, and let the heading carry what is true of all of them (see *Grouping* and *Group headings*). Analysis true of only some members is written once, on one of them, and **referenced** from the others (`see <id>`) — never restated, per *No repetition*. Dependencies are expressed the same way — see *Sequencing*. The same underlying change may appear on both — as a patch on #29, as a commit on #106 — linked by cross-reference, never by ownership.

*Sub-items:* a bullet that only elaborates its parent (a step, a caveat, a sub-task) is **not** a separate artifact and needs no verdict, checkbox or cross-reference of its own.

**Delegation — one way only: #106 → #29.** When a `devel` commit has a Terabithia twin, #106 **delegates it to #29** (verdict 5 below defines the marker). The substantive verdict, its reason, the checkbox and every supporting detail then live on the #29 line. **#29 never delegates to #106.**

[...]

PR #151, opened 2026-06-10
svcstatus: stop the trends-page locator redirect loop with SKIPLOC (TBT 44)

bonomani:
Backport of J.C. Cleaver's fix for the trends-page redirect loop in
locator-based (distributed) setups.

**Problem**: when svcstatus renders the `trends` column for a host whose
RRDs live on another server, it queries xymond_locator and redirects the
browser there — but the receiving server runs the same locator logic and
redirects again, looping forever.

**Fix**: the redirect URL gains `&SKIPLOC=1`; the receiving CGI parses it
(strictly value `1`) and renders locally instead of consulting the
locator — a one-hop guard.

**Lineage**: 4.4 branch `a0822c27`, shipped downstream as Terabithia patch
`44` (`xymon.trendsloop.patch`, tracked in #29 bucket 5 / #106 Part A).
Cherry-picked onto current main; applies cleanly.

**History**: this PR originally carried two more 4.4 backports (unique
info/trends/clientlog icons, extra vmstat/iostatx trends graphs). Both are
patch-identical to commits already inside #46, so they were dropped here
to leave #46 as their single carrier — this PR now carries only the fix
that #46 does not contain.

### DIFF
--- a/web/svcstatus.c
+++ b/web/svcstatus.c
@@ -40,6 +40,7 @@ static char *accessfn = NULL;
 static char *hostname = NULL;
 static char *service = NULL;
 static char *tstamp = NULL;
+static char *skiploc = NULL;
 static char *nkprio = NULL, *nkttgroup = NULL, *nkttextra = NULL;
 static enum { FRM_STATUS, FRM_CLIENT } outform = FRM_STATUS;
 STATIC_SBUF_DEFINE(clienturi);
@@ -105,6 +106,9 @@ static int parse_query(void)
 		}
 		else if (strcasecmp(cwalk->name, "NKTTEXTRA") == 0) {
 			nkttextra = strdup(cwalk->value);
+		}
+		else if ((strcasecmp(cwalk->name, "SKIPLOC") == 0)   && cwalk->value && strlen(cwalk->value) && (strcmp(cwalk->value, "1") == 0)) {
+			skiploc = strdup(cwalk->value);
 		}
 		else if ((strcmp(cwalk->name, "backsecs") == 0)   && cwalk->value && strlen(cwalk->value)) {
 			backsecs += atoi(cwalk->value);
@@ -303,7 +307,7 @@ int do_request(void)
 		strncpy(timesincechange, "0 minutes", sizeof(timesincechange));
 
 		if (strcmp(service, xgetenv("TRENDSCOLUMN")) == 0) {
-			if (locatorbased) {
+			if (locatorbased && !skiploc) {
 				char *cgiurl, *qres;
 
 				qres = locator_query(hostname, ST_RRD, &cgiurl);
@@ -312,7 +316,7 @@ int do_request(void)
 				}
 				else {
 					/* Redirect browser to the real server */
-					fprintf(stdout, "Location: %s/svcstatus.sh?HOST=%s&SERVICE=%s\n\n",
+					fprintf(stdout, "Location: %s/svcstatus.sh?HOST=%s&SERVICE=%s&SKIPLOC=1\n\n",
 						cgiurl, hostname, service);
 					return 0;
 				}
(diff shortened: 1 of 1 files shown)


bonomani:
## Cross-tree provenance (verified)

Confirmed where these three commits sit relative to `devel` and the Terabithia
downstream set, so this is a clean one-directional backport:

**On `devel` — yes (origin).** All three are J.C. Cleaver's 4.4/`devel` work. The
SHAs here differ only because #46 recreated them off `main`; the *content* is on
`upstream/devel`:
- `wwwfiles/gifs/trends.gif` (+ info/clientlog icons) — present
- iostatx/vmstat trends entries in `xymonserver.cfg.DIST` — present
- the trends/locator handling in `svcstatus.c` — present

So this brings `main` in line with `devel` for these items (the kind of
"already on devel, not yet on main" backport #106 tracks), not a divergence.

**On Terabithia — no.** The Terabithia spec carries *adjacent but distinct*
trends/vmstat patches (`Patch44 trendsloop`, `Patch73 vmstat-aix`,
`Patch76 pipevmstatfork`, `Patch246 vmstat_terminated`) — none of which is the
icons, the iostatx-graphs-on-trends, or the locator-skip change here. These
commits are **not** part of the Terabithia set tracked in #29, so there's no
cross-tracking overlap or conflict.

Net: clean `devel → main` backport, independent of #29.

bonomani:
Rescoped: trimmed to the single skiploc commit (rebased onto current main, now `74c888023`). The two dropped commits (icons `a644b44f6`, vmstat/iostatx `392656b25`) are patch-identical to commits carried by #46 and will land there — verified by patch-id. TBT `44` (`xymon.trendsloop.patch`) continues to be tracked by this PR.

bonomani:
**Source verification that the bug is real and the fix closes it** (traced on current `main`):

The trends branch (`web/svcstatus.c:306`) redirects whenever the CGI runs locator-based:

```c
qres = locator_query(hostname, ST_RRD, &cgiurl);
...
fprintf(stdout, "Location: %s/svcstatus.sh?HOST=%s&SERVICE=%s\n\n", cgiurl, ...);
```

Two facts make this fatal rather than merely awkward:

1. `locatorbased` is unconditional (`svcstatus.c:774`) — any svcstatus started with `--locator=` always takes the redirect branch, and the frontend on the RRD-owning server must run with `--locator` anyway to serve the hosts it does *not* own.
2. There is no self-detection anywhere: `lib/locator.c`'s `locator_query()` returns the registered CGI URL for the owning server with no "is that me?" concept, and the caller redirects on any non-NULL answer.

Consequence: the loop needs no second server. On the RRD-owning server itself, a trends request queries the locator, receives **its own URL**, and redirects to itself — indefinitely. Every multi-server chain also terminates in that same self-loop on the owning server. In the normal locator-based configuration the trends column is therefore unreachable (browser aborts with "too many redirects").

The `SKIPLOC=1` guard makes every topology one-hop, including the self-redirect case: first request bounces once carrying the flag, the second renders locally where the RRDs are. Degradation is graceful — with stale locator data the old code loops forever, the new code renders an empty trends page. The parameter is strictly validated (value must be `1`), initialized and freed; all it can do is suppress one internal redirect. The other `locatorbased` site in svcstatus (~line 649, client-data fetch) is an internal fetch, not a browser redirect — no loop possible, correctly untouched.

Severity note: this only affects distributed locator-based installations — which is why a bug this severe survived since 2019 — but for those setups it is the difference between the trends page working and not working at all.

bonomani:
#46 closed unmerged today, so the two backports this PR dropped no longer have a carrier.

The History section above records that the unique info/trends/clientlog icons and the extra vmstat/iostatx trends graphs were dropped here because they were patch-identical to commits inside #46, "to leave #46 as their single carrier". That carrier is gone.

Nothing is lost from the backlog: both are tracked on #106 as `3c299d3f4` and `d0803b0c6`, and both are now unchecked, so they read correctly as owed. But the reason they are not in this PR no longer holds, and it would be easy to inherit that decision without noticing it had expired.

Either resolution is fine — re-add them here, or let them ride a later web/packaging PR from #106's Part A. Worth deciding rather than leaving implicit.

One practical note for whoever takes `3c299d3f4`: it adds three new `.gif` files, so it needs a binary-capable patch (`git format-patch --binary`, or a cherry-pick). A text-only diff drops them silently and the change reads as applied when it is not.

PR #152, opened 2026-06-10
do_rrd: remember an RRD exists instead of stat()ing it every update (TBT 214, devel f106d8840)

bonomani:
`do_rrd.c` ran a `stat()` before every update to see whether the RRD existed. From J.C. Cleaver's Terabithia patch: remember the answer per cache item, so the check costs one syscall per flush rather than one per reading.

The flag has to be cleared wherever the file can go, and the correction commit is about the places it was not:

| | |
|---|---|
| `drophost` / `renamehost` | discard and rename never reach `flush_cached_updates()`, so `updcache_host_op()` clears the flag for every item it touches — otherwise a re-added host skips the create and writes into nothing |
| a deletion nothing announces | an admin removing one RRD after a droptest, `moverrd.sh`, `rrdreconcile`, a restore. Only the failed write can notice, so the readings are kept and the next update re-stats, recreates and writes them |
| that retention | applies **only** when the file is gone. Kept for every error, a reading RRDtool will never accept sits at the front of the list and fails every flush behind it |
| the reading that just arrived | joins the retained ones — it was handed to the update that failed, and the caller frees it on return |

`tests/rrd/rrd-recreate-after-drop.sh` drives the real worker through a drophost and through an external deletion, the second across a fifo so **one** worker sees the file disappear: a fresh worker starts with no cached flag and takes the ordinary stat-and-create path, which was never the one at risk.

**Not pinned, and worth saying:** the retention itself has no test. The window it protects is inside a single cache cycle — the flag is cleared at the end of every flush — and driving it from outside means timing the cache, which makes a race rather than a check. The narrowing above is by inspection.

Two commits: J.C. Cleaver's original, then one correction. Suite on a built tree, with `rrdtool` present so the gated halves run: 0 failed, 0 skipped.

**Overlaps in `flush_cached_updates()`.** Three other open PRs edit that one function at ranges colliding with this one's `@@ -270,13`: **#154** (stacked on this, so it rebases with it), **#349** (`@@ -258,18` — deletes the `utimes()` at `do_rrd.c:270`) and **#387** (draft, `@@ -265,9` — re-gates that same `utimes`, and says it will drop the gating once #349 lands).

Nothing here is superseded by any of them: each removes a *different* syscall from the per-update path — one `stat` here, two more in #154, one `utimes` in #349 — so the effects compose. Only the hunks need sequencing; whichever lands first, the rest rebase.

### DIFF
--- a/tests/rrd/rrd-recreate-after-drop.sh
+++ b/tests/rrd/rrd-recreate-after-drop.sh
@@ -1,0 +1,136 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/rrd/rrd-recreate-after-drop.sh
+#
+# Guard for the file-exists cache in do_rrd.c (TBT 214): remembering that an
+# RRD is present must not outlive the file.
+#
+# create_and_update_rrd() stat()s the RRD before every update to decide whether
+# to create it. Caching that answer per update-cache item removes a syscall per
+# value, but the flag then has to be cleared whenever the file can have gone --
+# and the paths that make it go are exactly the ones that flush without going
+# through the ordinary update: a drophost deletes the host's tree, a renamehost
+# moves it, and rrdcacheflush{all,host} commit an item at an arbitrary time.
+#
+# Left set, the next update skips the create, rrd_update fails on a file that
+# is not there, and flush_cached_updates() has already discarded the cached
+# readings -- so the data is lost rather than merely late. That is worse than
+# the stat it saves, which is why this is asserted rather than reasoned about.
+#
+# The scenario is the cheapest one that reaches it: report, drop, report again.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+require_bin XYMOND_RRD "xymond/xymond_rrd"
+
+work=$(mktempdir)
+ts=$(date +%s)
+mkdir -p "$work/tmp" "$work/rrd" "$work/home/etc"
+: > "$work/home/etc/analysis.cfg"
+: > "$work/hosts.cfg"
+
+status_msg() {  # status_msg <msg-timestamp>
+	printf '@@status|%s|127.0.0.1|origin|testhost|disk|%s|green||green|%s|0||0||%s|0|linux|/\n' \
+		"$1" "$(($1+1800))" "$ts" "$ts"
+	printf 'disk report\n'
+	printf '/dev/sda1 1000000 400000 600000 40%% /\n'
+	printf '@@\n'
+}
+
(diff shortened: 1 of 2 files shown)


bonomani:
## Self-review (deep analysis) — one bug found & fixed

**Fixed:** the `fileok` reset was placed after the flush error-return, so it only
ran on a *successful* flush. That broke the original per-update self-healing: if
an RRD file was removed/rotated while `fileok==1`, every flush would fail and the
file would never be re-stat'd or recreated — the RRD would be stuck until xymond
restarts. Now `fileok` is reset **after any flush attempt** (before the error
return), so a failed flush recreates the file on the next update. (The Terabithia
original had the same placement; this corrects it.)

**Checked and cleared:**
- *Stale `st`*: within `create_and_update_rrd`, `st` is not read after the cached
  guard. The earlier `stat()` (line ~292) is on the host **directory**
  (`rrddir/hostname`), a different path; the other `stat()` is in another function.
  Skipping the file stat is safe.
- *NULL deref*: both cache-lookup branches set `cacheitem` non-NULL (calloc /
  `xtreeData`) before the guard, matching the existing `cacheitem->tpl` deref just
  above. No new NULL risk.
- *Concurrency*: `fileok` is the same shared cache-item state as `valcount`/`vals[]`
  and inherits the cache's existing thread-safety assumptions — no new race class.

**Scope clarification:** there are two `stat()`s per update — the host **directory**
(line ~292) and the RRD **file** (the guard). This change caches only the **file**
check, so it removes the per-file stat (~one per `CACHESZ` updates instead of one
per update); the directory stat is unchanged.

bonomani:
A verification pass before merge found four; three taken, one is a limit I would rather state than paper over.

**The retention applied to every update error, and only one justified it.** A reading RRDtool will never accept — one below the minimum step — sat at the front of the list and failed every flush behind it, so the readings that followed were lost rather than the one that was already bad. Kept now only when the file is missing, which is the case the comment always described; dropped otherwise.

**The reading that had just arrived was not among those kept.** It was handed to the update that failed and freed by the caller on return, so the retry carried the older values and lost the new one. It joins them.

**The external-deletion case tested the wrong path.** It started a fresh worker after `rm`, and a fresh worker has no cached flag — it takes the ordinary stat-and-create path, which was never broken. It drives one worker across the deletion through a fifo now.

**What I could not pin, said plainly:** the retention itself has no test. The window it protects is inside a single cache cycle — the flag is cleared at the end of every flush — and driving that window from outside means timing the cache, which would make the test a race rather than a check. The narrowing is by inspection; the mutation that would prove it is the one I do not have.

The rename-path finding (old-host readings surviving in a cache item keyed by the old host) follows from the same retention and is narrowed by it, but I have not exercised it either.

Rebased on `f2b28fb75`. Suite on a built tree, with `rrdtool` present so the gated halves run: 74 passed, 0 skipped, 0 failed.

PR #154, opened 2026-06-11
do_rrd: check the host RRD directory only when creating a file (closes #153, TBT 187, devel f106d8840)

### LINKED ISSUE #153: do_rrd: the per-host directory stat() still runs on every RRD update
## Summary

`create_and_update_rrd()` in `xymond/do_rrd.c` does **two** `stat()` calls per RRD
update:

1. the per-host **directory** `rrddir/hostname` — to create it if missing
   (`xymond/do_rrd.c`, the `sprintf(filedir, "%s/%s", rrddir, hostname); if (stat(filedir, &st) == -1) mkdir(...)` block);
2. the per-RRD **file** `rrddir/hostname/rrdfn` — to create the RRD if missing.

PR #152 caches (2) in the update-cache item (`fileok`) so the file stat runs about
once per `CACHESZ` updates instead of every update. But (1), the **directory**
stat, still runs on **every** update.

## Why it matters

On a busy server a host has many RRD files, all updated continuously. The
directory `rrddir/hostname` is shared by all of that host's RRDs, yet it is
`stat()`'d once per *file* update — i.e. far more often than necessary.

## Possible improvement

Cache the directory-exists check the way #152 caches the file check, but keyed
**per host** (the directory is shared across a host's RRDs) rather than per
cache-item. As with #152, self-healing must be preserved: re-check periodically /
recreate the directory if a write later fails because it went missing.

## Notes

- Pure performance; behaviour-preserving if done carefully.
- Lower value than #152's per-file cache (one directory vs many files), and the
  per-host keying makes it a bit more involved — worth doing only if profiling
  shows the directory stat is significant.

Surfaced during the deep review of #152.

bonomani:
Stacked on #152, which caches the RRD-file-exists check. This is issue #153: the host directory was checked on **every** update, when it only has to exist where a file is about to be created — if the RRD is there, so is its directory.

From J.C. Cleaver's Terabithia patch, plus one correction commit.

| | |
|---|---|
| the host-directory path | built with `sprintf` into a static; `rrddir` and the hostname can overflow between them. `snprintf` into a local now, and the refusal the replaced route had moves with it |
| the cached file-exists flag | cleared wherever the file can go — the update path, `drophost` and `renamehost` through `updcache_host_op()`, and both `rrdcacheflush*` entry points |

`tests/rrd/rrd-recreate-after-drop.sh` drives the real worker through a drophost and through a deletion nothing announces, the second across a fifo so **one** worker sees the file disappear while holding the cached answer. Three things had to be right for that case to mean anything, and none of them were:

- it waited for the RRD to exist after one reading, but the file was already there from the case above — the wait returned at once and the deletion could land before the worker had read anything;
- the readings after the deletion have to reach a flush, where the missing file is noticed, and then an update that recreates it. `CACHESZ` is 12; two readings sat in the cache until shutdown;
- every message carried the same sequence number, so a worker fed more than one dropped the rest as duplicate updates — logged under `--debug` and nowhere else. `status_msg()` numbers them, which any test feeding a live worker will need.

Verified by mutation: with the flag-clearing removed from `flush_cached_updates()`, the case fails. It did not, in any earlier version.

Four commits: two are #152's underneath, two are this PR's — J.C. Cleaver's original and one correction. Suite on a built tree, `rrdtool` absent from `PATH`: 0 failed, 0 skipped.

**Overlaps in `flush_cached_updates()`.** Besides #152 underneath, two other open PRs edit that function at ranges colliding with the `@@ -270,13` hunk this inherits: **#349** (`@@ -258,18` — deletes the `utimes()` at `do_rrd.c:270`) and **#387** (draft, `@@ -265,9` — re-gates that same `utimes`, and says it will drop the gating once #349 lands).

Neither supersedes this: the two `stat()` calls removed here and the `utimes()` removed by #349 are different syscalls on the same per-update path, so the effects compose. Only the hunks need sequencing.

### DIFF
--- a/tests/rrd/rrd-recreate-after-drop.sh
+++ b/tests/rrd/rrd-recreate-after-drop.sh
@@ -1,0 +1,136 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/rrd/rrd-recreate-after-drop.sh
+#
+# Guard for the file-exists cache in do_rrd.c (TBT 214): remembering that an
+# RRD is present must not outlive the file.
+#
+# create_and_update_rrd() stat()s the RRD before every update to decide whether
+# to create it. Caching that answer per update-cache item removes a syscall per
+# value, but the flag then has to be cleared whenever the file can have gone --
+# and the paths that make it go are exactly the ones that flush without going
+# through the ordinary update: a drophost deletes the host's tree, a renamehost
+# moves it, and rrdcacheflush{all,host} commit an item at an arbitrary time.
+#
+# Left set, the next update skips the create, rrd_update fails on a file that
+# is not there, and flush_cached_updates() has already discarded the cached
+# readings -- so the data is lost rather than merely late. That is worse than
+# the stat it saves, which is why this is asserted rather than reasoned about.
+#
+# The scenario is the cheapest one that reaches it: report, drop, report again.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+require_bin XYMOND_RRD "xymond/xymond_rrd"
+
+work=$(mktempdir)
+ts=$(date +%s)
+mkdir -p "$work/tmp" "$work/rrd" "$work/home/etc"
+: > "$work/home/etc/analysis.cfg"
+: > "$work/hosts.cfg"
+
+status_msg() {  # status_msg <msg-timestamp>
+	printf '@@status|%s|127.0.0.1|origin|testhost|disk|%s|green||green|%s|0||0||%s|0|linux|/\n' \
+		"$1" "$(($1+1800))" "$ts" "$ts"
+	printf 'disk report\n'
+	printf '/dev/sda1 1000000 400000 600000 40%% /\n'
+	printf '@@\n'
+}
+
+# HOSTSCFG and an analysis.cfg of our own: without them the worker connects
+# out to a live xymond on 127.0.0.1:1984 for every message and, failing that,
+# reads the machine's real configuration instead of this fixture.
+run_worker() {  # run_worker <rrddir>
+	env XYMONHOME="$work/home" XYMONTMP="$work/tmp" HOSTSCFG="$work/hosts.cfg" \
+		"$XYMOND_RRD" --rrddir="$1"
+}
+
+rrd="$work/rrd/testhost/disk,root.rrd"
+
+# ---- a report creates the RRD ----------------------------------------------
+
+status_msg "$ts" | run_worker "$work/rrd" >"$work/w1.log" 2>&1 \
+	|| { cat "$work/w1.log" >&2; fail "the worker rejected the first report"; }
+# fail, not skip: require_bin above already covers "no worker", and this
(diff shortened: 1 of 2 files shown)


bonomani:
A verification pass before merge, run twice — and the first run is the finding.

Asked with the diff alone, it opened no file and answered "no defects" in 4.7k tokens. Asked again with the reading made the deliverable — *list every site that deletes, renames or moves an RRD, say whether it clears the flag and from which line* — it ran 40 commands, read 973 lines of `do_rrd.c`, and produced the list. Same code, same reviewer.

The list itself is clean: `xymon_rrd_update` clears at `do_rrd.c:318`; `drophost` and `renamehost` via `updcache_host_op` at `:655`; both `rrdcacheflush*` entry points through `:318`; `xymon_rrd_create` sets the flag rather than clearing it, which is correct. `build_rrd_filedir()` has two callers and both act on a refusal — `-1` from `create_and_update_rrd()`, `0` from `rrddatasets()`.

**The finding is against my test, and it is right.** The external-deletion case waits for `[ -f "$rrd" ]` after feeding one message through the fifo — but the file already exists from the previous case, so the wait returns instantly and the deletion can land before this worker has read anything. Then it creates normally, and the case passes with the flag-clearing reverted. If the timing goes the other way, only two more readings follow, below `CACHESZ`, so the flush lands at shutdown with no later update to recreate. Either way the case is timing-dependent.

**I have not fixed it, and I would rather say so than push something that looks fixed.** Deleting first and feeding 16 readings — enough to force a mid-stream flush and then a recreate — makes the case fail on current code, and I could not establish why: the worker reads the messages (it processes the shutdown that follows them) but reports one pending update at flush, not sixteen. That is either a real defect this test was hiding or a fault in how I am driving the worker, and I have not determined which. The test is back to the version that passes, and this stays open.

Rebased onto the current #152 (`1db70908a`), not onto main: it stacked the pre-review version of that PR. Four commits, two of them J.C. Cleaver's. Suite on a built tree, `rrdtool` absent from `PATH`: 74 passed, 0 skipped, 0 failed.

bonomani:
Correcting my last comment: I said I could not tell whether the failure was a real defect or a fault in how I drove the worker. It was my driving, and I can now say why.

Every message the test sends carries the same sequence number. A worker fed more than one drops the rest as duplicate updates — `do_rrd.c:499`, logged under `--debug` and nowhere else. So my sixteen readings arrived as one, and the RRD was never recreated. With distinct numbers: three flushes, no duplicates, recreated.

That third fault is what hid the other two:

- the case waited for the RRD to exist after feeding one reading, but the file was already there from the case above — the wait returned at once, and the deletion could land before the worker had read anything. It is deleted first now, so waiting for it to come back is what proves *this* worker created it.
- the readings after the deletion have to reach a flush, where the missing file is noticed, and then an update that recreates it. `CACHESZ` is 12; two readings sat in the cache until shutdown and were never written back. Sixteen now.

**Verified by mutation, which is what none of the previous versions could do:** with the flag-clearing removed from `flush_cached_updates()`, the case fails.

`status_msg()` numbers its messages, which any future test feeding a live worker will need.

Suite on a built tree, `rrdtool` absent from `PATH`: 74 passed, 0 skipped, 0 failed.

PR #157, opened 2026-06-11
client/darwin: fix APFS filesystem filter (false disk/inode PANIC) (closes #156)

### LINKED ISSUE #156: macOS client: filesystem filter inverted on APFS -> devfs/synthetic volumes cause false disk/inode PANIC
## Summary

On modern macOS (APFS, Catalina+), the Darwin client's `df`/`inode` filesystem
filter selects **nothing**, so `df` falls back to listing **all** filesystems —
including `devfs` (`/dev`) and `auto_home`, which are always **100% used** and
trip a false **disk/inode PANIC**.

## Root cause — the filter is inverted by APFS

`client/xymonclient-darwin.sh` builds the list with:

```sh
FILESYSTEMS=`mount | sed -E '/[\( ](nobrowse|afs|read-only)[ ,\)]/d;s/^.* on (.*) \(.*$/\1/'`
```

It drops `nobrowse`/`read-only` volumes, assuming those are synthetic. But on
APFS the **real** volumes carry exactly those flags:

```
/dev/disk3s1s1 on /                         (apfs, sealed, local, read-only, journaled)   <- read-only
/dev/disk3s5  on /System/Volumes/Data       (apfs, local, journaled, nobrowse, ... )       <- nobrowse  (the real data volume!)
devfs         on /dev                        (devfs, local, nobrowse)
... all /System/Volumes/* are nobrowse ...
```

**Every line matches**, so `FILESYSTEMS` is empty → `df -P -H`/`df -P -i` (no
args) list everything → `devfs` (100%) and `map auto_home` (100%) PANIC.

Observed live (macOS Tahoe), `disk` and `inode` both red:
```
/dev (100% used) has reached the PANIC level (95%)      # disk
devfs   691   691   0   100%   /dev                      # inode
```

## Scope: Darwin only

The other 7 clients exclude pseudo-FS natively and are unaffected:

| Client | filter | excludes devfs? |
|---|---|---|
| Linux | `df -Pl -x` (nodev types) | yes |
| FreeBSD | `df -t nonfs,…,devfs,…` | yes (explicit) |
| NetBSD/OpenBSD | `df -t nonfs,kernfs,procfs,…` | yes |
| OSF1 | `df -t noprocfs` | yes |
| Solaris/AIX | rootfstype logic / `df -Ik` | yes |
| **Darwin** | `mount\|sed` browse/read-only heuristic | **no — broken** |

macOS is the only client whose `df` has no type-exclusion flag, so it's the only
one using the brittle `mount` heuristic.

## Second bug

In the `[inode]` loop, subsequent filesystems are fetched with `df -P -H`
(blocks) instead of `df -P -i` (inodes) — masked today only because the list is
empty so the loop never runs.

## Fix direction

Select by filesystem **type** (real local disks: `apfs hfs exfat msdos`) and drop
the synthetic `/System/Volumes/*` internals, keeping `/` and
`/System/Volumes/Data`. Make it overridable via the same `XYMONCLIENT_FS_*_TYPES`
env vars #96 introduces for Linux, and fix the `-H`→`-i` inode bug.

Related: #49, #96 (Linux filesystem-filter configurability).

bonomani:
## Summary

Fixes the macOS (Darwin) client reporting **all** filesystems — including `devfs`
(`/dev`) and `auto_home`, which are always 100% used — so `disk` and `inode` go to
a false **PANIC**. Closes #156.

## Root cause

`client/xymonclient-darwin.sh` filtered the `df`/`inode` list by dropping
`nobrowse`/`read-only` volumes, assuming those are synthetic. On modern macOS
(APFS, Catalina+) the **real** volumes carry exactly those flags:

```
/dev/disk3s1s1 on /                   (apfs, sealed, local, read-only, journaled)   <- read-only
/dev/disk3s5  on /System/Volumes/Data (apfs, local, journaled, nobrowse, ...)        <- nobrowse  (real data vol)
devfs         on /dev                 (devfs, local, nobrowse)
... every /System/Volumes/* is nobrowse ...
```

Every line matched, so the list came out **empty** → `df -P -H`/`df -P -i` (no args)
listed everything → `devfs` (100%) and `map auto_home` (100%) PANIC.

## Fix

- Select by filesystem **type** (`apfs hfs exfat msdos`) — the approach the BSD
  clients already use — and drop the synthetic `/System/Volumes/*` internals,
  keeping `/` and `/System/Volumes/Data`.
- Overridable via the same `XYMONCLIENT_FS_INCLUDE_TYPES` / `XYMONCLIENT_FS_EXCLUDE_TYPES`
  env vars #96 introduces for the Linux client (shared contract; default = correct).
- Fall back to `/` rather than the unfiltered listing if nothing matches.
- Fix a latent bug: the `[inode]` loop fetched later filesystems with `df -P -H`
  (blocks) instead of `df -P -i` (inodes) — masked only because the list was empty.

## Validation

Tested against a real macOS Tahoe `mount`/`df` layout (the reporter's Mac mini):
the filter now yields exactly `/` and `/System/Volumes/Data`; `devfs`/`auto_home`
and the redundant synthetic volumes are gone, so `disk`/`inode` report the real
data volume (61%) and clear. `sh -n` clean. Scope is Darwin-only — the other 7
clients exclude pseudo-FS natively and are untouched.

Related: #49, #96.

### DIFF
--- a/client/xymonclient-darwin.sh
+++ b/client/xymonclient-darwin.sh
@@ -26,10 +26,26 @@ uptime
 echo "[who]"
 who
 
-FILESYSTEMS=`mount | sed -E '/[\( ](nobrowse|afs|read-only)[ ,\)]/d;s/^.* on (.*) \(.*$/\1/'`
+# macOS `df` has no filesystem-type filter, and modern APFS mounts the system
+# root read-only and the data volume nobrowse - so the old "skip nobrowse/
+# read-only" heuristic dropped the REAL volumes too, leaving an empty list, so
+# df then listed everything including devfs (/dev) at 100% -> false disk/inode
+# PANIC. Select by filesystem TYPE instead (real local disks) and drop
+# everything under /System/Volumes/ except the Data volume - nested internals
+# like /System/Volumes/Update/SFR/mnt1 and the Preboot cryptexes are apfs too.
+# Overridable via XYMONCLIENT_FS_INCLUDE_TYPES/EXCLUDE_TYPES in xymonclient.cfg.
+typepat=""
+for t in apfs hfs exfat msdos $XYMONCLIENT_FS_INCLUDE_TYPES; do
+	for x in $XYMONCLIENT_FS_EXCLUDE_TYPES; do [ "$t" = "$x" ] && t=""; done
+	[ -n "$t" ] && typepat="$typepat${typepat:+|}$t"
+done
+FILESYSTEMS=`mount | sed -nE "s#^.+ on (/.*) \\((${typepat})[,)].*#\\1#p" | awk '!/^\/System\/Volumes\// || /^\/System\/Volumes\/Data$/'`
+# Never fall back to the unfiltered df listing if the filter itself failed -
+# but honor an explicit exclusion that legitimately empties the list.
+[ -z "$FILESYSTEMS" ] && [ -z "$XYMONCLIENT_FS_EXCLUDE_TYPES" ] && FILESYSTEMS=/
 
 echo "[df]"
-(IFS=$'\n'
+[ -n "$FILESYSTEMS" ] && (IFS=$'\n'
  set $FILESYSTEMS
  df -P -H $1; shift
  while test $# -gt 0
@@ -39,12 +55,12 @@ echo "[df]"
  done) | column -t -s " " | sed -e 's!Mounted *on!Mounted on!'
 
 echo "[inode]"
-(IFS=$'\n'
+[ -n "$FILESYSTEMS" ] && (IFS=$'\n'
  set $FILESYSTEMS
  df -P -i $1; shift
  while test $# -gt 0
  do
-   df -P -H $1 | tail -1 | sed 's/\([^0123456789% ]\) \([^ ]\)/\1_\2/g'
+   df -P -i $1 | tail -1 | sed 's/\([^0123456789% ]\) \([^ ]\)/\1_\2/g'
    shift
  done) | awk '
 NR<2{printf "%-20s %10s %10s %10s %10s %s\n", $1, "itotal", $6, $7, $8, $9} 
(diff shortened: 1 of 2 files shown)


bonomani:
Superseded by #178, merged as 114008f57. Closing rather than rebasing.

Both patches fix #156, by different routes: this one selects by filesystem type (`apfs hfs exfat msdos`) and drops everything under `/System/Volumes/` except `Data`; #178 filters the `mount(8)` list on the `nobrowse`/`read-only` attributes and exempts the volume Apple flags `root data`.

#178 is the one to keep. It reaches the same result, adds the `XYMONCLIENT_FS_*` controls shared with the Linux, FreeBSD, NetBSD and OpenBSD clients (#170), and carries a regression test built from a real Apple-silicon mount table — which this PR does not have. Its assertions cover exactly what #156 reported: `devfs` dropped, the autofs map dropped, the data volume kept despite `nobrowse`, the sealed root dropped.

The two conflict on `client/xymonclient-darwin.sh` in the same `[df]` section, so resolving would have meant choosing one implementation anyway.

**Correction to an earlier version of this comment**, which suggested keeping this branch as a fallback if the attribute filter ever proved fragile. I have since run both filters over a mount table from a current Mac, and the type-based approach here has two blind spots that are present-day, not hypothetical:

| mount | #178 (attributes) | this PR (types) |
|---|---|---|
| external FAT/exFAT/NTFS, type `lifs` | kept | **dropped** |
| macFUSE volume, type `fusefs` | kept | **dropped** |
| sealed read-only `/` | dropped | kept — duplicates the same APFS container as `/System/Volumes/Data` |

macOS Ventura and later mount external media through LiveFiles as type `lifs`, which is not in the `apfs hfs exfat msdos` allowlist here, so a USB stick on a current Mac would silently go unmonitored — the mirror image of the bug this was written to fix. Same for NTFS-3G volumes, which appear as `fusefs`.

The weakness I ascribed to the attribute filter is real but conditional: if Apple stopped flagging the data volume `root data`, #178 would lose it while this branch would still find it by path. That is a future maybe against two current misses, and the merged version has a test that would catch the flag change on the next fixture update, whereas a new filesystem type triggers nothing at all.

So this is not a fallback to reach for as-is. If the attribute filter ever needs replacing, the idea is worth revisiting — the allowlist is not.

PR #158, opened 2026-06-11
xymonnet: support ntpdig/sntp/chronyd for the ntp test

### LINKED ISSUE #27: [Xymon] Patch: ntpdate option -p deprecated
```
Hi!

In Debian ntpdate from ntpsec package throws a warning that "-p is no
longer supported".  So it may be a good idea to remove the "-p 1" from
NTPDATEOPTS="-u -q -p 1"                        # Standard options to ntpdate
in xymond/etcfiles/xymonserver.cfg.DIST.

The attached patch solves this issue.

Greetings
Roland
-------------- next part --------------
Description: ntpdate obsoleted the "-p" option
Author: Carsten Leonhardt <[leo at debian.org](http://lists.xymon.com/mailman/listinfo/xymon)>
Bug-Debian: https://bugs.debian.org/1057044
Forwarded: no
Last-Update: 2023-11-28

--- a/xymond/etcfiles/xymonserver.cfg.DIST
+++ b/xymond/etcfiles/xymonserver.cfg.DIST
@@ -133,7 +133,7 @@
 FPING="@FPING@"					# Path and options for the ping program.
 FPINGOPTS="-Ae"					# Standard options to fping/xymonping
 NTPDATE="ntpdate"				# Path to the 'ntpdate' program
-NTPDATEOPTS="-u -q -p 1"			# Standard options to ntpdate
+NTPDATEOPTS="-u -q"				# Standard options to ntpdate
 TRACEROUTE="traceroute"                         # How to do traceroute on failing ping tests. Requires "trace" in hosts.cfg .
 TRACEROUTEOPTS="-n -q 2 -w 2 -m 15"		# Standard options to traceroute
 XYMONROUTERTEXT="router"			# What to call a failing intermediate network device.
```

bonomani:
Closes #27.

## Problem

ntpsec (default on Debian/Ubuntu, also Fedora/Alpine) ships `ntpdate` only as a wrapper around `ntpdig`, and the wrapper warns on Xymon's default `NTPDATEOPTS="-u -q -p 1"`; several current platforms package no `ntpdate` at all, so the ntp test is unusable out of the box. Separately, the offset the ntp test feeds into the `ntpstat` graph has always been stored in the wrong unit.

## Changes

- **Tool autodetection** in `run_ntp_service()`: first installed tool from a priority list, default **`ntpdig → ntpdate → sntp → chronyd -Q`**. ntpdig is probed first because ntpsec's `ntpdate` is just a wrapper around it (per @RolandRosenfeld); classic-ntp systems keep using ntpdate as before. `chronyd -Q` is query-only and unprivileged. Probing requires an executable *regular file* — `access(X_OK)` alone also accepts directories.
- **`NTPTOOL`** — admin-defined `|`-separated priority list; the first installed entry wins, so a single name forces a tool (e.g. `NTPTOOL="ntpdate|ntpdig"`). Unknown entries warn and skip. Legacy `SNTP`-forces-sntp kept (deprecated in favour of `NTPTOOL="sntp"`). New `NTPDIG`/`CHRONYD` (+ `*OPTS`) settings; `SNTP`/`SNTPOPTS` documented for the first time.
- **sntp invocation repaired**: timeout now passed as `-u` — classic ntp.org sntp (4.2.8) spells the unicast-response timeout `-u` and has no `-t`; the old `-d` was a delay only in ancient msntp and is a no-arg debug flag in everything current.
- **ntpstat RRD parser** (`do_net.c`): learns ntpdig / sntp 4.2.7+ / macOS sntp / chronyd output. The ntpsec `ntpdate` wrapper emits the ntpdig format, which the old branch (matched on the word "ntpdate") silently dropped — so the graph died on every ntpsec system. The dead pre-4.2.7 sntp branch is removed (the `+/-` parser covers it, detached sign included).
- **Offset unit fixed**: all these tools report **seconds**, but the DS is `offsetms` (graph labelled "Offset (ms)") — the network ntp test had understated drift 1000×. Now converted to ms before storage.
- **Out-of-bounds read fixed** (`do_ntpstat.c`): the `offset=` handoff from `do_net.c` *starts* the message, but the parser read `*(p-1)` — one byte before the buffer — and usually rejected it, so the offset never reached the RRD. The token is now accepted at message start.
- **Tests**: `tests/rrd/ntp-rrd-handoff.sh` compiles the real `do_net.c` + `do_ntpstat.c` (same include order as `xymond/do_rrd.c`) against stub RRD plumbing and drives every tool format through the full parser → handoff → RRD chain (ASan/UBSan when available); `tests/network/ntp-cmd.sh` pins tool selection + the per-tool timeout flags. Both follow the #101 suite contract (auto-discovered by `tests/testsuite`, quiet on success).
- **Docs**: `xymonnet(1)`, `xymonserver.cfg(5)`, `hosts.cfg(5)` (and synced HTML) updated.

## Behavior changes a reviewer should know

- **One-time 1000× step on existing `ntpstat` graphs** — old points were raw seconds, new ones are correct ms. This is the long-standing unit bug being fixed, but the step is visible on any site graphing the network ntp test.
- **ntpsec systems switch from the ntpdate wrapper to ntpdig** (same protocol/output, one less shell exec).
- **`SNTPOPTS` default `-u` → `""`** — the driver now supplies sntp's timeout as `-u <seconds>` itself (above), so the bare `-u` default is dropped to avoid duplicating it.

## Platform coverage (verified in containers / on real hardware)

| Platform | Tool used |
|---|---|
| FreeBSD, NetBSD; OpenBSD (`pkg_add ntp`) | ntpdate / sntp (base / ports) |
| Ubuntu 20/22, Debian 11, Arch, CentOS 7, Amazon Linux 2, openSUSE Leap 15 | ntpdate (classic) |
| Debian 12/13, Ubuntu 24.04, Fedora, EL 8/9/10 + EPEL | ntpdig (Debian 13 dropped classic ntp) |
| Alpine | ntpdig (ntpsec, no wrapper) |
| macOS | sntp (Apple) |
| Amazon Linux 2023, openSUSE Leap 16 / Tumbleweed | chronyd -Q |

Live-verified on **Debian 13** (ntpdig output/exit codes vs pool.ntp.org), **Amazon Linux 2023** (`chronyd -Q -t 10 'server … iburst'` exit 0 unprivileged, exit 1 on TEST-NET), and **macOS Apple Silicon** (`/usr/bin/sntp` `+0.009083 +/- 0.013184 …` — no timestamp prefix, parser anchors on `+/-`; failure exits 69). `ntp-rrd-handoff.sh` passes 12/12 (classic ntpdate, legacy sntp ± detached sign, ntpdig, Apple sntp, chronyd, ntpsec wrapper, ntpq/LARRD client formats, separator edge cases, failure inputs that must store nothing — including seconds→ms values); `ntp-cmd.sh` covers selection (default order under various installed sets, single-name forcing of missing/present tools, custom order, unknown entries, legacy `SNTP`, case-insensitivity). `xymonnet` and `do_rrd.o` compile clean with no new warnings.


RolandRosenfeld:
This PR looks really good to me (especially the expansion of the output parser and the documentation of the various tools and their preferred options), but I'm not fully sure, whether defining so much NTP based variables in combination with an algorithm, that defines their internal priority, is the correct way to go.

Maybe this is triggered because you prefer ntpdate over ntpdig in run_ntp_service() priority list.
Since on Debian ntpdate is only a shell wrapper over ntpdig, I'd prefer to use ntpdig if it's available, to avoid the useless ntpdate shell wrapper call.

I wasn't aware that switching between ntpdate and sntp was possible before.
But my intuition says, that it would be more strait forward to only have two variables NTP and NTPOPTS (or NTPDATE and NTPDATEOPTS for backward name compatibility), which can carry all variants of commands:
- ntpdate -u -q
- ntpdig
- sntp
- chronyd -Q
So it's explicitly defined, what variant is used to handle ntp.

On the other hand your PR brings the advantage, that the ntp mechanism (magically) always works (with any of these mechanisms), but there is no clean way to enforce one mechism (except SNTP) for example to prefer ntpdig over ntpdate).

What do you think about a mechanism, which allows deciding on build time what variant to use by default (it can be overridden by changing xymonserver.cfg by the xymon admin lateron).  Maybe something like the one for fping vs. xymonping?

Don't get me wrong, I'm not sure, what the better way is, but at the moment I tend to got with less variables or alternatively additional one, which allows to force one mechanism.

Comments welcome...

bonomani:
Thanks — both points taken, and implemented:

1. **ntpdig now leads the default order** (`ntpdig|ntpdate|sntp|chronyd`). You're right that on ntpsec systems the wrapper is a pointless indirection; and since classic ntp and ntpsec conflict as packages, ntpdig-first never changes the tool on classic-ntp systems.
2. **Forcing (and ordering) is now explicit**: `NTPTOOL` is a `|`-separated priority list — the first installed tool wins, so `NTPTOOL="chronyd"` forces one tool and `NTPTOOL="ntpdate|ntpdig"` defines your own preference order. The legacy "set `SNTP` to force sntp" behavior is kept, now deprecated in favor of `NTPTOOL="sntp"`.

On the single `NTP`/`NTPOPTS` variable: I went the other way because the invocation shapes differ per tool — chronyd needs `-Q … -t N 'server HOST iburst'` with the host inside a quoted directive, ntpdig/sntp take `-t N HOST`, classic ntpdate takes the host bare. A single command variable can't express where the host and timeout go without inventing a template syntax, which seemed worse than per-tool path/opts pairs. `NTPTOOL` hopefully gives the explicitness you wanted without that cost.

On build-time selection: fping fits that model because it's probed once at configure time (`build/fping.sh`) and baked into the generated `xymonserver.cfg` — both candidates are known when the package is built, and xymonping is bundled so there's always a fallback. The NTP tools are external packages that vary per host and can change after build, so a pre-built xymon package can't bake in the right answer — a package built where ntpsec is installed would hardcode ntpdig, then land on a chrony-only host. Runtime probing preserves the tool choice everywhere it can: classic-ntpdate systems keep running ntpdate, and an explicit `SNTP` setting still forces sntp. The one deliberate switch is ntpsec systems going from the ntpdate wrapper to ntpdig directly — same query, one less shell exec, and it fixes the offset graph that the wrapper's output format silently broke. A distro package can still pin its preferred default by shipping `NTPTOOL` in its `xymonserver.cfg` (e.g. `NTPTOOL="ntpdig|chronyd"`).

jezzaaa:
I'm late to the party here, but I thought I'd add a comment nevertheless, as it may influence future direction. And I might not have the right facts too, so I'd be happy to be corrected.

This patch seems to add complexity within Xymon C code (to auto-detect and parse different NTP clients) without adding (IMHO) fundamental features relating to monitoring, and also increasing complexity in the handling of a manually-defined NTP client. I'm not criticising these new features or their implementation. I'm just noting that the benefit they provide may not warrant the added complexity.

Additionally, it seems to me that the NTP protocol is simple enough that the NTP probe could be embedded into the Xymon code, and do away with the added complexity of scanning for a suitable NTP client, giving it the correct path and parameters, and parsing the result depending on the flavour. This would bring the relatively simple (s)ntp protocol inline with far more complex protocols that are implemented internally by xymonnet, such as the simple TCP services (telnet, snmp, etc), TLS services (telnets, pop3s), HTTP- and HTTPS-specific services (including cookie handling, setting host header for ) STARTTLS (LDAPS), and DNS. These are all far more complex than NTP, yet deemed important enough to implement in code rather than as a call-out to an external binary.

If the inclusion of these other protocols was from a question of efficiency, I can see that internalising the NTP probe might not fit into the same category. After all, most environments are likely to have several orders of magnitude more web/mail/DNS servers than NTP servers. So a call-out to an external binary for a few servers is not a significant burden. However, there are likely to be some enviornments where the number of NTP servers to check is larger than many of us typically experience, so an fork/exec every five minutes for dozens of NTP servers may have a measurable impact on a heavily-loaded and modestly-resourced Xymon server. On the other hand, if a fork/exec is never going to be a burden, then why not simply solve this problem with a wrapper that re-formats whatever tool is used into a standard format. We seem to be trying to see solve one problem in two different (and increasing complex) ways, where one solution is likely to work in all cases.

While I like it when features make things easier and more flexible for the sysadmin, the extra complexity that comes with added features needs to be balanced with the added cost/time of troubleshooting, and the added security risks, when code has an increased number of potential pathways to traverse. It's not clear what the balance should be in this case, but we should be cognisant that there is a balance to be considered here.

I have no authority here, so please consider this as just an opinion from an uninformed perspective. If it resonates, that's great. Otherwise, that's fine too. TBH I'm just grateful that people are working on this amazing product, esp the coordination that Bruno is giving to the project.

bonomani:
@jezzaaa Thanks for taking the time to step back like this — and you're right. An in-process probe is the clean answer, and the more I looked the more your instinct held up.

A bit of context first: the "ntp" test has *always* been a call-out (`ntpdate`, later `sntp`), so this PR wasn't adding the external-tool approach so much as restoring a test that now fails out-of-the-box — classic `ntpdate` is gone from current distros, which is what pushed it toward scanning for ntpdig/chrony. And roughly half of it was independent bug fixes (the offset was being stored 1000× too small, plus an out-of-bounds read in the RRD handoff). But you put your finger on the real thing: SNTP is a single 48-byte UDP exchange — simpler than the TLS/DNS probes xymonnet already does in-process — so the right move is to *internalize it* and stop scanning for a tool at all.

So I've gone and built a prototype of that. A built-in SNTP probe as the new default: no fork, no external program, and it works even when no NTP client is installed — which also removes the per-host fork/exec you flagged for larger fleets. The prototype is written and tested (the protocol math, the UDP transport, and the offset thresholds each have harnesses under AddressSanitizer) and builds/links cleanly. `ntpdate`/`sntp` stay as an opt-in fallback so nobody's existing config breaks, and the offset can now optionally drive yellow/red on drift.

Plan from here:

1. **One PR, replacing this one** — the built-in probe (default), with `ntpdate`/`sntp` kept as a fallback and optional offset thresholds. The offset bug fixes ride along in the same PR (as their own commit, so they're easy to review) since the probe builds on them and nobody's waiting on them separately.
2. **This PR I'll close as superseded** by it, so the added complexity you (rightly) questioned doesn't land.

*(And on the build side: the probe adds a couple of new source files, which I'll wire into the CMake migration as well as the current make build.)*

So — your read was correct, and it genuinely changed the direction here. I want to run a few more checks first, but I'll have it up as PR (1) very soon — and since you called it, you've earned the first crack at it: I'd genuinely value your review. Thanks for thinking out loud in the open; it's exactly the kind of feedback that makes this easier to get right.

jezzaaa:
Thanks for your work on this, and I'm impressed at the turn-around time for an interal SNTP implementation. Amazing! I'd be happy to test out the new build.

I certainly didn't want to force a new direction at the expense of existing users who expect things to continue to work a certain way. I like that you're maintaining backwards compatibility with ntpdate/sntp. It would be unfortunate if someone was relying on NTPDATE and other well-established variables in an existing installation, and they stopped working as a result of the addiition of an internal implementation.

The current default installation sets NTPDATE="ntpdate" which, as you have pointed out, is almost universally missing from modern Linux installations. Perhaps the default value for NTPDATE, in a fresh install of xymonserver.cfg from package or "make install", could be the null string (NTPDATE=""), so that the internal SNTP client could be run, yet any existing deployment with NTPDATE or SNTP defined would continue to use the external as expected.

On the other hand, I wonder if there are existing installations where NTPDATE has been set to the null string, specifically to disable the execution of any external command (forcing xymonnet to never attempt to run an NTP client, even of "ntp" is defined in hosts.cfg). In such cases, the approach I mentioned above would change the behaviour. I don't know if this scenario is even likely, or even if setting NTPDATE="" will prevent attempts at running ntpdate. But just something to consider. If this is likely to be a problem for some, maybe NTPDATE="internal" might be an alternative indicator?

Just some ideas. I really don't know enough about the code to have a reliable opinion on this.

bonomani:
Superseded by #181 (built-in SNTP probe for the `ntp` test), which fixes #27 and explicitly supersedes this PR. Closing.

PR #160, opened 2026-06-12
debian: refresh the packaging from Debian, without downstream patches

RolandRosenfeld:
The debian/ directory in this project is somewhat outdated.
jccleaver merged several changes from the Debian package here in the past, but timed passed and many changes, that happened in Debian didn't find their way here in in the meantime.

This PR tries to update the debian/ directory, based on current Debian package but without the several downstream applied patches and changes (see https://github.com/xymon-monitoring/xymon/issues/28).

I tried to reduce the Debian specials, but some were already applied years ago by jccleaver and others may be deep integrated between debian/rules and the (not applied) [12_hobbitvars.patch](https://salsa.debian.org/debian/xymon/-/blob/master/debian/patches/12_hobbitvars.patch).

Since I'm not sure, whether all this works correct now (and I didn't test this especially migrations between the Debian package and the upstream .deb package), this is a draft for the moment and my need more cleanup.

At least this PR should allow to build a deb source package using `debuild -us -uc -b` or to build a deb binary package using `debuild -us -uc -b`.

I didn't touch anything outside debian/, but we should also have a look at build/makedeb.sh.  This doesn't fully follow Debian standard build mechanisms (building the package simply in the source directory), but clones some sources to debbuild/ and builds there.  Not sure, whether this catches all sources or misses some.
It also doesn't handle the version in debian/changelog, which is used (instead of the given version from the parameter) during build, so maybe we should add
```bash
debchange -v "$REL" ''
debchange -r ''
```
before the dpkg-buildpackage and remove the `-rfakeroot`, which is no longer needed nowadays.


bonomani:
A note on two things in flight upstream that this packaging will meet, no action needed now.

**`ExecReload=/bin/kill -HUP $MAINPID` and the logrotate `rotate` reach xymonlaunch only.** It does not currently pass the signal on, so the daemons keep writing to the rotated files. #172 adds that relay (Terabithia 78): tasks marked `SENDHUP` in `tasks.cfg` get the HUP, and the shipped defaults mark `xymond`, `xymonproxy` and `xymonfetch` — the last two also learned to reopen a log they were given by the launcher rather than on their own command line. Merge chain is #219 → #172 → #173.

**#219 adds `XYMONRUNDIR`**, a configured directory for pidfiles and the `rrdctl.*` sockets, defaulting to the log directory. It is not the same content as this PR's `/var/run/xymon`, which holds the generated `*-include.cfg` files — but it does mean the packaging can point pidfiles and sockets at `/run/xymon` through the config rather than leaving them in `/var/log/xymon`.

PR #161, opened 2026-06-13
client: accept IFEXIST as an alias for OPTIONAL in FILE rules

bonomani:
## Summary

Accept `IFEXIST` as an alias for the existing `OPTIONAL` keyword in `FILE` analysis rules, so that a file check is skipped when the target file does not exist.

This is the upstreaming of Debian's `27_hobbit_files_ifexist` patch (tracked in #28), implemented as a one-line alias instead of a new flag.

## Why an alias rather than a new flag

Upstream already has exactly the requested behavior under `OPTIONAL` (`CHK_OPTIONAL`): the `FILE` check is skipped when the file is absent. Adding a separate `IFEXIST` flag would duplicate that logic with a second runtime path for no functional gain.

Routing `ifexist` to `CHK_OPTIONAL` means:
- no new flag bit, no new `check_file()` branch;
- the existing `dump_client_config()` and `check_file()` paths already handle it (a rule written with `ifexist` round-trips as `OPTIONAL`, which is the canonical spelling);
- existing Debian-derived configs using `IFEXIST` keep working unchanged.

## Supersedes #50

PR #50 implements the same feature with a dedicated flag:

```c
#define FCHK_IFEXIST  (1 << 32)
```

The `flags`/`chkflags` fields are `uint32_t` (valid bits 0–31; the existing flags already use up to `1 << 31`). `1 << 32` is undefined behavior on a 32-bit `int` and, with GCC constant-folding, evaluates to **0** — so `flags |= FCHK_IFEXIST` is a no-op, the dump never emits `ifexist`, and the `goto nextcheck` never fires. The feature silently does nothing as written. This alias approach avoids the flag-space problem entirely.

## Changes

- `xymond/client_config.c`: `FILE` rule parser accepts `ifexist` alongside `optional`.
- `xymond/analysis.cfg.5`: documents `IFEXIST` as an alias for `OPTIONAL`, and adds it to the `FILE` synopsis.

## Testing

`make xymond-build` compiles clean; `groff -man -z -ww xymond/analysis.cfg.5` lints clean. The behavior reuses the `CHK_OPTIONAL` path already exercised by the `OPTIONAL` keyword.

Refs #27, #28.

### DIFF
--- a/client/localclient.cfg
+++ b/client/localclient.cfg
@@ -193,7 +193,7 @@
 #             Example: Go red if the text "I/O error" or "read error" appears.
 #                 LOG %/var/(adm|log)/messages %(I/O|read).error COLOR=red
 #
-#    FILE filename [color] [things to check] [TRACK]
+#    FILE filename [color] [things to check] [OPTIONAL|IFEXIST] [TRACK]
 #             NB: The files you wish to monitor must be listed in a "file:..."
 #             entry in the client-local.cfg file, in order for the client to 
 #             report any data about them.
@@ -233,8 +233,11 @@
 #             - "MD5=md5sum", "SHA1=sha1sum", and so on for RMD160, SHA256, SHA512,
 #               SHA224, and SHA384 trigger a warning if the file checksum using the
 #               specified message digest algorithm does not match the one configured here.
-#               Note: The "file" entry in the client-local.cfg file must specify which 
+#               Note: The "file" entry in the client-local.cfg file must specify which
 #               algorithm to use as that is the only one that will be sent.
+#             - "OPTIONAL" skips this check entirely if the file does not exist
+#               (no warning is raised for the missing file). "IFEXIST" is an
+#               accepted alias, for compatibility with existing configurations.
 #
 #             "TRACK" causes the size of this file to be tracked in an RRD file, and
 #             shown on the graph on the "files" display.
--- a/docs/manpages/man5/analysis.cfg.5.html
+++ b/docs/manpages/man5/analysis.cfg.5.html
@@ -333,7 +333,7 @@ will just show up on the &quot;msgs&quot; column with status OK (green).
 <H2>FILES STATUS COLUMN SETTINGS</H2>
 
 <P>
-<B>FILE filename [color] [things to check] [OPTIONAL] [TRACK]</B>
+<B>FILE filename [color] [things to check] [OPTIONAL|IFEXIST] [TRACK]</B>
 
 <P>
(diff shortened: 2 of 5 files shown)


RolandRosenfeld:
Maybe 27_hobbit_files_ifexist.patch was created, because "OPTIONAL" isn't documented consistently and wasn't found, because of this?
We have the following points, where it should be documented:
- xymond/analysis.cfg.5 (done with this PR)
- xymond/etcfiles/analysis.cfg (neither mentions OPTIONAL nor IFEXIST, the old 27_hobbit_files_ifexist.patch added IFEXIST, maybe this PR should add both?)
- client/localclient.cfg (same as analysis.cfg)

@bonomani: Could you extend analysis.cfg and localclient.cfg with OPTIONAL/IFEXIST in the comments?

TODO (independent of this PR):
- (partly) sync contents of analysis.cfg and localclient.cfg
- Add a man page for localclient.cfg, which is installed when using localclient (maybe this is some "clone" of analysis.cfg.5)

bonomani:
Thanks @RolandRosenfeld — good catch on the root cause. You're right that `OPTIONAL` was documented in `analysis.cfg.5` but missing from the example configs admins actually read, which is very likely why the Debian `IFEXIST` patch was written in the first place.

Addressed in the latest push:

- **`xymond/etcfiles/analysis.cfg`** and **`client/localclient.cfg`** — added `OPTIONAL` (with `IFEXIST` as the alias) to the `FILE` "things to check" list and to the `FILE` synopsis line in both files.
- **`docs/manpages/man5/analysis.cfg.5.html`** — mirrored the `IFEXIST` line into the rendered HTML as a scoped hand-patch (just the 2 IFEXIST hunks). I deliberately did *not* run a full `man2html` regen here: the current `man2html` + source drift since the file was last generated (v4.3.30, 2019) would rewrite DOCTYPE, title casing, some bolding and the timestamp across the page — noise that affects every man page, not just this one, and is better done as a deliberate release-time regen.

Throughout I kept `OPTIONAL` as the canonical spelling and `IFEXIST` as the alias, consistent with the PR's design.

For your "TODO (independent of this PR)" — the `analysis.cfg`/`localclient.cfg` sync and the missing `localclient.cfg(5)` man page — I opened #162 to track those separately so they don't hold up this PR. While doing the above I noticed the `FILE` synopsis arg order already diverges between the two files (`[things to check] [color]` vs `[color] [things to check]`); details are in the issue.

RolandRosenfeld (approved):
Perfect now.

PR #163, opened 2026-06-13
build: let environment CFLAGS and LDFLAGS hardening flags survive and reach every binary

bonomani:
Ports the upstream-relevant part of Debian's `51_hardening-buildflags` (refs #28): make a distribution's hardening flags (`-D_FORTIFY_SOURCE`, `-fstack-protector`, PIE/RELRO) actually reach the shipped binaries. Two commits.

**1 — `CFLAGS` survive.** `build/Makefile.Linux` set `CFLAGS` with a plain `=`, overwriting any environment value. Split it:
- `CFLAGS ?= -g -O2` — optimisation/debug defaults the environment can override;
- `CFLAGS += <tirpc include, -D_REENTRANT, LFSDEF, OSDEF, warnings>` — mandatory project flags, always appended.

Byte-identical output when no env `CFLAGS` is set; env leads when it is. `make CFLAGS=…` on the command line still overrides entirely, as before — the environment is the supported injection path.

**2 — `LDFLAGS` reach every link.** Only 15 of 92 link rules referenced `$(LDFLAGS)` — patchy enough that the build had to pass `-no-pie` to dodge an `R_X86_64_32` relocation on Fedora/EL10. Thread `$(LDFLAGS)` through all 92, and forward `LDFLAGS` into the `client` sub-make in `build/Makefile.rules` (the one invocation that dropped it) so client binaries (`logfetch`, `msgcache`, …) are hardened too.

Scope: `CFLAGS` change is `Makefile.Linux` only (where distro hardening matters most); the same `?=`/`+=` can follow for other `Makefile.<OS>`. Preprocessor flags (`CPPFLAGS`, Debian's `_FORTIFY_SOURCE` channel) are dropped separately — follow-up #444; Fedora/EL fold FORTIFY into `CFLAGS`, so RPM is covered here.

Downstream: [xymon-rpm](https://github.com/xymon-monitoring/xymon-rpm) builds `main` unpatched — with this in, the spec just exports `%{optflags}`/`%{build_ldflags}` and drops its `-no-pie`/no-hardening workarounds. Merge order: **#163 (independent)** · runtime-directory stack **#219 → #172 → #173** (packaging #46 is downstream and separate).


bonomani:
I reviewed this, two rounds:

- **Commit 1 (`CFLAGS`):** clean, no findings — the `?=`/`+=` split preserves the existing defaults, appends the mandatory project flags to an environment `CFLAGS`, and leaves command-line-override semantics unchanged.
- **Commit 2 (`LDFLAGS`):** flagged that the `client` sub-make in `build/Makefile.rules` was the one invocation not forwarding `LDFLAGS`, so client binaries stayed unhardened — **fixed** in this branch. Also flagged that `CPPFLAGS` (Debian's FORTIFY channel) is never expanded → filed as #444, out of scope here since Fedora/EL fold FORTIFY into `CFLAGS`.

CI green on all variants (server, client×2, manpages).

RolandRosenfeld (approved):
This PR fully makes sense.
The first commit is a very good replacement of the old Debian 51_hardening-buildflags.patch.
The second commit also looks good to me.  It makes sense, that all binary building targets use LDFLAGS.  Unbelievable, that this wasn't the default before.

So I approve this pull request.

PR #167, opened 2026-06-13
ci: run the regression suite in every build leg

### LINKED ISSUE #166: CI: regression suite only runs in the server build leg; client legs build binaries but never test them
## Summary

The CI regression suite (`tests/testsuite`, RFC #97) runs inconsistently across build legs: the **server** build leg runs it, but the two **client** build legs build binaries and then run **no tests at all**. Meanwhile the separate build-free lane (`tests.yml`) runs only the source-reading subset. The net effect is that **client-variant binaries are built in CI but never exercised by the suite**, and the binary-driving tests (`require_bin`, e.g. `tests/server/xymongrep-filter.sh`) only ever execute in the server leg.

## Current behaviour

| Check | Builds? | Runs tests? | What runs |
| --- | --- | --- | --- |
| `build (server)` | server | ✅ yes | Full suite vs built server binaries (the only place binary tests actually execute) |
| `build (client, CT=client)` | client | ❌ no | nothing |
| `build (client CT=server)` | client | ❌ no | nothing |
| `tests / shell regression (tests.yml)` | no build | ✅ yes | source-only (binary tests skip — nothing's built) |

## Root cause

`.github/workflows/build.yml` gates the test step to the server leg:

```yaml
- name: Run regression tests against the build
  if: matrix.variant == 'server'
```

So the client legs compile their binaries and discard them without running a single test against them.

## Why it matters

- The client and localclient binaries get zero post-build test coverage in CI.
- Any future `require_bin` test for a client-only tool would silently never run.
- It reads as "tested" because the suite *is* wired up — it just doesn't execute where the relevant binaries exist.

## Proposed fix

Drop the `if: matrix.variant == 'server'` gate and run the suite in **every** build leg, scoping it per variant via `XYMON_VARIANT` (`need_variant` in `tests/lib/assert.sh`):

- server leg → `XYMON_VARIANT=server` (server + shared tests)
- client legs → `XYMON_VARIANT=client` / `localclient` (client + shared tests; server-only checks like `configure --server`, RRD, xymonnet, web CGI skip)
- `CONFTYPE=client` builds `LOCALCLIENT=yes`, so that leg scopes to `localclient`.

Each leg then runs its applicable tests against the binaries it just built, instead of skipping. The build-free `tests.yml` lane stays as the independent source-only signal that remains visible when the build itself is red.

bonomani:
**Stacked on #391** (#391 → #167 → #319 → #179 → #394); own commits: the last 4. Refs #166.

The suite runs only what a build contains, and every CI leg runs it.

- A test's directory is its area; one table maps areas to the variants that provide them. Tests filed by topic moved to the area of what they drive.
- Each CI leg names the build it made (`testvariant`: the three combinations of `CLIENTONLY` × `LOCALCLIENT`). The runner checks the label against the configured tree and refuses a mismatch; an unlabeled run (developer box, tarball, build-free lane) filters nothing.
- `build.yml` loses its server-only gate: a client leg now verifies the client it just built.

| leg | result |
|---|---|
| server | 86 passed, 0 skipped |
| client | 32 passed, 54 filtered |
| localclient | 33 passed, 53 filtered |

Rejected alternatives (`skip_env`, exempting `require_bin` callers from the filter) are recorded in `tests/README.md` so they are not re-derived.


SoundGoof:
## Findings

### Rebase and cover localclient `xymond_client`

PR #167 is 138 commits behind current `main` and needs
rebasing. The two PR commits test-rebase cleanly.

After rebasing, the localclient leg builds
`client/xymond_client`, but
`tests/server/analysis-file-ifexist.sh` only searches for
`xymond/xymond_client`. That test therefore skips despite a
compatible binary being available.

Running it with
`XYMOND_CLIENT=client/xymond_client` passes. Add a fallback
for the client path, similar to the `xymongrep` fallback in
this PR.

The test was added after this PR was opened, so rebasing is
required to expose and address this integration gap.

## Verification

- Test-rebase onto current `main` completed cleanly.
- Server: 37 passed, 0 skipped, 0 failed.
- Client: 18 passed, 19 skipped, 0 failed.
- Localclient: 18 passed, 19 skipped, 0 failed.
- All current GitHub checks are green.
- `git diff --check` passed.

bonomani:
Rebased and reshaped since. Your finding is fixed: `analysis-file-ifexist.sh` now passes in the localclient leg, driving the `client/xymond_client` that leg builds.

One correction — the `xymongrep` client fallback you pointed at was already on `main` at this branch's merge-base; rather than copying that probe into a second test, this PR deletes it. Per-variant paths live in one table (`variant_products()` in `tests/lib/assert.sh`), each CI leg names the build it made (`testvariant`), and the runner checks that label against the configured tree, refusing a mismatch.

Current legs: server 86 passed / 0 skipped, client 32 / 54 filtered, localclient 33 / 53 filtered — all green.

PR #169, opened 2026-06-13
tests: fix rrd-api-compat probe shadowing macOS <Availability.h>

### LINKED ISSUE #168: tests: rrd-api-compat probe shadows macOS <Availability.h> via -I lib (case-insensitive FS)
## Summary

`tests/buildsystem/rrd-api-compat.sh` fails to compile its probe on macOS because it puts `lib/` on the compiler's angle-bracket (`<...>`) include search path, where xymon's `lib/availability.h` shadows the macOS SDK's `<Availability.h>` on a case-insensitive filesystem (the names differ only in case).

## Symptom

When the suite runs on a macOS host, the probe compile fails with spurious errors:

```
lib/availability.h:46:26: error: unknown type name 'time_t'
.../MacOSX.sdk/usr/include/_time.h:157:16: error: missing ',' between enumerators
.../MacOSX.sdk/usr/include/_time.h:159:18: error: redefinition of enumerator '__OSX_AVAILABLE...'
##[error]tests/buildsystem/rrd-api-compat.sh FAILED (rc=1)
```

## Root cause

The probe is compiled with (`tests/buildsystem/rrd-api-compat.sh`):

```sh
"$CC" ... -I"$WORK" -I"$ROOT/lib" -c "$WORK/probe.c" ...
```

`-I"$ROOT/lib"` is only there so the **quoted** include `#include "rrd_api_compat.h"` resolves. But `-I` also adds `lib/` to the **angle-bracket** search path. The macOS SDK's own headers do `#include <Availability.h>` (e.g. `<time.h>` → `_time.h`), and on a case-insensitive filesystem that resolves to xymon's `lib/availability.h` instead of the SDK header — so the SDK's availability macros are undefined and compilation collapses.

## Scope

- **Test-harness only.** The real build is *not* affected: it reaches `availability.h` through quoted, relative includes (`include/libxymon.h` → `#include "../lib/availability.h"`) and never adds `lib/` to the angle-bracket path.
- **Latent in current CI**, because the build/test workflow is Ubuntu-only (case-sensitive FS → no collision). It bites any macOS developer running `./tests/testsuite`, and any future macOS CI lane. (It was first observed on a downstream branch that does run macOS lanes.)

## Fix

Use `-iquote "$ROOT/lib"` instead of `-I"$ROOT/lib"`: the probe reaches `rrd_api_compat.h` via a quoted include, so `-iquote` is sufficient and keeps `lib/` off the angle-bracket path, so `<Availability.h>` resolves to the SDK again. `$WORK` stays on `-I` because `rrd_api_compat.h` includes the mock `<rrd.h>`.

PR follows.

bonomani:
## What

In `tests/buildsystem/rrd-api-compat.sh`, compile the probe with `-iquote "$ROOT/lib"` instead of `-I"$ROOT/lib"`.

Fixes #168.

## Why

`-I"$ROOT/lib"` is only needed so the probe's **quoted** include `#include "rrd_api_compat.h"` resolves — but `-I` also drops `lib/` onto the **angle-bracket** (`<...>`) search path. xymon's `lib/availability.h` differs from the macOS SDK's `<Availability.h>` only in case, so on a case-insensitive filesystem (the macOS default) the SDK's own `<time.h>` → `<Availability.h>` resolves to xymon's header, and the probe fails to compile:

```
lib/availability.h:46:26: error: unknown type name 'time_t'
.../MacOSX.sdk/usr/include/_time.h:157:16: error: missing ',' between enumerators
```

`-iquote` satisfies the quoted include without polluting the angle-bracket path, so `<Availability.h>` resolves to the SDK again. `$WORK` stays on `-I` because `rrd_api_compat.h` includes the mock `<rrd.h>`.

## Scope / safety

- **Test-harness only.** The real build reaches `availability.h` via quoted relative includes and never adds `lib/` to the angle path, so it was never affected.
- Behavior-neutral on Linux (case-sensitive FS); fixes the failure on macOS regardless of build variant.

## Test

- `shellcheck tests/buildsystem/rrd-api-compat.sh` — clean (only the suite-wide SC1091 source-not-followed info).
- `./tests/buildsystem/rrd-api-compat.sh` on Linux: still `PASS` (both legacy and const argv APIs compile).

### DIFF
--- a/tests/buildsystem/rrd-api-compat.sh
+++ b/tests/buildsystem/rrd-api-compat.sh
@@ -72,8 +72,15 @@ EOF
 check() {                      # $1 = legacy|const   $2 = RRD_CONST_ARGS value
 	write_mock_rrd_h "$1"
 	printf '  %-7s API (RRD_CONST_ARGS=%s) ... ' "$1" "$2"
+	# -iquote (not -I) for lib/: the probe reaches rrd_api_compat.h via a
+	# quoted include, so -iquote suffices -- and it keeps lib/ off the angle-
+	# bracket (<...>) search path. With -I, lib/availability.h shadows the
+	# macOS SDK's <Availability.h> on a case-insensitive filesystem (the names
+	# differ only in case), so the SDK's own <time.h>/<stdio.h> pull in the
+	# wrong header and the probe fails to compile with bogus errors. $WORK
+	# stays on -I because rrd_api_compat.h includes the mock <rrd.h>.
 	if ! "$CC" -std=c99 -Wall -Wextra -Werror "-DRRD_CONST_ARGS=$2" \
-		-I"$WORK" -I"$ROOT/lib" \
+		-I"$WORK" -iquote "$ROOT/lib" \
 		-c "$WORK/probe.c" -o "$WORK/probe.o" 2>"$WORK/err"; then
 		echo "FAILED"
 		fail "rrd_api_compat.h does not compile for the $1 argv API:


bonomani:
## Evidence

### 1. Observed failure (macOS CI lane)

On a branch with macOS CI lanes, the probe compile fails verbatim:

```
/.../lib/availability.h:46:26: error: unknown type name 'time_t'
/.../lib/availability.h:47:30: error: unknown type name 'FILE'
/Applications/Xcode_16.4.app/.../MacOSX.sdk/usr/include/_time.h:157:16: error: missing ',' between enumerators
/Applications/Xcode_16.4.app/.../MacOSX.sdk/usr/include/_time.h:159:18: error: redefinition of enumerator '__OSX_AVAILABLE...'
##[error]tests/buildsystem/rrd-api-compat.sh FAILED (rc=1)
```

Note the error names `Availability.h` (capital, the SDK request) but the path is xymon's lowercase `lib/availability.h` — the case-insensitive-FS match. On the **same run**, the case-sensitive lanes (Linux, FreeBSD) all passed; only macOS failed.

### 2. Reproducible mechanism (no Mac needed — Linux, case-sensitive FS)

Emulating "SDK provides the angle-include target in a system dir, xymon ships a same-named header in `lib/`" with an exact-name file (the case difference is irrelevant to the *search-path* behavior being demonstrated):

```
# lib/availability.h  -> #error "SHADOWED ... reached via the <...> path"
# sysinc/availability.h -> #define REAL_SDK_AVAILABILITY 1   (stands in for the SDK header)
# probe.c -> #include <availability.h>   (as the SDK's own <time.h> does)

$ cc -I lib      -isystem sysinc -c probe.c   # BEFORE
lib/availability.h:1:2: error: #error "SHADOWED: ... reached via the <...> search path"   (exit 2)

$ cc -iquote lib -isystem sysinc -c probe.c   # AFTER
(exit 0)  # <availability.h> resolves to the real SDK header

$ cc -iquote lib -c probe_quoted.c            # AFTER still serves the quoted include
(exit 0)  # #include "rrd_api_compat.h" is still found
```

`-I` puts `lib/` on the angle-bracket path → shadows. `-iquote` keeps it off that path while still satisfying the probe's quoted include.

### 3. Real build is unaffected

`lib/` is never added to the angle-bracket (`-I`) path in the build; `availability.h` is reached only via quoted, relative includes:

```
$ git grep -nE '-I[^ ]*\blib\b' -- '*.h' 'Makefile*' build      # (no matches)
include/libxymon.h:67:#include "../lib/availability.h"
```

So this is a test-harness-only defect; shipped code on macOS was always fine.

SoundGoof (approved):
lgtm

PR #172, opened 2026-06-15
xymonlaunch: write a task's pidfile and forward HUP with SENDHUP (TBT 65/78/603/780, devel 3959aae5c/fa6ae1f94/266644346)

bonomani:
> **Runtime-directory stack `#219 → #172 → #173`** — this PR is the middle, stacked on #219. Separately, **#415** depends on it: the `tools/xymonlaunch.service` it generates carries `ExecReload=/bin/kill -HUP $MAINPID` — *"xymonlaunch will propagate -HUP as needed"* — and that propagation is this PR. Until it merges, `systemctl reload` is a no-op on a packaged install and logrotate cannot signal the daemons.

## xymonlaunch: PIDFILE + SENDHUP support (Terabithia backport)

Resolves the **SENDHUP chain** bucket of #29. Target 4.3.32. **Stacked on #219**, whose six commits are the first six here and disappear once it merges; `$XYMONRUNDIR` has to exist before the defaults below can name it.

Three original Terabithia patches applied as authored (`patch -p0`, offsets only, no fuzz or rejects), then three corrections:

- **TBT 65** — per-task `PIDFILE`: the child writes its pid, the launcher removes the file.
- **TBT 780** — the `SENDHUP` keyword: parser and dump, no relay yet.
- **TBT 78** — the relay itself, `$XYMONLAUNCH_LOGFILENAME` propagation, and `xymond_channel` rotation.

The correction ships the `PIDFILE`/`SENDHUP` defaults (Terabithia 603, the same three `SENDHUP` tasks devel names) and fixes what stopped them working. `[history]` keeps its own `--pidfile=`: that pid is xymond_history's, read by trimhistory, where `PIDFILE` records the direct child — xymond_channel.

**The pidfile was removed from a path the child never wrote.** A re-read clears every task's settings before parsing, so by the time `load_config()` decides what to kill, `twalk->pidfile` is no longer where the child wrote:

| the config is re-read and | before | after |
|---|---|---|
| the task is gone from it | the field is `NULL`, so the unlink written for this case never ran — the file stays, naming a pid the system may reuse | removed |
| its `PIDFILE` moved | the field holds the *new* path: the unlink hit a name not yet written | old removed, new written by the restarted task |

and nothing removed it when the task ended or when the launcher stopped. Both now do.

**A rotation could be lost or arrive nowhere.** `xymond_channel` reopened *after* the blocking wait — and the SIGHUP asking for the switch is what interrupts that wait, whose EINTR path continues straight back to it, so an idle channel kept writing to the rotated file until traffic arrived. `xymonproxy` reopened only with its own `--logfile=`, which a launcher-started task never has, and did it inside the signal handler; `xymonfetch` took a HUP as a config reload and has no logfile option at all. The devel half of `fa6ae1f9` is ported for the first, twelve lines teach the second — a deliberate divergence, since devel has shipped that `SENDHUP` line for years without it.

**The proxy removed a pidfile it never wrote.** This PR makes the launcher the process that writes and removes a task's pidfile, and changes the xymonproxy task to use it — while the binary still ran `if (pidfile) unlink(pidfile)` on the way out. `pidfile` is never NULL, it defaults to `/var/run/xymonproxy.pid`, and it is only ever *written* under `if (daemonize)`. So a proxy started with `--no-daemon`, which is how the shipped task runs it, deleted whatever that path named. Measured before the change, on a foreground proxy given a path it had not written:

```
DBG reached-tail pidfile=[.../p.pid] daemonize=0 unlink=0
pidfile: gone
```

The daemon case is not the same mistake and has to keep working: there the parent wrote that path holding the child's pid, so the child removing it on exit is the daemon cleaning up after itself. Both are now driven by a test.

Review then showed that guarding on the mode was the wrong test. If the daemonising parent's `fopen` fails it exits 0 regardless, and the child inherits `daemonize == 1` and unlinks the path anyway — so "we daemonised" is not "we own it". The pid is read back instead, which covers both cases and needs no flag; with it the mode test adds nothing and is gone. The failed write is now reported too, as `xymond` already does. The test's own premise was racy in the same way: it waits for `Listening`, which the proxy prints *before* `setup_signalhandler()`, so a signal in that window killed it outright and the assertion passed for the wrong reason — both halves now require `Caught TERM signal` in the log. The other two faults the same review found — `--pidfile=` parsed but written only when daemonising, and devel advertising the option in `--help` without parsing it — need a decision rather than a correction and are **#554**.

**The client launcher is started with `--env`, as the server's already is.** The launcher expands each task's pidfile path itself, so it must read what its tasks read; without it, it removed files nobody wrote and left the real ones. Every client task already loads that file.

| evidence | |
|---|---|
| tests | `xymonlaunch-pidfile-reload.sh` (live launcher: task deleted, path moved, task exits, launcher stops) · `channel-hup-rotate.sh` (real xymond_channel, rotation with no traffic) · `tasks-sendhup-contract.sh` (a `SENDHUP` task must name a binary that can rotate; the launcher must read its tasks' settings file) · `xymonproxy-pidfile-ownership.sh` (real binary, foreground and daemon) |
| mutations that fail them | the pidfile read from the cleared field · the unlink dropped from reap or shutdown · the reopen moved back after the wait · either port reverted · `--env` removed · the proxy's unlink left unconditional, and the same guard widened to exclude the daemon too |
| suite | 104 passed, 0 skipped, 0 failed |

**Not covered, said rather than implied:** *rotation* is driven for `xymond_channel` only — the xymonproxy and xymonfetch rotation ports are read and reviewed, not tested. (The proxy's pidfile ownership above is tested; its log rotation is not.) The launcher's log-switch flag is captured once per pass and cleared there, which narrows but does not close the window between reading and clearing it; that race and the equivalent one in the channel's wait belong together in a follow-up.

**Overlaps in `xymond/xymond_history.c` and `xymond/Makefile`.** Two collisions, both ordering rather than design:

- **#400 #401 #402** in `xymond_history.c` — both sides rewrite lines 91 and 104. Those three are one rebased series (each carries the one before it under new commit ids), so they go in that order; this then rebases once onto the result, rather than the series rebasing three times over a file this has changed.
- **#414** in `xymond/Makefile` — both sides rewrite lines 143, 202 and 203.

Nothing here is superseded by either: this writes a task's pidfile and forwards HUP, #400/#401/#402 confine the paths `xymond_history` writes, #414 changes what gets installed. Only the hunks need sequencing.


bonomani:
Now **stacked on #219** rather than parallel to it. The two shipped configs conflicted — `--pidfile=` on the `CMD` line against the `PIDFILE` keyword — and that conflict is resolved in the branch instead of at merge time: the keyword everywhere, `--pidfile=` kept on `[history]` alone, whose direct child is xymond_channel and not the daemon holding the pid.

The corrections since this was opened are in the body; each fails a test when reverted. Still gated on #219 for `$XYMONRUNDIR`.

bonomani:
## Pre-merge review

I reviewed this. Three findings — the P1 (a data-loss risk) is **reworked and CI-green**; two edge P2s left as-is.

**Reworked (P1 — the SIGHUP relay):**
The relay sent `kill(childpid, SIGHUP)` to *every* local worker. `xymond_filestore` and `xymond_distribute` install **no** SIGHUP handler, so a log rotation would terminate them and drop channel data. Removed the signal propagation: the channel still reopens its own log via `do_logswitch()`, and worker logs rotate through the **existing** `@@logrotate` message that `xymond` posts on every rotation (`posttoall`) and the channel forwards down the pipe — the mechanism every shipped worker already handles. (Confirmed only `[xymond]`/`[xymonproxy]`/`[xymonfetch]` carry `SENDHUP` in the shipped `tasks.cfg`, so channels are never signalled directly anyway.)

**Not changed (edge):**
- **[P2] per-task PIDFILE expansion** — xymonlaunch expands a task's `PIDFILE` in the parent; a child that overrides it via its own `ENVFILE`/`ENVAREA` can disagree, leaving a stale pidfile. Only bites `PIDFILE` values built from per-task env; the shipped tasks use the global `$XYMONRUNDIR`.
- **[P2] daemon self-managed pidfiles** — `xymonproxy`/`msgcache` also unlink a hardcoded internal pidfile; enabling those disabled tasks could remove an unrelated path. Pre-existing, only when enabled.

Also **rebased** from a 47-commit-stale base (which duplicated #219) onto the current #219 tip — now a clean 4-commit stack. CI green on all variants.

**Merge order:** #163 is independent; **#219 → #172** — this PR stacks on #219 and must merge after it.

bonomani:
Conflicts with two other open pull requests:

- #400 #401 #402 in `xymond/xymond_history.c`
- #414 in `xymond/Makefile`

Both apply cleanly to `main` on their own, so nothing here is broken. Whichever merges second needs a rebase before it can go in, and a rebase can cost it a fresh review. #400 #401 #402 are one rebased series and have to go in that order, so the cheapest sequence is the series first, then this rebased once onto the result.

PR #173, opened 2026-06-15
xymonlaunch: add DELAY and FAILDELAY to time a task's start and retry (TBT 125, devel 065cfec07)

bonomani:
## xymonlaunch: DELAY / FAILDELAY keywords (Terabithia backport)

Resolves the **launchfaildelay** bucket of #29. Target 4.3.32.

**Last in the runtime-directory stack `#219 → #172 → #173`.** Rebased onto the current #172. Only three commits here are its own — the two DELAY/FAILDELAY ones and a correction; the rest are #219's and #172's, carried because all three target `main`, so GitHub shows the cumulative diff rather than the stacked one. Reviewing it alone means reading `git diff backport/sendhup-chain...backport/launchfaildelay`, four files. The order is not a preference: `$XYMONRUNDIR` must exist before the pidfile defaults name it, and #172's SIGHUP correction is what keeps `PIDFILE`/`SENDHUP` from arriving in their buggy form.

`DELAY` sets the hold after a `NEEDS` dependency starts, replacing the hardcoded 5 s wait. `FAILDELAY` backs off a task that exits non-zero five times in a row and clears the count after that much uptime. Defaults 5 / 600 preserve today's behaviour, `FAILDELAY 0` disables the hold. TBT 125 is applied as authored — `patch -p0`, offsets only, no fuzz or rejects.

**One correction on top.** A re-read clears every task's settings before parsing, so a keyword removed from `tasks.cfg` reverts to its default. These two were set only where a task is *created*, never in that clearing loop:

| the line is removed and the config re-read | before | after |
|---|---|---|
| `DELAY 3600` | the task goes on waiting an hour, nothing in the file asking for it | the default 5 s |
| `FAILDELAY 86400` | the day-long back-off survives | the default 600 |

Two smaller ones came out of the same reading: `--dump` printed `DELAY` only when non-zero, so an explicit `DELAY 0` came back as the default 5, and `Postponing restart … for 0 seconds` was logged for a task with `FAILDELAY 0`, where the hold is disabled and nothing is postponed.

Reset to the defaults rather than to zero: for these two an absent keyword means "the default", where for the others it means "off". It costs no restart — both count into `changed` but not into `reload`, so the task keeps running with the new value.

**A refused read put back every field but these two.** A file caught mid-write is refused — *"Incomplete tasklist configuration - keeping the previous one"* — and keeping it means `restore_task()` putting back what the re-read cleared first. It restores fourteen fields and did not restore these. So a refused read left the defaults it had just installed: a task held by `DELAY 3600` became a task held by 5, with nothing in any config asking for it and the launcher having just said it kept the old settings. The reset above was right; the rollback had not been taught about it.

`tests/server/xymonlaunch-delay-reload.sh` drives a live xymonlaunch: a follower held back by `DELAY 3600` starts once the line is removed and the config re-read, and does *not* start when a mid-write config is refused. The premise is asserted first, and relative to the machine rather than to a number of seconds — a second dependent with no `DELAY` must have started in the same window, so the follower's absence is about `DELAY` and not about a slow poll. Reverting the two lines fails the final assertion; a `DELAY` never applied fails the premise. The refused-read case only bites if the refused config omits the `DELAY` line: one that still carries it sets the value again while parsing, so the rollback has nothing left to get wrong. The first version of that case did carry it and passed against the defect. Suite on a built tree: 105 passed, 0 skipped, 0 failed.

**Not covered, said rather than implied:** `FAILDELAY` takes the same reset with no behavioural assertion, since showing its back-off needs five consecutive failed starts — minutes of wall clock for one field. The man page comes from the devel port `065cfec0`; the original patch ships none.

### Suggested changelog entries (4.3.32)
```
* xymonlaunch: Add DELAY keyword for delaying initial task startup
* xymonlaunch: Add FAILDELAY keyword for configuring backoff period
  after too many (5) non-zero exit errors. Failure count is also
  cleared after this period passes without issue.
```


bonomani:
## Rebased onto the current #172

This branch was **47 commits stale** and re-bundled the whole runtime-directory stack: 8 commits, of which **6 were old copies of #219 + #172** — including the *pre-fix* versions of the SIGHUP relay (the worker-kill that dropped channel data) and the un-squashed #219. Only **2 were genuinely new** DELAY/FAILDELAY work.

Reworked to a clean **2-commit stack on the current #172 tip** (`3ebd43c6`), carrying only its own commits:

- `xymonlaunch: Add DELAY and FAILDELAY keywords (TBT 125)` — `common/xymonlaunch.c`
- `xymonlaunch: reset DELAY/FAILDELAY on a config re-read, and document them` — `xymonlaunch.c`, `tasks.cfg.5`, new `tests/server/xymonlaunch-delay-reload.sh`

It now inherits the **fixed** #172 (safe log rotation via the `@@logrotate` message path, no worker SIGHUP) and the **squashed** #219, so none of the earlier bugs are re-introduced. The six redundant commits are gone.

CI green on all variants (server, client×2, manpages).

**Merge order:** #163 independent; the runtime-directory chain is **#219 → #172 → #173** — this PR is last and must merge after #172.

Issue #170, opened 2026-06-14
Port the df/inode filesystem-filter options to the other xymonclient-*.sh scripts


bonomani:
Follow-up to #96 (env-var-controlled df/inode filesystem filter, #49), per @RolandRosenfeld's review.

#96 implements the filter only in `client/xymonclient-linux.sh`. The same options should be ported to the other still-used OS clients: **darwin, freebsd, netbsd, openbsd**.

### Scope (point 6)
Port the relevant subset of these settings to each `xymonclient-<os>.sh`:

- `XYMONCLIENT_FS_INCLUDE_TYPES` / `XYMONCLIENT_FS_EXCLUDE_TYPES`
- `XYMONCLIENT_FS_DF_LOCAL_ONLY`
- `XYMONCLIENT_FS_DF_TIMEOUT`

Notes:
- Each OS has a different `df` (flags, output columns, locality handling) and its own pseudo-/always-full filesystem set, so the defaults and the exclusion logic must be adapted per OS rather than copied verbatim.
- The linux client derives its `nodev` exclusion from `/proc/filesystems`, which the BSDs/macOS don't have — they need an OS-appropriate equivalent (or an explicit type list).

### Test coverage gap (point 7)
The current regression test drives `xymonclient-linux.sh` on every host (via `/proc/filesystems` + `df` mockups), so it validates the linux logic only and says nothing about the per-OS scripts actually used on macOS/BSD. As each OS client gains the options, add tests that exercise that OS's script (mocking its `df`/mount enumeration), so coverage matches the code actually run on that platform.

### Related
- #96 — the linux implementation this extends
- A separate follow-up will add a non-blocking df path for remote (NFS/CIFS) filesystems under `DF_LOCAL_ONLY=no`.

bonomani:
Started the port. One branch per OS (each = the #96 client filter + that OS's port), pushed to the `bonomani/xymon` fork:

- `feat/issue-170-fs-filter-freebsd` — `-t no<csv>` type list from INCLUDE/EXCLUDE_TYPES; nfs via `df -l` (DF_LOCAL_ONLY); DF_TIMEOUT wrapper; inode `-` (no-inode-limit) row drop; keeps the historical zfs-from-inode exclusion.
- `feat/issue-170-fs-filter-netbsd` — same, disk-only (NetBSD's client has no `[inode]` section).
[...]

PR #175, opened 2026-06-16
client/freebsd: env-var-controlled df/inode filesystem filter (#170)

bonomani:
Part of #170 — ports the df/inode filesystem-filter env vars to the FreeBSD client.

Adds `XYMONCLIENT_FS_INCLUDE_TYPES` / `_EXCLUDE_TYPES` / `_DF_LOCAL_ONLY` to
`xymonclient-linux.sh`'s FreeBSD sibling, building the df `-t no<csv>` exclude
list from a default set minus INCLUDE plus EXCLUDE (EXCLUDE wins). The inode
report also drops `zfs` (no usable inode counts). Remote mounts are hidden by
`df -l` (DF_LOCAL_ONLY=yes), not by type, so `no` can surface them. A df that
exits non-zero with no output is surfaced as a yellow failure marker, not read
as green. df is factored behind `run_df`/`emit_df` so the planned remote-df
sentinel (separate follow-up) hooks in additively.

Mock-tested in `tests/client/fs-filter-freebsd.sh` (argv construction,
Real FreeBSD df behaviour still to be verified on-platform.

Independent of the other #170 OS PRs (touches only `client/xymonclient-freebsd.sh`

**Update (on-platform verification):** running this branch on a real FreeBSD server surfaced that FreeBSD tmpfs reports `itotal = 2147483647` — the 2³¹−1 sentinel, identical on a 4GB and a 128GB machine. Not a limit (exhausting it would take ~9.5TB of RAM), so the metric carries no signal, and the `-` row-guard can't catch it because a number is printed. `e441c053c` excludes tmpfs from the **inode** report by type, next to zfs; the **disk** report keeps tmpfs (RAM-backed capacity is real, matching the Linux default) and mfs stays in both (UFS-in-RAM, genuine inode table). `INCLUDE_TYPES=tmpfs` remains the escape hatch, now pinned by a test.

### DIFF
--- a/client/xymonclient-freebsd.sh
+++ b/client/xymonclient-freebsd.sh
@@ -25,7 +25,10 @@ echo "[df]"
 # Exclude FS types via df "-t no<csv>": a default set, minus INCLUDE_TYPES, plus
 # EXCLUDE_TYPES. Remote (nfs) mounts are hidden by df -l (DF_LOCAL_ONLY=yes), not
 # by type, so DF_LOCAL_ONLY=no can surface them. The inode report also drops zfs
-# (no usable inode counts).
+# and tmpfs, whose inode counts carry no signal: zfs has no inode limit at all,
+# and FreeBSD tmpfs reports a constant itotal of 2^31-1 regardless of memory.
+# tmpfs/mfs are deliberately NOT excluded from the disk report: RAM-backed
+# capacity is real and fills up, matching the Linux client's default.
 : "${XYMONCLIENT_FS_INCLUDE_TYPES=}"
 : "${XYMONCLIENT_FS_EXCLUDE_TYPES=}"
 fs_excl_opt() {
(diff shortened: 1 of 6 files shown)


feld:
looks good at first glance but I need to test it directly to make sure I'm not overlooking any bash-isms as this will get executed with POSIX sh instead of bash on FreeBSD

client/xymonclient-freebsd.sh
```diff
@@ -21,19 +21,79 @@ uptime
...
+# (no usable inode counts).
+: "${XYMONCLIENT_FS_INCLUDE_TYPES=}"
+: "${XYMONCLIENT_FS_EXCLUDE_TYPES=}"
+fs_excl_opt() {
+	# $@ = extra per-report excludes (e.g. zfs for inode).
+	# noglob: treat a type like "tmp*" as a literal token, not a filename glob.
+	case $- in *f*) _restoreglob=no ;; *) _restoreglob=yes; set -f ;; esac
+	_l=""
+	# Default + per-report excludes, minus any INCLUDE_TYPES.
+	for _t in nullfs cd9660 procfs devfs linprocfs fdescfs autofs "$@"; do
```
feld, line 40: We also need to exclude tmpfs and mfs by default -- these are two memory filesystems that don't have real inodes either. Appears these were overlooked for a long time.

feld: keeping mfs after discussion about its characteristics (really UFS in memory, doesn't have crazy max inodes)

a 1GB mfs looks like this:

```
Filesystem 1K-blocks Used  Avail Capacity iused  ifree %iused  Mounted on
/dev/md1     1015324    8 934092     0%       2 131580    0%   /mnt
```

completely normal.

meanwhile a tmpfs looks like this:

```
Filesystem 1K-blocks Used   Avail Capacity iused      ifree %iused  Mounted on
tmpfs        4444980   12 4444968     0%      19 2147483628    0%   /tmp
```

completely impossible to use all those inodes, it takes many terabytes of memory

feld (approved):

PR #176, opened 2026-06-16
client/openbsd: env-var-controlled df/inode filesystem filter (#170)

bonomani:
Part of #170 — ports the df/inode filesystem-filter env vars to the OpenBSD client.

Adds `XYMONCLIENT_FS_INCLUDE_TYPES` / `_EXCLUDE_TYPES` / `_DF_LOCAL_ONLY`,
building the df `-t no<csv>` exclude list from a default set minus INCLUDE plus
EXCLUDE (EXCLUDE wins). Remote mounts are hidden by `df -l` (DF_LOCAL_ONLY=yes),
not by type, so `no` can surface them. A df that exits non-zero with no output is
surfaced as a yellow failure marker, not read as green. df is factored behind
`run_df`/`emit_df` so the planned remote-df sentinel (separate follow-up) hooks
in additively.

Mock-tested in `tests/client/fs-filter-openbsd.sh` (argv construction,
Real OpenBSD df behaviour still to be verified on-platform.

Independent of the other #170 OS PRs (touches only `client/xymonclient-openbsd.sh`

---
**Update (rebase + consistency pass):** rebased onto post-#175 main; test converted to the shared `fs-filter-common.sh` scaffolding and brought to family parity (INCLUDE/EXCLUDE leave-the-others asserts, invalid `DF_LOCAL_ONLY` warns).

**tmpfs is deliberately KEPT in both reports** — the opposite of FreeBSD's inode report, and source-verified: OpenBSD's `tmpfs` statfs reports memory-derived inode counts (`sys/tmpfs/tmpfs_vfsops.c`: `freenodes = MIN(limit remaining, avail mem / node size)`), a real exhaustion signal. FreeBSD's INT_MAX-capped `f_files` is its own divergence from this very code it ported. Pinned by test assertions so nobody "harmonizes" it into breakage.

**Open before un-drafting:** on-platform verification against real OpenBSD df/mount output (the #175 gate).

### DIFF
--- a/client/xymonclient-openbsd.sh
+++ b/client/xymonclient-openbsd.sh
@@ -20,17 +20,79 @@ uptime
 echo "[who]"
 who
 echo "[df]"
-df -k -tnonfs,kernfs,procfs,cd9660 | sed -e '/^[^ 	][^ 	]*$/{
+# --- filesystem filter (configurable; see xymonclient.cfg.DIST) --------------
+# Exclude FS types via df "-t no<csv>": a default set, minus INCLUDE_TYPES, plus
+# EXCLUDE_TYPES. Remote (nfs) mounts are hidden by df -l (DF_LOCAL_ONLY=yes), not
+# by type, so DF_LOCAL_ONLY=no can surface them.
+# tmpfs is deliberately kept in BOTH reports, unlike FreeBSD's inode report:
+# OpenBSD's tmpfs statfs reports memory-derived inode counts (f_files scales
+# with available RAM, sys/tmpfs/tmpfs_vfsops.c), a real exhaustion signal -
+# not FreeBSD's constant INT_MAX cap, which carries none.
+: "${XYMONCLIENT_FS_INCLUDE_TYPES=}"
+: "${XYMONCLIENT_FS_EXCLUDE_TYPES=}"
+fs_excl_opt() {
+	# noglob: treat a type like "tmp*" as a literal token, not a filename glob.
+	case $- in *f*) _restoreglob=no ;; *) _restoreglob=yes; set -f ;; esac
+	_l=""
+	# Default excludes, minus any INCLUDE_TYPES.
+	for _t in kernfs procfs cd9660; do
+		for _i in $XYMONCLIENT_FS_INCLUDE_TYPES; do [ "$_t" = "$_i" ] && continue 2; done
+		case " $_l " in *" $_t "*) ;; *) _l="$_l $_t" ;; esac
+	done
+	# EXCLUDE_TYPES last, so a type in both lists stays excluded (EXCLUDE wins).
+	for _t in $XYMONCLIENT_FS_EXCLUDE_TYPES; do
+		case " $_l " in *" $_t "*) ;; *) _l="$_l $_t" ;; esac
+	done
+	_csv=`echo $_l | tr ' ' ','`
+	[ "$_restoreglob" = yes ] && set +f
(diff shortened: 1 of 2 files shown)


SoundGoof:
## Findings

### Drop numeric zero-inode rows too

`client/xymonclient-openbsd.sh:94`

OpenBSD 7.9 reports filesystems without inode accounting as
`iused=0 ifree=0 %iused=100%`, not with `%iused=-`.

This was reproduced using a real CD9660 filesystem. It is
excluded by default, but setting
`XYMONCLIENT_FS_INCLUDE_TYPES=cd9660` exposes the row, which
the server interprets as 100% inode usage and can report red.

The test fixture uses `0 0 -`, so it misses the real OpenBSD
output. Match NetBSD's numeric guard:

`($6 + $7) <= 0 { next }`

Update the fixture to cover `0 0 100%`.

## BSD Comparison

- NetBSD excludes ZFS from inode collection and drops rows
  whose total inode count is zero.
- FreeBSD excludes both ZFS and `tmpfs` from inode
  collection.
- OpenBSD excludes CD9660 by default, but its configurable
  include list can expose CD9660's zero-total inode row.
- Disk reporting retains filesystems whose storage capacity
  remains meaningful.

## Verification

Tested directly on OpenBSD 7.9 GENERIC.MP using real FFS,
ext2fs, MSDOS, CD9660, and MFS filesystems.

- Default CD9660 exclusion passed.
- Custom filesystem exclusions passed.
- Exclude-over-include precedence passed.
- Explicit CD9660 inclusion reproduced `0 0 100%`.
- The PR script reproduced the unfiltered zero-total row.
- Stock OpenBSD 7.9 rejected `tmpfs` mounting as unsupported.
- Rebase onto current `main` and full build succeeded.
- Full suite: 38 passed, 0 skipped, 0 failed.

bonomani:
@SoundGoof thanks — fixed on your OpenBSD 7.9 finding. The inode report now drops zero-total rows via `($6 + $7) > 0` (NetBSD's numeric idiom) instead of the FreeBSD `%iused == "-"` text test, and the fixture uses the real `0 0 100%` shape, with a positive assert that a filesystem with real accounting is kept. Your tmpfs note is captured too — the retained-tmpfs path is moot since stock 7.9 won't mount tmpfs. I also wired `fs_excl_opt "$@"` so per-report excludes reach the `-tno` list, matching the FreeBSD sibling.

---

Release-notes text to stage on the `Changes` branch at merge — fold OpenBSD into the existing 4.3.31 fs-filter entry:

**RELEASENOTES (4.3.31):** add OpenBSD to the client list ("Linux, FreeBSD, NetBSD and macOS" → include OpenBSD) and to the `df -l` sentence ("The FreeBSD and NetBSD clients also pass -l to df" → "The FreeBSD, NetBSD and OpenBSD clients..."), so a remote non-nfs mount is hidden by default and `XYMONCLIENT_FS_DF_LOCAL_ONLY=no` restores it.

**Changes:**
> * Add a configurable df/inode filesystem filter to the OpenBSD client (#170, #176)

SoundGoof (approved):
No findings. The previous zero-inode-row issue is resolved.

## Verification

- Focused filesystem-filter test passed
- Full suite: 15 passed, 3 build-dependent skips, 0 failed
- Tested the exact script on OpenBSD 7.9
- Default, include/exclude, local-only, and invalid-value behavior passed
- POSIX shell parsing passed
- Current `main` merges cleanly
- `git diff --check` passed

Approved.

PR #177, opened 2026-06-16
client/netbsd: env-var-controlled df filesystem filter (#170)

bonomani:
Part of #170 — ports the df filesystem-filter env vars to the NetBSD client
(NetBSD has no `[inode]` section, so `[df]` only).

Adds `XYMONCLIENT_FS_INCLUDE_TYPES` / `_EXCLUDE_TYPES` / `_DF_LOCAL_ONLY`,
building the df `-t no<csv>` exclude list from a default set minus INCLUDE plus
EXCLUDE (EXCLUDE wins). Remote mounts are hidden by `df -l` (DF_LOCAL_ONLY=yes),
not by type, so `no` can surface them. A df that exits non-zero with no output is
surfaced as a yellow failure marker, not read as green. df is factored behind
`run_df`/`emit_df` so the planned remote-df sentinel (separate follow-up) hooks
in additively.

Mock-tested in `tests/client/fs-filter-netbsd.sh` (argv construction,
Real NetBSD df behaviour still to be verified on-platform.

Independent of the other #170 OS PRs (touches only `client/xymonclient-netbsd.sh`
+ its test). Docs are included: a NetBSD block in xymonclient.cfg.DIST and per-OS scoping in the xymonclient.cfg.5 entries (previously described as Linux-only).

---
**Update (rebase + consistency pass):** rebased onto post-#175 main; test converted to the shared `fs-filter-common.sh` scaffolding and brought to family parity (EXCLUDE leaves-the-defaults assert, invalid `DF_LOCAL_ONLY` warns).

**Why no `[inode]` handling:** verified end-to-end that NetBSD inode monitoring has never existed - the client script has no `[inode]` section and `xymond/client/netbsd.c` has no parsing for one - so this PR filters only the report that exists. Adding inode support is a separate two-sided feature; for whoever builds it: NetBSD's `df -i` exists, and NetBSD tmpfs reports memory-derived inode counts at source (`tmpfs_statvfs`, the computation OpenBSD inherited), so tmpfs would be KEPT there - FreeBSD's INT_MAX sentinel is FreeBSD-only divergence.

**Open before merge:** on-platform verification against real NetBSD df output (the #175 gate).

### DIFF
--- a/client/xymonclient-netbsd.sh
+++ b/client/xymonclient-netbsd.sh
@@ -20,10 +20,64 @@ uptime
 echo "[who]"
 who
 echo "[df]"
-df -P -tnonfs,kernfs,procfs,cd9660,null | sed -e '/^[^ 	][^ 	]*$/{
+# --- filesystem filter (configurable; see xymonclient.cfg.DIST) --------------
+# Exclude FS types via df "-t no<csv>": a default set, minus INCLUDE_TYPES, plus
+# EXCLUDE_TYPES. Remote (nfs) mounts are hidden by df -l (DF_LOCAL_ONLY=yes), not
+# by type, so DF_LOCAL_ONLY=no can surface them.
+: "${XYMONCLIENT_FS_INCLUDE_TYPES=}"
+: "${XYMONCLIENT_FS_EXCLUDE_TYPES=}"
+fs_excl_opt() {
+	# noglob: treat a type like "tmp*" as a literal token, not a filename glob.
+	case $- in *f*) _restoreglob=no ;; *) _restoreglob=yes; set -f ;; esac
+	_l=""
+	# Default excludes, minus any INCLUDE_TYPES.
+	for _t in kernfs procfs cd9660 null; do
+		for _i in $XYMONCLIENT_FS_INCLUDE_TYPES; do [ "$_t" = "$_i" ] && continue 2; done
+		case " $_l " in *" $_t "*) ;; *) _l="$_l $_t" ;; esac
+	done
+	# EXCLUDE_TYPES last, so a type in both lists stays excluded (EXCLUDE wins).
+	for _t in $XYMONCLIENT_FS_EXCLUDE_TYPES; do
+		case " $_l " in *" $_t "*) ;; *) _l="$_l $_t" ;; esac
+	done
+	_csv=`echo $_l | tr ' ' ','`
+	[ "$_restoreglob" = yes ] && set +f
+	[ -n "$_csv" ] && printf -- '-tno%s' "$_csv"
+}
+DFLOCALONLY="${XYMONCLIENT_FS_DF_LOCAL_ONLY:-yes}"
+case "$DFLOCALONLY" in
+	yes|no) ;;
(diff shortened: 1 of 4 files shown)


SoundGoof:
## Findings

### Document the NetBSD filesystem-filter contract — `client/xymonclient-netbsd.sh:24`

The new comment directs administrators to `xymonclient.cfg.DIST`, but that file documents these variables only for Linux and macOS.

The Linux documentation describes different defaults and Linux-specific `/proc/filesystems` behavior. It does not tell NetBSD users that:

- The default exclusions are `kernfs procfs cd9660 null`
- Exclusions are passed through NetBSD `df -t no<csv>`
- `XYMONCLIENT_FS_DF_LOCAL_ONLY=yes` adds `df -l`
- NetBSD has no inode report
- `EXCLUDE_TYPES` takes precedence over `INCLUDE_TYPES`
- A failed empty `df` result emits a yellow failure marker

The `xymonclient.cfg.5` manual also describes the variables as Linux settings.

Add a NetBSD section to the shipped configuration and manual in this PR. Otherwise the new public configuration interface is unavailable and potentially misleading to administrators.

## Verification

Tested on Fedora using mocked NetBSD `df` behavior.

- Exact-head suite: 15 passed, 3 skipped, 0 failed
- Rebased suite: 21 passed, 13 skipped, 0 failed
- Focused NetBSD filter test passed on both trees
- Negative control failed correctly against unpatched main
- POSIX shell syntax validation passed
- Test Bash syntax validation passed
- ShellCheck reported only expected source-following notices for the test
- NetBSD’s current `df(1)` manual confirms `-P`, `-l`, and `-t no<csv>` semantics
- The shared server parser confirms a headerless failure marker produces yellow

Additional probes confirmed:

- Defaults produce `-P -l -tnokernfs,procfs,cd9660,null`
- `DF_LOCAL_ONLY=no` removes `-l`
- Explicit exclusions override inclusions
- Invalid local-only values warn and fall back to `yes`
- Including every default type cleanly omits `-t`

Real execution against NetBSD’s `df` output was not available and remains an explicit verification gap in the PR description.

bonomani:
Addressed in 6eafc198b (still a single commit) — you're right that a public config surface shouldn't merge undocumented, so the docs moved into this PR instead of the planned follow-up.

**xymonclient.cfg.DIST**: a NetBSD block after the Linux one, covering the full contract from your list — default exclusions `kernfs procfs cd9660 null`, the `df -t no<csv>` mechanism, `-l` via `DF_LOCAL_ONLY` (with the same stale-remote-mount purple warning as Linux), EXCLUDE-wins precedence, no inode report, and the yellow marker on a failed empty df.

**xymonclient.cfg.5**: the INCLUDE/EXCLUDE entries no longer read as Linux settings — each now has an OS-neutral intro with explicit "On Linux:" / "On NetBSD:" paragraphs, structured so the pending macOS/FreeBSD doc additions can slot in as further per-OS paragraphs. The DF_LOCAL_ONLY entry was already OS-neutral and describes NetBSD's behavior exactly, so it's unchanged.

PR description updated to match (docs no longer deferred). The one open item remains the on-platform verification against real NetBSD df output, as flagged there. Manpage renders clean under groff -ww; focused test and suite unchanged and passing.

bonomani:
Superseded by #313, which extended this NetBSD filesystem-filtering approach to inode reports and is now merged.

PR #178, opened 2026-06-16
client/darwin: env-var-controlled df/inode filesystem filter (#170)

bonomani:
Part of #170 — ports the df/inode filesystem-filter env vars to the macOS client.

macOS df selects per path (not `-t`), so the filter is applied to `mount(8)`
output: drop the noise nobrowse/read-only attributes plus EXCLUDE_TYPES, keep
INCLUDE_TYPES (EXCLUDE wins). `DF_LOCAL_ONLY=yes` keeps only mounts mount(8)
flags `local` (MNT_LOCAL, the same flag `df -l` selects elsewhere), hiding remote
mounts (nfs, smbfs, afs, ...); `no` surfaces them. Empty output (every per-path
probe failed) is surfaced as a yellow failure marker, not read as green. The
per-path loop is factored behind `emit_df` and is the seam where the planned
remote-df sentinel (separate follow-up) will route non-local mounts.

Mock-tested in `tests/client/fs-filter-darwin.sh` (mount-based filter, inode
mode, empty-filter guard). Real macOS df behaviour still to be verified
on-platform.

Independent of the other #170 OS PRs (touches only `client/xymonclient-darwin.sh`

---
**Update (rebase + consistency pass):** rebased onto post-#175 main; test adopts `fsf_setup`/`fsf_extract` from the shared scaffolding. The runner stays deliberately custom: the snippet needs `bash` (macOS `/bin/sh` is bash; `IFS=$'\n'`), while `fsf_run` uses `/bin/sh` - commented in place so it doesn't get "simplified" back. Test coverage was already the family's most complete (invalid-value warn, precedence, literal-not-ERE, empty-filter guard, df-failure marker).

**Contract note, stated plainly:** on Darwin, `INCLUDE_TYPES`/`EXCLUDE_TYPES` operate on the mixed namespace of filesystem types *and* mount attributes as `mount(8)` prints them (`nobrowse`, `read-only`, `local`) - macOS df has no type selection, so the filter works on the mount list. Same variables, wider vocabulary; documented rather than pretended away.

**Open before un-drafting:** (1) on-platform verification against real macOS df/mount output (the #175 gate); (2) the APFS inode question - the tmpfs-sentinel analog: APFS's `df -i` free-inode figure is believed space-derived (ZFS-style, meaningless); measure `df -i /` on a real Mac and decide whether APFS joins the inode-report exclusion before this merges.

**Migration note (behavior change, the only removal-by-default in the #170 family):** the original script dropped only `afs` by name - **nfs and smbfs mounts were shown** in the disk report. The new default (`DF_LOCAL_ONLY=yes` keeps only `local`-flagged mounts) hides all remote mounts, consistent with df -l on the other clients. A macOS host that monitored an NFS/SMB mount through the disk report must set `XYMONCLIENT_FS_DF_LOCAL_ONLY=no` to keep it.

**Also fixed while porting (both test-pinned):** the original inode loop queried the 2nd+ filesystems with `df -P -H` (disk mode inside the inode report), and an all-excluded filter ran `df` with no path - listing every mount and defeating the filter. Latest push additionally sets noglob inside the per-path subshells so a mountpoint containing a glob character cannot expand against the working directory.

---
**Redesign for modern macOS (measured on a real Apple-silicon Mac, 2026-07).** Live `mount`/`df -i` output showed the previous defaults were pre-Catalina fossils: Apple now seals `/` read-only and flags every system volume - including the data volume holding all user files - `nobrowse`, so the default attribute drops excluded **every mount** on a modern Mac. (The original script masked this with its empty-list bug: a path-less `df` reported everything, filter defeated.) Three changes:

1. **`root data` exemption**: Apple marks the data volume with the `root data` attribute; it is exempted from the *default* drops, so the one volume that can actually fill is always reported. The sealed read-only root stays dropped (it cannot fill). An explicit `EXCLUDE_TYPES` entry still beats the exemption; `INCLUDE_TYPES=nobrowse` still resurrects the helper volumes.
2. **`apfs` excluded from the `[inode]` report**: measured `ifree` is container-derived - identical (3,780,253,960) across every volume of a container while `iused` varies, `%iused` pinned at 0% - the ZFS situation; no exhaustion signal exists. On an all-APFS Mac the inode list is legitimately empty and the section stays silent-empty (data, not failure).
3. **Empty disk list → yellow marker**: if a future macOS attribute change (or a config mistake) ever filters everything away again, the disk column goes visibly yellow instead of silently blank.

The test fixture is now the measured modern-Mac mount table (sealed root, nobrowse system volumes, root-data-flagged data volume) plus external/hfs/remote extras; new assertions pin the exemption, exclude-beats-exemption, the apfs-free inode report, the all-apfs silent-empty case, and the empty-list marker.

### DIFF
--- a/client/xymonclient-darwin.sh
+++ b/client/xymonclient-darwin.sh
@@ -26,29 +26,123 @@ uptime
 echo "[who]"
 who
 
-FILESYSTEMS=`mount | sed -E '/[\( ](nobrowse|afs|read-only)[ ,\)]/d;s/^.* on (.*) \(.*$/\1/'`
-
 echo "[df]"
-(IFS=$'\n'
- set $FILESYSTEMS
- df -P -H $1; shift
- while test $# -gt 0
- do
-   df -P -H $1 | tail -1 | sed 's/\([^ ]\) \([^ ]\)/\1_\2/g'
-   shift
- done) | column -t -s " " | sed -e 's!Mounted *on!Mounted on!'
+# --- filesystem filter (configurable; see xymonclient.cfg.DIST) --------------
+# macOS df selects per path, not with -t, so the filter is applied to the
+# mount(8) list: drop the noise nobrowse/read-only attributes plus EXCLUDE_TYPES,
+# keep INCLUDE_TYPES.
+: "${XYMONCLIENT_FS_INCLUDE_TYPES=}"
+: "${XYMONCLIENT_FS_EXCLUDE_TYPES=}"
+# DF_LOCAL_ONLY: "yes" (default) keeps only mounts mount(8) flags "local"
+# (MNT_LOCAL, what df -l selects elsewhere), hiding remote mounts (nfs, smbfs,
+# afs, ...); "no" surfaces them; invalid warns and uses yes. (df is per-path
+# here, where -l is ignored, so we filter mount(8) output instead.)
+DFLOCALONLY="${XYMONCLIENT_FS_DF_LOCAL_ONLY:-yes}"
+case "$DFLOCALONLY" in
+	yes|no) ;;
+	*) echo "xymonclient-darwin: invalid XYMONCLIENT_FS_DF_LOCAL_ONLY '$DFLOCALONLY', using yes" >&2; DFLOCALONLY=yes ;;
+esac
+# Escape ERE metachars (and the sed delimiter) so a type is matched literally in
+# the mount(8) filter, not as a regex (e.g. "fuse.*" must not match "fuseblk").
+_re_escape() { printf '%s' "$1" | sed 's,[]\^$.*+?()[{}|/],\\&,g'; }
+# noglob: treat a token like "tmp*" literally, not as a filename glob.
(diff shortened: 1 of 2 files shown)


bonomani:
**On-platform verification** (Apple-silicon Mac mini, macOS with sealed root + nobrowse system volumes — same machine as the `mount`/`df -i` measurements the redesign is built on). Default run of the branch script:

```
[df]
Filesystem    Size  Used  Avail  Capacity  Mounted on
/dev/disk3s5  995G  584G  387G   61%       /System/Volumes/Data
[inode]
[mount]
```

Exactly the designed behavior: the root-data exemption reports the one volume that can fill (61% - the real signal on this machine), the sealed read-only root and the nobrowse helpers are dropped, and the `[inode]` section is legitimately empty on an all-APFS system - no rows, no failure marker. Under the previous defaults this machine's mount table filtered to an empty list, i.e. the pre-redesign branch would have blanked disk monitoring here entirely.

PR #181, opened 2026-06-19
xymonnet: built-in SNTP probe for the ntp test (fixes #27, supersedes #158)

### LINKED ISSUE #27: [Xymon] Patch: ntpdate option -p deprecated

bonomani:
## Summary

Replaces the external `ntpdate`/`sntp` call-out for the `ntp` test with a **built-in, in-process SNTP probe** (RFC 4330/5905): one 48-byte UDP exchange on port 123, no fork, no external program — it works even when no NTP client is installed. The probe validates replies (origin-echo with a CSPRNG nonce + `connect()`-based source filtering) and retransmits on packet loss.

The port is fixed for the built-in path — the `ntp` test has no per-host port override. The probe passes the `"ntp"` service name to `udp_query_opt()`, which hands it to `getaddrinfo()`; that resolves it to 123/udp via `/etc/services` (present on essentially every system). This reads more clearly than a bare `123` at the call site and stays correct on the rare host that remaps the service.

The `ntp` test's clock offset is recorded by `do_net.c` into a **dedicated `ntp.rrd`** — DSes `offset` and `rootdist`, in milliseconds — and shown by a **new `[ntp]` graph**: the offset plotted as a line within its `offset ± root distance` (max-error) band, with a zero reference rule and a `cur/max/min/avg` legend. The legacy `time=ntpstat` test path is left untouched.

**Fixes #27.** **Supersedes #158** (the ntpdig auto-detect approach).

## Bundled standalone fixes (first 3 commits)

So the branch builds on its own, the first three commits are independent fixes that each have their own PR. Once they land in `main`, rebasing drops them (patch-id) and they vanish from the diff — no action here:

- **#185** — `do_ntpstat` out-of-bounds read (reached on every `offset=`-prefixed sample `do_net.c` forwards); also folds the two parse branches into an `ntpstat_offset()` helper
- **#197** — ntp offset recorded in ms (was ~1000× too low: seconds stored in a ms DS) on the `time=ntpstat` path
- **#191** — drop obsolete `ntpdate -p 1`

Review commits 4–7 for the probe work, recording + graph, and tests.

## Why the internal probe

This came out of @jezzaaa's review on #158: scanning for and parsing external NTP clients adds complexity without a fundamental monitoring benefit, and NTP is simple enough to implement in-process like Xymon's other built-in probes.

> … the NTP protocol is simple enough that the NTP probe could be embedded into the Xymon code, and do away with the added complexity of scanning for a suitable NTP client … This would bring the relatively simple (s)ntp protocol inline with far more complex protocols that are implemented internally by xymonnet, such as the simple TCP services … TLS … HTTP/HTTPS … STARTTLS (LDAPS), and DNS.
> … a fork/exec every five minutes for dozens of NTP servers may have a measurable impact on a heavily-loaded and modestly-resourced Xymon server.
> — @jezzaaa, [#158](https://github.com/xymon-monitoring/xymon/pull/158)

SNTP is a single UDP request/response — simpler than the TLS/DNS probes xymonnet already does in-process. Internalising it removes the per-host fork/exec, removes the runtime tool-scanning entirely, and makes the `ntp` test work out-of-the-box (classic `ntpdate` is gone from current distros, which is what broke the test).

## How the offset is recorded

The probe emits a status banner carrying `offset <seconds>` and a `+/- <seconds>` root distance (the RFC 5905 accuracy bound). `do_net.c` parses it — alongside the `ntpdate`/`sntp` output forms — validating each numeric token with `strtod()` end-pointer checks, so a malformed token (e.g. an error word) is **skipped, not recorded as a misleading `0`** (ntpdate's trailing-comma `offset <sec>,` still parses). Both values are recorded, in milliseconds, into a dedicated **`ntp.rrd`** (`setupfn("%s.rrd","ntp")`; DSes `offset`, `rootdist`; root distance stored `U` when a backend such as `ntpdate` omits it).

A new **`[ntp]` graph** draws the offset line inside its `offset ± root distance` band, with a grey zero rule and a conventional `cur/max/min/avg` legend for both series. `TEST2RRD` routes `ntp=tcp` (through `do_net_rrd`) and `GRAPHS_ntp="ntp.offset"` selects the graph — both shipped in `xymonserver.cfg.DIST` and the compiled-in `environ.c` fallback, and documented in `xymond_rrd.8`.

This is intentionally a **dedicated RRD**, not a reuse of `ntpstat.rrd`: the `ntp` test (network probe) and the `time=ntpstat` test (client-side `ntpq`) are separate columns, so giving the probe its own RRD avoids conflating them and lets the graph carry the root-distance band the probe uniquely provides. The `time=ntpstat` path (and the bundled **#197** ms fix for it) is unchanged.

## Comparison vs #158 (ntpdig auto-detect)

Both solve #27, keep external tools available, and parse the same offset banner; they are **mutually exclusive** (they conflict in `run_ntp_service`, `do_net.c`, `lib/environ.c`, `xymonserver.cfg.DIST`, manpages).

| | this PR (internal probe) | #158 (ntpdig) |
|---|---|---|
| Query mechanism | in-process SNTP (raw UDP, no fork) | external command runs the chosen tool |
| Default dependency | **none** — works with no NTP client | requires ntpdig/ntpdate/sntp/chronyd |
| Tool override | `NTPDATE` empty = built-in (default); else run that command | priority list, first installed wins |
| Spoof/loss resistance | CSPRNG origin nonce + `connect()` filter + retransmit | relies on the external tool |
| RRD / graph | **new `ntp.rrd`** (offset + root distance) + **`[ntp]` graph** | existing `ntpstat.rrd` |
| Reusable transport | new `udpquery.c/.h` (IPv4/IPv6, reusable by DNS/SNMP/RADIUS) | none |

Totals: this PR **+1879 / −110 (23 files)** — over half is tests (+998, 6 files); the probe + transport live in three new files (`ntpprobe.c`, `udpquery.c/.h`), and the recording + graph is `do_net.c` (+39/−13) plus `graphs.cfg`, `xymonserver.cfg.DIST`, `environ.c` and `xymond_rrd.8`.

## What's in it (7 commits)

1. **#185** — `do_ntpstat` out-of-bounds read fix, plus folding the two parse branches into an `ntpstat_offset()` helper (bundled; own PR).
2. **#197** — record the `time=ntpstat` offset in milliseconds (bundled; own PR).
3. **#191** — drop `ntpdate -p 1` (bundled; own PR).
4. **`udpquery`** — reusable single-shot UDP request/response transport (with its Makefile wiring): IPv4/IPv6, `connect()`-validated source, precise send/recv timestamps, optional opts (sub-second timeout, `MSG_TRUNC` detection). The wall-clock-step rejection is gated on whether timestamps are requested, so non-NTP probes can reuse it.
5. **built-in SNTP probe** (`ntpprobe.c`) — the probe is the default; `NTPDATE` holds an external `ntpdate` program to run instead (empty = built-in; on ntpsec, `ntpdig`), with a legacy explicit `SNTP` still forcing sntp. Anti-spoof nonce (`getrandom()` → `/dev/urandom` → portable fallback), retransmit with a short per-attempt timeout (dead server down in ~6 s, not ~29 s), overflow-safe offset math against a malicious/garbage server.
6. **record the offset + add the graph** — `do_net.c` matches the shared `offset `/`+/-` tokens (not the literal `ntpdate`), validates them with `strtod()`, and records offset + root distance into a dedicated **`ntp.rrd`**; adds the **`[ntp]` graph** (offset within its ±root-distance band, zero rule, `cur/max/min/avg` legend), the `ntp=tcp` / `GRAPHS_ntp` defaults (`xymonserver.cfg.DIST` + `environ.c`), and the `xymond_rrd.8` entry.
7. **tests** — the three ASan/UBSan harnesses (`udp-query`, `ntp-probe`, `ntp-offset-parse`) in their own commit, leaving commits 4–6 code-only.

## Backward compatibility

- The default flips from `ntpdate` to the built-in probe (empty `NTPDATE`) — the #27 fix; no config needed.
- External tools stay available via `NTPDATE` — set it to the command (e.g. `NTPDATE="ntpdate"`, `NTPDATEOPTS="-u -q"`, or `NTPDATE="ntpdig"`); a legacy explicit `SNTP` still forces sntp. Existing `NTPDATE="ntpdate"` configs are **honored, not overridden**.
- The `ntp` test now records to a new `ntp.rrd`, created on the first sample; the `time=ntpstat` test is unchanged. No RRD migration.

## Upgrade notes

- **New `ntp.rrd` + `[ntp]` graph.** The `ntp` test now records offset and root distance (ms) to a dedicated `ntp.rrd` and shows the new `[ntp]` graph. It starts fresh, so there's no historical step. (The bundled **#197** separately corrects the legacy `time=ntpstat` graph, whose `offsetms` DS used to store seconds — ~1000× low; existing ntpstat graphs show a one-time step where the corrected values begin.)
- **The built-in probe validates more strictly than `ntpdate`.** It enforces RFC 4330/5905 sanity (root distance ≤ 1 s, `leap ≠ 3`, stratum ≤ 15, an originate-timestamp echo, a non-zero reference timestamp). A few unusual-but-working servers that `ntpdate`/`sntp` tolerated — very distant / high-dispersion sources, middleboxes that don't echo the originate field, minimal SNTP or GPS appliances that send `reftime = 0` — may now report red. If you hit that, set `NTPDATE` to your `ntpdate`/`ntpdig` program to keep the old external-tool behaviour.

## Failure handling — why the probe retries, and what's left to config

A failed poll can mean two very different things — a single lost UDP datagram from a *healthy* server, or a *genuinely* unreachable one — and they are handled by three independent layers, only the **first** of which is new here:

1. **In-poll retransmission (new, in the probe).** Up to `NTP_PROBE_TRIES` (3) attempts of `NTP_PROBE_TRY_SEC` (2 s) each, **stopping on the first reply** — so a healthy server costs exactly one packet and a dead host is bounded at ~6 s. This filters transient UDP loss: a lost datagram is retried in-poll, so it never registers as a failure and the offset point is still recorded. (UDP is single-shot; without this, one dropped packet reports a reachable server as down.)
2. **Fast re-test (existing, `xymonnetagain`).** Failed network tests are re-run once a minute for up to 30 min (`frequenttestlimit`), so a short outage recovers in ~1 min instead of waiting for the 5-min cycle.
3. **Escalation (existing, `delayred`/`delayyellow`, or the deprecated `bad<test>`).** A genuine, sustained failure is paced green→yellow→red over a configurable delay, per host or globally via `DELAYRED`/`DELAYYELLOW` — e.g. `delayred=ntp:15`.

These are **sequential, not redundant**: layer 1 decides *whether a poll really failed*; layer 3 decides *how fast a real failure alarms*. Escalation alone can't tell a lost packet from an outage, so without the in-poll retries a healthy-but-lossy server would feed false failures into the escalation and could flicker to red.

**No new retry/escalation configuration knob is introduced.** This PR changes the default `NTPDATE` value to empty to select the built-in probe, but the tries (3) and per-try timeout (2 s) are protocol constants — an NTP RTT is tens of ms, so 2 s is ample, and `--cmdtimeout` (30 s, for fork/exec'd commands) is the wrong granularity for an in-process probe that already self-bounds at ~6 s. The escalation threshold is already configurable through the existing `delayred`/`delayyellow` (and the deprecated `bad<test>`) mechanisms, so no ntp-specific knob is added.

Summary of every lever involved:

| Concern | Mechanism | Where | Default |
|---|---|---|---|
| In-poll tries | `NTP_PROBE_TRIES` | compile-time constant (`ntpprobe.c`) | `3` |
| Per-try timeout | `NTP_PROBE_TRY_SEC` | compile-time constant (`ntpprobe.c`) | `2 s` |
| Total probe budget | tries × timeout, stop-on-success | self-bounded | ~6 s for a dead host |
| Fast re-test of failures | `xymonnetagain` / `frequenttestlimit` | `tasks.cfg` (`INTERVAL 1m`) + built-in | every 1 min, up to 30 min |
| Escalation (time-based) | `delayred` / `delayyellow`; global `DELAYRED` / `DELAYYELLOW` | `hosts.cfg` / `xymonserver.cfg` | off → immediate red |
| Escalation (count-based, **deprecated**) | `bad<TESTNAME>:x:y:z` (e.g. `badntp:1:1:2`) | `hosts.cfg` | off → immediate red |
| Soften red → yellow | add `ntp` to `NONETPAGE` | `xymonserver.cfg` | off |

The probe ships with no escalation default (like every other test: immediate red), leaving the green→yellow→red ramp entirely to the existing per-host/global mechanisms above.

## Concurrency and future direction

The probe queries hosts **serially**, exactly as the external path did (not a regression; the retransmit loop caps a dead host at ~6 s). No private NTP scheduler was added: protocol logic is in pure functions (`ntp_build_request`, `ntp_eval`) over the thin swappable `udp_query_opt` transport, so it can drop into the unified `conn_*` `select()` engine (`lib/tcplib.c`) when that drives xymonnet. The transport guarantees (source filtering, EINTR, retransmit, clock-step rejection) are harness-pinned.

## Hardening

- **No per-host shell-out on the default path** — the built-in probe replaces a `run_command()` fork/exec per host.
- **Untrusted input is bounds-checked** — the 48-byte reply is parsed behind a length gate at fixed offsets (no over-read on an oversized datagram); `recv` is capped by the caller's buffer; no network data reaches a format string.
- **Recorded values are validated** — the offset and root-distance tokens are parsed with `strtod()` end-pointer checks, so a non-numeric token is skipped rather than written as a misleading `0` to `ntp.rrd`.
- **Audited** for memory/fd leaks, double-free/close, uninitialised vars (`-Wmaybe-uninitialized`) and injection/format-string issues. One malicious-server signed-overflow path in the offset average was found and fixed (overflow-safe averaging). The nonce is CSPRNG-backed and layered on `connect()` source filtering; plain SNTP stays on-path-spoofable by design (this is a reachability probe, not a time source).

## Testing

Unit harnesses under AddressSanitizer + UndefinedBehaviorSanitizer, driving the real production code:

- **`ntp-probe-harness`** — packet build/validate/offset math: the NTP formula; rejection of origin-echo mismatch, leap=3, Kiss-o'-Death / out-of-range stratum, wrong mode, short packet, zero/reversed server timestamps; raw 64-bit `(int64_t)(a-b)` differencing across the 2036 era rollover; the retransmit-and-give-up loop.
- **`udp-query-harness`** — round-trip, ordered timestamps, closed-port error, a real timeout against a silent listener, sub-second timeout, truncation (detected vs silent), the NULL-timestamp (non-NTP) path, source filtering (reply from a wrong port dropped), and EINTR resilience.
- **`ntp-offset-parse-harness`** — the real `do_net.c` recording path across the built-in probe banner, `ntpdate` and `sntp`: offset + root distance scaled to ms into `ntp.rrd`, a **non-numeric offset recording nothing**, and the `do_ntpstat` `offset=`-at-start case (#183).
- **Live-verified** against Cloudflare, Google and two Swiss-pool servers — offsets agree within **~4 ms** across independent stratum-1/2/3 sources; the timeout and no-listener error paths return correctly.


bonomani:
@jezzaaa I'm assigning this to you — but only joking: no obligation at all. I'd love your eyes on it if you have the time and interest, but it's completely fine if you can't. Your earlier feedback already shaped this direction and was hugely appreciated.

On size: the production code is ~500 lines (~2× #158), but it's much simpler to *review*. #158 rewrites the parser shared by every tcp test (`do_net.c`); this PR deletes that shared branch (−54) and puts ~84% of its code in four small self-contained files, each with its own test and readable one commit at a time.

jezzaaa:
Some great work, particularly updates to the man page about the undocumented SNTP variables.

However, I have some concerns (possibly unfounded) about some aspects of the implementation. I'll put some more detail together later today, but in brief:
* The new probe appears to be (to my uneducated eyes) synchronous, in contrast to most xymonnet probes, including the current code that executes $NTPDATE, which are all asyncrhonous. With a 30-second timeout, a few NTP servers with issues could cause the probe to delay all other remaining xymonnet tests, and increase the xymonnet execution time.
* I think the creation of a new NTPTOOL variable adds complexity, and I think the code and the documentation would be simpler if the internal probe can be selected by an appropriate setting of the existing NTPDATE variable.
* The port number "123" should be obtained using getservbyname("ntp"...)
* I'm don't think the RRD file should switch from "ntpstat.rrd" to "ntp.rrd". This would cause existing historical data to stop updating, and new stats would have no history. I believe the rationale for this may be based on an incorrect assumption that client-side stats (in the clientlog) somehow provides NTP offset data.

jezzaaa (approved):
I'm not a C coder, so my review is for the intention of the code, as far as I can understand it from changes to documentation. It seems Ok to me.

PR #182, opened 2026-06-19
lib/matching: use strcasecmp for case-insensitive literal name matching

### LINKED ISSUE #192: alerts.cfg/analysis.cfg literal name matching is case-sensitive while host loading is case-insensitive (rules silently don't match)
## Summary

Xymon **loads and looks up hostnames case-insensitively**, but **matches them
case-sensitively** in config rules (`alerts.cfg`, `analysis.cfg`, …). The result: a host
is monitored correctly, yet alert/analysis rules that reference it with different
letter-case **silently never apply** — no error, no warning.

## The mismatch

Three matching paths, two of them already case-insensitive, one not:

| Path | Comparator | Case-insensitive? |
|---|---|---|
| Host tree load/lookup (`lib/loadhosts.c:318`) | `xtreeNew(strcasecmp)` | ✅ |
| Regex config patterns (`%…`) (`lib/matching.c:46,51`) | `PCRE2_CASELESS` | ✅ |
| **Literal config patterns** (`lib/matching.c` `namematch()`) | **`strcmp`** | ❌ |

So a host stored as `MyServer` (the host tree folds case) is **not** matched by a literal
`HOST=myserver` rule, because that path uses exact `strcmp`.

## Reproduction

1. `hosts.cfg`: define a host `MyServer`.
2. `alerts.cfg`: `HOST=myserver` … (any rule).
3. The rule never fires for that host. Switching the rule to exact case (`HOST=MyServer`)
   or to a regex (`HOST=%myserver`) makes it work — proving the literal path is the lone
   case-sensitive outlier.

Same applies to other literal dimensions routed through `namematch()`.

## Scope (verified)

`namematch()` is called from **42 sites in 2 files** (`lib/loadalerts.c`,
`xymond/client_config.c`). At **every** site the regex arg is compiled `PCRE2_CASELESS`
(via `compileregex()` / `multilineregex()`) — there is **no** call site where regex
matching is case-sensitive. So the literal `strcmp` path is inconsistent with the regex
path everywhere, not just for `HOST=`.

## Fix

PR #182 changes the two literal comparisons in `namematch()` from `strcmp` to
`strcasecmp`, aligning the literal path with the already-case-insensitive regex and
host-tree paths.

Note: this also makes literal matching case-insensitive for inherently case-sensitive
dimensions (filesystem/file paths, process names, IBM MQ object names). That is **not a
regression** — the regex path already case-folds those — but it does remove the last
case-sensitive matching option. Whether any of those dimensions *should* be matched
case-sensitively (MQ names in particular are case-sensitive by spec) is a separate,
larger question, tracked independently if desired.

StefCoene:
Fixes #192 — root-cause issue (case-insensitive host load vs case-sensitive literal match).

## Summary

- `namematch()` in `lib/matching.c` used `strcmp` for literal string patterns (`HOST=`, `SERVICE=`, `PAGE=`, `CLASS=`, etc. in `alerts.cfg`, `analysis.cfg` and similar config files), making them case-sensitive
- Regex patterns (with `%`-prefix) were already case-insensitive via `PCRE2_CASELESS`
- The host lookup tree in `loadhosts.c` already uses `strcasecmp` as its comparator
- This patch replaces the two `strcmp` calls in the literal-match path with `strcasecmp`, aligning all three code paths

## Test plan

- [ ] Verify `HOST=MyServer` in `alerts.cfg` matches a host defined as `myserver` in `hosts.cfg`
- [ ] Verify `HOST=myserver` matches `MYSERVER`
- [ ] Verify negation still works: `HOST=!MyServer` excludes the host regardless of case
- [ ] Verify comma-separated lists work: `HOST=myserver,OtherHost`
- [ ] Verify regex patterns (`HOST=%myserver.*`) are unaffected
- [ ] Run regression suite: `./tests/testsuite`

### DIFF
--- a/lib/matching.c
+++ b/lib/matching.c
@@ -100,11 +100,11 @@ int namematch(const char *needle, char *haystack, pcre2_code *pcrecode)
 
 		if (!found) {
 			if (*tok == '!') {
-				found = (strcmp(tok+1, needle) == 0);
+				found = (strcasecmp(tok+1, needle) == 0);
 				if (found) result = 0;
 			}
 			else {
-				found = (strcmp(tok, needle) == 0);
+				found = (strcasecmp(tok, needle) == 0);
 				if (found) result = 1;
 			}
 		}


bonomani (commented):
## Review — recommend merge, with two small follow-ups

Analyzed this end-to-end. The change is correct and, importantly, it doesn't *introduce* case-insensitivity to those config dimensions — it makes the **literal** match path agree with the **regex** path, which Xymon has already compiled case-insensitively for years. Details and verification below.

### Why it's a genuine fix, not just a tweak
- `namematch()`'s regex branch already matches case-insensitively — `compileregex()`/`multilineregex()` both pass `PCRE2_CASELESS` (`lib/matching.c:46,51`).
- The host tree is already case-insensitive — `lib/loadhosts.c:318 xtreeNew(strcasecmp)`.
- So the literal `strcmp` path was the lone exception: `HOST=MyServer` silently failing to match host `myserver` is a real, common footgun. This aligns the three paths.

### Verification we ran (blast-radius check)
`namematch()` is generic — also used for filesystems, files, process names, IBM MQ names, etc., which *can* be case-sensitive. So I enumerated **every** call site and checked how each one's regex is compiled:

| Caller | Sites | Regex compiled via | Case-insensitive? |
|---|---|---|---|
| `lib/loadalerts.c` | 12 (host/page/class/group/dg/service) | all `compileregex(spec+1)` | ✅ `PCRE2_CASELESS` |
| `xymond/client_config.c` | rest (incl. **fs / file / proc / MQ / log**) | all `setup_expr()` → `compileregex()` / `multilineregex()` | ✅ `CASELESS` (`+MULTILINE` for log) |

**42 call sites, 2 files, zero exceptions** — there is no caller anywhere that compiles a `namematch` regex case-sensitively (no raw `pcre2_compile` without `CASELESS`). So flipping `strcmp`→`strcasecmp`:
- removes the literal-vs-regex inconsistency at **every** site, and
- creates **no** new inconsistency (there's no site where regex matching is case-sensitive to disagree with).

This is why I'd **not** try to keep fs/file/MQ case-sensitive: there's no site where case-sensitive literal matching would even be self-consistent, and Windows disk/file targets actually want case-folding anyway.

### Code itself
Mechanically clean: both literal comparisons flipped (positive **and** negation), the `strcmp(haystack, "*")` wildcard check correctly left alone, builds trivially (`strcasecmp` is already used across the tree).

### Two small follow-ups before merge
1. **Document the scope** (PR body + a release-notes line): *"Literal config-name matching is now case-insensitive, matching the long-standing behavior of `%`-regex patterns."* Note it applies to **all** dimensions (host, page, class, group, service, **filesystem, file, process, MQ, log**) — not just `HOST=` — so the broader behavior is a deliberate, documented decision.
2. **Add one non-host test** (e.g. a `DISK=`/`SERVICE=` case check) so the wider effect is actually exercised, alongside the existing `HOST=` cases.

(No separate `devel`/4.4 change needed — that branch inherits `main`.)

bonomani (approved):
Approving — the fix is correct and well-scoped (verified: all 42 `namematch()` call sites already use `PCRE2_CASELESS` for regex, so this only aligns the literal path with the existing behavior; root cause documented in #192).

Two non-blocking follow-ups to fold in before merge (no code change needed):
1. Note in the description + release notes that this makes literal matching case-insensitive across **all** dimensions routed through `namematch()` (host, page, class, group, service, filesystem, file, process, MQ, log), not just `HOST=`.
2. Add one non-host test (e.g. a `DISK=` / `SERVICE=` case check) to the regression suite alongside the `HOST=` cases.

PR #185, opened 2026-06-20
xymond: fix out-of-bounds read in do_ntpstat offset parser

### LINKED ISSUE #183: do_ntpstat: out-of-bounds read when "offset=" is at the start of the message
`do_ntpstat_rrd()` finds the `offset=` token and inspects the character *before* it to confirm a field separator:

```c
p = strstr(msg, "offset=");
if (p && (isspace((int)*(p-1)) || (*(p-1) == ','))) {
```

When the token starts the message, `p == msg` and `*(p-1)` reads one byte before the buffer — out-of-bounds (UB). `isspace()` is also passed a possibly-negative `char`.

This is reached in normal operation: `do_net.c` records the ntp test offset by forwarding a message that begins with `offset=`, so `p == msg` on every `ntpdate`/`sntp` sample. The offset can also be silently dropped, since the out-of-bounds byte is rarely a separator.

bonomani:
Fixes #183.

The separator check before `offset=` only guarded against the token being the tail of a longer field name — which `ntpq -c rv` never emits, and which it couldn't recover from anyway (no retry on the next match). Remove it and parse the token directly, so the out-of-bounds read and the `isspace()` UB are gone by construction.

This also makes the `ntpdate`/`sntp` offset — forwarded by `do_net.c` as a message starting with `offset=` — record reliably.

While here, fold the two near-identical format branches (old LARRD `Offset:` line, `ntpq -c rv` output) into a small `ntpstat_offset()` helper.

### DIFF
--- a/xymond/rrd/do_ntpstat.c
+++ b/xymond/rrd/do_ntpstat.c
@@ -10,35 +10,28 @@
 
 static char ntpstat_rcsid[] = "$Id$";
 
+/* Find a marker in the message and read the number that follows it. */
+static int ntpstat_offset(const char *msg, const char *marker, double *offset)
+{
+	const char *p = strstr(msg, marker);
+	return p && sscanf(p + strlen(marker), "%lf", offset) == 1;
+}
+
 int do_ntpstat_rrd(char *hostname, char *testname, char *classname, char *pagepaths, char *msg, time_t tstamp)
 {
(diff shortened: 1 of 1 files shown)


bonomani:
The `do_ntpstat.c` offset-parser hardening here landed via #181 — `main`'s `do_ntpstat.c` is byte-identical to this PR's result (the `ntpstat_offset()` helper). The out-of-bounds read is fixed in `main`. Closing as already-merged (via #181).

PR #186, opened 2026-06-20
xymonnet: NUL-terminate cmdpath in run_rpcinfo_service

### LINKED ISSUE #184: xymonnet run_rpcinfo_service: strncpy leaves cmdpath potentially unterminated
## Problem

In `run_rpcinfo_service()`:

```c
p = xgetenv("RPCINFO");
strncpy(cmdpath, (p ? p : "rpcinfo"), sizeof(cmdpath));
```

`strncpy()` does not NUL-terminate the destination when the source is at least `sizeof(cmdpath)` bytes. A `RPCINFO` value of `PATH_MAX` or more leaves `cmdpath` unterminated, and the subsequent use as a C string reads past the buffer.

## Scope

Pre-existing latent bug on `main`, in the RPC path.

## Fix

Force termination after the copy:

```c
strncpy(cmdpath, (p ? p : "rpcinfo"), sizeof(cmdpath));
cmdpath[sizeof(cmdpath)-1] = '\0';
```

bonomani:
Fixes #184

`strncpy()` does not terminate the destination when the source is at least `sizeof(cmdpath)` bytes, so an `RPCINFO` value of `PATH_MAX` or longer left `cmdpath` unterminated and the later use as a C string read past the buffer.

This forces termination after the copy.

Pre-existing latent bug on `main`, in the RPC path.

Same idiom as Henrik's 2005 termination sweeps (`338496e18`, `6a6f92d12`), which fixed other strncpy sites but missed this one.

### DIFF
--- a/xymonnet/xymonnet.c
+++ b/xymonnet/xymonnet.c
@@ -1105,6 +1105,7 @@ void run_rpcinfo_service(service_t *service)
 
 	p = xgetenv("RPCINFO");
 	strncpy(cmdpath, (p ? p : "rpcinfo"), sizeof(cmdpath));
+	cmdpath[sizeof(cmdpath)-1] = '\0';	/* strncpy may not terminate */
 	for (t=service->items; (t); t = t->next) {
 		/* Do not run RPCINFO test if host does not resolve in DNS or is down */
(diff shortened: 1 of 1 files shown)


RolandRosenfeld (approved):
Looks good to me.

PR #190, opened 2026-06-21
xymond: drop non-finite offsets in do_ntpstat RRD handler

### LINKED ISSUE #189: do_ntpstat_rrd stores infinite clock offsets (inf) in the RRD, corrupting the graph
### Component
xymond / RRD handlers (`xymond/rrd/do_ntpstat.c`)

### Description
`do_ntpstat_rrd()` extracts the clock offset from the client `ntpstat` test
output (`ntpq -c rv`, or the legacy LARRD `Offset:` script) with
`sscanf("%f", …)`, which parses `nan`/`inf` as valid floats, then writes the
value into the RRD update string with `"%.6f"`.

RRDtool **accepts** such an update (no error). Verified against RRDtool 1.7/1.8
(librrd) through the C API:

- `offset=nan` → recorded as an UNKNOWN datapoint — harmless, equivalent to a
  gap (`U`);
- `offset=inf` / `offset=-inf` → stored as **real infinities** in the RRA, which
  propagate through AVERAGE/MAX consolidation and corrupt the graph and its
  aggregates.

So a malformed or hostile feed emitting `offset=inf` poisons `ntpstat.rrd`. The
network `ntp` path already rejects non-finite/out-of-range values via the shared
`parse_ntp_offset()` (`lib/ntpoffset.c`); the client-side `do_ntpstat_rrd()` does
not — an asymmetry.

### Reproduce
Feed an `ntpstat` status whose body contains `offset=inf`; the value is stored
as an infinity in `ntpstat.rrd` and shows up in the AVERAGE/MAX consolidated
rows. (`offset=nan` instead lands as a harmless UNKNOWN.)

### Proposed fix
A single post-parse finite-value check that drops the sample when
`!isfinite(offset)`, mirroring the network path. A standalone AddressSanitizer
harness pins the guard across the ntpq and LARRD offset formats.

### Relationship to the do_ntpstat OOB-read fix
Same function as the out-of-bounds-read fix, but a different and independent bug
(value validation vs. memory safety) on non-overlapping lines — the two changes
cherry-pick cleanly over each other, so they are kept as separate PRs.

bonomani:
Fixes #189.

`do_ntpstat_rrd()` reads the ntpstat clock offset (`ntpq -c rv`, or the legacy
LARRD `Offset:` script) with `sscanf("%f")`, which accepts `nan`/`inf` as valid
floats. RRDtool then accepts the resulting update — verified against RRDtool
1.7/1.8 (librrd) through the C API:

- `offset=nan` → recorded as a harmless UNKNOWN datapoint;
- `offset=inf` / `offset=-inf` → stored as **real infinities** that propagate
  through AVERAGE/MAX consolidation and corrupt `ntpstat.rrd` and its aggregates.

This guards the RRD write on `isfinite(offset)`, matching the finite-value
hardening already on the network `ntp` path (`parse_ntp_offset`,
`lib/ntpoffset.c`). A standalone AddressSanitizer harness pins the guard across
the `ntpq -c rv` and legacy LARRD offset formats.

Independent of the do_ntpstat out-of-bounds-read fix: it touches different lines
and the two cherry-pick cleanly over each other, so it is kept as a separate PR.

### DIFF
--- a/tests/rrd/ntpstat-nonfinite-harness.c
+++ b/tests/rrd/ntpstat-nonfinite-harness.c
@@ -1,0 +1,155 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later                                  */
+/*
+ * tests/rrd/ntpstat-nonfinite-harness.c
+ *
+ * Pins the finite-value guard in do_ntpstat_rrd() (xymond/rrd/do_ntpstat.c).
+ * The "ntpq -c rv" / LARRD offset is read with sscanf("%f"), which happily
+ * parses "nan"/"inf"; without a guard, "%f" then writes "nan"/"inf" into the
+ * RRD update string, which RRDtool rejects (or stores as NaN, poisoning the
+ * graph). The handler must drop a non-finite sample and record nothing.
+ *
+ * Drives the REAL do_ntpstat.c against stub RRD plumbing (the same shape as
+ * xymond/do_rrd.c uses), with no rrdtool and no socket. Built with
+ * AddressSanitizer when the toolchain supports it.
+ */
+
+#define _GNU_SOURCE
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <ctype.h>
+#include <math.h>
+#include <time.h>
+#include <limits.h>
+
+#ifndef PATH_MAX
+#define PATH_MAX 4096
+#endif
+
+/* ---- stub plumbing, mirroring xymond/do_rrd.c ---------------------------- */
+
+static char rrdvalues[4096];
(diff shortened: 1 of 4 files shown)


bonomani:
Closing: on reflection this is defensive/parity hardening, not a fix for an observed problem. `ntpq -c rv` / ntpstat never emit `nan`/`inf` (the offset is a finite measured value in ms), so the non-finite path is only reachable via a malformed/buggy client script or an injected status message (where message authentication is the real concern). The impact if it did occur is bounded — a few affected RRA bins, self-healing on the next finite sample. Not worth carrying upstream. Full analysis in #189.

PR #191, opened 2026-06-22
ntp test: drop obsolete ntpdate "-p 1" option (ntpsec compatibility)

bonomani:
ntpsec's `ntpdate` rejects the `-p 1` option, which breaks the `ntp` test on ntpsec-based distributions (the test errors out instead of returning an offset).

This drops `-p 1` from the `NTPDATEOPTS` default, leaving `-u -q`. The `-p` (number of samples) option is deprecated/unsupported in ntpsec, and a single sample was never needed for Xymon's offset reading.

Changed in all the places the default is documented/compiled:
- `lib/environ.c` — compiled-in default
- `xymond/etcfiles/xymonserver.cfg.DIST` — shipped config
- `common/xymonserver.cfg.5` + generated HTML manpage

No behaviour change for classic `ntpd`/`ntpdate`; `-u -q` works on both.

Refs #27


bonomani:
Landed via #181: `main` now ships `NTPDATEOPTS="-u -q"` (no `-p 1`; `-p` marked deprecated), so this change is already in `main` (`ecdbde678`). Closing as superseded.

Issue #180, opened 2026-06-19
Feature request: clientlog.d drop-in support for xymon-client scripts


SoundGoof:
## Feature request: clientlog.d drop-in support for xymon-client scripts

Please add a simple way to include custom data in the clientlog without editing the shipped `xymonclient-*.sh` scripts.
This would make local clientlog additions upgrade-safe and avoid patching the packaged client scripts after updates.

Suggested behavior:

* If a clientlog drop-in directory exists, run executable files from it.
* Append their stdout to the normal clientlog.
* Send errors to stderr.
* If one script fails, continue with the rest.
* Skip files that are not owned by root or are writable by group/others.
* Keep the behavior the same across the supported OS client scripts where possible.

Example directory:

```text
/etc/xymon/clientlog.d/
```

Example drop-ins:

```text
/etc/xymon/clientlog.d/iptables
/etc/xymon/clientlog.d/certificates
/etc/xymon/clientlog.d/site_inventory
```

Example output:

```text
[iptables]
table: filter
chain: INPUT
policy: ACCEPT
rule: -p tcp -m multiport --dports 22,80,443 -j ACCEPT
rule: -m conntrack --ctstate ESTABLISHED,RELATED -j ACCEPT

[certificates]
path: /etc/ssl/certs/example.pem
subject: CN=example.internal
issuer: CN=Example Internal CA
not_before: 2026-01-15T10:22:00Z
not_after: 2026-07-15T10:22:00Z

path: /etc/nginx/ssl/api.example.internal.pem
subject: CN=api.example.internal
issuer: CN=Example Internal CA
not_before: 2026-02-01T08:00:00Z
not_after: 2026-08-01T08:00:00Z
```

bonomani:
> **Note:** here is my analysis of this. @SoundGoof — could you review it and tell us which points actually match your need and which don't fit? Corrections welcome before we act on it.

---

## Analysis

**Short version: ~90% of this already exists in Xymon** via the `client/local/` drop-in mechanism — so editing `xymonclient-*.sh` is already unnecessary. Only one part of the request is genuinely new.

### What already exists
`client/xymonclient.sh` (the OS-independent wrapper that builds every client message) already runs drop-ins:

```sh
# See if there are any local add-ons
if test -d $XYMONHOME/local; then
    for MODULE in $XYMONHOME/local/*
    do
        if test -x $MODULE
        then
            echo "[local:`basename $MODULE`]" >>$MSGTMPFILE
[...]

PR #194, opened 2026-06-24
docs: document clientlog section naming in README-local (#193)

### LINKED ISSUE #193: docs: README-local should document the [local:<file>] clientlog section naming (follow-up to #180)
Follow-up to #180.

The `client/local/` drop-in mechanism already covers custom clientlog additions in an upgrade-safe way, but `client/README-local` does not document the two facts users actually need to find and use it:

1. **Section naming.** Each drop-in's stdout is emitted under a `[local:<filename>]` section in the client message (see `client/xymonclient.sh`). README-local never mentions this, so the data is hard to locate server-side.
2. **Self-emitted sections.** Because a drop-in's stdout is appended verbatim, a script may print its own `[section]` headers, which become separate sections server-side. This is the supported way to produce arbitrary clientlog sections without editing the shipped `xymonclient-*.sh` scripts.
3. **Discoverability.** The word *clientlog* does not appear in README-local, so users searching for clientlog extensions don't find it.

In #180, @SoundGoof explicitly asked for the section-prefix and clientlog mentions to be added to README-local. This issue tracks that docs-only change.

No code change required.

bonomani:
Closes #193. Follow-up to #180.

`client/local/` drop-ins are the supported, upgrade-safe way to add custom data to the clientlog, but `client/README-local` didn't document where that output ends up. This adds:

- each script's stdout appears under a `[local:<filename>]` section (per `client/xymonclient.sh`);
- a script may print its own `[section]` headers, which become separate sections server-side, so one drop-in can emit one or more sections;
- stderr is excluded from the message, and a failing script doesn't stop the others;
- the clientlog / upgrade-safe framing, for discoverability (the word *clientlog* was absent);
- a fix for a `direcly` typo.

Docs-only; no code change.

### DIFF
--- a/client/README-local
+++ b/client/README-local
@@ -2,13 +2,29 @@ This directory - the client/local/ directory - can be used to
 install Xymon client add-on scripts. The Xymon client will run
 all files in this directory that are executable, and include the
 output from each script in a separate section in the Xymon client
-message which is sent to the Xymon server.
+message (the "clientlog") which is sent to the Xymon server.
+
+This is the supported, upgrade-safe way to add custom data to the
+clientlog: you never need to edit the shipped xymonclient-*.sh
+scripts, so your additions survive a client upgrade.
+
+Each script's standard output is added under a section named after
+the script's filename, like this:
+
+    [local:myscript]
+    ...output of client/local/myscript...
+
+Because the script's output is included verbatim, a script may also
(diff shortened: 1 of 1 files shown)


SoundGoof (approved):
its easy to find with a search now

PR #197, opened 2026-06-28
xymond: record ntp test offset in milliseconds

### LINKED ISSUE #196: do_net: ntp test offset recorded in seconds into a milliseconds DS (graph ~1000x too small)
The network `ntp` test records its offset through `do_net.c`, which parses the ntpdate/sntp output (offset in **seconds**) and forwards it to `do_ntpstat_rrd`. That handler's DS is declared in **milliseconds** (`offsetms`) and stores the value unscaled.

So a 40 ms offset is recorded as `0.04`, and the NTP offset graph plots it ~1000x too small — flat near zero, effectively unusable.

The client `ntpstat` test feeds `ntpq -c rv` output, which is already in ms, so it is unaffected.

bonomani:
Fixes #196.

`do_net.c` forwards the network `ntp` test offset (from ntpdate/sntp, in seconds) to the ntpstat RRD handler, whose DS is milliseconds (`offsetms`), stored unscaled — so the recorded offset, and the NTP offset graph, were ~1000x too small. Scale seconds to milliseconds before recording.

Only the network `ntp` path is affected; the client `ntpstat` test (`ntpq -c rv`, already ms) is untouched.

Note: PR #158 rewrites this same `do_net.c` block and will need to rebase onto this.

### DIFF
--- a/xymond/rrd/do_net.c
+++ b/xymond/rrd/do_net.c
@@ -130,7 +130,9 @@ int do_net_rrd(char *hostname, char *testname, char *classname, char *pagepaths,
 		}
 		
 		if (offsetval) {
-			snprintf(dataforntpstat, sizeof(dataforntpstat), "offset=%s", offsetval);
+			/* ntpdate/sntp report the offset in seconds; the ntpstat RRD DS is
+			 * milliseconds (offsetms), so scale before recording. */
+			snprintf(dataforntpstat, sizeof(dataforntpstat), "offset=%.6f", atof(offsetval) * 1000.0);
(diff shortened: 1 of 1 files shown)


bonomani:
Superseded by #181, which rewrote the network `ntp` path: the offset is now recorded into its own `ntp.rrd` `DS:offset:GAUGE` in seconds (self-consistent), no longer forwarded to the ms-labeled `do_ntpstat` handler this PR patched — so the 1000x mismatch is structurally gone. Closing; see #196 for the verification note.

PR #200, opened 2026-06-30
xymond_rrd: free RRD-def tables at exit, build them once (TBT 237)

bonomani:
J.C. Cleaver's TBT 237 commit, kept as authored, plus one commit correcting it.

Three defects in the rework:

- the tables were rebuilt on every lookup instead of once, and the reverse handlers were left out of the teardown — so what it set out to free was still leaked;
- an empty definition list read one byte past its allocation. Both tables count entries with a do/while that dereferences before it tests, and an empty list is a 1-byte allocation holding only the NUL.

The test drives the real `lib/xymonrrd.c` under ASan through `find_xymon_rrd()`/`find_xymon_graph()` rather than a copy of the loop — @SoundGoof's point that a copy cannot stop the original drifting away from it. The contrast that shows it can still see the fault uses the same source with the guard removed. ASan is probed separately: folding the probe into the harness build reported any compile error at all as "no ASan" and passed the test.

**Merge with a merge commit, not a squash** — a squash collapses J.C.'s commit into one authored by me and loses the attribution, which is what the two-commit shape exists for (same as #334).

Full suite: 68 passed, 0 failed, 0 skipped.

### DIFF
--- a/lib/xymonrrd.c
+++ b/lib/xymonrrd.c
@@ -36,27 +36,16 @@ static const char *metafmt = "<RRDGraph>\n  <GraphType>%s</GraphType>\n  <GraphL
 
 
 /*
- * Define the mapping between Xymon columns and RRD graphs.
- * Normally they are identical, but some RRD's use different names.
+ * Free the RRD-definition tables built by rrd_setup().
+ * Mainly so daemons can release them on shutdown (clean valgrind exit).
  */
-static void rrd_setup(void)
+void rrd_destroy(void)
 {
-	static int setup_done = 0;
-	SBUF_DEFINE(lenv);
-	char *ldef, *p, *services;
-	SBUF_DEFINE(tcptests);
-	int count;
 	xymonrrd_t *lrec;
 	xymongraph_t *grec;
 
-
-	/* Do nothing if we have been called within the past 5 minutes */
-	if ((setup_done + 300) >= getcurrenttime(NULL)) return;
-
-
-	/* 
-	 * Must free any old data first.
-	 * NB: These lists are NOT null-terminated ! 
+	/*
+	 * NB: These lists are NOT null-terminated !
 	 *     Stop when svcname becomes a NULL.
 	 */
 	lrec = xymonrrds;
@@ -69,6 +58,8 @@ static void rrd_setup(void)
 		xfree(xymonrrds);
 		xtreeDestroy(xymonrrdtree);
 	}
+	xymonrrds = NULL;
+	xymonrrdtree = NULL;
 
 	grec = xymongraphs;
 	while (grec && grec->xymonrrdname) {
@@ -77,6 +68,30 @@ static void rrd_setup(void)
 		grec++;
 	}
 	if (xymongraphs) xfree(xymongraphs);
+	xymongraphs = NULL;
+}
+
+
+/*
+ * Define the mapping between Xymon columns and RRD graphs.
+ * Normally they are identical, but some RRD's use different names.
+ */
+static void rrd_setup(void)
+{
+	SBUF_DEFINE(lenv);
+	char *ldef, *p, *services;
+	SBUF_DEFINE(tcptests);
+	int count;
+	xymonrrd_t *lrec;
+	xymongraph_t *grec;
+
+
+	/*
+	 * The tables are derived entirely from environment variables (TEST2RRD,
+	 * GRAPHS) and the compiled-in TCP service list, all fixed at startup -
+	 * so they never change while we run. Build them once.
+	 */
+	if (xymonrrdtree != NULL) return;
 
 
 	/* Get the tcp services, and count how many there are */
@@ -152,8 +167,6 @@ static void rrd_setup(void)
 		grec++;
 	}
 	xfree(lenv);
-
-	setup_done = getcurrenttime(NULL);
 }
 
 
--- a/lib/xymonrrd.c
+++ b/lib/xymonrrd.c
@@ -35,6 +35,10 @@ static const char *xymonlinkfmt = "<table summary=\"%s Graph\"><tr><td><A HREF=\
 static const char *metafmt = "<RRDGraph>\n  <GraphType>%s</GraphType>\n  <GraphLink><![CDATA[%s]]></GraphLink>\n  <GraphImage><![CDATA[%s&amp;graph=hourly]]></GraphImage>\n</RRDGraph>\n";
 
 
+/* The TCP service list the tables were built from - lifecycle state shared
+ * by rrd_setup() (rebuild-if-changed) and rrd_destroy() (final release). */
+static char *builtfrom = NULL;
+
 /*
  * Free the RRD-definition tables built by rrd_setup().
  * Mainly so daemons can release them on shutdown (clean valgrind exit).
@@ -43,6 +47,8 @@ void rrd_destroy(void)
 {
 	xymonrrd_t *lrec;
 	xymongraph_t *grec;
+
+	if (builtfrom) { xfree(builtfrom); builtfrom = NULL; }
 
 	/*
 	 * NB: These lists are NOT null-terminated !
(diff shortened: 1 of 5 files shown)


SoundGoof:
## Findings

### Preserve external handlers for reverse tests

`xymond/xymond_rrd.c:374`

The new early `break` prevents `update_rrd()` from running for every reverse test.

Previously, `find_xymon_rrd()` returned `NULL`, but `update_rrd()` fell back to the test name. This allowed matching `--extra-tests` entries to invoke `--extra-script`.

A harness using the production `do_rrd.o` confirmed that this fallback invokes the configured external handler. The new branch bypasses it completely.

Keep built-in RRD suppression for reverse tests without skipping matching external handlers.

bonomani:
Confirmed and fixed (folded into the correction commit; the TBT 237 origin commit stays intact, authored to Japheth Cleaver).

The built-in RRD suppression already lives in `find_xymon_rrd()` (returns NULL for reverse tests), so the early `break`'s only effect was to drop the `extids`/`exthandler` dispatch. Removed.

Self-review of the same rework also caught that build-once froze runtime config — `init_tcp_services()` reloads `protocols.cfg` on change — so the freeze is replaced by the original 5-minute throttle plus rebuild-on-change, and two empty-`TEST2RRD`/`GRAPHS` underflows are guarded.

SoundGoof:
## Findings

### Release `builtfrom` during final teardown — `lib/xymonrrd.c:89`

The updated reload logic introduces:

```c
static char *builtfrom = NULL;
```

It is freed when the TCP service list changes, but the final `rrd_destroy()` call cannot access this function-local static.

Shutdown therefore frees the RRD tables while retaining the last `builtfrom` allocation. This contradicts the PR’s clean-Valgrind shutdown objective.

A focused harness linked against the real updated library found:

- Updated PR after `rrd_destroy()`: 2 tracked copies remained
- Review-only ownership correction: 1 tracked copy remained

The baseline copy belongs to the existing `init_tcp_services()` cache. The additional copy is `builtfrom`.

Move the snapshot into lifecycle state accessible to `rrd_destroy()`, then free and reset it there. Remove the separate rebuild-path free to avoid a double free.

## Verification

Tested the updated exact PR head and both commits applied to current `main`.

- Exact-head suite: 11 passed, 0 skipped, 0 failed
- Rebased suite: 30 passed, 3 skipped, 0 failed
- Empty `TEST2RRD` and `GRAPHS` no longer underflow
- Dynamic service-list comparison is retained
- The production `do_rrd.o` harness confirmed external handlers run again
- The earlier reverse-test handler finding is resolved
- The allocation probe confirmed the retained `builtfrom` copy
- The ownership correction reduced live tracked copies from 2 to 1

Shutdown ownership was verified with an allocator probe linked around the real production library.

bonomani:
Fixed (folded into the correction commit), CI green.

`builtfrom` is now file-scope lifecycle state (still static): `rrd_destroy()` frees and resets it to NULL, which also forces a rebuild after any teardown. The rebuild path's separate free is dropped — `rrd_destroy()` runs first there and owns the release.

Reproduced your delta with a glibc mtrace probe around the real library: exactly one fewer live allocation at exit, the service-list strdup (matches your 2→1). The `init_tcp_services()` cache stays deliberately untouched (netservices ownership hazard, TBT 213).

SoundGoof:
## Findings

### Handle empty lists before counting commas — `lib/xymonrrd.c:158`

When `GRAPHS` is empty, `lenv` contains only its terminating NUL. The loop still calls:

```c
strchr(lenv + 1, ',')
```

This starts one byte beyond the allocation, causing undefined behavior.

A guard-page harness linked against the exact production library reliably segfaulted. GDB confirmed the fault at `lib/xymonrrd.c:158`.

Return a zero-entry table without scanning when `*lenv == '\0'`. Apply the same correction to the equivalent loop at line 133.

Add a regression test for empty `GRAPHS`.

## Resolved Findings

- Reverse tests can reach matching external handlers again.
- `builtfrom` is released and reset by `rrd_destroy()`.

## Verification

- Full server build succeeded
- Test suite: 11 passed, 0 failed
- Guard-page probe reproduced the fault
- Current `main` merges cleanly
- `git diff --check` passed

bonomani:
Confirmed and fixed (folded into the correction commit).

Both comma-count loops skip the scan when `*lenv == '\0'`, leaving `count` at 0 — `calloc(count+1)` still provides the zeroed terminator and `strtok()` returns NULL at once, so an empty list yields a clean zero-entry table. The same guard covers the `TEST2RRD` loop, reachable with `TEST2RRD=""` and an empty TCP service list.

Regression test `tests/rrd/rrddef-empty-list.sh` binds to both guarded loops and proves the property under ASan (guarded loop clean and counting zero, original loop flagged). Verified failing before the fix and passing after.

SoundGoof:
### Finding

#### Keep declarations before executable statements

`lib/xymonrrd.c:48-51`

`rrd_destroy()` now executes the `builtfrom` cleanup before declaring
`lrec` and `grec`. This breaks the codebase's C90-compatible declaration
style.

The PR fails:

```text
error: ISO C90 forbids mixed declarations and code
[-Werror=declaration-after-statement]
```

Current `main` passes the identical compile check. Move both declarations
above the `builtfrom` cleanup.

## Verification

- Empty-list fix correctly guards both production loops.
- Real built-library driver passed empty `TEST2RRD` and `GRAPHS` lookup,
  destroy, and rebuild paths.
- Synthetic merge built successfully.
- Post-build suite: 47 passed, 2 sanitizer-dependent skips, 0 failed.
- Current GitHub checks are green.
- Merge into current `main` is clean.

bonomani:
Fixed (folded into the correction commit): declarations moved back above the `builtfrom` cleanup; `-Werror=declaration-after-statement` passes on the branch again, and the empty-list regression test still passes.

SoundGoof:
## Findings

### Low: Do not treat every harness compilation failure as missing ASan

`tests/rrd/rrddef-empty-list.sh:81`

Any compilation failure exits successfully with “cc lacks ASan.” This includes errors in the generated C source, invalid compiler options, or missing headers, allowing a broken test to pass.

Probe ASan support separately, then fail if the harness itself does not compile. Existing sanitizer tests fall back to an unsanitized build and fail when that also fails:

- [`tests/rrd/ntp-offset-parse.sh`](https://github.com/xymon-monitoring/xymon/blob/main/tests/rrd/ntp-offset-parse.sh)
- [`tests/network/ntp-probe.sh`](https://github.com/xymon-monitoring/xymon/blob/main/tests/network/ntp-probe.sh)

## Residual Risk

The runtime test executes a handwritten copy of the vulnerable loop rather than production `xymonrrd.c`. The exact source-text assertions bind the test to formatting but cannot prevent the copy from diverging behaviorally.

Using production code is straightforward and consistent with existing tests. Compile `xymonrrd.c`, link against the configured library, and trigger setup through `find_xymon_rrd()` and `find_xymon_graph()` with empty definitions.

Relevant production-harness examples:

- [`tests/web/html-log-custom-graphs.sh`](https://github.com/xymon-monitoring/xymon/blob/main/tests/web/html-log-custom-graphs.sh) links a focused harness against the real `libxymoncomm.a`.
- [`tests/web/html-log-custom-graphs-harness.c`](https://github.com/xymon-monitoring/xymon/blob/main/tests/web/html-log-custom-graphs-harness.c) exercises the real RRD lookup functions.
- [`tests/lib/build-worker.sh`](https://github.com/xymon-monitoring/xymon/blob/main/tests/lib/build-worker.sh) obtains configured linker flags through `make`.

This should add only a few seconds in configured build jobs and can skip immediately when no configured tree is available.

## Verification

- Configured server build succeeded.
- Full suite: 12 passed, 0 skipped, 0 failed.
- Production-library empty-list and teardown harness passed.
- Synthetic merge with current `main` is clean.
- Previous reverse-handler, `builtfrom`, empty-list, and C90 findings are resolved.

bonomani:
Both addressed.

**Production code, not a copy.** The test now compiles the real `lib/xymonrrd.c` under ASan and reaches `rrd_setup()` the way every caller does — the first `find_xymon_rrd()`/`find_xymon_graph()` builds both tables. `xymonrrd.c` is compiled in rather than taken from `libxymoncomm.a`, so the loop under test is instrumented and the mutant is swappable; naming the object first leaves the archive copy unused.

One detail worth recording, since it decides whether the case is reached at all: the TCP service list is only empty with an **empty** `protocols.cfg` present. With no file, `init_tcp_services()` falls back to its built-in defaults and the list is not empty.

The contrast that shows the harness can still see the fault now uses the same production source with the guard taken back out, so it cannot drift from production either.

**The ASan gate.** Probed on its own with `asan_usable()`, which builds *and runs* a probe. A harness that does not compile now fails:

```
FAIL: the harness does not compile against production xymonrrd.c
```

where it previously exited successfully as "cc lacks ASan".

One thing I changed on my own account: the mutant is generated with `awk`, not `python3` — the BSD hosts you test on do not have it, and it would have skipped or failed there.

`upstream/main` merged in first; the branch predated `asan_usable()`. Full suite: 63 passed, 0 failed, 0 skipped.

SoundGoof (dismissed):
No findings.

The latest revision resolves the previous test-harness concerns. It compiles
the production `xymonrrd.c`, probes ASan separately, and proves the test detects
the unguarded empty-list regression.

Validation:
- Linux: 68 passed, 0 skipped, 0 failed.
- FreeBSD 14.3: 68 passed, 0 skipped, 0 failed.
- NetBSD 9.4: 68 passed, 0 skipped, 0 failed.
- NetBSD 11.0: 68 passed, 0 skipped, 0 failed.
- OpenBSD 7.9: 66 passed, 2 unsupported-ASan skips, 0 failed.
- The prospective merge with current `main` builds and tests cleanly.

The earlier reverse-handler, lifecycle ownership, empty-list, C90, and
test-reliability findings are resolved.

bonomani:
Restructured to the two-commit shape used for the other TBT ports: J.C. Cleaver's commit as authored, plus one "correct and cover". The two merge-from-main commits are gone.

Needs a fresh approval on `f2144870a` — the force-push dismissed the last one. And it wants a **merge commit**: a squash would drop J.C.'s authorship.

SoundGoof (approved):
No findings.

The branch was rebased onto current `main` and condensed to two commits.
The production implementation is byte-for-byte identical to the revision
already reviewed and tested across Linux and four BSD systems. The regression
test changes are documentation-only.

Validation:
- Exact rebased head on Linux: 68 passed, 0 skipped, 0 failed.
- FreeBSD 14.3: 68 passed, 0 skipped, 0 failed.
- NetBSD 9.4: 68 passed, 0 skipped, 0 failed.
- NetBSD 11.0: 68 passed, 0 skipped, 0 failed.
- OpenBSD 7.9: 66 passed, 2 unsupported-ASan skips, 0 failed.
- `git diff --check` passes.
- The final PR head remains `f2144870`.

The earlier reverse-handler, lifecycle ownership, empty-list, C90, and
test-reliability findings are resolved.

Issue #201, opened 2026-07-01
xymond: stale status and its alert resurrected from checkpoint — drop/rename not checkpointed


feld:
I had a server installed with Debian. I have some automated checks in analysis.cfg for CLASS=linux

I reinstalled the server as FreeBSD. Xymon client started reporting data to the Xymon server.

The Xymon server is now alerting that some PROC checks associated with CLASS=linux are not running. But this client should be recognized as CLASS=freebsd now...

I dropped the server to let it start fresh with new data for this client, but not even that changed the server's belief that this is a CLASS=linux client.

bonomani:
Initial hypothesis — a stuck class from a client-side override or a `hosts.cfg CLASS:` line — **ruled out below**; kept only for the diagnostic @feld's reply refers to:

```sh
xymon 127.0.0.1 "xymondboard host=THEHOST fields=hostname,XMH_OS,XMH_CLASS"
```

See the root-cause comment below.

feld:
The result I'm getting from that xymondboard command is showing freebsd. The client isn't reporting any class at all, I validated that with tcpdump, and there's nothing in my hosts.cfg files that sets/overrides a class for anything at all.

I have found a solution to the issue, but this means there's a bug lurking we need to hunt down: 

I stopped the Xymon server, deleted the contents of `/usr/local/www/xymon/server/tmp` (which has things like cached alerts), and started the Xymon server again. I restarted the Xymon client of the affected server to make it immediately send new data, and the problem is now gone -- it's being recognized as CLASS=freebsd as it's picked up those PROC checks and is no longer stuck in an alerting state.

So the issue was either related to the `alert.chk` or `xymond.chk` files which both contained information relevant to the faulty alert.

bonomani:
## Root cause + resolution

Not class resolution (@feld confirmed OS=freebsd, no client class, no `hosts.cfg CLASS:`). What was stuck is a **stale status + its alert**, persisted in `server/tmp/*.chk` — which is why deleting them fixed it:

1. Class `linux` → `freebsd`: the client stops producing the old class's column.
2. The last red status and its active alert were persisted in `xymond.chk` / `alert.chk`, so the alert kept repeating as RED.
[...]

PR #203, opened 2026-07-02
RFC/WIP: xymond: retire stale-purple statuses that the config no longer produces (#201)

bonomani:
Draft / RFC for bug **B** of #201 — **no behavioural change yet**, just a documented scaffold at the fix site to agree the approach.

**Problem:** when a client's applicable ruleset changes (e.g. class `linux` -> `freebsd` after an OS reinstall), a test the old class emitted stops being reported. It ages to purple in `check_purple_status()`, and because purple is in the default `ALERTCOLORS` (`red,yellow,purple`) it **keeps paging forever**, with no removal path — unlike the `H_SUMMARY` case just above it, which is dropped. Combined with bug A (#… drop-checkpoint PR) it also survives restarts.

**This PR** adds only a TODO at the exact spot describing the proposed fix; it ships no code change, so it is safe to sit as an RFC.

**Proposed fix (opt-in, default off — existing installs rely on purple = "client died"):**
- add `PURPLERETIRE=<seconds>` (0 = disabled);
- when a non-summary log has been purple longer than that, drop it like the `H_SUMMARY` case **and** post an `@@droptest` to the page channel so `xymond_alert` clears the active alert (freeing the log alone would not) — cleanest via `handle_dropnrename(CMD_DROPTEST,…)`, but that mutates `hwalk->logs` mid-walk so the loop cursor must be advanced first.

**Open question for maintainers:** opt-in purple-retire as above, or instead make a class change actively invalidate the old class's columns in `xymond_client` (no timeout). Happy to implement whichever is preferred.

Refs #201.

### DIFF
--- a/xymond/xymond.c
+++ b/xymond/xymond.c
@@ -5177,6 +5177,26 @@ void check_purple_status(void)
 						if (!lwalk->dismsg) lwalk->dismsg = strdup(cause);                                         
 					}
 
+					/*
+					 * Issue #201: flag a client-column reconciliation problem.
+					 * If the host is still sending client data (its last client
+					 * report is at/after this column's expiry) yet this column has
+					 * gone purple, the client has stopped producing a column it
+					 * used to send - typically retired by a config or class change
+					 * (e.g. linux -> freebsd after a reinstall). Such a column ages
+					 * to purple and keeps alerting with no removal path. We do NOT
+					 * drop it here (a manual "xymon drop" remains the fix, made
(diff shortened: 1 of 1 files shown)


bonomani:
Superseded by #204. This branch was renamed from `fix/issue-201-retire-stale-purple` to `feat/issue-201-warn-stale-client-columns` to reflect that it is a feature (a non-destructive reconciliation *warning*), not a fix — which closed this PR. The updated approach lives in #204; the restart-resurrection fix for #201 remains #202.

PR #204, opened 2026-07-02
xymond: log when a client stops sending a report section (#201)

### LINKED ISSUE #201: xymond: stale status and its alert resurrected from checkpoint — drop/rename not checkpointed

bonomani:
> **Follow-up enhancement — not the fix for #201.** The fix is #202 (drop-checkpoint). This PR does not close #201.

## What

When a client stops producing a column it used to send (e.g. an OS reinstall flips the class `linux` → `freebsd`), the column ages to purple and keeps alerting with no removal path, indistinguishable from a dead client. This logs the observable cause, non-destructively.

## How

The full client report already arrives in `handle_client()` and is stored per host. xymond tracks the section names (`[cpu]`, `[df]`, …) of each collector's stored message and logs **one line when a section disappears** between that collector's consecutive reports (+71 lines in `xymond.c`, nothing removed):

```
Client report change: host 'x' no longer sends section(s) [mdstat] - columns fed by
them will go purple although the host keeps reporting; if intentional (e.g. an
OS/class change), retire them with 'xymon drop x <testname>'
```

That line tells an orphaned column apart from a dead client: a dead client stops reporting entirely (the diff never fires), and a stopped feeder like xymonnet doesn't change the client report. No section→column mapping is attempted — xymond doesn't have that knowledge; the vanished section names point at the affected columns.

## Properties

- **Event-scoped**: logged once, when the change happens — no sweep, no re-assertion, no flags on status logs. Section additions never log.
- **Per-collector**: diffs one collector's consecutive reports only, so a slow sub-collector purged between its runs (`MAX_SUBCLIENT_LIFETIME`) never triggers it.
- **Strict section grammar**: a header must be exactly `[name]` alone on its line, matching what logfetch guarantees — bracket-prefixed data lines (Apache-style log content) can never appear as phantom sections.
- **Non-destructive**: nothing dropped, recolored, or silenced; `xymon drop HOST TEST` remains the resolution (durable via #202).
- **No protocol change, no client-side change**; hosts without stored client data (`--no-clientlog`) are untouched — the diagnostic is inert there, as there is nothing to diff.

## Known limitation

Section baselines live in memory: a change that happens while xymond is down, or predates a restart, is not logged.

## Testing

- Builds clean (full `make` on Linux, no new warnings); CI green (server + both client builds + shell regression).
- Standalone harness over the verbatim function: no-change/additions silent; class flip logs one line naming all vanished sections; sub-collector changes tagged with the collector id; malformed headers and bracket-prefixed data lines ignored.
- **Validated end-to-end on a live xymond**: two identical linux client reports stay silent; a linux → freebsd class flip logs exactly one line naming `[free], [mdstat], [msgs:/var/log/messages]`.


SoundGoof (commented):
## Review

###  Expired collector baselines can produce false warnings

At `xymond/xymond.c:2427`, the matching collector’s timestamp is refreshed before `update_clientsections()` and before stale entries are purged.

If that collector returns after `MAX_SUBCLIENT_LIFETIME`, its old message is replaced and its timestamp renewed, but its old `sections` baseline remains. The subsequent purge can no longer recognize it as expired. A removed section is therefore reported even though the previous report should have expired.

This contradicts the PR’s stated property that slow collectors purged between runs never trigger the diagnostic.

Purge or invalidate an expired matching entry before updating its timestamp and comparing sections. Treat the returning report as a new baseline.

Add a regression covering:

1. Collector sends section `[old]`.
2. More than `MAX_SUBCLIENT_LIFETIME` elapses.
3. The collector returns without `[old]`.
4. No vanished-section warning is emitted.
5. A section removed between two non-expired reports still emits exactly one warning.

### Verification

- Server build completed successfully.
- Regression suite: 14 passed, 0 failed.
- `git diff --check` passed.
- All four GitHub CI checks are green.

bonomani:
Confirmed and fixed. You were right that it contradicted the PR's own stated property — a collector purged between runs is exactly the case the diagnostic is meant to stay quiet about.

The refresh is now a named function that drops the section list when the entry has outlived its window, so the returning report becomes the baseline. Named rather than inlined for a practical reason: waiting out `MAX_SUBCLIENT_LIFETIME` is not something a test can do, so the decision has to be drivable with timestamps of the test's own.

The PR had no test at all — there is one now, extracting the real `clientmsg_refresh()` and `update_clientsections()` from `xymond.c` (both static, no library form) and covering your steps plus one more:

| Case | Expected |
|---|---|
| returns after expiry without `[old]` | 0 warnings |
| section removed between two live reports | exactly 1 |
| removal *after* a return from expiry | 1 again |

The third is not in your list and I added it deliberately: without it, "quiet after expiry" could be satisfied by never comparing again at all, which would pass the first two and silently disable the feature.

Reverting the drop to a plain timestamp refresh gives `expected '0', got '1'` on the first case.

SoundGoof:
# Review

## Medium: The harness cannot run with configured NetBSD libraries

[tests/server/client-sections-expiry.sh:56](https://github.com/xymon-monitoring/xymon/blob/74a1be1fc428e8cefbdf2ef9ce46f738d4a84ae8/tests/server/client-sections-expiry.sh#L56)
links with `PCRELIBS` but omits the configured `RPATHOPT`.

On NetBSD 11.0, PCRE2 is installed under `/usr/pkg/lib`. The harness
links successfully, but execution fails with:

```text
Shared object "libpcre2-8.so.0" not found
```

The pipeline at
[tests/server/client-sections-expiry.sh:62](https://github.com/xymon-monitoring/xymon/blob/74a1be1fc428e8cefbdf2ef9ce46f738d4a84ae8/tests/server/client-sections-expiry.sh#L62)
then masks that loader failure with `|| true` and reports a misleading
warning count of zero.

Carry the configured runtime linker flags into the harness link and
check the harness exit status separately from `grep`. Adding NetBSD's
generated `-Wl,-rpath,/usr/pkg/lib` made all scenarios pass with the
expected counts `0, 1, 1`.

## Validation

- Linux suite: 60 passed, 3 expected skips, 0 failed.
- ASan and UBSan: all focused scenarios clean.
- Live xymond: removals log once, unchanged reports stay silent, and
  collector baselines remain separate.
- FreeBSD 14.3 and OpenBSD 7.9: focused regression passes.
- NetBSD 11.0: fails without configured rpath; passes after supplying it.
- All five upstream CI checks are green.

bonomani:
Self-review before this goes further. Parking it rather than polishing: three of the findings are design-level, not finish.

### The diagnostic watches a different set from the one that feeds columns

`update_clientsections()` implements a second notion of "section". `splitmsg_r()` (`xymond_client.c:128-153`) — the splitter every `xymond/client/*.c` handler actually consumes — accepts a `[` at offset 0, and ends a name at `strcspn(sectname, "]\n")`. The new parser only matches `\n[`, and skips a header whose `]` is not followed by end-of-line. So a section that genuinely feeds a column can be invisible to the tracker, and when it later disappears nothing is logged — the case #214 exists to catch.

Fix: use `splitmsg_r`, not a parallel parser.

### It will warn every cycle on healthy hosts

`[vmstat]` is emitted only when the background `vmstat 300 2` file has landed, and the client deletes it after reading (`client/xymonclient-linux.sh:210-212`; same shape for `[iostatcpu]`/`[iostatdisk]` on SunOS, `[sar]` on IRIX). Every cycle that races the background job drops the section, logs the full warning with its `xymon drop` hint, and the next cycle brings it back. There is no state suppressing a repeat.

Fix: report a vanished section once, not once per cycle.

### It also warns about text that is not a section

`COL_CLIENT` (`xymond.c:3813`, `3889`) exists to "send client data from a standard BB utility" — the body is whatever the sender wrote, as are `[local:<module>]` sections. Any `[something]` line alone becomes a tracked section, and its absence next time produces a warning naming a column that never existed.

Fix: track only the real client channel (`4809`).

### Mechanical, once the above is settled

`grabstrbuffer()` returns the whole 4096-byte allocation (`strfunc.c:36,79`), so `cwalk->sections` holds 4 KB per collector entry — ~20 MB at 5000 hosts; `strdup` the content instead. `newstrbuffer()` can return NULL and both call sites dereference it. The parser NUL-terminates the retained `cwalk->msg` in place with an `addtobuffer()` — which `realloc`s unchecked — between the two writes. `strtok` should be `strtok_r`, as the rest of this file already is. The missing-section search rescans the whole set per token.

Tests: the `MAX_SUBCLIENT_LIFETIME` guard is vacuous (the token also appears in the `clientmsg_refresh` block it extracts beside it); the "exactly once, not once per section" assertion counts lines while the implementation emits one line and the scenario removes one section, so it cannot fail; the purge loop — and therefore both new `xfree(czombie->sections)` sites — is not exercised; the `sed` on `$ROOT/Makefile` runs unguarded under `set -e`.

One reported item did not survive checking: the hardcoded `-lpcre2-8` fallback is not a divergence — `analysis-ds-firstmatch.sh:66` carries the identical line.

### Attribution

This cites #201 in the code, the comments, the tests and the branch name. #201 is closed and is the checkpoint-resurrection bug fixed by #202; this work is #214, which is open. To be repointed.

Issue #107, opened 2026-05-27
Review checklist: PR #44 custom graphs — blockers + #38/#46 collisions


bonomani:
Tracking checklist to verify **PR #44** ("Fix custom graphs", `Fixes #31`) before
it can be trusted or merged. It is a single **draft** commit (`261e4643`, +348/−170
over 8 files) that the author flagged as untested. This issue decomposes that one
commit into checkable items.

Refs: **PR #44** (draft, custom-graphs fix) · **#31** (the underlying bug report) · **#20** (related `showgraph` RRD-filename matching bug — unanchored match, e.g. `conn` matches `proxyconn`; fix in PR #139, same matching subsystem, still open). **PR #38 merged 2026-06-05** (closes #30).

The 20 tasks are split by whether they touch code that **collides elsewhere**
(verified by real merges, re-run 2026-07-02):

- **Part 1 — independent** (15): touch files nothing else conflicts with →
  reviewable / verifiable **now**.
- **Part 2 — same-file collisions, order now settled** (5): `web/svcstatus-trends.c`
  conflicts with **`main` itself** (#38 landed 2026-06-05, the recommended order) →
  resolved during #44's rebase, keeping main's loop-grow. `web/svcstatus.c`'s
  colliding change (`skiploc`, `a0822c27`) is carried by **#151 alone** (#46
  dropped `web/svcstatus.c` on 2026-07-03; #205 was a short-lived duplicate,
  closed). The rework (#206) does not touch `web/svcstatus.c` at all, so the
  formerly documented adjacent-line combination is void. **No open merge-order
  question remains.**

Only 2 of #44's 8 files actually conflict; `web/showgraph.c` shares a file with #46
but **auto-merges clean**, so it stays in Part 1.

Verification key (reviewed 2026-05-27; states + merges re-verified 2026-07-02 —
#44 head unchanged at `261e4643`, so every per-code finding below still stands):
✅ pass · ⚠️ finding (fix before merge) · ❌ blocking · ⏳ can't verify yet.
(Part 2's *collision* reconciliation is still gated on merge order; the bullets
below verify #44's own change in each file, independent of that.)

---

### Blocking — build / rebase

- [ ] **PCRE1 → PCRE2**: `count_rrd_files_for_graph()` uses `pcre_compile` /
  `pcre_exec` / `pcre_free` (PCRE1). `main` is PCRE2-only (since #5 / #89) — this
[...]

PR #206, opened 2026-07-02
htmllog: render custom GRAPHS_<service> graphs without a default graph (fixes #31, supersedes #44)

### LINKED ISSUE #31: Custom RRD graphs do not really work as documented

bonomani:
Fixes #31. **Supersedes #44** — keeps its diagnosis (Mark Felder co-authors the fix commit), drops the RRD-counting approach: a deep review of a corrected revision of it showed the count-gate regressed locator/distributed setups, filestore static pages, TEST2RRD-mapped columns and `--multigraphs` paging ([record in #107](https://github.com/xymon-monitoring/xymon/issues/107#issuecomment-4866693882)).

## The fix — additive only (+56/−16 in `lib/htmllog.c`)

The `GRAPHS_<service>` block sat inside the `(rrd && graph)` guard, so a custom graph on a column with **no default graph definition** never rendered. The guard is widened; such a column gets a stack-local gdef for the link. Nothing is gated on filesystem state, and the legacy `linecount`/locator/filestore paths are untouched — **no page that rendered graphs before can lose them.**

Six commits, one subject each, ordered preparation → fix (every line written once; all six build individually):
1. `strtok` no longer mutates the live `GRAPHS_<service>` environment value (second render saw a truncated list);
2. entries render **verbatim** through a stack-local gdef (`disk.sda1`/`tcp.smtp` reach the URL untouched — prefix-resolving them rewrote the link or tripped the `tcp:` special-case), the `GRAPHS` `name::N` split size is now honoured on status pages, and the shared graph table is never written (the legacy mutation corrupted every later lookup in long-running processes, and leaked) — this also makes the loop NULL-safe for a missing default graph, preparing:
3. the #31 guard fix — the guard-move itself, ~15 lines; set-but-empty value counts as not set;
4. hardening — the value echoed into the `<!-- GRAPHS_… -->` comment is HTML-quoted;
5. the regression test — runner-conformant (skips on an unbuilt tree, compile failure = FAIL with log), covering every behaviour above;
6. `howtograph.html` — the missing `GRAPHS_<columnname>` section plus combined-vs-split rendering (`<!-- linecount=N -->`, `GRAPHS name::N`).

## Validation

Harness through the real `generate_html_log()`; full suite 15/15, CI green. Live differential against a running xymond: upstream `svcstatus.cgi` renders **no** graph section on a custom column (#31 reproduced), this branch renders it — and with real RRD files the whole chain works (combined image via `showgraph.cgi`, automatic listing and count on the trends page, which needed no code change).

### DIFF
--- a/docs/howtograph.html
+++ b/docs/howtograph.html
@@ -242,6 +242,39 @@ Save the file, and when you click on the <b>trends</b> column you should
 see the slab graph at the bottom of the page.
 
 
+<h2>Show the graph on a status page: GRAPHS_&lt;columnname&gt;</h2>
+<p>To display graphs on a specific status page - including a custom column
+that has no default graph of its own - set GRAPHS_&lt;columnname&gt; in
+~xymon/server/etc/xymonserver.cfg to a comma-separated list of graph
+definitions:</p>
+
+<pre><tt>
+	GRAPHS_smart=&quot;smart-temp,tcp:smtp&quot;
+</tt></pre>
+
+<p>Each name needs a matching graphs.cfg section; <tt>bundle:instance</tt>
+(or <tt>bundle.instance</tt>) shows a single instance of a multi-file graph,
+like the smtp response time from the [tcp] section above. The setting
+<i>overrides</i> the column's default graph, so include that in the list to
+keep it (GRAPHS_cpu=&quot;la,smart-temp&quot;). Adding the graph to GRAPHS is
+only needed for the <b>trends</b> column - the status page works from
+GRAPHS_&lt;columnname&gt; alone.</p>
+
+<p>By default the page shows one image containing every RRD file the graph's
+FNPATTERN matches. To split it into groups of instances the way the
+<b>disk</b> column does, put the instance count in the body of the status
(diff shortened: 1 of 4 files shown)


feld (approved):
Built Xymon from main branch with these patches and deployed on my server, checked my custom graphs, and they worked. Looks good!

PR #207, opened 2026-07-02
Custom graphs on mapped, dotted and bundled columns (stacked on #206)

bonomani:
**Stacked on #206** — the first four commits are #206's; review the last two here. Merge #206 first, then this rebases to just them.

## What

The behaviour #206's counting mechanism needs to not regress existing pages — deliberately kept out of #206 so the repair-vs-behaviour boundary stays reviewable:

- **Dotted graph names** (`tcp.http`) resolve to their `graphs.cfg` section the way `showgraph.c` does (primary = after the `.`/`:`, fallback = before), instead of silently dropping from the trends page.
- **Bundle filtering**: new optional `svcfilter` argument to `count_rrd_files_for_graph()` — when one service of a bundle section (`tcp`/`ncv`/`devmon`) is rendered, count only that service's RRD files (matching showgraph's `wantsingle`), not the whole bundle.
- **TEST2RRD fallback**: htmllog falls back to the resolved RRD name when the visible service has no `graphs.cfg` match (`cpu` → `la.rrd`, `procs` → `processes.rrd`), so standard columns keep their graphs.

## Tests

Extends #206's suite with the feature assertions: svcfilter bundle counting, dotted primary/fallback resolution, and a TEST2RRD-mapped host. Full `make` clean, every commit builds, suite passes at the tip.


bonomani:
Closing: the deep review of the counting approach (see #206) showed the count-based gate regresses locator setups, filestore static pages and multigraphs paging — beyond what this PR's mapping layer repaired. #206 is now a minimal additive fix that needs no counting, which removes this PR's premise. Branch kept for reference.

PR #211, opened 2026-07-02
svcstatus: do not crash when SCRIPT_NAME is not set (fixes #209)

### LINKED ISSUE #209: svcstatus.cgi segfaults when SCRIPT_NAME is unset (strlen on NULL getenv)
`web/svcstatus.c` `parse_query()` runs `strlen(getenv("SCRIPT_NAME"))` unchecked when building `clienturi`. A web server always sets `SCRIPT_NAME`, but any other invocation — debugging from a shell, a test harness, a misconfigured wrapper — segfaults (`SIGSEGV`, `si_addr=NULL`, in libc `strlen`) before producing a single line of output, which makes the crash needlessly hard to diagnose (found the hard way while validating #206 end-to-end: the CGI died silently until traced with a preloaded backtrace handler).

**Fix** (one line, validated live — `rc=1` with a proper error page instead of `rc=139`): treat a missing `SCRIPT_NAME` as empty. Branch: `bonomani:fix/svcstatus-script-name-null`.

bonomani:
Fixes #209. One-line fix: `parse_query()` ran `strlen()` on `getenv("SCRIPT_NAME")` unchecked — a web server always sets it, but any other invocation (shell debugging, test harnesses, misconfigured wrappers) segfaulted before producing any output. A missing `SCRIPT_NAME` is now treated as empty.

Validated live: the same invocation that segfaulted (`rc=139`, `SIGSEGV` at NULL in libc `strlen`) now exits `rc=1` with a proper error page.

### DIFF
--- a/web/svcstatus.c
+++ b/web/svcstatus.c
@@ -151,6 +151,7 @@ static int parse_query(void)
 		char *hostquoted = htmlquoted(hostname);
 
 		req = getenv("SCRIPT_NAME");
+		if (!req) req = "";	/* Only set when invoked as a CGI - do not crash outside one */
 		SBUF_MALLOC(clienturi, strlen(req) + 10 + strlen(hostquoted));
 		strncpy(clienturi, req, clienturi_buflen);
 		p = strchr(clienturi, '?'); if (p) *p = '\0'; else p = clienturi + strlen(clienturi);


feld (approved):
Pretty sure I've hit this when trying to manually execute things to debug and it's very frustrating when it feels like all of Xymon is inaccessible without CGI. Thanks.

PR #217, opened 2026-07-03
client_config: honor first-match for DS (RRD dataset) rules (#32)

### LINKED ISSUE #32: analysis.cfg DS rules do not honor "first match"
This bug has probably persisted so long because of #31 causing people to give up even trying to do things with custom graphs. Once you can make custom graphs and begin writing alerts with Xymon instead of a more complicated TSDB solution like Prometheus you will quickly encounter this bug 😄

I have a patch for this that I will submit as a PR soon

bonomani:
Closes #32.

## Problem

`analysis.cfg` is documented as first-match (`analysis.cfg.5`):

> The entire file is evaluated from the top to bottom, and the first match found is used. So you should put the specific settings first, and the generic ones last.

Every threshold type honors this — `get_disk_thresholds()` and its inode/proc/port/load siblings walk to the **first** pattern-matching rule and use it. `check_rrdds_thresholds()` did not: it looped over **every** matching `DS` rule and emitted a `modify` for each, so a later wildcard rule was not shadowed by an earlier specific one — both applied:

```
DS conn tcp.conn.rrd:sec >0.5 COLOR=red HOST=web1   # web1 tolerates more
DS conn tcp.conn.rrd:sec >0.1 COLOR=red             # general default
```

On `web1` both evaluated, so the general `>0.1` rule wrongly fired alongside the specific `>0.5` override — duplicate/conflicting `modify conn` lines.

## Fix

Track the `(column, dataset, colour)` targets already decided in this check pass and skip later rules for the same target, so the first matching rule wins — as for disk/inode/proc/load. The target is recorded as soon as a rule *applies* to it (its dataset is present in the RRD and has a value), **even when that rule's threshold does not trigger** — matching the disk semantics, where the first matching rule defines the check rather than falling through to a more general rule.

Two reasons the key is finer than disk's simple "return on first match":
- One DS check pass legitimately fires rules for *different* `(column, dataset)` targets in the same RRD, so the loop must continue across targets.
- **Severity (colour) is part of the key**, so a specific rule shadows only later rules of the **same** colour. A yellow threshold and a red threshold on one dataset both still fire (multi-level DS alerting keeps working — that is the only way DS expresses two levels), while a specific-vs-general pair of the same colour no longer both apply.

## Status

Compiles clean. A regression harness around `check_rrdds_thresholds()` (config + rule tree + dataset names) would be a good follow-up; I can add one if wanted.

### DIFF
--- a/xymond/client_config.c
+++ b/xymond/client_config.c
@@ -3085,9 +3085,11 @@ strbuffer_t *check_rrdds_thresholds(char *hostname, char *classname, char *pagep
 	rrdtplnames_t *tpl;
 	double val;
 	void *hinfo;
+	strbuffer_t *seen;
 
 	if (!resbuf) resbuf = newstrbuffer(0);
 	clearstrbuffer(resbuf);
+	seen = newstrbuffer(0);
 
 	hinfo = hostinfo(hostname);
 	rule = getrule(hostname, pagepaths, classname, hinfo, C_RRDDS);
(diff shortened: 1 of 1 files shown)


feld (approved):
I compiled and deployed Xymon Server from git main branch with this patch applied and expected to get alerts for a few rules if this was broken and I'm not getting alerts, so this is working as expected.

PR #219, opened 2026-07-04
build: keep pidfiles and control sockets in their own XYMONRUNDIR (TBT 3)

bonomani:
Xymon writes its pidfiles and its `rrdctl.*` control sockets into the log directory. `XYMONRUNDIR` gives them one of their own, defaulting to that same log directory so nothing moves unless it is set. Point it at `/run/xymon` and the runtime files land where the FHS wants them.

J.C. Cleaver's Terabithia patch, applied as authored, then five correction commits. **First in the runtime-directory stack `#219 → #172 → #173`**: without it `$XYMONRUNDIR` is undefined and the pidfile paths #172 ships expand to `/xymond.pid` — measured.

## Setting it

```
XYMONRUNDIR=/run/xymon XYMONCLIENTRUNDIR=/run/xymon-client ./configure --server
```

Both are prompted for interactively too. `/run` is not the default, and the reason is not backward compatibility: it is a root-owned tmpfs, a source installation runs as an unprivileged user that cannot create a directory there, and `xymon.sh` stops with an error when it cannot create this one. The log directory is the only one certain to exist and be writable when the server first starts. Packages run as root and can — **#415** ships the `tmpfiles.d` fragment (`d /run/xymon 0755 xymon xymon`) and says in its own description that it waits on this PR to become useful. The two were written to fit together.

## Two directories, not one

The client running on the server needs its own. `xymonlaunch` expands a task's pidfile path in the *parent*, which holds one value per variable name, and the server's launcher starts both sides — so one name cannot carry both paths. `XYMONRUNDIR` is the server's, `XYMONCLIENTRUNDIR` the client's, and `xymonserver.cfg` defines both.

This split is introduced here, not inherited. An earlier version of this description said it followed the existing `XYMONSERVERLOGS`/`XYMONCLIENTLOGS` pair; that was wrong — `xymonserver.cfg` defines no `XYMONCLIENTLOGS`, only the legacy `BBCLIENTLOGS`, which points at the *server's* logs.

No C changes for the split: all six programs that name the variable are server-side, and `xymonlaunch` names it nowhere. A client config still setting only the generic name keeps working — the new name defaults to it, the idiom the table already uses for `XYMSRV`.

## What the original patch left behind

It moves the files; the readers stayed put. Each of these was found by a separate verification pass, and each has a test that fails when the behaviour is reverted.

| | before | after |
|---|---|---|
| `rrdcachectl` | composed the socket path from `$XYMONTMP` while `xymond_rrd` had moved it — the flush utility looked in an empty directory | reads `$XYMONRUNDIR`, bounded |
| `rrdcachectl` again | aliases `xgetenv`→`getenv`, so it never saw the `XYMONRUNDIR`→`XYMONLOGDIR` default `lib/environ.c` gives the socket *writer*; an unexported variable composed a `"(null)/…"` path | explicit fallback chain |
| `xymon.sh reload` / `rotate` | looked under the log directory, so with `XYMONRUNDIR=/run/xymon` reload answered *"xymond not running"* and rotate signalled nobody | reads `@XYMONRUNDIR@`, its own launcher pidfile included, and creates the directory first — `/run` is gone after a reboot |
| `xymon.sh` unknown command | printed usage to stdout and fell through to `exit 0`, so an invalid argument reported success | `>&2` and `exit 2` |
| `xymond`'s default pidfile | unbounded `sprintf()`, still naming the log directory | `$XYMONRUNDIR`, bounded, and the write is skipped rather than `fopen(NULL)` |
| `xymond_history --pidfile=` | bounded on the default path, `strcpy()` on the one the shipped `tasks.cfg` actually passes | bounded |
| `XYMONCLIENTRUNDIR` ordering | defined before `XYMONCLIENTHOME`; `loadenv()` expands each line as it reads it, so it resolved to `/logs` with an `xgetenv: Cannot find value` error | home first |
| an unreconfigured tree | a top-level Makefile predating this PR defines neither variable, so every placeholder is substituted with nothing: `/xymond.pid`, `/rrdctl.<pid>`. As root those are created rather than refused, and `xgetenv()` cannot rescue it — it falls back on a *missing* value, not an empty one | `ifeq ($(strip …),)` in `build/Makefile.rules`. `?=` is not enough: it assigns only when a variable is undefined |
| `xymond` shutdown | the bounded default above can leave `pidfile` NULL, and shutdown still called `unlink(pidfile)` on it — the write guard's own reasoning, not applied to the matching removal | guarded the same way |
| four manual pages | described `$XYMONTMP` as the socket directory and `$XYMONSERVERLOGS` as the pidfile one | say what the code does |

`XYMONCLIENTRUNDIR` was also the one variable a packager could not set without patching a shipped file — it was in neither configure script. It now takes the same route as `XYMONRUNDIR`: prompt or environment, Makefile line, `@placeholder@`, sed clause.

Its default is resolved at configure time (`$XYMONTOPDIR/client/logs`) rather than left as the reference `$XYMONCLIENTHOME/logs`. That indirection cannot survive the build: `Makefile.rules` passes these values to the sub-makes as double-quoted shell words, so the recipe's own shell expands it — measured, the generated file read `XYMONCLIENTRUNDIR="/logs"`. Every other path in `xymonserver.cfg` is already a literal fixed at configure time; this line was the exception. **Moving `XYMONCLIENTHOME` after installation now means moving this too**, which the comment says.

## Deliberate, not overlooked

- **`runclient.sh` keeps its own launcher pidfile beside its log.** It never sources `xymonclient.cfg` — it derives `XYMONCLIENTHOME` from `dirname $0` and hardcodes `logs/` for its writability check — so it has no access to `XYMONCLIENTRUNDIR`. Giving it access changes the client startup path for every install, which is its own change with its own risk. Nothing breaks: the launcher's pidfile stays where everything already looks for it.
- **`xymon.sh` bakes `@XYMONRUNDIR@` rather than resolving it at run time**, and its `mkdir` is create-or-fail. Both are the design `tests/buildsystem/xymon-sh-pidfile-dir.sh` enforces: `start` must not succeed with no pidfile.
- **Upgrade migration of an existing `tasks.cfg`** is a separate build/install concern, left as a known limitation.
- **`xymonfetch`** is passed a `--pidfile=` it does not implement; that predates this change, and #172 removes the flag from the shipped task.

## Verification

| | |
|---|---|
| tests | `xymond-rrd-ctlsocket-path.sh` (real `xymond_rrd` and `rrdcachectl` meeting on one socket) · `xymon-sh-pidfile-dir.sh` (both ends name the same directory, each side its own, both settable from the environment, every placeholder substituted, and neither left empty by an old Makefile) · `testsuite-stdin-isolation.sh` |
| mutations that fail them | the socket path back on `XYMONTMP` · the control script back on the log directory · the `mkdir` dropped · the client path back on the server's name · `xymonserver.cfg` not defining the client's · configure not reading `XYMONCLIENTRUNDIR` from the environment, or not writing it to the Makefile · the Makefile not substituting its placeholder · either empty-value guard removed |
| end to end | configured both ways and generated the file — default gives `/usr/lib/xymon/client/logs`, `XYMONCLIENTRUNDIR=/run/xymon-client` gives that, no `@VAR@` left |
| CI | `build.yml` now passes both directories to configure. It passed neither before, so every build tested only the case where `XYMONRUNDIR` equals the log directory — the one case this PR is not about |
| suite | 100 passed, 0 skipped, 0 failed |

The placeholder-substitution assertion is not decoration: it caught the `"/logs"` defect above while it was being written.

**The runner is hardened here too**, because writing these tests found the fault: `tests/testsuite` walked its list with `while read … done <<EOF`, so the list was on its own stdin — the `rrdcachectl` call above read it and cut a 75-test run to 51, reported as *51 passed, 0 failed*. Each test now gets stdin from `/dev/null`.

## Reviewed

Roland Rosenfeld raised four points in review; all four are answered above — the `build.yml` example, the reason for the default, and the three `XYMONCLIENTRUNDIR` questions. The fourth concerned `xymonproxy`: its `--pidfile=` is parsed but written only when daemonising, so the shipped `--no-daemon --pidfile=…` never produced one. Measured on the built binary:

```
--no-daemon --pidfile=X  ->  pidfile NOT written
--daemon    --pidfile=X  ->  pidfile WRITTEN
```

#172 replaces it with a launcher-written `PIDFILE` entry, and now also stops the proxy unlinking a pidfile it never wrote. The remaining binary-level faults are **#554**.


bonomani:
Superseded: folded into the pull request description (see *What the original patch left behind*).

bonomani:
Superseded: the one real finding from this pass (`XYMONCLIENTRUNDIR` defined before `XYMONCLIENTHOME`) is in the pull request description, under *What the original patch left behind*. The items judged by-design are under *Deliberate, not overlooked*.

bonomani:
Superseded: both fixes from this pass (`xymon.sh` reporting success on an unknown command, and `rrdcachectl` composing a `"(null)/…"` socket path) are in the pull request description, under *What the original patch left behind*.

bonomani:
Superseded: the empty-`XYMONRUNDIR` defect is now fixed — `build/Makefile.rules` defaults both runtime directories when the top-level Makefile predates them, and `xymon-sh-pidfile-dir.sh` fails if either comes out empty. The `runclient.sh` decision is stated in the description under *Deliberate, not overlooked*.

RolandRosenfeld:
I'm currently testing this PR (may take some time until I notice all potential issues).

Maybe you should add XYMONRUNDIR=/run/xymon to .github/workflow/build.yml as an suggested example (in my first test I forgot to add this to debian/rules and wondered, why the pid files are still created in /var/log/xymon).

Is it really a good idea to use $XMONLOGDIR (and not /run/xymon or /var/run/xymon) as the default for XYMONRUNDIR? Is this for backward compatibility or what's the reason for this?

Now I try to understand, how XYMONCLIENTRUNDIR is intended to work:
- in client/xymonclient.cfg.DIST it is set to $XYMONCLIENTLOGS
- in xymond/etcfiles/xymonserver.cfg.DIST it is set to $XYMONCLIENTHOME/logs
- in lib/environ.c it is set to $XYMONRUNDIR
This sounds a little inconsistent to me.  I don't see, why we need two different variables for client and server and why the client variable cannot been set via configure script/environment variable (which implies, that for Debian packaging I still have to apply patches to set this to my preferred location).

As a followup to this PR: xymonproxy has its pidfile hardcoded as /var/run/xymonproxy.pid (can be overridden via command line option), maybe this should be changed to $XMONRUNDIR, too (Hmmm, on a server with proxy I use `xymonproxy --pidfile=/var/run/xymon/xymonproxy.pid` no pid file exists in /var/run/xymon, maybe another issue, that should be checked, but this is also 4.3.30!).

This is an interim result, this will take more time...

bonomani:
All four points are answered in the pull request description, which has been rewritten to carry them: the `build.yml` example, the reason the default is the log directory (with #415 as the other half), and the three `XYMONCLIENTRUNDIR` questions — including the one you were right about, that it could not be set without patching a shipped file.

Two things worth pulling out rather than leaving you to find them in the diff:

- **A behaviour change.** `XYMONCLIENTRUNDIR` no longer defaults to the reference `$XYMONCLIENTHOME/logs` but to a path fixed at configure time. The indirection cannot survive the build — `Makefile.rules` passes these values as double-quoted shell words, so the recipe's own shell expanded it and the generated file read `XYMONCLIENTRUNDIR="/logs"`. Moving `XYMONCLIENTHOME` after installation now means moving this too.
- **Your `xymonproxy` observation is a real bug, reproduced.** `--pidfile=` is written only inside `if (daemonize)`, while the shipped task passes `--no-daemon` *and* `--pidfile=`, so it never produced one. #172 already replaces it with a launcher-written `PIDFILE`, and now also stops the proxy unlinking a pidfile it never wrote. The remaining faults in the binary are #554.

The PR has also been rebased onto current `main`; its previous CI run was from 29 August.

bonomani:
Superseded: see the end of the pull request description.

PR #221, opened 2026-07-05
xymond_filestore: harden update_file(), confine the notes and HTML log names (TBT 45 + devel a70282c11)

bonomani:
Hardens the three writers in xymond_filestore against symlink/hardlink attacks, path escapes, buffer overflow and silent write loss, with a test. Two commits: guard the `update_file()` handle + notes name; then harden all writers, confine both name halves, refuse leading-dot/backslash names, and fix an uninitialised HTML field.

The writers built a predictable temp/target name and opened it with `fopen()`, which follows symlinks; names came off the channel unvalidated. Each writer now unlinks then opens `O_CREAT|O_EXCL|O_NOFOLLOW` (stops both a planted symlink and a hardlink), builds bounded names, confines each component with `name_confined()` (refuse `""`, leading `.`, `/`, `\`), and latches a partial write instead of renaming a fragment over the good file.

| case | before | after |
|---|---|---|
| symlink/hardlink at the temp name | payload written through to the target | refused (`O_EXCL`+`O_NOFOLLOW`) |
| notes ID `,victim` (→ `.victim`) then note `victim` | the `.victim` note deleted by `victim`'s temp unlink | leading-dot refused |
| notes ID of 4592 bytes | `sprintf` overflow onto another note's name | refused |
| `@@renamehost oldhost link/moved`, `@@renametest … a/../../escaped` | renamed out of the directory through a symlink | confined |
| host `../evil` under `--status --html` | wrote `$XYMONVAR/evil.conn.html` | confined |
| test name `a\b` on an SMB/CIFS `$XYMONVAR` | escaped via `\` | refused |
| negative "time since change" | uninitialised `timestr` printed into the page | terminated |
| failed `fwrite`/`fclose` | fragment renamed over the good file | latched, temp removed |
| `basename()` twice in one `snprintf` | `conn.conn` on NetBSD/OpenBSD | refusal, no `basename()` |

**Test:** `tests/xymond/filestore-notes-path.sh` drives the real worker over each case by mutation, each with a positive control (an ordinary note/page/disable/rename must still happen). The oversized-ID case builds under ASan, since a stack overflow is invisible to a plain build.

### DIFF
--- a/tests/xymond/filestore-notes-path.sh
+++ b/tests/xymond/filestore-notes-path.sh
@@ -1,0 +1,720 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/xymond/filestore-notes-path.sh
+#
+# Regression guard for the two update_file() defects in xymond_filestore.
+#
+# 1. The @@notes handler built its destination as basename(filedir)/hostname
+#    -- a cwd-relative base, from the basename of the configured directory --
+#    with the host part taken verbatim from the message. xymond deliberately
+#    does not validate the ID on notes messages.
+#
+# 2. update_file() wrote through its fopen() handle without checking it. A
+#    destination that cannot be opened for writing therefore reached
+#    fwrite(msg, len, 1, NULL) and crashed the worker. A "." ID reaches
+#    exactly that: update_file() writes to "$dir/.$base", so "." lands on
+#    the directory itself, which fopen(..., "w") refuses.
+#
+# Workers read their messages from stdin (xymond_worker.c), so this drives
+# the real binary and checks the filesystem.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+# shellcheck source=tests/lib/build-worker.sh
+. "$(dirname "$0")/../lib/build-worker.sh"
+
+ROOT=$(find_root)
+
+# build_xymond_worker resolves GNU make, the configured compile and link flags
+# and the library set for us. Hand-rolling those meant literal `make`, which
+# BSD make cannot parse on the generated GNU Makefiles, so this test failed
+# before reaching an assertion on FreeBSD, NetBSD and OpenBSD.
+work=$(mktempdir)
+build_xymond_worker "$work" xymond_filestore \
+	xymond/xymond_filestore.c xymond/xymond_worker.c
+
+mkdir -p "$work/etc"
+cp "$ROOT/xymond/etcfiles/xymonserver.cfg.DIST" "$work/etc/xymonserver.cfg"
+
+setup() {
+	rm -rf "$work/var"
+	mkdir -p "$work/var/notes" "$work/var/sibling" "$work/var/logs" "$work/var/html"
+	printf 'CANARY-OUTSIDE-NOTES\n' >"$work/var/CANARY.txt"
+	printf 'sibling content\n'      >"$work/var/sibling/keepme"
+}
+
+notes() {  # notes <id>
+	local rc
+	set +e
+	printf '@@notes#1|1785830400|10.0.0.99|%s\nNOTE PAYLOAD\n@@\n' "$1" \
+	| XYMONHOME="$work" XYMONVAR="$work/var" \
+		"$work/xymond_filestore" --env="$work/etc/xymonserver.cfg" \
+		--notes --dir="$work/var/notes" >/dev/null 2>>"$work/worker.log"
+	rc=$?
+	set -e
+	# 0 is the normal exit when the input pipe closes. A crash (139) is how
+	# the unchecked fopen() handle failed, so it must not read as "refused".
+	[ "$rc" -le 1 ] || fail "xymond_filestore exited $rc (crash?) on notes ID '$1'"
+}
+
+# Sanity first: if nothing is ever written here, every "did not escape"
+# assertion below holds vacuously.
+setup
+: >"$work/worker.log"
+notes "realhost"
+[ -f "$work/var/notes/realhost" ] \
+	|| { cat "$work/worker.log" >&2; fail "a legitimate notes message was not stored -- the checks below would pass vacuously"; }
+assert_contains "NOTE PAYLOAD" "$(cat "$work/var/notes/realhost")" "notes payload not stored"
+
+# No ID may put a file outside the configured notes directory, and none may
+# crash the worker.
+for id in ".." "." "../evil" "/etc/passwd" "a/b" "//"; do
+	setup
+	notes "$id"
+	escaped=$(find "$work/var" -maxdepth 1 -type f ! -name 'CANARY.txt' | head -1)
+	[ -z "$escaped" ] || fail "notes ID '$id' wrote outside the notes directory: $escaped"
+	[ -f "$work/var/CANARY.txt" ] || fail "notes ID '$id' clobbered a file outside the notes directory"
+	[ -f "$work/var/sibling/keepme" ] || fail "notes ID '$id' touched a sibling directory"
+done
+
+# ---- an oversized ID is refused, and one that fits exactly does not overflow -
+#
+# The notes ID has no length check anywhere before this worker: xymond accepts
+# a 64 MiB message, the notes channel carries 256 KiB, and posttochannel()
+# copies the whole ID through.
+#
+# Two different sizes fail in two different ways. A grossly long one must be
+# refused outright, leaving the directory as it was rather than truncated into
+# some other note's name.
+#
+# 5000 bytes overflows the PATH_MAX buffer the name is built into, so the notes
+# handler's snprintf bound refuses it before the filesystem is touched -- and it
+# does so on every platform, since 5000 exceeds PATH_MAX everywhere (1024 on the
+# BSDs, 4096 on Linux). An earlier version built a >3900-byte scratch directory
+# here to force that; it was never used by the message below and only broke the
+# BSD lanes, where mkdir of such a path fails under set -e (@SoundGoof).
+before=$(ls "$work/var/notes" | sort | tr '\n' ' ')
+notes "$(printf 'a%.0s' $(seq 1 5000))"
+after=$(ls "$work/var/notes" | sort | tr '\n' ' ')
+
+assert_equal "$before" "$after" \
+	"an oversized notes ID changed the notes directory -- it must be refused, not truncated into some other note's name"
+assert_file_exists "$work/var/CANARY.txt" \
+	"an oversized notes ID reached outside the notes directory"
+
+# The overflow @SoundGoof reported under ASan (WRITE of size 4097) is not that
+# one, and asserting on a refusal cannot show it either way: the write is past
+# a stack buffer, which a plain build survives silently. Every caller builds
+# the destination with a bounded snprintf() into a PATH_MAX buffer, so 5000
+# bytes never reaches update_file(). What does reach it is a destination that
+# fits PATH_MAX *exactly*: update_file() inserts a '.' before the basename, so
+# the temporary name is one byte longer than the name it is derived from, and
+# at PATH_MAX-1 that byte is one too many. Accounting for it is the whole of
+# the fix, so the case is worth pinning even though it needs a second build.
+if asan_usable; then
+	asanwork=$(mktempdir)
+	# A wrapper rather than CC="cc -fsanitize=...": build_xymond_worker
+	# resolves $CC with command -v, which cannot see a multi-word string.
+	cat >"$asanwork/cc-asan" <<EOF
+#!/bin/sh
+exec ${CC:-cc} -fsanitize=address -g "\$@"
+EOF
+	chmod +x "$asanwork/cc-asan"
+	CC="$asanwork/cc-asan" build_xymond_worker "$asanwork" xymond_filestore \
+		xymond/xymond_filestore.c xymond/xymond_worker.c
+
+	mkdir -p "$asanwork/etc" "$asanwork/var/notes"
+	cp "$work/etc/xymonserver.cfg" "$asanwork/etc/xymonserver.cfg"
+
+	# The buffers are PATH_MAX from <limits.h>, so ask the compiler instead of
+	# assuming 4096: a wrong guess would quietly stop reaching the overflow and
+	# leave this half asserting nothing.
+	printf '#include <limits.h>\n#include <stdio.h>\nint main(void){printf("%%d\\n", PATH_MAX);return 0;}\n' \
+		>"$asanwork/pathmax.c"
+	"${CC:-cc}" -o "$asanwork/pathmax" "$asanwork/pathmax.c" 2>/dev/null \
+		|| fail "cannot determine PATH_MAX"
+	pathmax=$("$asanwork/pathmax")
+
+	notesdir="$asanwork/var/notes"
+	# "<notesdir>/<id>" of exactly PATH_MAX-1: the longest destination that is
+	# not refused, hence the one whose derived ".<id>" no longer fits.
+	idlen=$(( pathmax - 1 - ${#notesdir} - 1 ))
+	[ "$idlen" -gt 0 ] || fail "the work directory is too deep to build a PATH_MAX-1 notes path"
+
+	set +e
+	printf '@@notes#1|1785830400|10.0.0.99|%s\nNOTE PAYLOAD\n@@\n' \
+		"$(printf 'a%.0s' $(seq 1 "$idlen"))" \
+	| XYMONHOME="$asanwork" XYMONVAR="$asanwork/var" \
+		"$asanwork/xymond_filestore" --env="$asanwork/etc/xymonserver.cfg" \
+		--notes --dir="$notesdir" >/dev/null 2>"$asanwork/asan.log"
+	rc=$?
+	set -e
+
+	if grep -q AddressSanitizer "$asanwork/asan.log"; then
+		sed -n '1,12p' "$asanwork/asan.log" >&2
+		fail "a notes ID whose destination fills PATH_MAX exactly overflowed the temporary name"
+	fi
+	# ASan aborting without leaving its banner in the log -- a plain crash, or a
+	# report written elsewhere -- would slip past the grep. The normal exit is 0
+	# (pipe closed); anything past 1 is the overflow crashing the worker.
+	[ "$rc" -le 1 ] || fail "the exact-fit notes worker exited $rc (crash?) with no ASan banner"
+else
+	printf 'note: %s cannot build and run ASan binaries, the overflow itself is unverified\n' \
+		"${CC:-cc}" >&2
+fi
+
+
+# ---- the "--status --html" role writes a second file, from the same name ----
+# It inserts the hostname raw into "$htmldir/<host>.<test>.html", so a '/'
+# in the name lands the file outside htmldir. Shipped DISABLED in tasks.cfg
+# ("storestatus"), but the code path is the same class as the one above.
+status_html() {  # status_html <hostname> [testname]
+	local rc
+	set +e
+	printf '@@status#1|1785830400.000000|10.0.0.99|origin|%s|%s|0|green|flags|green|1785830000|0||0||0|0\nSTATUS BODY\n@@\n' "$1" "${2:-conn}" \
+	| XYMONHOME="$work" XYMONVAR="$work/var" \
+		"$work/xymond_filestore" --env="$work/etc/xymonserver.cfg" \
+		--status --dir="$work/var/logs" --htmldir="$work/var/html" \
+		>/dev/null 2>>"$work/worker.log"
+	rc=$?
+	set -e
+	[ "$rc" -le 1 ] || fail "xymond_filestore exited $rc (crash?) on status '$1'/'${2:-conn}'"
+}
+
+setup
+status_html "realhost"
+[ -f "$work/var/html/realhost.conn.html" ] \
+	|| { cat "$work/worker.log" >&2; fail "a legitimate status HTML log was not written -- the checks below would pass vacuously"; }
+# Both halves of the name must survive into it. basename(3) may return storage
+# that the next call overwrites -- NetBSD and OpenBSD do -- so building the
+# name with two calls inside one snprintf() left both arguments naming the same
+# string and stored realhost/conn as conn.conn (measured before basename() was
+# retired for refusal). Kept: any future reduction of the components has to
+# preserve both halves, whatever it is built from.
+[ -f "$work/var/logs/realhost.conn" ] \
+	|| fail "the raw status log was not written as <host>.<test>: $(find "$work/var/logs" -type f)"
+[ ! -f "$work/var/logs/conn.conn" ] \
+	|| fail "both name components came out as the test name -- basename() results aliased"
+
+# ---- a negative "time since change" must not print uninitialised bytes ------
+# update_htmlfile()'s timestr is function-scoped and handed to
+# generate_html_log() unconditionally, which prints "Status unchanged in <it>"
+# when it is non-empty. The block that fills it runs only for a non-negative
+# delta; a changetime later than the log time (a clock step) makes the delta
+# negative and skips it, so an uninitialised timestr leaks into the served
+# page. Two messages in one worker run: the first (large positive delta) dirties
+# the stack slot, the second (negative delta) must still print no such line.
+setup
+{
+	printf '@@status#1|1785830000.000000|10.0.0.99|origin|realhost|oldone|0|red|f|red|1700000000|0||0||0|0\nBODY1\n@@\n'
+	printf '@@status#2|1785830000.000000|10.0.0.99|origin|realhost|negdelta|0|red|f|red|1785830400|0||0||0|0\nBODY2\n@@\n'
+} | XYMONHOME="$work" XYMONVAR="$work/var" \
+	"$work/xymond_filestore" --env="$work/etc/xymonserver.cfg" \
+	--status --dir="$work/var/logs" --htmldir="$work/var/html" >/dev/null 2>>"$work/worker.log"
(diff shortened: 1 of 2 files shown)


SoundGoof:
this needs the same thing as #314

SoundGoof:
## Findings

###  The test name still permits path traversal

`xymond/xymond_filestore.c:327,363-364`

The HTML fix confines only `hostname`. It inserts `testname` unchanged into
both the raw status and HTML paths.

I sent a status with test name `../../../escaped` through the real worker.
It created:

```text
$XYMONVAR/escaped
$XYMONVAR/escaped.html
```

Both are outside the configured `logs` and `html` directories. No upstream
status-channel validation rejects the slash-bearing test name.

Confine both hostname and test name before constructing either filename.
Add hostile test-name cases to the regression test.

### The temporary file follows symlinks outside the destination

`xymond/xymond_filestore.c:49-59`

`update_file()` uses the predictable temporary name `.<basename>` and opens
it with `fopen()`, which follows an existing symlink.

I created:

```text
notes/.victim -> ../CANARY
```

and sent a note for `victim`. The canary outside `notes` was replaced with
the note payload. The symlink was then renamed to `notes/victim`.

Use a temporary-file operation that refuses symlinks, such as
`open(..., O_NOFOLLOW | O_CREAT | O_EXCL, ...)`, and add a seeded-symlink
case.

### The new test does not run on BSD

`tests/xymond/filestore-notes-path.sh:28-54`

The script invokes literal `make` and hand-builds compiler and linker flags.
It fails before reaching any assertion on FreeBSD 14.3, NetBSD 11.0, and
OpenBSD 7.9 because BSD make cannot parse the generated GNU Makefiles.

Current `main` already provides `tests/lib/build-worker.sh`, which uses
`require_gnu_make`, `$XYMON_MAKE`, `xymon_cflags`, and configured libraries.
Rebase and use that helper.

The branch is currently 150 commits behind `main`; its local suite contains
only 19 tests.

## Validation

- Existing focused test passes on Linux.
- Branch-local suite: 18 passed, 1 expected skip, 0 failed.
- Focused test fails before assertions on all three BSDs.
- Test-name traversal and temporary-symlink overwrite reproduced.
- `git diff --check` passes.
- The PR merges cleanly.

bonomani:
All three addressed in `4696ca874`, with the branch brought up to date with `main` first.

### The temp file follows symlinks — confirmed, fixed

Reproduced exactly as you described. `notes/.victim -> ../CANARY`, send a note for `victim`, and the canary outside `notes` holds `NOTE PAYLOAD` while the symlink has been renamed to `notes/victim`.

Opened with `O_NOFOLLOW` now; the same attack fails the open with `ELOOP`, leaves the canary intact and writes nothing.

I did not add `O_EXCL`. A leftover temp file from a crashed run is legitimate and should still be truncated the way it is today — `O_NOFOLLOW` is what defeats the attack, and `O_EXCL` would turn a crash into a permanently stuck worker.

### The test name — real gap, five sites, but I could not make it escape

You are right that it was inserted unconfined while only the hostname was confined, and it is worse than the two sites in your note: **status, data, enadis, droptest and the HTML log** all build `"<dir>/<host>.<test>"`. All five now go through one helper that confines both components, which also converts the remaining `sprintf()` calls to `snprintf()`.

But I could not reproduce the file creation. With `filedir=$XYMONVAR/logs` and hostname `myhost`, `../../../escaped` produces `logs/myhost.../../../escaped`, whose first component is `myhost..` — nonexistent, so the open fails with ENOENT and nothing is written. I tried `../../../escaped`, `/../../escaped`, `./../../escaped`, a slash-bearing hostname and an empty hostname; none wrote anything outside the configured directories. The `<host>.<test>` join appears to be protecting it by accident.

So I have fixed it as hardening, and I would rather say plainly that its test cases pass with or without the change — the seeded-symlink case is the one that fails on the unfixed worker.

**What did your reproduction use?** If it created `$XYMONVAR/escaped` there is an input shape I have not found, and that would make this a live traversal rather than a latent gap — worth knowing which it is.

### The test does not run on BSD — fixed

Replaced the hand-rolled build with `build_xymond_worker()`, which resolves GNU make, the configured compile and link flags and the library set. That was the same defect as #340 and #346 in a third place.

Added hostile test-name cases and a seeded-symlink case. Suite: 58 passed, 0 skipped, 0 failed on Linux; shellcheck on the test drops from 4 findings to 2.

I cannot run the BSDs, so the portability half is unverified here — if you re-run it there, that is the part I would want confirmed.

SoundGoof:
## Findings

### High: `basename()` aliases both filename components on NetBSD and OpenBSD

[xymond/xymond_filestore.c:61](https://github.com/xymon-monitoring/xymon/blob/4696ca874f60940811e24eae8b74b4de30147564/xymond/xymond_filestore.c#L61-L64)

Some BSD `basename()` implementations return shared static storage. Calling it
twice within one `snprintf()` therefore makes both arguments point to the last
result.

On both NetBSD 11 and OpenBSD 7.9, a status for host `realhost` and test `conn`
was stored as:

- `logs/conn.conn`
- `html/conn.conn.html`

The added test fails its legitimate-status assertion on both systems.

Store each `basename()` result in independent buffers before calling
`snprintf()`.

### High: A failed write still replaces the valid destination

[xymond/xymond_filestore.c:109](https://github.com/xymon-monitoring/xymon/blob/4696ca874f60940811e24eae8b74b4de30147564/xymond/xymond_filestore.c#L109)
[xymond/xymond_filestore.c:124](https://github.com/xymon-monitoring/xymon/blob/4696ca874f60940811e24eae8b74b4de30147564/xymond/xymond_filestore.c#L124-L138)

An `fwrite()` failure is logged but does not prevent `rename()`. `fclose()` can
then succeed because the stream has no buffered data left, allowing the
partial temporary file to replace the previous valid file.

With `RLIMIT_FSIZE` set to 1024 bytes, the exact worker reported `File too
large`, exited successfully, and replaced an existing valid note with a
1024-byte fragment.

Latch failures from `fwrite()`, `fprintf()`, and `ferror()`, remove the
temporary file, and skip the rename.

## Validation

All five GitHub checks are green. Debian passed the focused test three times
and the full suite passed 55 tests with 3 unrelated skips. FreeBSD passed the
focused test. NetBSD 11 and OpenBSD 7.9 fail because of the filename regression
above.

bonomani:
Both confirmed and fixed.

**`basename()` aliasing.** Each result is copied out before the other call is made. Nothing to reproduce here — glibc returns a pointer into the argument — so the guard is a pair of assertions that are inert on Linux and decisive on the BSDs: the raw log must be `realhost.conn`, and `conn.conn` / `conn.conn.html` must not exist. Stated as such in the test, so nobody reads a green Linux run as evidence.

**Partial file renamed into place.** Failures from `fwrite()`, `fprintf()` and `ferror()` are latched, and a temp file that was not written whole is removed instead of renamed. Driven with `ulimit -f 1` and `SIGXFSZ` ignored — an ignored disposition survives `exec`, so the worker sees `EFBIG` instead of dying. Reverting the latch gives exactly the defect you described:

```
FAIL: a write that could not complete replaced the previous note with a fragment:
expected 'NOTE PAYLOAD', got 'XXXXXXXX...'  (1024 bytes)
```

One thing came along with the first fix and is worth flagging rather than burying: copying the basenames into arrays of their own let gcc see the worst case and emit two `-Wformat-truncation` warnings, which this tree builds clean of. I resolved it by checking `snprintf()`s result rather than silencing it — a truncated name is as likely to be another host's or another test's, so nothing is written under it and the empty path fails at `open()`, which every caller already reports. Build is warning-free again.

bonomani:
Restructured to the two-commit shape this repo uses for TBT ports: J.C. Cleaver's original, then a single correction. It had drifted to two correction commits as your second round of findings came in — same content, same tree, just folded together so the original stays first and the corrections read as one answer.

Both rounds are in that message: the `O_NOFOLLOW` temp file, the five unconfined name sites and the raw hostname in the HTML role, the `basename()` storage reuse that produced `conn.conn` on the BSDs, and the failed `fwrite()` whose fragment was renamed over a good file.

Tree unchanged by the restructure. Suite green.

SoundGoof:
### An oversized notes ID overflows `tmpfn`

`xymond/xymond_filestore.c:482` silently truncates the notes destination into
the `PATH_MAX`-sized `logfn`. `update_file()` then inserts an additional dot
with an unbounded `sprintf()` at line 104 into another `PATH_MAX` buffer.

This is reachable through production `xymond`, assuming the sender is
authorized for notes:

- Network input accepts messages up to 64 MiB.
- Notes IDs have no length validation, only a nonblank check.
- The notes channel defaults to 256 KiB.
- `posttochannel()` copies the complete ID with `%s` into that channel.

A 5000-byte notes ID is therefore accepted and forwarded to
`xymond_filestore`. Sending one through the exact worker under ASan produced:

    ERROR: AddressSanitizer: stack-buffer-overflow
    WRITE of size 4097
    #2 update_file xymond/xymond_filestore.c:104
    #3 main xymond/xymond_filestore.c:484

Reject truncated notes paths and construct the temporary name with checked
bounds. `update_file()` must account for the extra dot for every caller, since
a destination of `PATH_MAX - 1` bytes cannot fit in `tmpfn` after that prefix
is added. Add an oversized notes-ID case to the regression.

bonomani:
Fixed, and the input you reported is not the one that overflows.

`update_file()` now builds `tmpfn` with a bounded `snprintf()` and refuses a name that does not fit; the notes role, the one caller not going through `build_logfn()`, checks its own `snprintf()` the same way.

On the regression case: 5000 bytes does not reach the overflow once the bounds are in place — every caller builds the destination into a `PATH_MAX` buffer, so an oversized ID is refused upstream of `update_file()`. What reaches it is a destination of exactly `PATH_MAX-1`: the `.` makes the derived name one byte longer, which is your `WRITE of size 4097`. The test asserts both — 5000 bytes refused with the notes directory unchanged, and the exact-fit case against a second worker built under ASan (`asan_usable()`, like the other sanitizer tests). Restoring the unbounded `sprintf()` fails it; with the 5000-byte case alone it passed, which is why the exact-fit one is there. `PATH_MAX` is read from the compiler rather than assumed.

One thing I will not claim: the notes-side refusal is not independently observable. A truncation always fills the buffer to `PATH_MAX-1`, which the `tmpfn` bound then rejects on its own, and both builds report the dropped note — mutating the notes check out leaves the test green. It is kept for the reason `build_logfn()` has it: a caller establishing that its own name fits, rather than the property resting on `sizeof(tmpfn) == sizeof(logfn)` in another function.

Amended into the correction commit; structure unchanged — J.C. Cleaver's original first, one correction on top.

SoundGoof:
## Blocking finding

The HTML writer retains the same predictable temporary-file symlink
vulnerability that this PR fixes in `update_file()`.

`update_htmlfile()` still performs:

```c
tmpfn = malloc(strlen(fn) + 5);
sprintf(tmpfn, "%s.tmp", fn);
output = fopen(tmpfn, "w");
...
rename(tmpfn, fn);
```

Anyone able to create an entry in the HTML output directory can plant
`<host>.<test>.html.tmp` as a symlink. `fopen()` follows it, truncates and writes
the target with the daemon's privileges, and `rename()` then moves the symlink
into place as the HTML status file.

I reproduced this through the real PR worker:

1. Created `html/realhost.conn.html.tmp -> ../CANARY.txt`.
2. Sent an ordinary status through `--status --html`.
3. The worker exited 0.
4. `CANARY.txt` was replaced with generated HTML.
5. `realhost.conn.html` became the attacker-planted symlink.

Apply the same no-follow and checked-write treatment to `update_htmlfile()`.
Add this case beside the existing `.seeded` notes test. The HTML writer should
also check `fclose()` and `rename()` before publishing the generated file.

## Validation

Reviewed exact head `5d028e67f0e4362f5abd4c8fea632032c43411ea`.

The submitted focused real-worker test passes, including its ASan path where
available. The additional HTML symlink probe reproduces the finding.

All five GitHub checks are green. The full suite reports 71 passed, 3 skipped,
and 0 failed.

bonomani:
Confirmed and fixed. Reproduced here first, exactly as you describe: a symlink at `html/realhost.conn.html.tmp` had the worker rewrite `CANARY.txt` with generated HTML, exit 0, and publish the symlink as `realhost.conn.html`.

`update_htmlfile()` now opens `O_NOFOLLOW` like `update_file()`, and a page that was not written whole is removed rather than renamed — `fclose()` and `rename()` are both checked. A legitimate page still writes and leaves no `.tmp` behind.

That it survived two rounds is the part worth naming: I fixed the writer the review pointed at and did not ask which other writer in the same file had the same shape. The `.seeded` case existed and its sibling did not.

**Test**, beside `.seeded` as you asked, plus the write that cannot finish:

| mutation | result |
|---|---|
| `fopen()` put back | `FAIL: a symlink planted at the HTML temp name let the page overwrite a file outside htmldir` |
| `fclose()` check dropped | `FAIL: an HTML page that could not be written whole replaced the served one with a fragment` |

One thing measured rather than assumed: on the short-write path the failure surfaces at `fclose()`, so dropping `ferror()` alone changes nothing there. `ferror()` covers the other shape — a failure already reported with the buffer emptied, where `fclose()` has nothing left to say — and this test does not reach it. Said rather than implied.

Rebased on `2b2272eb9`. Structure unchanged: J.C. Cleaver's original, then one correction. Suite on a built tree: 74 passed, 0 skipped, 0 failed.

bonomani:
Rather than wait to be told about the next one, I put the question to a second reviewer: which *other* writers in this file have the shape you found? Four more, three of them real, all measured before being taken.

**Reachable from the network, not the config.** `renamehost` and `renametest` built both names with `sprintf()` from message fields, bypassing `build_logfn()` entirely — and those fields arrive on the channel. Measured on this branch: an oversized rename target aborts the worker.

```
FAIL: xymond_filestore exited 134 on an oversized rename target -- the name was built without a bound
```

Both are bounded now and the `rename()` result reported; the `drophost` sweep beside them is bounded too.

**`update_enable()` is the third writer with your shape.** It creates the disable-file for its timestamp alone, which is still a truncation of whatever a planted symlink points at. Measured: without `O_NOFOLLOW` the canary outside the directory is emptied, with it intact.

**`build_logfn()`'s refusal was being handed on.** An empty name reached `update_htmlfile()`, which built `".tmp"` from it and wrote that in whatever directory the worker happened to be running from — visible in the log as `Cannot rename .tmp to :`. Only the rename check from the last round kept it from staying there.

Also: a temp file created by `open()` but unusable after `fdopen()` is removed rather than left.

Not taken from that round: a notes destination of exactly `PATH_MAX-1` being accepted by the caller and then refused by `update_file()` was reported as a defect. It is the bound working as intended — the temp name is one byte longer than the destination.

| mutation | result |
|---|---|
| `O_NOFOLLOW` dropped from `update_enable()` | `FAIL: a symlink planted at the disable-file name let the worker truncate a file outside its directory` |
| the rename bound removed | `FAIL: xymond_filestore exited 134 …` |

The rename case asserts the exit status as well as the files, because against an unbounded build the worker dies — and a test that only looked at what was left afterwards would have described a dead process as a refusal.

Suite on a built tree: 74 passed, 0 skipped, 0 failed.

bonomani:
A fourth round, and this one landed on the fixes themselves rather than beside them. Three real, two not — all measured.

**Every name was shortened where it should have been refused.** Each component is taken in a fixed copy before `basename()`, and a copy that did not fit was silently truncated. The cut lands wherever the length says: after a `/` it leaves a short, *valid* basename that is not the one asked for — so the assembled-path bound I added sees nothing wrong and the write lands on another file.

```
notes ID: 4088×'a' + "/victim" + 500×'X'   (4592 bytes)
before:  notes/victim  ==  PAYLOAD-INTRUS      <- an existing note, overwritten
after:   notes/victim  ==  NOTE-LEGITIME       <- refused, nothing written
```

`build_logfn()` has the same cut for the status/data/enadis components, driven the same way through a host name. Both refuse now.

**`renametest` traverses.** Its destination was built from `newtestname` raw, and unlike the hostnames beside it nothing turns its dots into commas — it is the one field where a `/` and a `..` both survive. `a/../../escaped` moved the log clean out of the configured directory. Measured, then confined through `basename()` like every other component.

**Two reported and not taken.** A rename to `../evil` does *not* traverse: the dot-to-comma pass above it makes that `,,`, measured as a failed rename with the original left in place. And `O_NOFOLLOW` does not stop a **hard link** planted at the temp name — that is correct, and the answer to it is `O_EXCL`, which this change deliberately does not use because a leftover temp file from a crashed run is legitimate and must still be truncated. Worth recording as a known limit rather than pretending otherwise.

| mutation | result |
|---|---|
| `newtestname` used raw again | `FAIL: a rename target containing '..' moved the log out of the configured directory` |
| the notes copy-fit check removed | `FAIL: an oversized notes ID was truncated onto another note's name and overwrote it` |
| the `build_logfn()` copy-fit check removed | `FAIL: an oversized status hostname was truncated onto another host's log` |

The traversal case comes with an ordinary rename beside it — without that, it passes on a worker that renames nothing at all.

Suite on a built tree: 74 passed, 0 skipped, 0 failed.

bonomani:
Fifth round, and the findings are narrowing — two in the code, two about the tests, all four fair.

**renamehost was confined only half way, and that half is mine.** `renametest` got `basename()` in the last round; its sibling did not, so the target still went in raw. The dot-to-comma pass over it stops `..`, but not `/`:

```
@@renamehost oldhost -> link/moved     (link is a symlink to ../outside)
before:  outside/moved.conn            <- the log left the configured directory
after:   logs/moved.conn               <- confined
```

**The `utime()` that stamps a status with its expiry was unchecked.** That stamp is what the readers go by, so a silent failure publishes a file that lies about when it goes stale.

**Both test criticisms are right, and they are the criticism I made of my own work one round earlier.** I added an ordinary-rename control beside the renametest traversal case and did not do the same for the other two: the enadis symlink assertion and the oversized-renamehost assertion both pass on a worker that stopped writing or renaming altogether. Both have their positive control now, each verified by mutation — an enadis path forced to fail creation fails the control, a raw `newhostname` fails the escape case.

Suite on a built tree: 74 passed, 0 skipped, 0 failed.

Worth saying after five rounds: what these passes keep finding is not the defect you first reported but its siblings — the writer next to the writer, the field next to the field, the assertion without its control. That is a better argument for the review than any single finding.

bonomani:
One more push since my last comment, with **no behaviour change**: `git diff 012ce3ce5 535305189` touches comments only (20 insertions, 46 deletions, all inside `/* */`). The code comments had grown across five rounds and were repeating the commit message; they are down from 100 lines to 74 for 243 added lines of C.

The PR body is rewritten too — it still described the state three rounds ago. It now carries one evidence table and the two limits stated plainly: no `O_EXCL`, so a hard link at the temp name is not stopped, and `build_logfn()`'s refusal is asserted at two of its six call sites.

Suite on a built tree after the trim: 74 passed, 0 skipped, 0 failed.

bonomani:
Your HTML finding is fixed, and asking *which other writers in this file have the same shape* turned up five more. Four review rounds, merged here; the PR body carries the full state.

**Your finding.** `update_htmlfile()` opened its temp file with `fopen()`. Reproduced: a symlink at `html/realhost.conn.html.tmp` had the worker rewrite `CANARY.txt` with generated HTML, exit 0, and publish the symlink as the page. `O_NOFOLLOW` now, and a page not written whole is removed rather than renamed.

**Its siblings, found by looking rather than waiting:**

| | |
|---|---|
| `update_enable()` | the third writer with the same shape — it creates the disable-file for its timestamp alone, which still truncates whatever a planted symlink points at |
| `renamehost` / `renametest` | built both names with `sprintf()` from **channel fields**, bypassing `build_logfn()`. An oversized target aborted the worker (`exited 134`); `link/oldhost` and `a/../../escaped` moved logs out of the directory |
| every name | taken in a fixed copy before `basename()`, and a copy that did not fit was silently shortened. A cut after a `/` leaves a short, *valid* name that is not the one asked for: measured, a 4592-byte notes ID **overwrote the note called `victim`** |
| `build_logfn()`'s refusal | acted on at five of its six call sites |
| the log buffer, `fclose()`, `rename()`, `utime()` | leaked or unchecked |

**And four of my own test cases held for the wrong reason** — the more useful half of the review: an oversized ID longer than `NAME_MAX` (the filesystem refused it before this code decided anything), two write-failure cases that swallowed the worker's exit status, escape cases with no positive control, and a rename traversal case using a clean hostname while the bug was in the other field.

Every fix above fails the test when reverted. Not taken: `update_file()` receiving `""` builds `"."`, not a creatable name, so those five sites failed safely — the guards are for the contract. Not addressed: `O_EXCL`, deliberately, so a hard link at the temp name is not stopped.

SoundGoof:
## Review
###  Confinement aliases malformed names onto legitimate files

`build_logfn()` confines components using `basename()` rather than rejecting
components containing `/`.

This allows different wire identifiers to resolve to the same stored file.
For example:

```text
a/victim -> victim
```

I reproduced three consequences on the current head:

- Note `a/victim` replaced the existing note `victim`.
- Status host `a/victim` replaced both `victim.conn` and its HTML page.
- `renamehost oldhost a/victim` replaced `victim.conn` and removed
  `oldhost.conn`.

The existing tests verify that malformed names cannot escape the configured
directory, but they do not verify that existing files inside it remain
untouched.

Reject components containing `/` rather than reducing them with `basename()`.
Add collision regressions with an existing legitimate destination.

### Low: Patch fails `git diff --check`

`xymond/xymond_filestore.c:505` introduces trailing whitespace.



The focused `filestore-notes-path.sh` regression passes, as do all five CI
checks.

bonomani:
@SoundGoof Both your findings are fixed, and reviewing for the same shape elsewhere turned up more. Rebased on main and force-pushed (head `4eef9a5ec`); full suite 87/0/0, `git diff --check` clean. Each fix below fails the test when reverted.

**Your review:**
- **Aliasing (`a/victim` → `victim`).** `build_logfn()` no longer reduces with `basename()`; a component holding `/` is refused outright (`name_confined()`), at all four sites that reduced — status/data/notes/enadis/droptest via the helper, and both rename targets. That also retired the NetBSD/OpenBSD `basename(3)` static-storage trap. Collision regressions seed the legitimate destination and assert it survives.
- **`git diff --check`.** Clean.

**Found while widening the same question (real-probe review):**
- `name_confined()` also refuses `""`, `.` and `..` — a `.` test name had built a stray `realhost..`. Reachable only where the pass leaves a literal dot (notes ID, test names).
- Empty **source** hostname in `drophost`/`renamehost`: reachable when the field is not last (`@@drophost|..||extra`), the match-lead became `.` and unlinked the `.host.test` temp files. Both sources are `name_confined()` now.
- `O_NOFOLLOW` stopped a planted **symlink** but not a **hardlink** (a hardlinked temp wrote straight through to the linked file — reproduced at the notes temp, the HTML temp, and the disable-file name). And the data role appended to a crash-leftover temp and published the stale bytes. All three writers now `unlink()` then create with `O_EXCL`, so what is opened is always a file this run made and a leftover is dropped rather than reused.

Not taken, deliberately: the post-open rename/`utime` TOCTOU (needs `renameat`/`O_TMPFILE`; the plant bug is closed) and dropping the publish on a `utime` failure (keeps a staler file for a near-zero trigger). Commit structure unchanged — J.C. Cleaver's original plus the one correction on top.

SoundGoof:
### Findings

#### Medium: The regression fails before its assertions on every BSD

[tests/xymond/filestore-notes-path.sh:92](https://github.com/xymon-monitoring/xymon/blob/f3f0c1dfba2d85cbef10377ea6539cde6ab50b74/tests/xymond/filestore-notes-path.sh#L92-L98)

The test unconditionally constructs a directory path at least 3,900 bytes
long. All four tested BSD systems report `PATH_MAX=1024`, so `mkdir` fails
under `set -e`:

```text
mkdir: ...: File name too long
```

This prevents every subsequent security assertion from running on those
platforms.

The deep directory is not used by the 5,000-byte notes message or its
assertions. Removing that setup preserves the intended test and allows the
complete regression to pass. Alternatively, derive any required depth from
the compiler's `PATH_MAX` with enough room for the final component.

### Summary

No production-code findings.

The confinement checks cover status, data, notes, disable, drop, and rename
messages. The create-new temporary files and atomic replacement prevent the
documented symlink, hard-link, stale-temp, and partial-write failures.

### Validation

Reviewed latest head `f3f0c1df`.

- `git diff --check` passed
- Linux full suite: 84 passed, 3 skipped, 0 failed
- Unmodified focused test passed on Debian 13
- Unmodified focused test failed during setup on all four BSDs
- With only the unrelated deep-directory setup removed, the focused test
  passed on:
  - NetBSD 9.4
  - NetBSD 11.0
  - OpenBSD 7.9
  - FreeBSD 14.3
- Exact production head built successfully on all five native guests

bonomani:
@SoundGoof The BSD finding is fixed, and a later review round turned up three more, all in `92aebabbf` (head rebased on main; suite green).

**Your finding — deep-directory setup fails on BSD.** Fixed. The >3900-byte scratch dir is gone; the oversized case now sends the 5000-byte ID directly, which exceeds `PATH_MAX` on every platform (1024 BSD / 4096 Linux), so the bounded `snprintf()` refuses it before any filesystem call. The exact-fit overflow leg (`PATH_MAX-1`) reads `PATH_MAX` from the compiler rather than assuming 4096.

**Found in a follow-up review, each with a regression that fails on the unfixed worker:**
- `name_confined()` now refuses a **leading dot**, not just `.`/`..`: a stored note `.victim` is the exact temp name `update_file()` derives for `victim`, so writing `victim` `unlink()`ed the `.victim` note. Reachable as wire `,victim` (xymond uncommafies the notes ID).
- It also refuses `\`, which xymond's new-testname charset admits and which separates directories on an SMB/CIFS `$XYMONVAR`.
- `update_htmlfile()`'s `timestr` was read uninitialised for a negative "time since change" (changetime past logtime) and leaked stack bytes into the served page; terminated now.

Commit structure is now two: the handle/notes guard, then the writer hardening with these folded in — so this is a force-push over the `f3f0c1df` you last reviewed.

SoundGoof (approved):
lgtm

PR #222, opened 2026-07-05
etcfiles: optional .d include dirs for every server and client config (TBT 5/6)

bonomani:
## What

Generalizes the existing `tasks.d` drop-in directory to **all** the config files. Each shipped config now ends with:

```
optional directory @XYMONHOME@/etc/<name>.d
```

covering `alerts.d`, `analysis.d`, `combo.d`, `graphs.d`, `client-local.d`, `rrddefinitions.d`, `hosts.d`, `xymonserver.cfg.d`, and client-side `xymonclient.cfg.d` / `clientlaunch.d`.

Packagers and admins can drop config fragments into these directories without editing the shipped file — the same pattern `tasks.d` already provides, now everywhere.

## Mechanism (all pre-existing upstream)

- `directory` reads every file in the dir (as `tasks.cfg` does today); `optional` makes a missing dir silently ignored — both already implemented in `lib/stackio.c`.
- Five configs that shipped without a template step (`analysis`, `combo`, `graphs`, `client-local`, `rrddefinitions`) gain one so `@XYMONHOME@` is substituted, and are renamed to `.DIST` to match every other generated config.
- The install rule creates the new `.d` directories next to `tasks.d`.

## Origin / credits

Terabithia patches `5 xymon.breakout-directories` (server cfgs) and `6 xymon.envdirs` (env files), J.C. Cleaver — also on devel as `bff5b5ef5`, which uses the same `optional directory` form (adopted here). Closes the `.d defaults` item in #28; refs #29 (TBT 5/6), #106.

## Testing

- Full server build + `tests/testsuite`: 14/14.
- Generated configs carry the substituted absolute paths (verified for all 10).
- End-to-end: a host in `hosts.d/extra.cfg` is picked up by the parser; a missing `.d` directory produces no error.

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -27,14 +27,14 @@ COMMONTOOLS=xymon xymoncmd xymongrep xymoncfg xymondigest
 all: $(PROGRAMS) $(COMMONTOOLS) xymonclient.cfg clientlaunch.cfg $(EXTRATOOLS)
 
 xymonclient.cfg: xymonclient.cfg.DIST
-	cat xymonclient.cfg.DIST | sed -e 's!@XYMONHOSTIP@!$(XYMONHOSTIP)!g' >xymonclient.cfg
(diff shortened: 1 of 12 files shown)


feld (approved):
Really glad to see this consistency here

PR #223, opened 2026-07-06
stackio: accept 'source FILE' and '. FILE' as aliases for 'include' (TBT 59)

bonomani:
Shell-syntax config files (`xymonserver.cfg`, `xymonclient.cfg`) are read both by xymon and by shell scripts that source them. Shell's own include directives are `source FILE` and `. FILE`; accepting them as aliases for xymon's `include` lets one such file pull in another and have *both* loaders follow it.

| line | before | after |
|---|---|---|
| `source FILE` / `. FILE` | ignored by xymon | followed, like `include` |
| `sourcehost ...` at line start | — | not an include (separator required) |
| `netincludes ...` | **followed as an include** (bare prefix compare) | not an include |
| `directory` | unchanged | unchanged — no shell equivalent |

The `netinclude`/`dispinclude` match was the pre-existing half: it compared a bare prefix, so `netincludes` was followed. It now requires the same trailing space or tab as every other keyword.

**Docs:** `hosts.cfg.5`, where the include directives already live, plus the point worth stating plainly — the aliases work in *every* file read through the stacked parser, not only shell-syntax ones. `xymonserver.cfg.5` carries the use case and the warning that the `optional` prefix is Xymon-only and would break a shell sourcing the same file.

**Test:** the xymongrep contract test, through the real `load_hostnames()`/`stackfgets()` path — `source FILE`, `. FILE`, `source<TAB>FILE`, `optional source` of a missing file staying silent, and `sourcehost ...` asserted on stderr staying free of `Cannot open include file`, which is what the original bare-prefix patch would have logged.

From Terabithia patch 59 (J.C. Cleaver, `xymon.shellinclude`), with the match tightened. TBT 59 is an RPM patch with no upstream commit, so there is no original to place first. Tracked in #29.

### DIFF
--- a/common/hosts.cfg.5
+++ b/common/hosts.cfg.5
@@ -52,8 +52,18 @@ filename (if it begins with a '/'), or a relative filename -
 relative file names are prefixed with the directory where
 the main hosts.cfg file is located (usually $XYMONHOME/etc/).
 
-You can nest include tags, i.e. a file that is included 
+You can nest include tags, i.e. a file that is included
 from the main hosts.cfg file can itself include other files.
+
+The directives "source filename" and ". filename" (dot, space,
+filename) are accepted as aliases for "include". They exist so a
+configuration file written in shell syntax \- e.g.
+.I xymonserver.cfg(5)
+\- can use the shell's own include directives and be followed both
+by a shell sourcing the file and by the Xymon config parser. Note
+that the aliases are recognised in every Xymon configuration file
+read through this parser, not only shell\-syntax ones: a line whose
+first word is "source" or "." is always treated as an include.
 
 .IP "dispinclude filename"
 Acts like the "include" tag, but only for the xymongen tool.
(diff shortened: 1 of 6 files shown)


feld:
Where in the docs/manpages can we put this? I think the behavior should be clearly outlined somewhere

bonomani:
Documented in `239cba11c`: **hosts.cfg.5** (the manpage that owns `include`/`directory`/`optional`) describes the aliases, including the point worth stating plainly - they are recognised in every config file read through the stacked parser, not just shell-syntax ones. **xymonserver.cfg.5** gets the use-case paragraph (a shell-sourced env file following the same reference), with a note that the `optional` prefix is Xymon-only and would break a shell sourcing the same file.

Also added regression coverage (`21c5f9e6a`) to the xymongrep contract test: `source`/`.`/`source<TAB>` includes, `optional` composing with both aliases, and a line-start `sourcehost ...` not being misparsed as an include (asserted on stderr staying free of the parser's "Cannot open include file" warning - what the original bare-prefix patch would have logged).

SoundGoof (approved):
lgtm

PR #224, opened 2026-07-06
xymon: XYMON_TIMEOUT env sets the CLI request-timeout default (TBT 129, guarded)

bonomani:
## What

The `xymon` command-line tool has a compiled-in 15s request timeout (`XYMON_TIMEOUT` in sendmsg.h), overridable per call with `--timeout=N`. This lets an `XYMON_TIMEOUT` **environment variable** set that default, so it can be configured once (e.g. in `xymonserver.cfg`, sourced into the tool's environment) instead of adding `--timeout=` to every call. `--timeout=` still wins (parsed afterwards).

## Safety

Non-positive / non-numeric values are **ignored**, keeping the 15s default:

```c
char *envtimeout = getenv("XYMON_TIMEOUT");
if (envtimeout && (atoi(envtimeout) > 0)) timeout = atoi(envtimeout);
```

This matters: `atoi()` returns 0 for empty/garbage input, and a 0 timeout makes `select()` block **forever** (`sendmsg.c`: `(timeout ? &tmo : NULL)`). So a stray or mistyped `XYMON_TIMEOUT` must not silently disable the timeout and hang the tool.

## Credits

Terabithia patch `129 xymon.clienttimeout` (J.C. Cleaver), which used `atoi()` unguarded; corrected here to reject invalid values.

## Testing

- Full server build; `tests/testsuite` 14/14.
- Runtime (debug output of the timeout used): `XYMON_TIMEOUT=60` → 60; `XYMON_TIMEOUT=abc` → 15; `XYMON_TIMEOUT=` → 15.

Refs: #29 (TBT `129`).

### DIFF
--- a/common/xymon.1
+++ b/common/xymon.1
@@ -27,7 +27,9 @@ server.
 
 .IP "\fB\-\-timeout\fR=\fIN\fR"
 Specifies the timeout for connecting to the Xymon server, in
-seconds. The default is 5 seconds.
+seconds. The default is 15 seconds; an XYMON_TIMEOUT environment
+variable overrides that default (see ENVIRONMENT below), and this
+option overrides both.
 
 .IP "\fB\-\-response\fR"
 The xymon utility normally knows when to expect a response
@@ -489,6 +491,15 @@ file "statusmsg.txt". Instead of providing it on the command-line,
 pass it via stdin to the xymon command:
 
    $ cat statusmsg.txt | $XYMON $XYMSRV "@"
+
+.SH ENVIRONMENT
+.IP "\fBXYMON_TIMEOUT\fR"
+Sets the default request timeout, in seconds, overriding the compiled-in
+15 - so it can be set once (e.g. in
+.IR xymonserver.cfg (5))
+rather than passing \-\-timeout= on every call. A non-positive or
+non-numeric value is reported on stderr and ignored. A \-\-timeout=
+option on the command line takes precedence.
 
 .SH "SEE ALSO"
 combostatus(1), hosts.cfg(5), xymonserver.cfg(5), xymon(7)
--- a/common/xymon.c
+++ b/common/xymon.c
@@ -38,6 +38,15 @@ int main(int argc, char *argv[])
 	sendreturn_t *sres;
 	int wantresponse = 0, mergeinput = 0, usebackfeedqueue = 0;
 
+	/* An XYMON_TIMEOUT environment variable overrides the compiled default.
+	   Ignore empty/non-numeric/non-positive values (atoi() would yield 0,
+	   and a 0 timeout means select() blocks forever) - keep the default then.
+	   A --timeout= option on the command line still wins, being parsed below. */
+	{
+		char *envtimeout = getenv("XYMON_TIMEOUT");
+		if (envtimeout && (atoi(envtimeout) > 0)) timeout = atoi(envtimeout);
+	}
+
 	for (argi=1; (argi < argc); argi++) {
 		if (strcmp(argv[argi], "--debug") == 0) {
 			debug = 1;
--- a/common/xymon.c
+++ b/common/xymon.c
@@ -38,6 +38,30 @@ int main(int argc, char *argv[])
 	sendreturn_t *sres;
 	int wantresponse = 0, mergeinput = 0, usebackfeedqueue = 0;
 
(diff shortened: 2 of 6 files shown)


SoundGoof:
## Findings

###  Parse `XYMON_TIMEOUT` with checked conversion — `common/xymon.c:47`

`atoi()` accepts numeric prefixes and has undefined behavior outside the `int` range.

Runtime testing showed:

- `60junk` becomes `60`
- `1e2` becomes `1`
- `4294967297` becomes `1`
- `8589934652` becomes `60`

Malformed configuration can therefore silently select a short positive timeout instead of retaining the default.

Use the checked integer parsing pattern added in [PR #308](https://github.com/xymon-monitoring/xymon/pull/308), specifically `parse_minimum_free()` in `xymond/xymond_history.c`.

###  Document the new environment variable

Current `main` documents only `--timeout=N`.

Neither the `xymon` manual nor the shipped configuration documentation mentions `XYMON_TIMEOUT`.

Document its valid range, malformed-value behavior, compiled default, and that `--timeout=` takes precedence.

## Verification

Tested the exact PR head and the patch rebased onto current `main`.

- Exact-head suite: 13 passed, 1 skipped, 0 failed
- Rebased suite: 29 passed, 4 skipped, 0 failed
- Valid `XYMON_TIMEOUT=60` worked
- `--timeout=7` correctly overrode the environment
- Removing the new parsing block restored the default timeout of 15 for malformed cases

bonomani:
Both findings addressed; the branch is also restructured into two commits to keep attribution honest: 817cf256b is the original Terabithia patch applied verbatim (author Japheth Cleaver, from the 4.3.30 SRPM), and 6410d3ba7 is the correction on top.

**Checked conversion**: the env value now goes through the #308 idiom — `errno = 0` + `strtol`, reject on no-digits (`endp == value`), trailing junk (`*endp != '\0'`), `ERANGE`, non-positive, or > INT_MAX — with an `errprintf` on rejection. (`parse_minimum_free()` itself is static in xymond_history.c on the still-open #308, so the pattern is replicated rather than called.) One deliberate divergence from #308: out-of-range values are rejected keeping the default rather than clamped, per this PR's contract — clamping a timeout to INT_MAX seconds would be worse than ignoring it. Runtime-verified via the timeout echoed in the send-failure report: `60` → 60; `60junk`, `1e2`, `4294967297`, `8589934652`, `abc`, empty, `-5`, `0` → all keep 15 with a stderr warning; `--timeout=7` still beats `XYMON_TIMEOUT=9`. The unguarded commit reproduces your cases (`2junk` → 2, `abc` → 0).

**Documentation**: xymon.1 gains an ENVIRONMENT section (valid range, malformed-value behavior, compiled 15s default, `--timeout=` precedence) — and its `--timeout` entry's stale "default is 5 seconds" claim is corrected to 15 while there; xymonserver.cfg.5 gains an XYMON_TIMEOUT entry; xymonserver.cfg.DIST a commented example. Both manpages render clean under `groff -ww`.

Suite on the branch: 14 passed, 0 failed; CI green.

common/xymon.1
```diff
@@ -534,6 +536,17 @@ pass it via stdin to the xymon command:
...
    $ cat statusmsg.txt | $XYMON $XYMSRV "@"
 
+.SH ENVIRONMENT
+.IP XYMON_TIMEOUT
+Overrides the compiled-in default request timeout (15 seconds), so the
+timeout can be configured once - e.g. in
+.I xymonserver.cfg(5)
+- instead of adding \-\-timeout= to every call. The value must be a
+positive integer number of seconds; anything else (empty, garbage,
+trailing junk like "60junk", zero, negative, or out-of-range values) is
```
SoundGoof: I think this can be more concise

bonomani:
@SoundGoof tightened the xymon(1) ENVIRONMENT entry per your note - dropped the invalid-value enumeration to "A non-positive or non-numeric value is reported on stderr and ignored." Also rebased onto current main (resolving the xymonserver.cfg.5 conflict with #303's bolding, and fonting the two new labels to match) and regenerated the manual-page HTML for the new gate.

SoundGoof (approved):
## review

No findings. The previous parsing and documentation concerns are resolved.

## Verification

- Exact head `bd2fc95e` built successfully
- Test suite: 44 passed, 0 failed
- Valid, malformed, overflow, and override cases behaved correctly
- Manpages rendered successfully
- `git diff --check` passed

Approved.

PR #225, opened 2026-07-06
xymond_rrd: env-tunable buffering for the external RRD processor (devel f1682a661 = TBT 513, corrected)

bonomani:
Ports the external-RRD-processor buffering feature (devel `f1682a661` = #29 TBT `513`)
to `main`, with the bugs it ships with on devel/4.4 and in the Terabithia RPMs fixed.

| | |
| --- | --- |
| `d3c5d6e3` | the original, unmodified — Japheth "J.C." Cleaver, 2015-11-22, with `cherry picked from` provenance |
| `534a7b9b` | everything we changed — fixes, manpage, tests |

## Feature

`xymond_rrd` `fflush`es the `--processor` stream after every line — a write syscall per
RRD update. Two env vars let you buffer instead: `RRD_EXTPROC_BUFSIZ` (stdio buffer size,
default the system `BUFSIZ`) and `RRD_EXTPROC_DOFLUSH` (flush every write, today's
behaviour).

## Fixes

1. **Teardown discarded the buffer.** `shutdown_extprocessor()` closed the raw fd and
   NULLed the `FILE*` without `fclose()` — leaking the FILE and, once the stream is
   buffered, dropping unwritten data on every restart and on exit. Now `fclose()`, with
   `close()` as fallback if `fdopen()` failed. This is what makes the feature safe to
   enable.
2. **The `setvbuf`-failure diagnostic was itself a crash** — `cmd` and `n` swapped against
   the format, so `%d` took a `char*` and `%s` an `int`. Reordered, size cast for `%lu`.
3. **`setvbuf()` ran on an unchecked `fdopen()` result.** Guarded.
4. **Port decision:** the original moved the "External processor started" line to
   `logprintf`, which on `main` is `#define logprintf(...) printf(__VA_ARGS__)` — a bare
   stdout write. It stays on `errprintf` with the rest of the file's diagnostics.

The original's `Changes` entry targeted devel's 4.4-alpha section; `main` has had no
unreleased section since 4.3.30, so there is nowhere for it to land.

## Behaviour change

The default flips from flush-per-line to buffered, matching `f1682a661`:

- **Clean shutdown loses nothing** — SIGTERM/SIGINT reach `shutdown_extprocessor()`, whose
  `fclose()` flushes. That is fix 1 working.
- **Processor death loses the buffer.** Measured on a pipe mimicking this code (SIGPIPE
  caught, 20 buffered updates, reader gone): `fclose` returns `-1`, versus `0` with nothing
  pending under flush-per-line. Previously at most one line was in flight; now up to
  `RRD_EXTPROC_BUFSIZ`.
- **Delivery is no longer prompt.** At ~50 bytes per update, ~160 accumulate before a
  flush — immaterial for RRD graphing, material if the processor feeds a live consumer.

`RRD_EXTPROC_DOFLUSH` restores per-line flushing. **Whether buffered should be the default
on `main` is worth deciding deliberately** rather than inheriting from devel.

`RRD_EXTPROC_BUFSIZ` is unvalidated — `atol()` gives `0` for `yes`, `64` for `64K`,
`SIZE_MAX` for `-1`, and glibc's `setvbuf` accepts all of them silently. Left alone here;
validating it is a behaviour change, not a port fix.

## Documentation and tests

`main` already carries most of devel's `4ec669eb`; the gap was the paragraph describing the
two variables. Written fresh, not ported — devel calls `RRD_EXTPROC_BUFSIZ` the *"pipe
buffer size (on systems which allow this)"*, and it is the stdio buffer set by `setvbuf` on
a stream this process owns. The new text adds what devel omits: that buffering delays
delivery, and what is lost when the processor dies. HTML regenerated, `mandoc -T lint`
clean.

Two tests in `tests/rrd/`, each proven to fail on the defect it guards:

| test | defect reintroduced |
| --- | --- |
| `extproc-teardown-flush.sh` | `FAIL: the external processor received nothing: the buffered update was discarded at teardown` |
| `extproc-doflush.sh` | `FAIL: RRD_EXTPROC_DOFLUSH did not deliver the update while xymond_rrd was still running` |

Both drive the real `xymond_rrd` with `--no-rrd` and a recording processor, forcing
`RRD_EXTPROC_BUFSIZ=65536` so nothing can arrive by the buffer filling. Neither asserts the
negative direction (that the default has *not* delivered yet) — that is the absence of an
event, and bounding it with a wall clock is a race.

## Verified

- Suite 55 passed / 0 skipped / 0 failed on a built server tree (53 before).
- SIGPIPE is caught in `xymond_rrd`, so the new `fclose()` into a broken pipe returns EOF
  rather than killing the worker; the reload path clears its flag *after*
  `setup_extprocessor()`, so a SIGPIPE from the teardown flush is swallowed, not looped.
- Rebased onto current `main` and rebuilt. The branch had been cut from `6f5f87861`, before
  this file gained its path-truncation hardening, so it had never compiled against the code
  it merges into.

Refs: #29 (TBT `513` = `f1682a661`)

### DIFF
--- a/docs/manpages/man8/xymond_rrd.8.html
+++ b/docs/manpages/man8/xymond_rrd.8.html
@@ -352,6 +352,25 @@ Section: Maintenance Commands (8)<br/>Updated: Version 4.3.30: 4 Sep 2019<br/><a
     nearly suitable for injection using <i>rrdupdate(1)</i> and easily
     transformable to other formats. If the process exits, xymond_rrd will
     re-launch it automatically.</p>
+<p class="Pp">The stream is fully buffered, so an update is normally written to
+    the processor only once the buffer fills. On a busy server that is
+    invisible; on a quiet one an update may wait a long time before it is
+    written. Set <b>RRD_EXTPROC_DOFLUSH</b> in <i><a href="../man5/xymonserver.cfg.5.html">xymonserver.cfg</a>(5)</i> to
+    flush after every update instead, at the cost of one write per update -- use
+    it when the processor forwards to something that expects a live stream.
+    <b>RRD_EXTPROC_BUFSIZ</b> sets the size of the buffer, in bytes; the default
+    is the system's BUFSIZ.</p>
+<p class="Pp">The buffer reaches the processor only if the processor is still
+    running when the stream is closed. That holds on an orderly stop: xymond_rrd
+    flushes the stream to the processor before exiting, so a normal shutdown
+    loses nothing.</p>
+<p class="Pp">If the processor exits first, everything still buffered is lost.
+    xymond_rrd notices only on its next write to the stream -- which, with the
+    stream buffered, is the write that fills the buffer -- and it is that late
+    in re-launching the processor as well. Up to <b>RRD_EXTPROC_BUFSIZ</b> bytes
+    of updates can go to a processor that is already gone; with
+    <b>RRD_EXTPROC_DOFLUSH</b> set the same failure costs at most a single
+    update.</p>
 <p class="Pp"></p>
 <p class="Pp"></p>
 </section>
--- a/tests/rrd/extproc-doflush.sh
+++ b/tests/rrd/extproc-doflush.sh
@@ -1,0 +1,98 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/rrd/extproc-doflush.sh
+#
+# RRD_EXTPROC_DOFLUSH must make --processor data arrive as it is produced,
+# not when the buffer fills or the processor is torn down.
+#
+# The default is a buffered stream, so on a quiet server an update can sit
+# unwritten for a long time -- fine for feeding an RRD store, not fine for a
+# processor forwarding to something that expects a live stream. DOFLUSH is
+# the documented escape hatch (xymond_rrd.8), and it is worth a guard: it is
+# one boolean in setup_extprocessor(), and losing it would be invisible
+# until an operator noticed their feed had gone quiet.
+#
+# The assertion is made while xymond_rrd is still running -- stdin is held
+# open through a fifo -- because at teardown the buffered stream is flushed
+# too (tests/rrd/extproc-teardown-flush.sh covers that path), so a run that
+# waits for exit cannot tell the two modes apart.
+#
+# Deliberately only the positive direction. Asserting that the *default*
+# has not delivered yet means asserting the absence of an event, which on a
+# loaded runner is a wall-clock race, and a flaky test is worse than none
+# (tests/README.md).
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+require_bin XYMOND_RRD xymond/xymond_rrd
+
+work=$(mktempdir)
+mkdir -p "$work/home/etc" "$work/tmp" "$work/rrd"
+: > "$work/home/etc/analysis.cfg"
+: > "$work/hosts.cfg"
+
+cat > "$work/processor" <<EOF
+#!/bin/sh
+while IFS= read -r line; do printf '%s\n' "\$line" >> "$work/received.txt"; done
+EOF
+chmod +x "$work/processor"
+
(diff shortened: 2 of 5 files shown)


SoundGoof:
## Review

The implementation appears correct, but it needs documentation and focused regression coverage before merge.

### Document the changed default

`--processor` now uses full buffering instead of flushing every update. Operators cannot discover the latency change, tune it, or restore the old behavior because these controls are absent from shipped documentation.

The wording already exists on `devel`:

- [`f1682a66`](https://github.com/xymon-monitoring/xymon/commit/f1682a6615e0b34dc36f04d8787c3100959f1e45) added buffering on 2015-11-22.
- [`4ec669eb`](https://github.com/xymon-monitoring/xymon/commit/4ec669eb2c46df8ae3bb83b93e2d22947e36dbdc) documented it on 2016-01-23.
- Neither commit is part of `main`.
- This PR ports the code from `f1682a66`, with fixes, but omits the later documentation.

Port the relevant wording from `4ec669eb` into `xymond/xymond_rrd.8`

### Add behavioral regression coverage

Add a focused test covering:

- Default buffering of short writes.
- Immediate delivery with `RRD_EXTPROC_DOFLUSH`.
- Flushing pending data during `shutdown_extprocessor()`.
- Safe recovery when the processor exits.
- Invalid `RRD_EXTPROC_BUFSIZ` values.

### Verification

- Clean merge against current `main`
- Server build succeeded
- Existing suite: 14 passed, 0 failed
- Editor diagnostics and `git diff --check`: clean
- GitHub CI: 4/4 checks green

bonomani:
Thanks — both points were right. Addressed in `534a7b9b`.

**Docs.** `main` already carries most of `4ec669eb`: the `--processor` block and the whole `SENDING METRIC DATA` section are there. The gap was one paragraph, the one describing the two variables. I wrote it fresh rather than porting, because devel's wording is wrong on the substance — it calls `RRD_EXTPROC_BUFSIZ` the *"pipe buffer size (on systems which allow this)"*, and it is the stdio buffer set by `setvbuf` on a stream this process owns; nothing touches the pipe's kernel buffer. Devel also omits what an operator most needs: that buffering delays delivery, and what is lost when the processor dies. Both are in the new text. HTML regenerated, `mandoc -T lint` clean.

**Tests.** Two in `tests/rrd/`, each proven to fail on the defect it guards, not just to pass:

| test | defect reintroduced |
| --- | --- |
| `extproc-teardown-flush.sh` | `FAIL: the external processor received nothing: the buffered update was discarded at teardown` |
| `extproc-doflush.sh` | `FAIL: RRD_EXTPROC_DOFLUSH did not deliver the update while xymond_rrd was still running` |

For the first I reverted `shutdown_extprocessor()` to its bare `close()`, rebuilt, and confirmed the processor receives nothing — that is the guard for the data loss. Both drive the real `xymond_rrd` with `--no-rrd` and a recording processor, forcing `RRD_EXTPROC_BUFSIZ=65536` so nothing arrives by the buffer filling. `extproc-doflush.sh` holds stdin open through a fifo and asserts while the worker still runs, since at teardown both modes deliver. Suite: 55 passed, 0 failed.

**Three of your five covered; the other two deliberately not.** Asserting the default has *not* delivered yet is asserting the absence of an event — a wall-clock race on a loaded runner, and a flaky test is worse than none. Invalid `RRD_EXTPROC_BUFSIZ` I measured instead of testing: `atol()` gives `0` for `yes`, `64` for `64K`, `SIZE_MAX` for `-1`, and glibc's `setvbuf` accepts all of them silently, so there is nothing to assert until validation exists — a follow-up, not something to smuggle into a port. Processor-exit recovery is deterministic and worth having; I left it out to keep this to one addition, happy to add it here if you'd rather it land together.

**Two corrections to your verification**, both from reviewing the previous branch: it merged cleanly but was **based on `6f5f87861`** and had never been compiled against current `main`, which has since gained path-truncation hardening in this file — it is now on `a5876a995`. And "14 passed" is an unbuilt tree, where every binary-driving test skips; built, it is 53 before this PR and 55 after.

The PR is also restructured since your review: `f1682a661` is its own commit with authorship preserved, everything of ours in one on top.

SoundGoof (approved):
lgtm

PR #230, opened 2026-07-09
Add HOSTDESCURL template to link the DESCR description on the info page

SoundGoof:
Adds one optional xymonserver.cfg template to turn the description part of a hosts.cfg `DESCR:Hosttype:Description` tag into a link on the host's info page.

**Problem**
`DESCR:linux:CIID1234` renders as plain-text rows on the info page, with no way to make the description a clickable link (CMDB entry, wiki page, inventory system) short of embedding raw HTML in every host line.

**Solution**
`HOSTDESCURL`, following the existing COLUMNDOCURL/HOSTDOCURL convention: a printf-style `%s` format string, unset/empty (the default) = no link, output byte-for-byte identical to today.

```
HOSTDESCURL="https://cmdb.example.com/ci/%s"
```

The substituted value is urlencode()d since DESCR text is free-form. Values already containing raw HTML (`<`) are left untouched, so the existing per-host raw-HTML passthrough — now documented in hosts.cfg.5 along with the first-colon split rule — keeps working and takes precedence per host.

**Changes**
- environ.c: register the HOSTDESCURL default (empty).
- svcstatus-info.c: wrap the description in a link when the template is set and the value carries no raw HTML.
- xymonserver.cfg.DIST, xymonserver.cfg.5, hosts.cfg.5: documentation.

**Testing** (by the author)
Clean build; testsuite 14/14; `groff -man -z` clean on both man pages; manual rendering of the info page with template set / unset / raw-HTML DESCR / DESCR without a description part.

**Review history**: the PR originally added a second template (`HOSTTYPEURL`) for the host-type part; it was dropped during review as having no concrete use case — the type is a category, not an identifier. See the review thread.

### DIFF
--- a/common/hosts.cfg.5
+++ b/common/hosts.cfg.5
@@ -290,6 +290,9 @@ text that will be shown on the "Info" column page; this can
 e.g. be used to store information about the physical location
 of the device, contact persons etc. If the text contain whitespace,
 you must enclose it in double-quotes, e.g.  DESCR:"switch:4th floor Marketing switch"
+The "Hosttype" and "Description" values can be turned into links on the
+info page by setting the HOSTTYPEURL and HOSTDESCURL formatting strings in
+.I xymonserver.cfg(5).
 
 .IP "CLASS:Classname"
 Force the host to belong to a specific class. Class-names are used
--- a/common/hosts.cfg.5
+++ b/common/hosts.cfg.5
@@ -290,9 +290,14 @@ text that will be shown on the "Info" column page; this can
 e.g. be used to store information about the physical location
 of the device, contact persons etc. If the text contain whitespace,
 you must enclose it in double-quotes, e.g.  DESCR:"switch:4th floor Marketing switch"
-The "Hosttype" and "Description" values can be turned into links on the
-info page by setting the HOSTTYPEURL and HOSTDESCURL formatting strings in
-.I xymonserver.cfg(5).
+"Hosttype" and "Description" are split at the \fBfirst\fR colon, so
+"Hosttype" must not itself contain a colon.
+HOSTTYPEURL and HOSTDESCURL in
+.I xymonserver.cfg(5)
+can turn these values into links on the info page. The info page renders
+this text unescaped, so raw HTML placed directly in "Description" (e.g.
+an \fB<a href=...>\fR link) is also rendered as a link, independently of
+HOSTDESCURL, and can be used to link different hosts to different targets.
 
 .IP "CLASS:Classname"
 Force the host to belong to a specific class. Class-names are used
--- a/common/hosts.cfg.5
+++ b/common/hosts.cfg.5
@@ -292,9 +292,9 @@ of the device, contact persons etc. If the text contain whitespace,
 you must enclose it in double-quotes, e.g.  DESCR:"switch:4th floor Marketing switch"
 "Hosttype" and "Description" are split at the \fBfirst\fR colon, so
 "Hosttype" must not itself contain a colon.
-HOSTTYPEURL and HOSTDESCURL in
+HOSTDESCURL in
 .I xymonserver.cfg(5)
-can turn these values into links on the info page. The info page renders
+can turn the description into a link on the info page. The info page renders
 this text unescaped, so raw HTML placed directly in "Description" (e.g.
 an \fB<a href=...>\fR link) is also rendered as a link, independently of
 HOSTDESCURL, and can be used to link different hosts to different targets.
--- a/common/xymonserver.cfg.5
+++ b/common/xymonserver.cfg.5
@@ -347,6 +347,24 @@ becomes a formatting string for the documentation URL. E.g. for the host
 Default: Not set, so host documentation will be retrieved from the
 XYMONNOTES directory.
 
+.IP HOSTTYPEURL
+Format string used to build a link for the host-type part of a
+.I hosts.cfg(5)
+"DESCR:Hosttype:Description" tag, shown on the host's info page. If set,
+this string becomes a formatting string for the link URL, e.g.
+HOSTTYPEURL="https://wiki.example.com/systems/%s" will turn a host-type of
+"linux" into a link to "https://wiki.example.com/systems/linux".
+Default: Not set, so the host-type is shown as plain text.
+
+.IP HOSTDESCURL
+Format string used to build a link for the description part of a
+.I hosts.cfg(5)
+"DESCR:Hosttype:Description" tag, shown on the host's info page. If set,
+this string becomes a formatting string for the link URL, e.g.
+HOSTDESCURL="https://cmdb.example.com/ci/%s" will turn a description of
+"CIID1234" into a link to "https://cmdb.example.com/ci/CIID1234".
+Default: Not set, so the description is shown as plain text.
+
 
 .SH SETTINGS FOR SENDING MESSAGES TO XYMON
 .IP XYMSRV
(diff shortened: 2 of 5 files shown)


bonomani (changes requested):
Thanks for this - it's a clean, well-documented contribution: opt-in, byte-identical output when unset, follows the COLUMNDOCURL/HOSTDOCURL convention, and it even urlencode()s the substituted value, which the precedent in lib/links.c doesn't. The code itself is fine.

First, some project context so the timing is transparent: right now the priority is triaging and absorbing the accumulated patches that exist across the distributions (Debian, Terabithia, ...) - features and fixes that real deployments already run. That alone is an enormous amount of work, and it comes before adding new features; and once it's done, the plan is to step back and rethink the overall shape of things so Xymon can grow on solid ground rather than by accretion. So a new-feature PR like this one may sit for a while, and the design bar for adding permanent config surface is deliberately high at the moment - none of which is a judgement of the contribution itself.

With that said, the design question this PR raises:

**1. Global-only may be the wrong shape for this need.** The templates are one URL pattern for the whole installation. Fleets pointing at more than one system (two CMDBs, a wiki for some hosts) get nothing - and a future per-host mechanism would be a hosts.cfg tag (e.g. `DESCRURL:...`) carried through XMH, i.e. a different design that would make these env vars dead weight. Worth asking now: is per-host the shape this feature actually wants?

**2. There is an existing (undocumented) per-host mechanism - which is both a conflict for this PR and a competing solution.** The info page emits the DESCR text unescaped, so raw HTML in DESCR renders today, and some sites use exactly that to get per-host links already. To be precise about what it can and cannot do: the parser splits type from description at the **first** colon in the value, so HTML works only in the **description** part (`DESCR:"linux:<a href='https://cmdb/ci/1234'>CIID1234</a>"` - colons after the first are kept). The **type** part cannot carry a normal URL at all: `https:` contains a colon, so the split lands inside the URL and garbles both rows; only colon-free forms (scheme-relative `//wiki/...` or a relative path) work there. Fragile and undocumented, but in use and fully per-host.

The interaction: with such a DESCR in place and `HOSTDESCURL` set globally, the host's HTML gets urlencoded into the template's %s - a garbled nested link - so a mixed fleet can't flip the global on. If the template approach stays, it needs either a guard (skip wrapping when the value contains `<`) or at minimum a docs warning to clean out raw-HTML DESCRs first.

**3. Small observation for context**: the type/description colon split is consumed only by svcstatus-info.c - every other DESCR consumer (tooltip, confreport, xymongen) treats it as one opaque string. Not an objection, just why the feature is info-page-only no matter what.

So there are really three shapes on the table: **(a)** this PR's global templates, plus the markup guard + docs warning; **(b)** a per-host `DESCRURL:` hosts.cfg tag - more plumbing, serves more setups, I can help with the XMH side; **(c)** no new code at all - document the existing HTML passthrough (with its first-colon caveat) as the supported per-host mechanism. Interested in your take on which shape fits the need you have.

SoundGoof:
I do not see this as competing with the existing per-host configuration, but rather as a complementary option.

We have a couple of thousand nodes and very long CMDB-generated host lines. Duplicating the same configuration on every host entry would significantly increase the size of `hosts.cfg` and introduce a large amount of unnecessary duplication.

Smaller installations, or installations requiring more per-host flexibility, can continue using the existing host-level configuration. Larger corporate environments may instead benefit from a uniform global default with less configuration in each host entry. Both approaches could therefore be used together, depending on the scale and requirements of the installation.

I also do not think it is generally a good idea to introduce yet another host field solely for this purpose, especially when the same result can be achieved in a more centralized and scalable manner.

I will implement a per-host option to disable this behavior, allowing host-specific HTML links to take precedence when needed.

I am also aware of where the current development effort is focused, so I am trying to keep this proposal relatively small and non-intrusive while still addressing the use case for larger installations.

bonomani:
Thanks for the follow-up commit — the raw-HTML guard and the hosts.cfg.5 documentation of the passthrough resolve the interaction concern from my review.

I also want to say this clearly: your reply shows you fully understand the project's current situation — the absorb-first phase, the high bar for new permanent config surface, and why you kept the proposal small and non-intrusive. That understanding is exactly why I think a small, contained feature like the one you propose should be implemented rather than parked.

Two questions before settling the final shape.

**1. Could the existing mechanisms fit your need?**

Reviewing the surrounding code, I found Xymon already has two host-documentation link mechanisms — one documented, one not:

- **`HOSTDOCURL`** (documented in xymonserver.cfg.5): global template, `%s` = the **hostname**. `HOSTDOCURL="https://cmdb.example.com/host/%s"` gives every host a CMDB link on its info page ("Notes:" row) today, one config line, no code. This fits if your CMDB can resolve a CI by hostname — including via a search endpoint.
- **`DOC:` hosts.cfg tag** (fully functional but absent from every man page): a **per-host** URL, rendered as a "Documentation:" row on the info page. It supports `%s`→hostname substitution (`lib/loadhosts.c`), inherits through the `.default.` host (fleet-wide default, per-host override), and as an XMH item it's also exported to alert scripts. For example:

  ```
  0.0.0.0    .default.   # DOC:https://cmdb.example.com/host/%s
  10.0.0.1   web1        # DESCR:linux:CIID1234 DOC:https://cmdb.example.com/ci/CIID1234
  10.0.0.2   sw-floor4   # (no DOC: tag - inherits the default, %s becomes "sw-floor4")
  ```

  `web1`'s info page links to its CI record; `sw-floor4` falls back to the fleet-wide hostname URL. Since your hosts.cfg is CMDB-generated, your generator could emit the per-host form directly.

If either covers you, this PR shrinks to documentation. My guess is they don't quite — your link key is the CI-ID, not the hostname, and `DOC:` repeats the URL base on every host line, which is the duplication you want to avoid — but I'd rather ask than assume.

**2. If they don't fit: could you drop `HOSTTYPEURL` and keep only `HOSTDESCURL`?**

Making the two fields symmetrical is a natural instinct (I suspect that's why both are in the PR), but they are different kinds of values. The description is an **identifier**: it resolves to a record, so a link is its natural, stable rendering. The type is a **category**, and categories in monitoring configs are typically inputs to *decisions* — routing, grouping, rule matching (what `CLASS` does in analysis.cfg) — rather than link targets. The practical value is also thin: a type like `linux` or `switch` would send every such host to the same generic page — very general information that rarely adds real value on a specific host's info page. And hard-wiring "the type renders as a link through one global template" freezes today's accidental shape of that field (a single free-form token) into permanent config surface: if the type field ever needs to do real category work — multiple types per host, or a more elaborate decision keyed on it — `HOSTTYPEURL` either constrains that evolution or becomes dead weight. And the cost asymmetry favors waiting: adding a type link later, if a real use case shows up, is a twenty-minute PR; removing a shipped variable is a breaking change. So my ask is: `HOSTDESCURL` exactly as you wrote it, guard included, and `HOSTTYPEURL` out — unless you have a real need for it that neither `DOC:` nor `HOSTDOCURL` can cover.

SoundGoof:
I can drop HOSTTYPEURL. I added it because it was convenient to implement alongside HOSTDESCURL, rather than because of a specific requirement.

The CIID link is the part I need

Thank you also for pointing out the undocumented DOC: functionality. I will open a separate PR to document it.

bonomani (approved):
Both review asks landed: the raw-HTML guard (with the passthrough now documented in hosts.cfg.5) and the removal of HOSTTYPEURL. Final shape is one opt-in template, byte-identical output when unset, urlencoded substitution, per-host raw HTML taking precedence. Verified the final diff: guard logic correct, urlencode() uses the library's growing static buffer (no leak), snprintf bounded, config-as-format matches the HOSTDOCURL/COLUMNDOCURL precedent. Thanks for the responsive iteration — merging squashed.

PR #237, opened 2026-07-10
xymond_rrd: refuse an overlong XYMONTMP instead of overflowing sun_path (#236)

### LINKED ISSUE #236: xymond_rrd: unchecked sprintf of $XYMONTMP into sun_path aborts startup when the path is long
## Symptom

`xymond_rrd` dies instantly at startup with:

```
*** buffer overflow detected ***: terminated
Aborted (core dumped)
```

No Xymon error message is logged, so there is no hint of the cause. Every trend stops updating.

## Cause

When building the cache-flush control socket, the address is composed with an unchecked `sprintf` into `sun_path`, which is only ~108 bytes on Linux (as small as 92-104 elsewhere):

https://github.com/xymon-monitoring/xymon/blob/main/xymond/xymond_rrd.c#L256

```c
memset(&ctlsockaddr, 0, sizeof(ctlsockaddr));
sprintf(ctlsockaddr.sun_path, "%s/rrdctl.%lu", xgetenv("XYMONTMP"), (unsigned long)getpid());
```

Any `XYMONTMP` longer than ~92 characters (108 minus `/rrdctl.` and a pid) overflows the buffer. With `_FORTIFY_SOURCE` (default on most distros) that is an immediate abort; without it, a silent stack smash of the neighbouring `sockaddr_un` fields.

Found while driving `xymond_rrd` from a test sandbox with a deep `XYMONTMP`; `strace` shows the abort fires right after the signal-handler setup, i.e. at this line. Deep paths are not exotic - containers, CI sandboxes and per-user tmpdirs easily exceed 92 characters.

## Repro

```sh
XYMONTMP=$(mktemp -d /tmp/$(head -c 100 /dev/zero | tr '\0' x)-XXXX)   # any path > ~92 chars
echo -n | XYMONTMP=$XYMONTMP xymond_rrd --rrddir=/tmp/rrd --no-cache
# => *** buffer overflow detected ***: terminated
```

## Notes

The peer side of this socket already handles the problem: `web/showgraph.c:154` checks the composed path against `sizeof(myaddr.sun_path)` and reports `rrdctl files located in XYMONTMP with too long pathname` instead of overflowing. The bind side in `xymond_rrd` needs the same treatment: `snprintf` plus a length check, and a clear `errprintf` naming `XYMONTMP` before exiting (or falling back to running without the control socket, since it only serves cache-flush requests).

bonomani:
## Problem

`xymond_rrd` composes its cache-control socket path (`$XYMONTMP/rrdctl.<pid>`) with an unchecked `sprintf` into `sun_path`, which is only ~108 bytes on Linux (as small as 92 on other platforms). A deep `XYMONTMP` — containers, CI sandboxes, per-user tmpdirs — overflows it: with `_FORTIFY_SOURCE` the daemon aborts at startup with no logged cause; without it, the overflow silently smashes the neighbouring sockaddr fields and the daemon keeps running (observed: failed `chmod()`, exit 0). Every trend stops updating either way.

Closes #236.

## Fix (single commit)

- Bound the composition with `snprintf`; on overflow, exit through the same path as the adjacent `socket()`/`bind()` failures, mirroring the sender-side guard in `web/showgraph.c`.
- The diagnostic reports the offending `XYMONTMP`'s length and the **actual** limit for it — `sun_path` minus the `/rrdctl.<pid>` suffix, which shifts with the pid's digit count (e.g. `237 characters; max 94 with this process ID`) — not the raw `sun_path` size, so an admin shortens to a length that actually fits.
- Documented the constraint (platform-dependent, ~92–107 bytes for the complete path) in `xymonserver.cfg.5` (XYMONTMP entry) and `xymond_rrd.8` (ENVIRONMENT).

## Regression coverage

`tests/server/xymond-rrd-ctlsocket-path.sh` drives the built worker with a 120-character `XYMONTMP` (asserts clean exit 1 naming XYMONTMP; the pre-fix binary dies on SIGABRT or corrupts silently) and with an ordinary one (clean start, EOF shutdown). Skips on the bare-tree lane, executes post-build. Full suite: 16 passed, 0 failed.

### DIFF
--- a/common/xymonserver.cfg.5
+++ b/common/xymonserver.cfg.5
@@ -181,6 +181,14 @@ Default: $XYMONSERVERROOT/server/ .
 
 .IP XYMONTMP
 Directory used for temporary files. Default: $XYMONHOME/tmp/
+.br
+Note:
+.I xymond_rrd(8)
+creates its cache-control Unix socket ("rrdctl.<pid>") in this
+directory, and Unix socket paths have a small, platform-dependent
+length limit (around 92 to 107 bytes for the complete path, depending
+on the OS). xymond_rrd refuses to start when XYMONTMP is too deep for
+the socket path to fit, so keep this directory path short.
 
 .IP XYMONWWWDIR
 Directory for Xymon webfiles. The $XYMONWEB URL must map to this directory.
--- a/tests/server/xymond-rrd-ctlsocket-path.sh
+++ b/tests/server/xymond-rrd-ctlsocket-path.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/server/xymond-rrd-ctlsocket-path.sh
+#
+# Regression guard for issue #236: xymond_rrd composed its cache-control
+# socket path with an unchecked sprintf into sun_path (~108 bytes), so any
+# XYMONTMP longer than ~92 characters overflowed the buffer and aborted the
+# daemon at startup ("*** buffer overflow detected ***") with no logged
+# cause. It must instead refuse to start with a clear error naming the
+# problem - and keep starting normally with an ordinary XYMONTMP.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+require_bin XYMOND_RRD "xymond/xymond_rrd"
+
+work=$(mktempdir)
+
+# A path safely past sun_path (108 bytes on Linux, as small as 92 elsewhere)
+longtmp="$work/$(printf 'x%.0s' $(seq 1 120))"
+mkdir -p "$longtmp" "$work/rrd" "$work/tmp"
+
+# Overlong XYMONTMP: a clean refusal (exit 1 + message), not a SIGABRT (134)
+rc=0
+out=$(echo -n | XYMONTMP="$longtmp" XYMONHOME="$work" \
+	"$XYMOND_RRD" --rrddir="$work/rrd" --no-cache 2>&1) || rc=$?
+[ "$rc" -eq 1 ] || fail "expected clean exit 1 on overlong XYMONTMP, got $rc: $out"
+assert_contains "XYMONTMP is too long" "$out" "overlong XYMONTMP refused with a clear error"
+
+# An ordinary XYMONTMP still starts and shuts down cleanly on EOF
+rc=0
+out=$(echo -n | XYMONTMP="$work/tmp" XYMONHOME="$work" \
+	"$XYMOND_RRD" --rrddir="$work/rrd" --no-cache 2>&1) || rc=$?
+[ "$rc" -eq 0 ] || fail "expected clean exit 0 with short XYMONTMP, got $rc: $out"
+
+pass "xymond_rrd refuses an overlong XYMONTMP instead of aborting"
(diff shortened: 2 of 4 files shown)


SoundGoof:
## Findings

### [P2] Do not report 107 as the maximum `XYMONTMP` length — `xymond/xymond_rrd.c:264`

The diagnostic reports:

> XYMONTMP is too long for a socket path (max 107 characters)

However, 107 is the maximum payload of the complete `sun_path`, not the maximum length of `XYMONTMP`. The generated `/rrdctl.<pid>` suffix also occupies that buffer.

On Fedora, a 92-character `XYMONTMP` succeeded while 93 characters failed and reported a maximum of 107. The precise limit also changes with the number of digits in the process ID.

This can mislead an administrator into shortening the path to a value that the daemon still rejects. Either omit the numeric maximum or report the actual limit after accounting for the generated suffix.

### Document the control-socket path constraint

Current `main` documents `XYMONTMP` only as the directory used for temporary files.

Neither `xymonserver.cfg.5` nor the `xymond_rrd` documentation explains that `xymond_rrd` creates an `rrdctl.<pid>` Unix socket there and cannot start when the complete path exceeds the platform’s `sun_path` capacity.

Add this constraint to the `XYMONTMP` configuration documentation or the `xymond_rrd` manual. Describe it as platform-dependent rather than documenting 107 as a universal `XYMONTMP` limit.

## Verification

Tested on Fedora using the exact PR head and both commits applied to current `main`.

- Exact-head suite: 16 passed, 0 skipped, 0 failed
- Rebased suite: 31 passed, 3 skipped, 0 failed
- Focused regression test passed on both builds
- Ordinary `XYMONTMP` startup and EOF shutdown succeeded
- Overlong `XYMONTMP` exited 1 with the new diagnostic
- Boundary test confirmed 92 characters succeed and 93 fail on this run
- Negative control against the original implementation failed as expected
- Shell syntax validation passed
- ShellCheck reported only the expected sourced-file informational warning

The locally compiled Fedora binary did not use the distribution RPM hardening flags, so the original implementation did not abort through `_FORTIFY_SOURCE`. It continued after overflowing the socket address, reported a failed `chmod()`, and exited 0.

This confirms that the regression test detects the non-fortified failure mode as well as the documented fortified abort.

bonomani:
Both findings addressed in 10ad193b6 (squashed to a single commit), CI green.

**Diagnostic**: rather than omitting the number, the error now reports the actual limit for XYMONTMP itself — sun_path minus the `/rrdctl.<pid>` suffix, computed per run so it tracks the pid's digit count — plus the offending value's own length: `XYMONTMP is too long for a socket path (237 characters; max 94 with this process ID)` (94 = 107 − 8 − 5 on that run, consistent with your 92-with-a-7-digit-pid boundary). No more guessing game for the admin.

**Documentation**: the XYMONTMP entry in xymonserver.cfg.5 and a new XYMONTMP entry in xymond_rrd.8's ENVIRONMENT section both describe the `rrdctl.<pid>` socket and the limit as platform-dependent (~92–107 bytes for the complete path), per your note not to enshrine 107 as universal.

The regression test needed no change — its assertion doesn't pin the number — and it passes against the new message; suite 16/0. Also folded your non-fortified observation (silent sockaddr corruption, failed chmod, exit 0) into the PR description, since it shows the pre-fix failure mode is worse than the fortified abort the issue described.

SoundGoof (approved):
## Review

No findings.

The socket-path bounds check, diagnostic, documentation,
and regression coverage look correct.

Builds using Xymon's legacy `snprintf` fallback retain the
previous overflow behavior. This only affects old systems
without native `snprintf`

## Verification

- Rebased cleanly onto current `main`; patch unchanged.
- Full build succeeded.
- 92-character path accepted; 93 rejected correctly.
- Full suite: 37 passed, 0 skipped, 0 failed.

Issue #234, opened 2026-07-10
Dead graph links remain


feld:
#206 has been merged and #44 (was draft) is now closed, but the dead graph link problem remains:

<img width="996" height="455" alt="Image" src="https://github.com/user-attachments/assets/80754bf6-6770-48f7-9625-690d287c05f9" />

My original method in #44 included RRD counting by trying to match filenames via FNPATTERN / EXFNPATTERN instead of guessing by counting lines from the status. It worked, but it might not be ideal.

bonomani:
Trying to pin down which divergence your page actually hits before picking the fix. From the screenshot the paging emitted 4 image slots but only 5 RRD files had data behind them - there are a few different ways to get there, and they need different fixes. Three quick facts about skeletor would settle it:

1. How many lines does the disk status show (the raw df section on the status page)?
2. `ls -l ~xymon/data/rrd/<hostname>/disk*` - how many files, and are their mtimes current?
3. Are `RRDDISKS` or `NORRDDISKS` set in your xymonserver.cfg?

Separately: #175 ports the df/inode filesystem-type filter (`XYMONCLIENT_FS_INCLUDE_TYPES` / `_EXCLUDE_TYPES` / `_DF_LOCAL_ONLY`) to the FreeBSD client, and its one open item is verification against real FreeBSD df output. If you could run that branch on skeletor it would both validate it on-platform and show how much of your df noise a type-level filter removes in practice - filtered types never reach the status at all, so lines and RRD files stay in sync for whatever passes.

Depending on your answers to 1-3, the remainder (if any) is what a server-side paging fix has to cover.

feld:
Remember, these are filesystems not physical disks. With ZFS I create and destroy them regularly. The path with the most churn is excluded

1. 111 lines including the df header
2. 160 rrd files, 99 current mtime
3. yes, `NORRDDISKS="^(.*poudriere.*)"` to filter out temporary filesystems that get created and destroyed regularly

feld:
I've tested #175 and it made no difference here. I suspect because I'd already filtered out all the filesystems that shouldn't get graphed. I did leave a note about two additional types that need to be excluded from inodes

PR #241, opened 2026-07-11
client: XYMONCLIENT_FS_EXCLUDE_MOUNTS drops mounts by glob (#170)

bonomani:
## What

A new client-side filter in the #170 family: `XYMONCLIENT_FS_EXCLUDE_MOUNTS` takes whitespace-separated shell glob patterns matched against the **mount point**; matching filesystems are dropped from the disk and inode reports entirely - no status line, no trending, no alerts. Default empty: output is byte-identical unless set.

```sh
XYMONCLIENT_FS_EXCLUDE_MOUNTS="/poudriere/* /var/lib/docker/*"
```

## Why

The existing type filters cannot express "exclude these ZFS datasets, keep those" - on a ZFS host every mount shares one type, so high-churn build mounts (poudriere, docker) cannot be filtered by type at all (surfaced while diagnosing #234). The server-side `NORRDDISKS` covers the *other* semantic - visible on the status page but untrended - and is documented as the contrast in every doc this PR touches.

## Implementation

- Linux and FreeBSD clients (the two with the fs-filter base on main); NetBSD/OpenBSD/Darwin adopt the variable when their #170 type-filter PRs land.
- Final stage of each report, after the split-line join (and the inode reformat on FreeBSD), so the mount point is always the last field; the header row is unconditionally kept.
- `case` glob matching under noglob (a `*` pattern never expands against the working directory), shell builtins only - the restricted-PATH regression scenario fails on any new external command, and did exactly that to an early `cat`-based draft.
- Documented in `xymonclient.cfg.DIST` and `xymonclient.cfg.5`.

## Testing

Per-OS regression cases in `fs-filter-linux.sh` / `fs-filter-freebsd.sh`: default-off, single and multiple patterns, both reports, non-matching rows kept, header survival (including under a bare `*`), noglob behavior. Full suite: 17 passed, 0 failed.

### DIFF
--- a/client/xymonclient-freebsd.sh
+++ b/client/xymonclient-freebsd.sh
@@ -57,6 +57,38 @@ case "$DFLOCALONLY" in
 	*) echo "xymonclient-freebsd: invalid XYMONCLIENT_FS_DF_LOCAL_ONLY '$DFLOCALONLY', using yes" >&2; DFLOCALONLY=yes ;;
 esac
 DFLOCAL=""; [ "$DFLOCALONLY" = yes ] && DFLOCAL="-l"
+# XYMONCLIENT_FS_EXCLUDE_MOUNTS: shell globs matched against the mount point
+# (last df column); matching filesystems are dropped from the disk and inode
+# reports entirely. Default: empty. (NORRDDISKS on the server keeps them
+# visible and only skips trending.)
+: "${XYMONCLIENT_FS_EXCLUDE_MOUNTS=}"
+# Header row always kept; shell builtins only; noglob as in fs_excl_opt.
+fs_filter_mounts() {
+	if [ -z "$XYMONCLIENT_FS_EXCLUDE_MOUNTS" ]; then
+		while IFS= read -r _fmline; do
+			printf '%s\n' "$_fmline"
+		done
+		return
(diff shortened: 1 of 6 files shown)


bonomani:
Closing: this duplicates a long-standing server-side feature. `DISK filesystem IGNORE` / `INODE filesystem IGNORE` in analysis.cfg (PCRE patterns supported, e.g. `DISK %^/poudriere/ IGNORE`) already provide the exact semantic this PR adds: the ignored filesystem is dropped from the disk status column entirely (xymond_client.c builds the status from the filtered df text), and since xymond_rrd's disk handler runs on the status channel, no RRD is ever created either - no status line, no trending, no alerts. That also covers the #234 churn motivation.

Local-mode clients are covered too: xymonclient.sh --local pipes the report through `xymond_client --local --config=localclient.cfg` on the client, same code path, same IGNORE support.

What remains unique here is only the reduction of raw [df] bytes in the client message - not worth new config surface. The one salvageable piece is documentation: the NORRDDISKS (visible, untrended) vs DISK IGNORE (excluded entirely) contrast is under-documented, which is why this PR exists at all. I'll do that as a separate doc patch.

PR #243, opened 2026-07-12
docs: clarify RRD graph settings in xymonserver.cfg (GRAPHS ::N split size, GRAPHS_<column> scope, legacy disk filters)

bonomani:
Documentation only, no behaviour change. The RRD graph settings in `xymonserver.cfg.5` and the matching comments in `xymonserver.cfg.DIST` were terse and easy to misread — especially the `GRAPHS` `::N` split size, which setting owns it, and the status of `NORRDDISKS`/`RRDDISKS`.

**Three roles, now stated up front:**

- `TEST2RRD` — maps a column to the RRD graph name used for storing data and for the column's default status-page graph.
- `GRAPHS_<COLUMNNAME>` — display only; it collects no data and replaces the column's default graph, so the original must be re-listed if still wanted.
- `GRAPHS` — graph list and order on the trends page, plus graph-display metadata such as the split size.

**`::N` split size (belongs in `GRAPHS`):**

- `name::N` caps the number of RRD **files** drawn per image once a graph is split (default 5). Files can hold several DS, each drawn as a curve, so an image may show more than N lines — a 2-DS graph at `::2` draws 4.
- It applies wherever links for that graph are generated: trends page, the column's `TEST2RRD` default graph, and `GRAPHS_<COLUMNNAME>` entries naming the same graph.
- Status pages split only when an instance count is available: a `<!-- linecount=N -->` marker in the status message, or automatic counting for `--multigraphs` columns (built-in default: disk, inode, qtree, quotas, snapshot, TblSpace, if_load).
- The legacy middle field (`name:x:N`) is documented as parsed-but-ignored; a host `TRENDS` tag with an explicit graph list (e.g. `disk:disk1|disk2`) ignores N and uses the default of 5.

**`NORRDDISKS`/`RRDDISKS`** are now documented as deprecated legacy settings, retained for old configurations: they affect RRD graph tracking only and do not remove already-tracked RRD files. The usual replacement is a `DISK`/`INODE ... IGNORE` rule in `analysis.cfg(5)` when a filesystem should be dropped from status, alerts and graphs.

Man page and DIST comments were updated together so they stay consistent; the man page renders cleanly with `groff`.

### DIFF
--- a/common/xymonserver.cfg.5
+++ b/common/xymonserver.cfg.5
@@ -540,19 +540,27 @@ The graph names must match graph definitions from
 or instance names handled by those definitions.
 
 .IP GRAPHS
-List of graph names that should be shown on the "trends" column,
-in display order. The matching RRD files are selected by the
-corresponding graph definition in
+List of graph names known to the graph display code. On the
+"trends" column, this list also defines which graphs are shown
+and in what order. The matching RRD files are selected by the
+corresponding graph definitions in
 .I graphs.cfg(5).
 Each entry is either "name" or "name::N": "name" is the graph name,
 and N is the maximum number of RRD files drawn per graph image after
 the graph has been split; the default is 5 per image. The double colon
 is literal, so "mrtg::1" draws one file per image.
+N counts RRD files, not lines: each RRD file may hold several data sets
+(DS), each drawn as its own curve, so an image showing N files can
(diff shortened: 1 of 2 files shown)


feld (approved):
Thank you, this should help clear up some confusion.

PR #245, opened 2026-07-12
htmllog: mirror NORRDDISKS/RRDDISKS when sizing disk graph paging (#234)

### LINKED ISSUE #234: Dead graph links remain

bonomani:
## Problem

A disk-family status page sizes its graph paging by counting status lines, but `NORRDDISKS`/`RRDDISKS` filter those filesystems at RRD-write time only - the counter never knew, so the page requested trailing graph slices with no files behind them: permanently broken images for every visible-but-untrended filesystem.

Field case (#234, verified numbers): 110 lines counted, ~11 held back by `NORRDDISKS="^(.*poudriere.*)"`, 99 RRD files - 2-4 dead slots on every page view. Closes #234 for the reported configuration.

## Fix

Apply the same filters in the counter - no filesystem access, pure in-memory string work: same patterns (compiled once, as in `do_disk.c`), same subject (the mount point - the line's last field, only when it starts with '/', so the df header and decorations are never touched), same precedence (exclude wins; with an include pattern set, only matching filesystems count). Gated to the do_disk column family (disk, inode, qtree, quotas, snapshot, TblSpace); other multigraph columns untouched.

Counter and writer now derive from one truth - the property that already makes `analysis.cfg` `DISK ... IGNORE` self-consistent. No new state, no reconciliation channel, byte-identical behavior when the variables are unset.

Caveat (in the code comment): assumes the rendering host shares the RRD host's `xymonserver.cfg` settings - true wherever the RRD directory is local.

## Testing

Harness drives the real `generate_html_log()` once per pattern scenario (patterns are process-cached): unfiltered (20 lines -> 4x5 slots), NORRDDISKS (15), RRDDISKS-only (15), both with exclude-wins (19 -> rebalanced 4-per-slot). Fails on unpatched code at exactly the filtered-count assertions. Full suite 19/19.

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -57,6 +57,9 @@ $(XYMONCLIENTCOMMLIB): $(XYMONCLIENTCOMMLIBOBJS)
 loadhosts.o: loadhosts.c loadhosts_file.c loadhosts_net.c
 	$(CC) $(CFLAGS) -c -o $@ loadhosts.c
 
+htmllog.o: htmllog.c
+	$(CC) $(CFLAGS) $(PCREINCDIR) -c -o $@ $<
+
 eventlog.o: eventlog.c
 	$(CC) $(CFLAGS) $(PCREINCDIR) -c -o $@ $<
 
--- a/lib/htmllog.c
+++ b/lib/htmllog.c
@@ -23,6 +23,9 @@ static char rcsid[] = "$Id$";
 #include <sys/types.h>
 #include <sys/stat.h>
 
+#define PCRE2_CODE_UNIT_WIDTH 8
+#include <pcre2.h>
+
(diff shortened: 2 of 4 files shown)


bonomani:
Closing by maintainer decision: the legacy NORRDDISKS/RRDDISKS machinery stays entirely untouched, paging behavior included. The generic NORRD/RRDONLY filter (#246) is being restructured to stand alone - paging-aware for its own entries only.

Issue #229, opened 2026-07-08
RFC: unify status damping — close the DURATION gaps, deprecate badconn/badTEST and delayred/delayyellow


bonomani:
# RFC: Unify status damping — deprecate `badconn`/`badTEST` and `delayred`/`delayyellow` in favor of alert-layer criteria

*(Revised after the discussion below and a source-verification pass; outcomes are folded into the sections they concern.)*

## Summary

Xymon ships **four coexisting mechanisms** for damping unstable or short-lived red/yellow statuses, in different layers, none documented against the others:

| Mechanism | Since | Layer | Semantics | Status |
|---|---|---|---|---|
| `badconn:x:y:z` / `badTEST:x:y:z` | Big Brother / fping.sh era | xymonnet (source) | count-based (consecutive probe failures), network tests only, with a `clear` phase | deprecated 2011, still fully functional |
| `delayred=` / `delayyellow=` | 2011 (`efd36f4d4`) | xymond (state of record) | time-based (consecutive minutes in color), all tests | active |
| `DURATION>` / `DURATION<` | 2004 (`ac4d163c3`) | xymond_alert (notification) | time-based (age of the alert event) | active |
| flap suppression (`--flap-count`/`--flap-seconds`, opt-out `noflap`) | 4.3.0 | xymond (state of record) | oscillation-based (more than N changes within M seconds pins the most severe color), all tests | active |

This RFC closes the alert-layer gap (color age), moves the display/SLA concern to read time, and completes the deprecation started in 2011. `DURATION` and flap suppression **stay** — the latter with a `lastchange` bookkeeping fix (see shared foundation); the two hosts.cfg tag families are replaced and deprecated.

## Problems with the status quo

1. **`delayred` mutates the state of record** to solve an alerting problem: the color is rewritten in `handle_status()` (`xymond/xymond.c:1676-1694`), so the damping leaks into history, events and SLA — the deferral window is recorded as **yellow**, one incident can page twice (yellow rule, then red rule), and suppressed blips vanish from the `WARNSTOPS` outage count, so instability-based SLAs undercount.
2. **The interim display is three-phase and surprised users for 14+ years.** Depending on `delayyellow` (unset / smaller than `delayred` / larger), a deferred red shows yellow immediately, previous → yellow → red, or previous → red. Undocumented until PR #228; recurring mailing-list question (2012, 2015, 2016).
[...]

PR #250, opened 2026-07-15
xymond: make lastchange the true hold-time of the recorded color (issue #229, step 2)

bonomani:
Step 2 of the rollout plan in #229, and a standalone fix.

`lastchange` was reset whenever a test stopped reporting: a host down for two hours that came back green showed as green for 0 minutes. `xymond_alert` derives `eventstart` from it, so an outage that dropped out of sight also restarted its DURATION clock. It now means the hold-time of the recorded color.

## What changed

- The flap detector's ring of change times is split out of `lastchange`, which was doing both jobs.
- A gap opens whenever the recorded color is xymond's rather than the reporter's — invented because no data arrived (purple, or clear for a downed conn test or a `dialup` host), rewritten by `get_hts()` for DOWNTIME, or rewritten in `handle_status()` by a disable, a change delay, the flap pin or a modifier — and closes on the first report whose color is the reporter's own. That bookkeeping sits outside the status-change path, which a report that changes nothing never reaches.
- Bridging is bounded to 4× the test's report validity, two hours at the default, so a host absent for a month cannot return claiming a month-long hold-time. A gap past that window is not checkpointed: nothing can bridge it, and while an override is in force nothing closes it.
- Purple stays out of the flap color pin on both sides. Recorded, it outranks green/clear/blue and would rewrite every report from a resumed test back to no-data until the flap window expired.
- The state rides in a new `@@XYMONDCHK-V1|.flapstate.` record. The status record is positional and its reader errors on unknown fields, so appending would cost an older xymond the whole board on a rollback; it stays byte-identical.

## Behaviour changes

For a release note, against current `main`.

1. `lastchange` is the hold-time of the recorded color, and **can move backwards** — `xymondboard ... lastchange>N` misses a bridged recovery. Both `xymon.1` definitions say so.
2. `alerts.cfg` DURATION spans a bridged gap, so rules that never fired because their target kept vanishing can now fire.
3. A flapping test that stops reporting is shown purple, not rewritten back to its last real color.
4. Flap suppression survives a restart: the ring is restored and the flag re-derived from it, so a stale checkpoint cannot resurrect it.
5. `xymond.chk` gains `.flapstate.` records. An older xymond skips them and loads the board normally, losing only the flap and hold-time state.

Unchanged: `flapinfo`'s two timestamps, the number of color changes that triggers flap detection, and the three consumers computing `now - lastchange` as an age.

## Tests

Each assertion verified by reintroducing the defect and watching the test name it. `xymond-holdtime-gap.sh` drives the gap paths on a real xymond, from a restored checkpoint and on a running daemon; `xymond-checkpoint-roundtrip.sh` save and restore, including a gap a minute old (kept) against one a day old (dropped); `xymond-holdtime-bridge.sh` both halves of the bridge rule; `xymond-flap-purple-pin.sh` the pin over every color pair.

Suite 57 passed / 0 skipped / 0 failed. Nothing exercised `lastchange` or xymond's checkpointing before this, which is how both reached review unnoticed.

Refs #229.

### DIFF
--- a/common/xymon.1
+++ b/common/xymon.1
@@ -206,7 +206,8 @@ Status color (green, yellow, red, blue, clear, purple)
 .IP "\fBtestflags\fR"
 For network tests, the flags indicating details about the test (used by xymongen).
 .IP "\fBlastchange\fR"
-Unix timestamp when the status color last changed.
+Unix timestamp when the current status color began, i.e. the hold-time of
+the color shown. See "Hold-time and no-data gaps" below.
 .IP "\fBlogtime\fR"
 Unix timestamp when the log message was received.
 .IP "\fBvalidtime\fR"
@@ -292,6 +293,21 @@ and with the following inequalities:  >= > <= < = !=
 
 These filters are: lastchange, logtime, validtime, acktime, disabletime
 
+Hold-time and no-data gaps
+
+\fBlastchange\fR is the hold-time of the color shown: when that color
+began, not when a message last moved it. If a test stops reporting and
+later resumes with the same color, the silence counts as hold-time and the
+value is unchanged across it - the status did not move while it was out of
+sight. A gap longer than four times the report validity is not bridged, and
+a resume with a different color starts a new hold-time. The gap is measured
+from when the status went stale, itself one report validity after the last
+report.
+
+This value can therefore move backwards, which matters for the
+\fBlastchange\fR filter above: a poller asking for changes since its last
+look will not see a bridged recovery.
+
 The response is one line for each status that matches the CRITERIA,
 or all statuses if no criteria is specified. The line is composed of
 a number of fields, separated by a pipe-sign. You can select which
@@ -307,7 +323,8 @@ Status color (green, yellow, red, blue, clear, purple)
 .IP "\fBflags\fR"
 For network tests, the flags indicating details about the test (used by xymongen).
 .IP "\fBlastchange\fR"
-Unix timestamp when the status color last changed.
+Unix timestamp when the current status color began, i.e. the hold-time of
+the color shown. See "Hold-time and no-data gaps" below.
 .IP "\fBlogtime\fR"
 Unix timestamp when the log message was received.
 .IP "\fBvalidtime\fR"
--- a/docs/manpages/man1/xymon.1.html
+++ b/docs/manpages/man1/xymon.1.html
@@ -252,7 +252,8 @@ Section: User Commands (1)<br/>Updated: Version 4.3.30: 4 Sep 2019<br/><a href="
   <dd>For network tests, the flags indicating details about the test (used by
       xymongen).</dd>
   <dt id="lastchange"><a class="permalink" href="#lastchange"><b>lastchange</b></a></dt>
-  <dd>Unix timestamp when the status color last changed.</dd>
+  <dd>Unix timestamp when the current status color began, i.e. the hold-time of
+      the color shown. See &quot;Hold-time and no-data gaps&quot; below.</dd>
   <dt id="logtime"><a class="permalink" href="#logtime"><b>logtime</b></a></dt>
   <dd>Unix timestamp when the log message was received.</dd>
   <dt id="validtime"><a class="permalink" href="#validtime"><b>validtime</b></a></dt>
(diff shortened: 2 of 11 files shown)


SoundGoof:
this needs a rebase, and also be added to a milestone

bonomani:
Rebased onto current `main` (`99eeb476a`), still a single commit, no conflicts.

On the milestone: I would rather not pin one. It is step 2 of #229 but stands on its own, so put it wherever it fits your release planning.

SoundGoof:
# Review

## High: A returning report retroactively changes the gap limit

[xymond/xymond.c:1750](https://github.com/xymon-monitoring/xymon/blob/99eeb476a1ebc2a71b61739346052a108b773fbb/xymond/xymond.c#L1750)
replaces `log->validity` with the returning report's validity before
[xymond/xymond.c:2007](https://github.com/xymon-monitoring/xymon/blob/99eeb476a1ebc2a71b61739346052a108b773fbb/xymond/xymond.c#L2007)
uses it to evaluate an existing gap.

For example, a `status+5` test that disappears for one hour and returns
with `status+240` bridges the gap using a sixteen-hour limit, even though
the test was expected every five minutes when the gap opened. Reversing
the validities incorrectly rejects the same one-hour gap.

A live-daemon probe reproduced both outcomes. Preserve the validity that
applied when the gap opened and use that value when closing it.

## Medium: A non-default gap can be lost across two restarts

The `.flapstate.` checkpoint record at
[xymond/xymond.c:5041](https://github.com/xymon-monitoring/xymon/blob/99eeb476a1ebc2a71b61739346052a108b773fbb/xymond/xymond.c#L5041)
stores the gap timestamps and color, but not its validity. Restoration
assigns `defaultvalidity` at
[xymond/xymond.c:5400](https://github.com/xymon-monitoring/xymon/blob/99eeb476a1ebc2a71b61739346052a108b773fbb/xymond/xymond.c#L5400).

Consequently, the next checkpoint filters the restored gap using the
default two-hour window at
[xymond/xymond.c:5031](https://github.com/xymon-monitoring/xymon/blob/99eeb476a1ebc2a71b61739346052a108b773fbb/xymond/xymond.c#L5031).

A probe restored a three-hour-old `status+240` gap and stopped the daemon
before another report. The newly written checkpoint dropped the gap,
although its original bridge window was sixteen hours. A second restart
therefore cannot recover the hold-time.

Checkpoint the validity associated with an open gap. Add a two-restart
case using non-default validity.

## Medium: Both extracted-function tests omit configured include flags

[tests/server/xymond-holdtime-bridge.sh:61](https://github.com/xymon-monitoring/xymon/blob/99eeb476a1ebc2a71b61739346052a108b773fbb/tests/server/xymond-holdtime-bridge.sh#L61)
and
[tests/server/xymond-flap-purple-pin.sh:50](https://github.com/xymon-monitoring/xymon/blob/99eeb476a1ebc2a71b61739346052a108b773fbb/tests/server/xymond-flap-purple-pin.sh#L50)
compile `libxymon.h` without the configured compiler/include flags.

Both tests fail with `pcre2.h: file not found` on FreeBSD 14.3,
NetBSD 11.0, and OpenBSD 7.9, where PCRE2 is under `/usr/local/include`
or `/usr/pkg/include`. The configured Makefiles already contain those
paths.

Use the configured compilation flags. Supplying the omitted paths through
`CPATH` allowed all four focused tests to pass on each BSD.

## Low: Checkpoint restoration ignores NOFLAP

[xymond/xymond.c:5277](https://github.com/xymon-monitoring/xymon/blob/99eeb476a1ebc2a71b61739346052a108b773fbb/xymond/xymond.c#L5277)
restores `log->flapping` from the saved flag and ring age without applying
`isset_noflap()`.

When `NOFLAP` is added before restarting from a flapping checkpoint,
`xymondboard flapinfo` still reports `1/...` until another status arrives.
A checkpoint written before that report preserves the stale flag again.

Apply the loaded host's `NOFLAP` configuration during restoration and add
a restart test that inspects `flapinfo` before sending another status.

## Validation

Linux: 54 passed, 3 expected skips, 0 failed.  
FreeBSD 14.3, NetBSD 11.0, OpenBSD 7.9: all focused behavioral tests pass
once the missing configured include paths are supplied.

bonomani:
All four addressed in code. Test coverage is uneven, so I am splitting the answer accordingly.

**High — the gap window.** The validity in force when the silence began is now recorded with the rest of the gap state and used to close it. The gap bookkeeping is extracted into `update_gapstate()` so a test can drive it; reaching it through a live daemon means waiting out a report validity. Both directions of your example are covered, and putting the window back on the returning report reproduces exactly the two outcomes you described:

```
FAIL a 1h gap opened at status+5 is not bridged by a returning status+240: want 0, got 1
FAIL a 1h gap opened at status+240 is bridged although the return says status+5: want 1, got 0
```

**Medium — the checkpointed validity.** Fixed. It rides as a suffix on the `gapstart` field rather than as a field of its own: the record is parsed by position, so a new field would shift the flap-change list and misread every checkpoint written before this change. `strtol()` stops at the `:`, so an older file still restores — with the default it had then, which is the behaviour it already had.

**Low — NOFLAP on restoration.** Fixed; `isset_noflap()` is applied when the flag is re-derived from the restored ring.

**Medium — configured include flags.** Both tests use `xymon_cflags` now instead of a hand-rolled `-I` list.

**What I did not do**, rather than let it be assumed: the two-restart case and the `flapinfo`-before-next-status case you asked for are *not* written. Those two fixes are code-only for now. Both need a live daemon stopped and restarted twice, which none of the existing harnesses do — the round-trip test drives one restart. I would rather say so than let the green suite imply they are guarded.

SoundGoof (approved):
No remaining findings.

All four previous findings are addressed:

- Gap bridging uses the validity active when the gap opened.
- Non-default validity survives repeated checkpoint restarts.
- Extracted harnesses use configured compiler and include flags.
- Restored flap state respects `NOFLAP`.

## Validation


All five focused tests passed on Linux.

Independent runtime probes also passed:

- A three-hour-old `status+240` gap survived two checkpoint cycles
  with its `:240` validity intact.
- A flapping checkpoint restored under `noflap` reported
  `flapinfo=0/...` before receiving another status.

The corrected extracted-function tests passed on:

- FreeBSD 14.3
- NetBSD 11.0
- OpenBSD 7.9

This includes `xymond-holdtime-bridge`,
`xymond-flap-purple-pin`, and `xymond-gap-validity` on each system.

All five upstream CI checks are green.

Issue #254, opened 2026-07-15
Parsing of analysis.cfg for DS rules compiles unrelated regex rules


feld:
When we get RRD data from a client there's a separate parsing of `analysis.cfg` to find the DS rules that match. I discovered this when I found that the RRD rules were not obeying "first match" behavior in #32 

However, the issue I encountered in #253 shows that this unique pass of `analysis.cfg` for RRD DS alert rules is causing it to attempt to compile any regex found in the file even if it's not a DS rule which is how it ended up in the rrd log files. This seems wasteful and undesirable.

bonomani:
Dug into this. The mechanism confirmed: there is no separate DS-only pass — `xymond_rrd` calls the full `load_client_config()` (xymond_rrd.c:330) because `do_rrd.c` needs `check_rrdds_thresholds()`, and `setup_expr()` eagerly compiles every `%`-pattern at parse time. Since three processes load the file (xymond_client + the rrd-data and rrd-status instances), a broken pattern prints in exactly the three logs you saw in #253.

Two findings that reframe the severity:

1. **The CPU waste is negligible** — `load_client_config()` has a `stackfmodified()` guard, so patterns recompile only when analysis.cfg actually changes, not on every 600s reload tick. (The duplication is also symmetric: xymond_client loads DS rules it never evaluates.)
2. **The actual pain is misattribution**: the pcre error carried no file/line, and landing in the rrd logs it reads like an RRD problem. #255 fixes that — the error now reports the config line, plus a hint when the pattern looks whitespace-truncated (the real root cause in #253: the tokenizer splits on spaces, so `%(Synchronization failed|...)` was cut at the space before ever reaching pcre — that's worth its own issue as a config-grammar limitation).

On skipping non-DS rules in the rrd path: doable (a rule-type mask on the loader), but it's per-consumer special-casing that the unified rule engine in #218 would delete again — one namespace, one engine, loaded once. Given the mtime guard already removes the recurring cost, I'd fold this into #218 rather than build a DS-only parse path.

Issue #253, opened 2026-07-15
pcre compile '(Synchronization' failed (offset 16): missing closing parenthesis


feld:
I'm seeing these in the `clientdata.log`, `rrd-data.log`, and `rrd-status.log` files. I do not have any idea of the root cause but they first started showing up on 2026-07-09

feld:
I just remembered I added an alert for a log line and "Synchronization failed" was a specific part of the regex.

This error message is not about "Synchronization failed", it's telling me that `%(Synchronization failed|... other strings here)` failed to compile. It's telling me that the issue was found at:

`(Synchronization`

It just so happens "failed" is the next word in the string I was trying to match on. So if you see this error about pcre compile errors, the key is within the single quotes.

The fix on my end is to actually change it to `%(Synchronization[[:space:]]failed|...)`

The fact that this error is showing up in those 3 log files is interesting. It should only be showing up in `clientdata.log` as this has nothing to do with RRDs. Perhaps there's something reading/parsing the analysis.cfg for RRD purposes (custom DS alert rules? I know `analysis.cfg` is handled separately there) and this forces it to try to compile this regex expression. But it should just be skipping everything in the `analysis.cfg` that isn't a DS rule. This seems like unnecessary duplication and we could optimize this.

Issue #218, opened 2026-07-03
RFC: unify threshold evaluation into one metric/rule engine


bonomani:
## Motivation

Thresholding is spread across mechanisms that look alike from the user's seat but run in different places on different data:

| Mechanism | Runs in | Input | Effect |
|---|---|---|---|
| `analysis.cfg` DISK/PROC/MEM/LOAD/... | xymond_client | parsed client-report text | colors the column natively |
| `analysis.cfg` DS | xymond_rrd | the value being written to an RRD | emits a `modify` overriding the column afterwards |
| `alerts.cfg` | xymond_alert | column color changes | routing only |

This is accretion, not design: each threshold hook was added at whatever pipeline stage the data happened to be reachable. Symptoms: DS ignored the documented first-match contract for years (#32, fixed by #217); DS-based checks race the natively-computed status via `modify`; text formatting (`&N/&V/&L/&U`) exists only in the DS engine; suppression idioms differ per rule type (`DISK ... IGNORE` vs nothing for DS). Any metric visible at two stages (disk: report text *and* RRD) can be thresholded two subtly different ways.

The rule **vocabulary** is already unified — every rule type shares `HOST=`/`EXHOST=`/`CLASS=`/`PAGE=`/`TIME=` filters and (after #217) first-match semantics. What is fragmented is metric addressing and the evaluation machinery.

## Proposed model

Everything is a named metric, judged by one engine:

1. **One metric namespace.** Every number the server learns gets a name (`disk./usr.pct`, `conn.sec`, `smart-temp.sda`). This half-exists: the RRD layer already sees all numeric data as `(host, rrdfile, dataset)`, and the native analyzers compute a number before coloring.
2. **One rule form.** `THRESHOLD <metric-pattern> <condition> COLOR=<c> TEXT=<t>` + the existing shared filters. `DISK /usr 90 95` compiles to two THRESHOLD rules on `disk./usr.pct`; `DS` compiles to a THRESHOLD on an RRD-sourced metric. **Existing config files keep parsing forever** — they become frontend syntax over one engine.
3. **One evaluation point per metric — where it is born.** Report-parsed metrics are judged in xymond_client (native color, as today); graph-only metrics at RRD-write. Both call the same `evaluate(metric, value)` engine, so semantics (first-match per severity, text formatting, suppression) are identical everywhere, and the DS modify-race disappears for anything with a native path.
[...]

PR #255, opened 2026-07-15
analysis.cfg: report the line of an invalid pattern, hint at whitespace truncation

bonomani:
Follow-up to #253 (resolved by hand there, but the errors were un-diagnosable for days): make pattern-compile failures self-locating. The whitespace-truncation grammar limitation behind it is tracked in #256; the DS-only-parsing question in #254 stays open (recommended home: #218).

When a `%pattern` in analysis.cfg fails to compile, the pcre error printed only the (possibly truncated) pattern — no file, no line. And because every process that loads analysis.cfg compiles all patterns, the same error lands in `clientdata.log`, `rrd-data.log` and `rrd-status.log`, which made #253 look like an RRD problem.

This PR:

- **`lib/matching.c`** — adds `compileregex_ext()`, which is `compileregex_opts()` plus optional out-parameters for the pcre2 error code and offset; `compileregex_opts()` becomes a call through it. Both out-parameters take NULL, so no existing caller changes and the message is still formatted in one place. (`xymond/rrd/do_disk.c` and `web/showgraph.c` currently call `pcre2_compile()` directly for the same reason — not converted here.)
- **`xymond/client_config.c`** — prints the config line next to the pcre error (`Invalid pattern '%(Synchronization' at line 2 ...`), following the parser's existing "at line %d" convention; and when pcre2 reports `MISSING_CLOSING_PARENTHESIS` or `MISSING_SQUARE_BRACKET`, adds a hint about the usual cause — the config tokenizes on whitespace, so a pattern written with a literal space is silently truncated, and `[[:space:]]` is the supported spelling. Any other compile error gets the location only.

Gating on pcre2's own error code rather than scanning the pattern text is what @SoundGoof's review asked for, and the counterexample is real: `%[(]{2,1}` contains no whitespace and fails on its quantifier, but a delimiter-counting scan sees one unmatched `(` (it is a class literal) and offers the whitespace hint. Verified against pcre2 directly:

```
FAIL  (Synchronization   err=114  missing closing parenthesis
FAIL  bad{2,1}proc       err=104  numbers out of order in {} quantifier
FAIL  [(]{2,1}           err=104  numbers out of order in {} quantifier
FAIL  foo[a-z            err=106  missing terminating ] for character class
OK    [[:space:]]foo
```

With this, the #253 scenario is self-diagnosing:

```
pcre compile '(Synchronization' failed (offset 16): missing closing parenthesis
Invalid pattern '%(Synchronization' at line 1 (hint: if the pattern contains a space, the config splits tokens on whitespace - write [[:space:]] instead)
pcre compile 'bad{2,1}proc' failed (offset 7): numbers out of order in {} quantifier
Invalid pattern '%bad{2,1}proc' at line 2
pcre compile '[(]{2,1}' failed (offset 7): numbers out of order in {} quantifier
Invalid pattern '%[(]{2,1}' at line 3
```

(line 4 of that fixture is `%Synchronization[[:space:]]Failed` — the supported spelling — and is silent.)

`tests/server/analysis-pattern-error-context.sh` pins all four cases through `xymond_client --dump-config`. It fails against `main` (no location at all) and against the first revision of this PR (false hint on `%[(]{2,1}`). Full suite: 51 passed, 3 skipped, 0 failed.

Two caveats worth stating rather than hiding:

- The reported number is the parser's running line count, and `stackfgets()` follows `include` transparently, so with an included analysis.cfg it spans files instead of restarting per file. That is the same caveat the neighbouring `Syntax error line %d` / `Unknown token ... at line %d` messages already carry; giving `stackio` a current-filename accessor is a separate change.
- A `LOG ... match=%pattern` rule calls `setup_expr()` twice (multiline + single-line), so a bad pattern there is reported twice. Pre-existing — the pcre error was already doubled — and untouched here.

The architectural point in #254 (rrd modules loading all rule types) is left alone deliberately — that belongs to the unified-rule-engine work in #218 rather than a DS-only parse path.

### DIFF
--- a/lib/matching.c
+++ b/lib/matching.c
@@ -23,7 +23,13 @@ static char rcsid[] = "$Id$";
 
 #include "libxymon.h"
 
-pcre2_code *compileregex_opts(const char *pattern, uint32_t flags)
+/*
+ * As compileregex_opts(), but also hands back the pcre2 error code and offset
+ * so a caller can react to *why* a pattern failed - the message itself is
+ * still logged here, so callers only interested in that can keep using the
+ * simpler wrappers below. Both out-parameters are optional (pass NULL).
+ */
+pcre2_code *compileregex_ext(const char *pattern, uint32_t flags, int *errcode, PCRE2_SIZE *erroffset)
 {
 	pcre2_code *result;
 	char errmsg[120];
@@ -35,10 +41,20 @@ pcre2_code *compileregex_opts(const char *pattern, uint32_t flags)
 	if (result == NULL) {
 		pcre2_get_error_message(err, errmsg, sizeof(errmsg));
 		errprintf("pcre compile '%s' failed (offset %zu): %s\n", pattern, errofs, errmsg);
+		if (errcode) *errcode = err;
+		if (erroffset) *erroffset = errofs;
 		return NULL;
 	}
 
+	if (errcode) *errcode = 0;
+	if (erroffset) *erroffset = 0;
+
 	return result;
+}
+
+pcre2_code *compileregex_opts(const char *pattern, uint32_t flags)
+{
+	return compileregex_ext(pattern, flags, NULL, NULL);
 }
 
 pcre2_code *compileregex(const char *pattern)
--- a/lib/matching.h
+++ b/lib/matching.h
@@ -19,6 +19,7 @@
 
 extern pcre2_code *compileregex(const char *pattern);
 extern pcre2_code *compileregex_opts(const char *pattern, uint32_t flags);
+extern pcre2_code *compileregex_ext(const char *pattern, uint32_t flags, int *errcode, PCRE2_SIZE *erroffset);
 #ifdef PCRE_FIRSTLINE
 #define firstlineregex(P) compileregex_opts(P, PCRE_FIRSTLINE);
 #define firstlineregexnocase(P) compileregex_opts(P, PCRE_CASELESS|PCRE_FIRSTLINE);
--- a/tests/server/analysis-pattern-error-context.sh
+++ b/tests/server/analysis-pattern-error-context.sh
@@ -1,0 +1,105 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/server/analysis-pattern-error-context.sh
+#
+# Regression guard for the diagnostics emitted when a %pattern in analysis.cfg
+# fails to compile (#253).
+#
+# Two things are pinned here.
+#
+# 1. Location. The pcre error alone printed only the (possibly truncated)
+#    pattern, so an operator saw the same context-free line in clientdata.log,
+#    rrd-data.log and rrd-status.log with nothing saying which config line
+#    produced it. The parser now adds "at line N", following the convention
+#    already used by the neighbouring syntax errors.
+#
+# 2. The whitespace hint, and its gate. analysis.cfg is tokenized on
+#    whitespace, so a pattern written with a literal space is silently cut at
+#    the space; when the cut lands inside a group or a character class the
+#    remnant fails to compile. That specific cause gets a hint pointing at
+#    [[:space:]]. The gate is pcre2's own error code -- MISSING_CLOSING_
+#    PARENTHESIS / MISSING_SQUARE_BRACKET -- not a scan of the pattern text,
+#    because counting delimiters cannot tell a group from a class literal.
+#
+# The fixture is four PROC rules, one per line, so the asserted line numbers
+# are also a check that the parser reports the right one:
+#
+#   line 1  %(Synchronization Failed        truncated at the space -> hint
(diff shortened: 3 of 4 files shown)


SoundGoof:
### Use PCRE2’s compile error instead of partially parsing regex syntax

[xymond/client_config.c](xymond/client_config.c#L428-L436) counts delimiters without understanding regex context. For example:

```text
PROC %[(]{2,1}
```

The `(` is a literal inside `[]`, but the scanner counts it as an unmatched group delimiter. PCRE2 reports the actual error, `numbers out of order in {} quantifier`, while this code incorrectly adds the whitespace-truncation hint.

Existing Xymon code already demonstrates the stronger approach:

- [lib/matching.c](lib/matching.c#L26-L40) receives `err` and `errofs` from `pcre2_compile()`, then obtains PCRE2’s diagnostic with `pcre2_get_error_message()`.
- [lib/matching.h](lib/matching.h#L20-L29) provides the shared regex compilation API that can be extended without changing existing callers.
- [xymond/rrd/do_disk.c](xymond/rrd/do_disk.c#L39-L55) retains the PCRE2 error code and offset when reporting invalid configured patterns.
- [web/showgraph.c](web/showgraph.c#L951-L963) follows the same direct-PCRE diagnostic pattern.

Add an extended shared compilation function that optionally returns the PCRE2 error code and offset. Keep `compileregex()` and `compileregex_opts()` as compatibility wrappers. `setup_expr()` can then add the whitespace hint only for errors such as:

- `PCRE2_ERROR_MISSING_CLOSING_PARENTHESIS`
- `PCRE2_ERROR_MISSING_SQUARE_BRACKET`

Add behavioral coverage for:

- `%(Synchronization Failed`: line and hint shown.
- `%bad{2,1}proc`: line shown, no hint.
- `%[(]{2,1}`: line shown, no hint.
- A valid pattern using `[[:space:]]`: no diagnostic.

This uses Xymon’s existing PCRE2 error-handling pattern and avoids maintaining a partial regex parser.

bonomani:
You're right, and the counterexample reproduces exactly as you described. Checked `%[(]{2,1}` against pcre2 directly:

```
FAIL  (Synchronization   err=114  missing closing parenthesis
FAIL  bad{2,1}proc       err=104  numbers out of order in {} quantifier
FAIL  [(]{2,1}           err=104  numbers out of order in {} quantifier
FAIL  foo[a-z            err=106  missing terminating ] for character class
OK    [[:space:]]foo
```

The `(` is a class literal and the failure is the quantifier, but the delimiter scan saw one unmatched group and offered the whitespace hint for a pattern containing no whitespace. `114`/`106` separate the two classes cleanly, so the scanner is gone.

Done as you suggested:

- `compileregex_ext(pattern, flags, int *errcode, PCRE2_SIZE *erroffset)` in `lib/matching.c`, with `compileregex_opts()` reduced to a call through it. Both out-parameters are NULL-able, so no existing caller changes and the message stays formatted in one place. I kept the offset even though `setup_expr()` only needs the code — `do_disk.c` and `showgraph.c` call `pcre2_compile()` directly for the (code, offset) pair, so a wrapper that hands back only half of it would not be able to absorb them later. Not converting those here.
- `setup_expr()` gates the hint on `PCRE2_ERROR_MISSING_CLOSING_PARENTHESIS` / `PCRE2_ERROR_MISSING_SQUARE_BRACKET`; anything else gets the location only.
- `tests/server/analysis-pattern-error-context.sh` covers your four cases through `xymond_client --dump-config`. It fails against `main` (no location at all) and against the previous revision of this PR (false hint on `%[(]{2,1}`), so it is a real guard in both directions. Full suite: 51 passed, 3 skipped, 0 failed.

Also rebased onto current `main` — the branch was 140 commits behind, and the shell test harness has grown a lot since.

Two limits I left in place and documented rather than fixed, in case you disagree:

- `cfid` is the parser's running line count and `stackfgets()` follows `include` transparently, so with an included analysis.cfg the reported number spans files instead of restarting per file. Same caveat the neighbouring `Syntax error line %d` and `Unknown token ... at line %d` messages already carry. Fixing it means giving `stackio` a current-filename accessor, which touches every config loader in the tree — separate change, if it is wanted at all.
- `LOG ... match=%pattern` calls `setup_expr()` twice (multiline + single-line), so a bad pattern there reports twice. Pre-existing: the pcre error was already doubled.

SoundGoof (approved):
## Review

No remaining findings.

The update resolves my previous finding. Hint selection now uses PCRE2's
error code rather than delimiter counting. As a result:

- `%[(]{2,1}` receives only the location diagnostic
- an actually truncated pattern receives the `[[:space:]]` hint
- unrelated PCRE errors receive no whitespace hint
- valid `[[:space:]]` patterns remain accepted

### Validation

| Platform | Result |
|---|---|
| Fedora 43 | Full suite: 51 passed, 3 expected skips, 0 failed |
| FreeBSD 14.3 | Focused real-binary test passed |
| NetBSD 11.0 | Focused real-binary test passed |
| OpenBSD 7.9 | Focused real-binary test passed |

Issue #252, opened 2026-07-15
Availability reports: inconsistent up/down semantics between 24x7 and REPORTTIME modes, WARNSTOPS blind spots, doc gaps


bonomani:
Source-verified inventory of problems in the availability-report engine (`lib/availability.c:parse_historyfile()`), found while working on #229's C1 (`MINOUTAGE`). This issue is the carrier for the existing quirks; #229 C1 stays the proposal for the new knob and deliberately inherits these semantics unchanged.

## How the two modes treat each color today

| Color | 24x7 availability | REPORTTIME availability | Counts as outage (`*stops`) |
|---|---|---|---|
| green | up | up | no |
| yellow | up | up | no |
| clear | up | **down** | no |
| blue (disabled) | up | **SLA clock paused** | no |
| purple (no data) | up | **SLA clock paused** | no |
| red | down | down | yes (SLA mode: only during service hours) |

24x7: `fullavailability = 100 - pct(red)` (`availability.c:494`). REPORTTIME: denominator is green+yellow+red+clear time only, and `reportavailability = 100 - pct(red) - pct(clear)` (`availability.c:500-507`).

## Problems

1. **`clear` flips meaning between modes.** It counts as *up* in a 24x7 report and as *down* in a REPORTTIME report — same history, opposite verdicts. Whichever is intended, one of them is wrong, and neither is documented.
2. **No-data (purple) is invisible to every SLA clause.** In 24x7 mode it counts as up; in SLA mode the clock pauses; it never counts as an outage (`fullstops`/`reportstops` only increment for `color > COL_YELLOW`, i.e. red — purple sorts below yellow in the enum). A host whose client dies for a week reports a perfect month. Pausing the clock is a defensible policy; *up* is hard to defend; the mode inconsistency is not.
3. **Empty data ⇒ perfect SLA.** No history in the window → `CLEAR` at 100% (`availability.c:424-429`); REPORTTIME window with zero overlap with the report period → green, 100% (`availability.c:509-513`). Combined with (2), "we didn't measure" is indistinguishable from "it never failed".
4. **The 24x7 denominator starts at the first history entry, not at the report start** (`repinfo->reportstart = starttime`, `availability.c:436`): a host first monitored on the 25th gets its monthly percentage computed over 5 days, silently. Related to (3).
[...]

PR #257, opened 2026-07-16
docs: document WARNSTOPS, correct WARNPCT threshold description (#252 items 5+6)

bonomani:
Items 5 and 6 of #252 — documentation only, no behavior change.

- **WARNSTOPS** (implemented since the Big Brother era, documented nowhere except xymon-xmh.5's circular "Value of the WARNSTOPS tag"): documented in hosts.cfg.5 — more than N outages during the reporting period turns the report cell red regardless of the availability percentage (`lib/availability.c:511`); outages count within `REPORTTIME` service hours only; the count is displayed in parentheses next to the percentage; there is no global default — the per-host tag is the only way to enable it, and `.default.` inheritance works (verified: the generic fallback in `lib/loadhosts.c:245-252` covers all XMH items), so it is added to the `.default.` tag list.
- **WARNPCT**: the old text said green means "100% available"; the code colors green above the configurable green level (default 99.995, `XYMONREPGREEN`) and yellow between the warning threshold (default 97, `XYMONREPWARN`) and the green level (`lib/availability.c:515-517`). The entry now describes both thresholds and their global settings.

`groff -man -z` clean.

Refs #252.

### DIFF
--- a/common/hosts.cfg.5
+++ b/common/hosts.cfg.5
@@ -564,27 +564,33 @@ that cross into/out of a REPORTTIME interval.
 If no REPORTTIME is given, the standard 24x7 calculation is used.
 
 .IP WARNPCT:percentage
-Xymon's reporting facility uses a computed availability threshold
-to color services green (100% available), yellow (above threshold,
-but less than 100%), or red (below threshold) in the reports.
+Xymon's reporting facility colors each service in the availability
+report by comparing the computed availability percentage against two
+thresholds: green when the availability is above the green level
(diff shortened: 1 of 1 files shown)


SoundGoof:
i think this one needs a onceover after merge of warnstops from another pr

bonomani:
Rebased on main to resolve the conflict with f3ff2a8b3 (interim WARNSTOPS entry): this text now replaces it, and the .default. tag-list line is taken from main unchanged.

SoundGoof (approved):
lgtm

PR #259, opened 2026-07-16
xymonnet: remove the legacy SNTP/SNTPOPTS external call-out (#199)

### LINKED ISSUE #199: Remove SNTP/SNTPOPTS: the external sntp path fails against every current sntp, and never recorded offsets
## Summary

Remove (or at minimum deprecate) the `SNTP` / `SNTPOPTS` configuration variables and the external `sntp` call-out path of the `ntp` test. The path was written for the original `msntp`, **fails outright against every current sntp implementation**, never recorded an offset even when it did run, and is redundant with the built-in SNTP probe (#181) and the `NTPDATE` escape hatch.

## The path, and why it is dead (verified 2026-07-16)

When `SNTP` is explicitly set, `run_ntp_service()` runs (`xymonnet/xymonnet.c:1108`):

```
sntp $SNTPOPTS -d <cmdtimeout-1> <server>     # defaults: SNTP="sntp", SNTPOPTS="-u"
```

That is **msntp syntax** ([msntp(1)](https://man.cx/msntp): `-d <secs>` = total-runtime deadline, `-u` = unprivileged port). Against the two implementations actually shipped today:

- **ntp classic 4.2.8 `sntp`** ([docs](https://www.ntp.org/documentation/4.2.8-series/sntp/)): `-d` is `--debug-level` and takes **no argument**, so the number leaks through as a bogus host; `-u` still exists but now takes an argument (unicast timeout in seconds) — it swallows the following `-d` and the invocation errors out.
- **ntpsec `ntpdig`** (also installed as `sntp` on e.g. Debian; [docs](https://docs.ntpsec.org/latest/ntpdig.html)): `-u` does not exist at all → immediate usage error.

So on any current distro the command fails before querying anything; the test goes red regardless of the server's health. (No hang: xymonnet's own `run_command(..., extcmdtimeout)` bound at `xymonnet.c:1114` still applies — only the in-tool deadline that msntp's `-d` used to provide is gone.)

**The offset side never worked at all.** The `" secs"` branch in `xymond/rrd/do_net.c` (line 115) is dead code: tested empirically against real output of msntp, sntp 4.2.6, sntp 4.2.8 and ntpdig, in real status-message framing and bare — it matches none of them. Its tokenizer expects a 7-token line (single date token, detached `+`/`-` sign), which no tool ever printed, and in a status message it consumes the leading `green <date> ...` line first. The built-in probe and ntpdate record offsets via the separate `"offset "` branch, which works.

## Proposal

Since the path is provably non-functional (command fails on both current implementations; offset recording never worked), plain removal is defensible and simpler than a deprecation cycle:

1. Remove the `SNTP`/`SNTPOPTS` branch from `run_ntp_service()` and the defaults from `lib/environ.c:106-107`.
2. Remove the dead `" secs"` parser branch from `xymond/rrd/do_net.c`.
3. `xymonnet.1`: replace the `SNTP`/`SNTPOPTS` entries with a removal note — use the built-in probe (default), or `NTPDATE=ntpdig` / `NTPDATE=sntp` for an external client. That escape hatch also covers anyone genuinely still running msntp (`NTPDATE=msntp`). Also drop "sntp" from the `--cmdtimeout` help text (`xymonnet.c:2290`).

`NTPDATE`/`NTPDATEOPTS` stay: in the shipped design `NTPDATE` is the selector between the built-in probe (empty) and an external client.

**Separate follow-up (unchanged):** rename `NTPDATE` → `NTPTOOL` with a back-compat alias, since the value is "which external NTP client to exec", not ntpdate specifically. Avoid the bare name `NTP` (collides with the hosts.cfg test tag).

## Relationship to other work

Follow-up to #181 (built-in SNTP probe, merged) and #197 (offset recording in ms, merged) — this is the last piece of the ntp-test consolidation.

bonomani:
Closes #199.

Removes the legacy `SNTP` / `SNTPOPTS` external call-out from the `ntp` test. Full analysis with sources in #199; the short version:

- The invocation (`sntp $SNTPOPTS -d <timeout> <host>`) is **msntp syntax**. On ntp classic 4.2.8, `-d` is `--debug-level` (no argument — the timeout leaks through as a bogus host) and `-u` (the `SNTPOPTS` default) now takes an argument and swallows the following `-d`; on ntpsec's `ntpdig`, `-u` doesn't exist. **The command fails outright on every current implementation.**
- The offset side **never worked at all**: the `" secs"` parser branch in `do_net.c` expects a 7-token line no tool ever printed, and in a status message it tokenizes the leading `green <date> ...` line first. Verified empirically against msntp, sntp 4.2.6, sntp 4.2.8 and ntpdig output, framed and bare: zero matches. Dead code since inception.
- No capability is lost: the built-in SNTP probe (#181) is the default, and `NTPDATE` remains the escape hatch to any external client (`ntpdig`, `sntp`, or `msntp` for a genuine holdout).

## Change

- `xymonnet/xymonnet.c`: drop the `SNTP` branch from `run_ntp_service()`; drop "sntp" from the `--cmdtimeout` help text.
- `lib/environ.c`: drop the `SNTP`/`SNTPOPTS` defaults.
- `xymond/rrd/do_net.c`: remove the dead `" secs"` parser branch (the `"offset "` branch serving the built-in probe and ntpdate is untouched).
- `xymonnet.1`: replace the `SNTP`/`SNTPOPTS` entries with a removal note pointing at the built-in probe / `NTPDATE`.
- `tests/rrd/ntp-offset-parse-harness.c`: the old sntp positive case (which only passed with a synthetic single-token date no tool prints) becomes two negative cases — legacy and modern sntp-style output must record nothing.

Net: −33 lines.

## Validation

- `tests/rrd/ntp-offset-parse.sh` passes (11 checks, ASan/UBSan): probe banner and ntpdate offsets still recorded in ms, root distance still recorded/`U`, sntp-style output records nothing, #183 `offset=` regression case intact.
- Full `./configure --server && make` build clean; `xymonnet --help` shows the updated text; `strings` on both binaries confirms no `SNTPOPTS` residue.

### DIFF
--- a/common/hosts.cfg.5
+++ b/common/hosts.cfg.5
@@ -1019,7 +1019,8 @@ test to end up with a green status, all lookups must succeed.
 Check for a running NTP (Network Time Protocol) server on this
 host. By default this uses a built-in SNTP probe and needs no external
 program. Set NTPDATE to the ntpdate program (on ntpsec, "ntpdig") to use
-that instead.
+that instead, and adapt NTPDATEOPTS to the chosen client (its default is
+ntpdate syntax).
 
 .IP "\fBrpc\fR[=\fIrpcservice1\fR,\fIrpcservice2\fR,...]"
 Check for one or more available RPC services. This check is indirect
--- a/common/xymonserver.cfg.5
+++ b/common/xymonserver.cfg.5
@@ -712,8 +712,10 @@ no external program. On ntpsec, where ntpdate is unavailable, set this to "ntpdi
 Default: "" (built-in probe)
 
 .IP "\fBNTPDATEOPTS\fR"
-Options for the ntpdate program (used only when NTPDATE is set).
+Options for the NTPDATE program (used only when NTPDATE is set).
 Default: "\-u \-q"
+The default is ntpdate syntax; when NTPDATE is set to a different client
+(such as "ntpdig"), clear or adapt these options to ones that client accepts.
 
 .IP "\fBRPCINFO\fR"
 Defines the
--- a/docs/manpages/man1/xymonnet.1.html
+++ b/docs/manpages/man1/xymonnet.1.html
@@ -82,10 +82,10 @@ Section: User Commands (1)<br/>Updated: Version 4.3.30: 4 Sep 2019<br/><a href="
   </dd>
   <dt id="cmdtimeout"><a class="permalink" href="#cmdtimeout"><b>--cmdtimeout</b>=<i>N</i></a></dt>
   <dd>This option sets a timeout for the external commands used by the RPC and
-      traceroute tests and by the external NTP backends (ntpdate/sntp). The
-      built-in NTP probe is not governed by it: the probe bounds itself with a
-      short, fixed retransmission budget (a few attempts), so it can never hang
-      and does not need an external kill timer.
+      traceroute tests and by the external NTP backend (NTPDATE). The built-in
+      NTP probe is not governed by it: the probe bounds itself with a short,
+      fixed retransmission budget (a few attempts), so it can never hang and
+      does not need an external kill timer.
     <p class="Pp"></p>
   </dd>
   <dt id="concurrency"><a class="permalink" href="#concurrency"><b>--concurrency</b>=<i>N</i></a></dt>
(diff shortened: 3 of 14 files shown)


SoundGoof:
## Findings

### Document tool-specific `NTPDATEOPTS`

[`xymonnet/xymonnet.1:558-570`](https://github.com/xymon-monitoring/xymon/blob/d31b761c9dfee4fadc2d7605b4d4db831e9a9e5d/xymonnet/xymonnet.1#L558-L570)

The migration note recommends `NTPDATE=ntpdig` or `NTPDATE=sntp`, but xymonnet always appends the shipped `NTPDATEOPTS="-u -q"`.

This produces:

```text
ntpdig -u -q <target>
sntp -u -q <target>
```

`ntpdig` supports neither option. Current ntp.org `sntp` requires an argument after `-u` and does not support `-q`. Both commands therefore fail before querying the server.

Document that `NTPDATEOPTS` must be cleared or configured for the selected client. Update the duplicate guidance in `common/xymonserver.cfg.5`, `common/hosts.cfg.5`, and `xymond/etcfiles/xymonserver.cfg.DIST`.

### Reconcile external-client RRD behavior with its documentation

[`tests/rrd/ntp-offset-parse-harness.c:81-86`](https://github.com/xymon-monitoring/xymon/blob/d31b761c9dfee4fadc2d7605b4d4db831e9a9e5d/tests/rrd/ntp-offset-parse-harness.c#L81-L86)

The new test explicitly requires modern `sntp`/`ntpdig` output to produce no RRD sample, while `xymond_rrd(8)` says external `sntp` offsets are graphed.

This also weakens the claim that `NTPDATE` is an equivalent escape hatch to any external client: the service check may work, but its offset and root-distance graph disappears.

Either parse supported `sntp`/`ntpdig` output or document that only the built-in probe and compatible `ntpdate` output populate `ntp.rrd`.

## Validation

- PR commit cherry-picks cleanly onto current `main` (`eea531bd`).
- Focused NTP parser test passed.
- Full suite: 30 passed, 3 unrelated build-dependent skips, 0 failed.
- Patch whitespace check passed.

bonomani:
Both findings addressed in the amended commit (725b43ac0) — docs-only, no code or test changes, CI green.

**NTPDATEOPTS**: every site that recommends pointing NTPDATE at another client now says the shipped `-u -q` default is ntpdate syntax and must be cleared or adapted to the chosen client — `xymonnet.1` (both the NTPDATEOPTS entry and the removal note), `xymonserver.cfg.5`, `hosts.cfg.5`, and `xymonserver.cfg.DIST`.

**RRD behavior vs docs**: resolved by documenting, not parsing. `xymond_rrd.8` no longer claims external sntp offsets are graphed; it now states that only the built-in probe and ntpdate-style "offset" output feed `ntp.rrd`, and sntp/ntpdig output records nothing (the `xymonnet.1` removal note cross-references this). The sntp graph path never worked, so nothing regresses; a parser for modern sntp/ntpdig output would be new feature surface beyond this removal's scope — happy to open a follow-up issue if there's demand for it.

SoundGoof:
rebase this one

SoundGoof (approved):
lgtm

PR #260, opened 2026-07-17
xymond_client: emit an exact graph-paging linecount for disk/inode

bonomani:
## What

The `disk` and `inode` status pages page their graphs by counting status-body lines — a heuristic ("one df line per filesystem") that miscounts whenever the body contains the df header or lines that never became RRDs. Wrong count → empty or mis-split graph pages.

`unix_disk_report()` / `unix_inode_report()` already know the exact number: they iterate the filesystems and emit one RRD per shown one. So they now state it, embedding

```
<!-- linecount=N -->
```

into the status body (N = filesystems shown, after analysis.cfg IGNORE rules, never the df header).

## Why this is safe

- The consuming side has been on main for years: `lib/htmllog.c` honors an explicit `linecount` before its line-counting heuristic (heritage of Francesco Duranti's hobbit-perl-client, which embeds the same comment in its own statuses). This change just makes Xymon's own report generator use the existing convention.
- No client-side change: the unix client still sends raw df output; the count is added server-side by xymond_client, the one component that knows both the filesystem list and the IGNORE rules.
- The comment is invisible on the rendered page; older servers ignore it. No new configuration.
- A column that states its count no longer needs to sit in the `--multigraphs` list.

## Validation

- `./configure --server && make` clean on current main.
- New regression test `tests/server/disk-linecount-hint.sh` passes (drives the real report generators and checks the emitted count against the shown filesystems).

### DIFF
--- a/tests/server/disk-linecount-hint.sh
+++ b/tests/server/disk-linecount-hint.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/server/disk-linecount-hint.sh
+#
+# unix_disk_report()/unix_inode_report() emit "<!-- linecount=N -->" into the
+# disk/inode status, N = the number of filesystems shown (one RRD file each),
+# so the status page pages the graph precisely instead of re-counting df
+# lines. The functions live in the xymond_client daemon and only output
+# through the message channel, so - like svcstatus-trends-overflow.sh - this
+# is a source guard: it asserts the fix is present and correctly shaped
+# (counted per shown filesystem, emitted before the df output), which is what
+# survives future edits. The renderer half (htmllog honouring an explicit
+# linecount override) is exercised behaviourally elsewhere.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+ROOT=$(find_root)
+FILE="$ROOT/xymond/xymond_client.c"
+
+[ -f "$FILE" ] || skip "xymond/xymond_client.c absent"
+src=$(cat "$FILE")
+
+# The count is incremented once per shown (non-ignored) filesystem - so it
+# equals the RRD-file count and excludes IGNORE'd mounts and the df header.
+assert_contains "if (!ignored) fscount++;" "$src" \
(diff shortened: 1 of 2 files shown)


bonomani:
Converting to draft after a closer look — this hint has no observable
effect in practice, and is superseded by the self-describing METRICS
work.

**No functional divergence from the existing count.** In every real
case the emitted `linecount=N` equals what the renderer's existing
line-counting heuristic already computes:

- Normal `df`: both give the same number.
- Wrapped `df` lines (long LVM/NFS device names) — the one case where
  text line-counting could diverge — never reach the server: the client
  normalizes them first (`df -P`, and the explicit un-wrap `sed` in the
  client scripts, e.g. `xymonclient-netbsd.sh`). So the server always
  sees one line per filesystem.
- `analysis.cfg DISK ... IGNORE`: reduces both `fscount` and the shown
  table lines, so the hint and the heuristic stay in agreement.
- `NORRDDISKS`/`RRDDISKS`: both over-count identically (the filtered fs
  is still shown and counted) — no difference between the two.

So the hint fixes no miscount; its only value is architectural
(decoupling the renderer from disk-report format knowledge). That
benefit is invisible today (the input already arrives normalized) and
doesn't let the fragile heuristic be removed (it must stay for clients
that don't emit the hint).

**Superseded by METRICS.** A self-describing `<!--XYMON METRICS: disk`
block already carries one instance line per filesystem, i.e. the same
count, declared at the source and respected by the paging logic. Once
paging takes the count from the block, this separate `linecount` hint
is fully redundant.

Parking as draft rather than merging a second, weaker count mechanism.
Reconsider only if a concrete case surfaces where the renderer's
heuristic actually miscounts on input the client did not normalize.

bonomani:
Closing in favor of #246, which supersedes this from the server side: generate_html_log() now derives the paging count from xymon_graph_count() — the actual, stale-excluded RRD file set the graph will draw — for every column, not just disk/inode. That count stays correct under NORRDDISKS and the new RRDEXCLUDE/RRDINCLUDE filters, where this PR's producer-side filesystem count would drift (the client cannot know which of its filesystems the server filtered out of trending). The client-side `<!-- linecount=N -->` hint would only matter on the rrdcount<0 fallback path (locator-based setups); if that gap proves real it can be revisited as a follow-up to #246.

PR #261, opened 2026-07-17
showgraph: make the RRD instance sort a total order

bonomani:
`rrd_name_compare()` picked its mode per pair — numeric when both keys of that pair parsed as integers, `strcmp` otherwise. On a mixed set that is intransitive, which is undefined behaviour for `qsort(3)`.

Measured: the three permutations of `{9, 10, 1a}` produced three different "sorted" orders.

| | before | after |
|---|---|---|
| `9 < 10` | numeric | — |
| `"10" < "1a"` | text | — |
| `"1a" < "9"` | text | — |
| result | three orders for one set | one deterministic order |
| `007` vs `7` | compared equal, order unspecified | tie broken on the text form |

Mixed sets are real: interface instances can be `0`, `1`, `eth0` on the same host. Display order and the first/count graph paging built on it depended on `readdir()` order.

The mode is now decided once for the whole set — which is what the comparator's own comment ("See if the keys are all numeric") always promised. Homogeneous sets sort exactly as before.

**Test:** all six permutations of a mixed set must sort identically, numerically-equal distinct keys must have one order, plain integers unchanged. It drives the real `rrd_set_sort_mode()` rather than a copy of it — @SoundGoof showed the earlier version kept passing with the production pre-scan replaced by `rrd_keys_all_numeric = 0`; that mutation now fails two of the three checks.

### DIFF
--- a/tests/web/showgraph-sort-total-order.sh
+++ b/tests/web/showgraph-sort-total-order.sh
@@ -1,0 +1,88 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/web/showgraph-sort-total-order.sh
+#
+# rrd_name_compare() must be a total order. The old per-pair mode choice
+# (numeric only when BOTH keys of a pair were numeric) was intransitive on
+# mixed key sets (9 < 10 < "1a" yet "1a" < 9), which is undefined behaviour
+# for qsort(3) - the sorted order depended on readdir() order. And distinct
+# keys must never compare equal ("007" vs "7"), or their order is unspecified.
+#
+# This extracts the REAL comparator (and its mode flag) from web/showgraph.c,
+# wraps it in a driver replicating the caller's all-numeric pre-scan, and
+# asserts: every input permutation of a mixed set sorts identically, and
+# numerically-equal distinct keys have one deterministic order.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+ROOT=$(find_root)
+SRC="$ROOT/web/showgraph.c"
+[ -f "$SRC" ] || skip "web/showgraph.c absent"
+
+CC=${CC:-cc}
+command -v "$CC" >/dev/null 2>&1 || skip "no C compiler available (CC=$CC)"
+
+work=$(mktemp -d)
+trap 'rm -rf "$work"' EXIT
+
+# Extract the comparator and its mode flag from the real source.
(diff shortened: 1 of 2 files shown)


SoundGoof:
## Review: Request changes

### Fall back when sanitizer runtimes are unavailable

`tests/web/showgraph-sort-total-order.sh:84`

The test unconditionally links with ASan/UBSan. On GCC 15.3.1 without the optional runtime packages, linking fails and makes the full suite fail. Existing sanitizer tests retry without sanitizer flags.

Add the same fallback used by `tests/network/ntp-probe.sh` and related tests.

### Exercise production mode selection instead of duplicating it

`tests/web/showgraph-sort-total-order.sh:41`

The harness reimplements the caller’s set-wide pre-scan. I replaced the production pre-scan in `generate_graph()` with `rrd_keys_all_numeric = 0`; all assertions still passed.

Extract or invoke the real mode-selection path so the regression test fails if production returns to pairwise or lexical-only sorting.

## Verification

- Clean synthetic merge into current `upstream/main`.
- Configured build completed, including `showgraph.cgi`.
- Post-build suite: 44 passed, 0 skipped, 1 failed.
- The sole failure was the new test’s missing sanitizer-runtime fallback.
- `git diff --check` passed.

bonomani:
Both fixed.

**Mode selection.** You were right that the replica made the test blind — the pre-scan is now a named function, `rrd_set_sort_mode()`, which the test extracts and calls alongside the comparator instead of reimplementing. Your mutation (replace the real pre-scan with `rrd_keys_all_numeric = 0`) now fails two of the three checks:

```
FAIL numeric: got '007 10 7 ' want '007 7 10 '
FAIL: got '10 2 ' want '2 10 '
```

**Sanitizer runtime.** Now goes through `asan_usable()`, which builds *and runs* a probe before committing to `-fsanitize=`, falling back to a plain build. Verified with a `cc` wrapper that rejects `-fsanitize`: the test passes on the fallback instead of taking the suite down.

The branch predated both helpers, so `upstream/main` is merged in first. That also brought `mktempdir()`, which replaces a hand-rolled `EXIT` trap here — it had been overwriting the shared cleanup, which is the pitfall `asan_usable()`'s own comment warns about.

SoundGoof (approved):
No blocking findings.

The two previous findings are resolved:

- The regression now exercises the production `rrd_set_sort_mode()` function.
- Sanitizer unavailability falls back to a plain compiler build.


- `git diff --check` passes.
- The focused regression passes.
- Forcing production mode selection to lexical ordering makes the regression
  fail on both numeric-order assertions.
- A compiler wrapper that rejects sanitizer flags confirms the plain-build
  fallback passes.

PR #262, opened 2026-07-17
docs: name the GRAPHS ::N setting by its unit - instances per image

bonomani:
Follow-up coherence fix to #243, in two commits:

**1. docs** — the new xymonserver.cfg documentation called the third `GRAPHS` field the "per-image split size". "Split size" names the operation rather than the unit; say what the value counts: **the maximum number of instances (RRD files) shown per graph image**. The two sentences that said only "RRD files"/"files" now use "instances (RRD files)" consistently. The `::N` syntax is unchanged; "splitting" as a verb stays.

**2. lib,web** — rename the `xymongraph_t.maxgraphs` field to `maxinstancesperimage`. The identifier has never been public (no config file, man page or DIST comment ever says "maxgraphs" — users only write the anonymous third `GRAPHS` field), and the name is wrong: it does not cap graphs — the number of images is derived and unbounded — it caps instances per image. Pure mechanical rename (11 occurrences, 4 files), no behaviour change; verified with a full `./configure --server && make`.

### DIFF
--- a/common/xymonserver.cfg.5
+++ b/common/xymonserver.cfg.5
@@ -523,7 +523,7 @@ for storing data and for the column's default status-page graph.
 GRAPHS_<COLUMNNAME> changes only what graphs are shown on that
 column's status page. GRAPHS defines the graph list and order on
 the "trends" page, and also stores graph-display metadata such
-as the per-image split size.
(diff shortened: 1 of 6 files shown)


SoundGoof (approved):
lgtm

PR #263, opened 2026-07-19
tree: xtree lifecycle fixes - tsearch destroy leak + array-variant deleted-record iteration

bonomani:
Two xtree lifecycle bugs in the fallback (array) variant and the POSIX
binary-tree (tsearch) variant. Both are pre-existing on `main`, found
during unrelated review work; neither has another PR upstream.

## 1. tsearch `xtreeDestroy` leaks the whole tree

The `HAVE_BINARY_TREE` variant of `xtreeDestroy()` frees only the tree handle, leaking every internal tsearch node and record wrapper of the destroyed tree.

**Why now, when no default build compiles that variant:** the genconfig probe for the POSIX binary-tree path is commented out on `main`, so today this is latent. But it is **not** hypothetical:

- **devel runs it live** — its probe is enabled, so any glibc/BSD devel build uses the tsearch variant as the running code.
- **#106 tracks bringing that to main** (the portability items `d56377dc`/`1c93aea9`), and TBT patch `61` (issue #29) is the same enablement. The moment that lands, this leak fires on every `xtreeDestroy` call site.

**The fix:** the canonical POSIX teardown — delete the root record until the tree is empty (the `*(rec **)root` node access POSIX documents): `tdelete` frees the internal nodes, we free the record wrappers. No allocation during teardown, no scratch state. Keys and userdata stay caller-owned, unchanged.

Compared to devel's answer (`tdestroy(tree->root, free)` **plus a probe guard** that disables the whole tsearch path on platforms without GNU `tdestroy`), this is pure POSIX: enabling the binary-tree path needs **no platform gate**, and the site-specific teardown patch tracked as TBT `235` (flagged in #29 as messy: GNU-only, valgrind residue, risky `free(zombie->key)`) becomes unnecessary.

## 2. array variant visits deleted records (second commit)

`xtreeDelete` only tombstones a slot in the fallback (array) variant, and the iterators did not all honor that — this one bites the DEFAULT build, today:

- **`xtreeFirst`** returned slot 0 unconditionally, handing back userdata the caller had already freed when the first-sorting entry was a tombstone. `xymond_rrd`'s update-cache eviction walks the whole tree and hits this once the lowest-keyed entry is evicted.
- **`xtreeNext`** read `entries[pos].deleted` **before** its bounds check — an out-of-bounds slot read when the trailing entries are all tombstones.

Both now skip tombstones with the bounds check first.

## Tests

- `tests/server/xtree-destroy.sh`: compiles `lib/tree.c` **both ways** (one config.h shim per variant, so no built tree is needed) and runs an insert/find/delete/destroy workload under AddressSanitizer's leak checker. With the destroy fix reverted the tsearch variant reports ~48 kB leaked from a 1000-record tree.
- `tests/server/xtree-iterate-deleted.sh`: same both-ways scheme, walks a tree with deleted head/middle/tail entries under ASan and dereferences every visited record so a stale pointer trips the sanitizer. With the iteration fix reverted the array variant aborts.

Both variants run clean with the fixes. Full suite: 20/20 on this branch.

Refs #106 (portability cluster), #29 (TBT `61`/`235`/`240`).

### DIFF
--- a/lib/tree.c
+++ b/lib/tree.c
@@ -60,7 +60,19 @@ void *xtreeNew(int(*xtreeCompare)(const char *a, const char *b))
 
 void xtreeDestroy(void *treehandle)
 {
-	free(treehandle);
+	xtree_t *tree = treehandle;
+
+	if (!tree) return;
+	/* No tdestroy() outside glibc, so delete the root record until the
+	 * tree is empty (*(rec **)root is the POSIX-documented node layout):
+	 * tdelete frees the internal nodes, we free the record wrappers.
+	 * Keys and userdata belong to the caller, as always. */
+	while (tree->root) {
+		treerec_t *rec = *(treerec_t **)tree->root;
+		tdelete(rec, &tree->root, xtree_i_compare);
+		free(rec);
+	}
+	free(tree);
 }
 
 xtreeStatus_t xtreeAdd(void *treehandle, char *key, void *userdata)
--- a/lib/tree.c
+++ b/lib/tree.c
@@ -324,11 +324,15 @@ xtreePos_t xtreeFind(void *treehandle, char *key)
 xtreePos_t xtreeFirst(void *treehandle)
 {
 	xtree_t *mytree = (xtree_t *)treehandle;
+	xtreePos_t pos = 0;
 
 	/* Does tree exist ? Is it empty? */
 	if ((treehandle == NULL) || (mytree->treesz == 0)) return -1;
 
-	return 0;
+	/* Skip deleted records: their userdata is stale (usually freed) */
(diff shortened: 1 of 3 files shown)


SoundGoof (approved):
No blocking findings.

- Clean merge into current `main`.
- POSIX and FreeBSD confirm the portable `tdelete()` destruction pattern.
- Iterator regression test fails against unpatched `main` and passes with this PR.
- Configured server build completed successfully.
- Post-build suite: 45 passed, 2 skipped, 0 failed.
- Both variants passed focused heap-perturbation tests.
- All four GitHub checks are green.
- `git diff --check` passed.

PR #264, opened 2026-07-19
xymond_rrd: make the drophost deletion unraceable, flush the cache before a rename

### LINKED ISSUE #290: xymond_rrd: channel hostname used as a path component without confinement -- @@drophost deletes $RRDDIR recursively
## Summary

`xymond_rrd` turns the hostname carried in a channel message straight into a path component, in two handlers. One filtered it with `basename()`, which returns `".."` for `".."` and so confines nothing; the other did not filter at all. Since the first of them deletes **recursively**, an unconfined component destroys the RRD directory — the historical data.

Same defect class as #283, different worker. Split out because the trigger and the blast radius are not the same.

## The two call sites

| handler | effect | filtering |
|---|---|---|
| `@@drophost` (`xymond_rrd.c:418`) | `dropdirectory("$RRDDIR/<host>", 1)` — recursive delete | `basename()` |
| `@@renamehost` (`:440-441`) | `rename("$RRDDIR/<old>", "$RRDDIR/<new>")` | **none** |

## Reproduction

Workers read their messages on stdin (`xymond_worker.c`), so the binary can be driven directly. Against the unpatched worker:

```console
$ ls $V/rrd
realhost
$ printf '@@drophost#1|1785830400|10.0.0.99|..\n@@\n' \
  | ./xymond_rrd --env=xymonserver.cfg --rrddir=$V/rrd --no-rrd --processor=/bin/cat --debug
Starting to remove directory $V/rrd/..
Recurse into $V/rrd/../rrd
Removing file $V/rrd/../rrd/realhost/tcp.rrd
Removing directory $V/rrd/../rrd/realhost
Removing directory $V/rrd/../rrd
Removing directory $V/rrd/..
$ ls $V/rrd
ls: cannot access '$V/rrd': No such file or directory
```

`@@renamehost` with a new name of `../escaped` likewise creates a directory outside `$RRDDIR`.

One measurement worth recording, because it made an early version of the regression test pass for the wrong reason: a *sibling file* of `$RRDDIR` often survives this. `dropdirectory()` unlinks entries while `readdir()` is still walking them, which is undefined for entries not yet returned, so a sibling can simply be skipped. `$RRDDIR` itself and sibling *directories* are destroyed deterministically, because directories are recursed into. Any test here should assert on those, not on a canary file.

## Reachability

Lower than #283, and honestly so:

- Both handlers act on **admin-originated** messages. `drophost` and `renamehost` are xymond commands, gated by `--admin-senders=127.0.0.1,$XYMONSERVERIP` in the shipped `tasks.cfg`. A remote reporting client does not reach them.
- For the name to be attacker-chosen at all, `--ghosts=allow` is needed (not the default, not in the shipped `tasks.cfg`).

So this is hardening rather than a demonstrated remote attack. What makes it worth fixing anyway is `net_worker_run()` (`xymond_rrd.c:243`): in a distributed setup the worker listens on its own ip:port and takes messages from remote xymond instances. Its trust boundary is not "our own xymond", so the component has to be confined in the worker — no entry-side check in xymond could cover that input.

## Fix

Route both through `safe_basename()` (#147) and drop the message when nothing is left, rather than letting an empty component collapse the path onto `$RRDDIR` itself — for a recursive delete that would mean wiping every host's data instead of one.

PR: #287.

## Related

- #283 — same class in `xymond_hostdata`, fixed by #284
- #221 — `xymond_filestore`; `basename()` **is** sufficient there, since the component names a file followed by a suffix rather than a directory to delete

bonomani:
`@@drophost` forked the host directory deletion, so a message read while the child was still emptying it recreated the directory and the child's final `rmdir()` failed — orphaned RRD files for a host that no longer exists.

**Delete synchronously instead.** Message processing is sequential, so the deletion completes before the next message is read and nothing can land inside it. That is a property of the ordering, not of any timing assumption: no window to choose, no table to keep, no timestamps to compare. The tree is one host's RRDs, and the worker already blocks on `rrd_update` for every sample it writes.

Two cache fixes travel with it, each standing alone:

- **drophost** discards the host's pending updates before the files go, so a later flush cannot write them back into the deleted directory;
- **renamehost** flushes pending updates into the old-named files before they move. Without this they are lost on *every* rename — the cache is keyed on the old path and nothing rewrites those keys. Attempted only when the directory is still there: the stock install runs one `xymond_rrd` per channel over a shared `rrddir` and xymond posts the marker to both, so the second worker finds it already moved and discards quietly, as it did before.

`rrdcache_drop_host()` keeps the cache records and clears only their pending values — they are persistent by design and each owns its `cacheitem->tpl`, so deleting them would leak a template per drop.

**Both names go through `safe_basename()`**, which refuses `/`-bearing and the degenerate `.`, `..` and `""` outright rather than reducing them, with `snprintf()` and a truncation check. This closes **#290**, a pre-existing hole: `basename("..")` is `".."`, so `drop ..` appended it to `rrddir` and handed the parent of the whole RRD tree to a recursive delete — and xymond accepts `drop` from anyone who can reach port 1984.

Pre-existing, not fixed here: `rrdcacheflushhost()` matches `"/host/fn"`-shaped keys against the bare hostname `rrdcachectl` sends, so it is a silent no-op. Deserves its own issue.

**Test.** Message ordering matches the real channel (xymond stamps `metadata[1]` at post time and delivers in that order), so nothing depends on a stream the channel cannot produce. The pre-rename flush is pinned by ordering against the worker's shutdown-flush line, since the summary line prints from the match count whether or not anything was written.

Mutations caught: either name via `basename()`, drop skips the empty-name refusal, no pre-rename flush, drop does not discard, flush removed. Not caught and left that way: the rename empty-name refusal is redundant with the kernel, which returns `EINVAL` for the paths it would let through.

Suite: 58 passed, 0 skipped, 0 failed. Test runtime 0.1s.

Fixes #290.

### DIFF
--- a/tests/server/xymond-rrd-drophost.sh
+++ b/tests/server/xymond-rrd-drophost.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/server/xymond-rrd-drophost.sh
+#
+# The drophost straggler race: @@drophost forks the host directory
+# deletion, but messages for that host already queued in the channel can
+# still arrive afterwards and recreate RRD files inside (or after) the
+# dying directory - leaving a half-resurrected host behind. The writer
+# now keeps a drop barrier that discards messages for a recently dropped
+# host and purges its cached updates before the deletion starts.
+#
+# The test forces the losing interleaving: the deletion completes, THEN
+# the straggler arrives - without the barrier the recreated directory
+# has nothing left to clean it up.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+require_bin XYMOND_RRD "xymond/xymond_rrd"
+
+work=$(mktempdir)
+mkdir -p "$work/rrd" "$work/tmp"
+ts=$(date +%s)
+
+status_msg() {  # status_msg <statusts> -- one disk status for testhost
+	printf '@@status|%s|127.0.0.1|origin|testhost|disk|%s|green||green|%s|0||0||%s|0|linux|/\n' \
+		"$1" $(($1+1800)) "$ts" "$ts"
+	printf 'disk report\n'
+	printf '/dev/sda1 1000000 400000 600000 40%% /\n'
+	printf '@@\n'
+}
+
+{
+	status_msg "$ts"
+	printf '@@drophost|%s|127.0.0.1|testhost\n@@\n' "$ts"
+	# Give the forked deletion time to FINISH before the straggler
+	# arrives (without the delay the child's rm usually runs last and
+	# hides the recreation by timing luck).
+	sleep 2
+	status_msg $((ts+1))
+} | env XYMONHOME="$work" XYMONTMP="$work/tmp" \
+	"$XYMOND_RRD" --rrddir="$work/rrd" --no-cache 2>/dev/null
+sleep 1		# any still-running forked deletion
+[ -e "$work/rrd/testhost" ] \
+	&& fail "straggler recreated the dropped host directory: $(ls "$work/rrd/testhost")"
+
+# A host dropped longer ago than the barrier window resumes normally -
+# verified indirectly: a FRESH host (never dropped) is unaffected.
+{
+	printf '@@status|%s|127.0.0.1|origin|otherhost|disk|%s|green||green|%s|0||0||%s|0|linux|/\n' \
+		$((ts+2)) $((ts+1802)) "$ts" "$ts"
+	printf 'disk report\n'
+	printf '/dev/sda1 1000000 400000 600000 40%% /\n'
+	printf '@@\n'
+} | env XYMONHOME="$work" XYMONTMP="$work/tmp" \
+	"$XYMOND_RRD" --rrddir="$work/rrd" --no-cache 2>/dev/null
+[ -d "$work/rrd/otherhost" ] || fail "unrelated host's updates must not be barriered"
+
+# renamehost: pending CACHED updates must flush into the old-named files
+# before the rename moves them (rrdcacheflushhost cannot do this: it
+# expects "/host"-shaped keys and rate-limits; the old call was a no-op).
+rm -rf "$work/rrd"; mkdir -p "$work/rrd" "$work/tmp"
+{
+	status_msg "$ts"
+	printf '@@renamehost|%s|127.0.0.1|testhost|newhost\n@@\n' "$ts"
+} | env XYMONHOME="$work" XYMONTMP="$work/tmp" \
+	"$XYMOND_RRD" --rrddir="$work/rrd" --debug >"$work/dbg.log" 2>&1
+[ -d "$work/rrd/newhost" ] || fail "rename did not move the host directory"
+[ -e "$work/rrd/testhost" ] && fail "old directory survived the rename"
+grep -q "flushed and dropped 1 entries for host testhost" "$work/dbg.log" \
+	|| fail "pending update not flushed before the rename: $(grep -i updcache "$work/dbg.log")"
+
+echo "OK $(basename "$0")"
--- a/tests/server/xymond-rrd-drophost.sh
+++ b/tests/server/xymond-rrd-drophost.sh
@@ -1,0 +1,174 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/server/xymond-rrd-drophost.sh
+#
+# The drophost straggler race: @@drophost forks the host directory
+# deletion, but messages for that host already queued in the channel can
+# still arrive afterwards and recreate RRD files inside (or after) the
+# dying directory - leaving a half-resurrected host behind.
+#
+# The writer decides "straggler" by message time: a message at or before the
+# drop's own timestamp was already on its way, anything later means the name
+# is in use again. Every assertion here is therefore driven by the timestamps
+# in the messages, not by sleeping - the only sleep is the one that lets the
+# forked deletion finish, which is what makes the losing interleaving happen.
+#
+# The barrier table is process-local, so assertions about it have to be made
+# inside one worker lifetime: a second invocation starts with an empty table
+# and would accept anything.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+require_bin XYMOND_RRD "xymond/xymond_rrd"
+
+work=$(mktempdir)
+ts=$(date +%s)
+
+status_msg() {  # status_msg <msg-timestamp> [hostname]
+	printf '@@status|%s|127.0.0.1|origin|%s|disk|%s|green||green|%s|0||0||%s|0|linux|/\n' \
+		"$1" "${2:-testhost}" $(($1+1800)) "$ts" "$ts"
+	printf 'disk report\n'
+	printf '/dev/sda1 1000000 400000 600000 40%% /\n'
+	printf '@@\n'
+}
+
+run_worker() {  # run_worker <rrddir> [extra args...]
+	local dir=$1; shift
+	env XYMONHOME="$work" XYMONTMP="$work/tmp" \
+		"$XYMOND_RRD" --rrddir="$dir" "$@"
+}
+
+mkdir -p "$work/tmp"
+
+# ---- the race itself: a message older than the drop must not resurrect it ----
+# Its own run: in the combined stream below a legitimate later message
+# recreates the directory, so "does it exist at the end" could not tell a held
+# barrier from a broken one.
+rrd1="$work/rrd1"; mkdir -p "$rrd1"
+{
+	status_msg "$ts"
+	printf '@@drophost|%s|127.0.0.1|testhost\n@@\n' $((ts+10))
+	# Let the forked deletion FINISH before the straggler arrives. Without
+	# it the child's rm usually runs last and hides the recreation by timing
+	# luck - the losing interleaving is the one worth pinning.
+	sleep 2
+	status_msg $((ts+5))		# generated before the drop: a straggler
+} | run_worker "$rrd1" --no-cache 2>/dev/null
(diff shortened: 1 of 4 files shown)


SoundGoof:
# Review

## Findings

### P1: Failed filesystem renames still discard updates and rename locator state

`xymond/xymond_rrd.c:493-497`

The rename path drops the host cache, activates the old-host barrier, and updates
locator state without checking whether `rename()` succeeded.

I reproduced this with an occupied destination. Both filesystem trees remained,
confirming the rename failed, but the following valid status for the old hostname
was discarded by the five-minute barrier. Locator mode would additionally record
a rename that never happened on disk.

Check `rename()` first and only activate the barrier and update locator state
after success. The pre-rename cache flush still needs to happen before the files
move, but destructive cache removal should follow the result.

### P2: The regression test does not exercise isolation or expiry

`tests/server/xymond-rrd-drophost.sh:49-59`

The “fresh host unaffected” check starts a new `xymond_rrd` process after the
drop process exits. Since `recentdrops` is process-local, the new process has no
barrier state. It therefore tests neither another hostname during an active
barrier nor expiry after 300 seconds.

As a sensitivity check, I changed the barrier to reject every hostname forever
after the first drop. The unchanged test still passed.

Keep the drop, unrelated-host update, and expiry update in one worker lifetime,
with a practical way to shorten or control the expiry window for testing.

## Validation

- Current CI: green
- Full Linux suite: 58 passed, 0 skipped, 0 failed
- Changed test passed on OpenBSD 7.9, FreeBSD 14.3, and NetBSD 11.0
- Adversarial failed-rename reproduction confirmed the P1 behavior

bonomani:
Both findings confirmed — I reproduced each independently before fixing, and both were real.

### P1: failed rename

Reproduced with an occupied destination: both trees remain, and the next valid `testhost` status is discarded. Worth adding a detail that makes it worse rather than better — **this PR is what made it harmful**. `rename()` was already unchecked on `main`; unchecked it was harmless, because nothing followed it. Putting a 300 s barrier and a locator update behind it turns the same failure into a five-minute blackhole for a live host plus a locator record for a directory that is not there. That is my regression, not a pre-existing one.

Fixed as you described, keeping the ordering distinction you drew: the cache flush stays *before* the move — the values land in the files they were cached for whether or not the rename then works — while `note_hostdrop()` and `locator_rename_host()` run only on `rename() == 0`, with the failure logged.

(My first attempt to reproduce this was itself vacuous: I looked for an RRD under `testhost` afterwards, but the pre-rename message had already created one. Re-running with an empty source directory gives the real answer.)

### P2: the test proved nothing

Reproduced with your exact sensitivity check — barrier rejects every hostname forever after the first drop, rebuild, test still `OK`. The cause is the one you identified: the "unrelated host" block ran in a *second* `xymond_rrd` process, where `recentdrops` is empty. The comment above it also claimed to verify expiry, which it never did.

Rebuilt so the drop, the unrelated host and the expiry all run inside one worker lifetime. That needs a shorter window than 300 s, so `xymond_rrd` gained `--drop-barrier=N` — a testing hook, undocumented on purpose since the default is the behaviour. **If you would rather not have new CLI surface for this, say so**: the alternative is dropping the expiry assertion and stating in the test that expiry is unverified, which I would rather do than keep a check that cannot fail.

One thing that caught me while rebuilding it: the straggler case needs its own worker run. In the combined stream the post-expiry message legitimately recreates the directory, so "does it exist at the end" cannot distinguish a held barrier from a broken one — the same trap the original fell into, one level down.

### Every assertion is now measured

| mutation | assertion that fails |
| --- | --- |
| barrier rejects every host, forever | unrelated host was dropped |
| barrier never rejects | straggler recreated the directory |
| barrier never expires | host did not resume |
| failed rename still barriers | old host's next update discarded |
| no pre-rename flush | cached update not flushed |

Also rebased onto current `main` — the branch was 143 commits behind, so its local suite was running 19 tests rather than 58. Suite now 58 passed / 0 skipped / 0 failed.

Thanks for both of these. The sensitivity check is the part I should have run myself.

bonomani:
Reworked — and I owe you a correction.

**The `if (rename(...) == 0)` guard I added last round was wrong.** The stock install runs one `xymond_rrd` per channel over a shared `rrddir`, and xymond posts the rename marker to both; the second worker gets `ENOENT` and so never armed its barrier. My fix for your P1 reopened the race in half the installs.

**The window itself was the deeper defect**, present in the revision you reviewed: nothing lowered a barrier, so any name legitimately back inside 300s was blackholed — rename-and-rename-back, drop-and-re-add, even `rename WEB1 web1` (the table uses `strcasecmp`). The barrier now decides by message time and lifts itself, which removes the window, the constant, and the `--drop-barrier` flag I added for testability. That also answers my own question about new CLI surface: there is none.

With that rule, arming on a failed rename is harmless, so it is unconditional again and the locator is told either way.

Your P2 fix went further too: building the mutation matrix showed my rewritten test drove one worker (so "rename must arm the barrier" passed with the call deleted) and used a plain hostname where `basename()` is a no-op. Both covered now, and both cache branches run with the cache on — with `--no-cache` the discard path had no coverage at all.

Eight mutations, all caught. Suite 58/0/0.

bonomani:
Reworked again — the message-time barrier I described above was wrong, and a review pass caught why.

`metadata[1]` is the channel *post* time and delivery is FIFO, so any message read after the drop marker is stamped at or after it. My rule discarded the at-or-before set, which is empty in practice — the barrier was aimed at the wrong messages entirely, and my test only looked right because it fed an out-of-order stream xymond cannot produce.

So the barrier is gone. The deletion is synchronous, which closes the race by ordering rather than by any timing model. Both ingest paths are now byte-identical to `main`, which also removes the `--processor` suppression and the failed-rename blackhole you and the pass flagged. What remains is the cache discard, the pre-rename flush, and `safe_basename()` confinement on both names.

That confinement closes a pre-existing hole worth knowing about independently of this PR: `basename("..")` is `".."`, so `drop ..` handed the parent of the whole RRD tree to a recursive delete, unauthenticated on port 1984.

Test is 11s -> 0.1s and no longer depends on an impossible message order.

SoundGoof:
## Findings

### The second stock RRD worker discards cached updates on rename

`xymond/xymond_rrd.c:505-520`

The stock configuration runs `rrdstatus` and `rrddata` against one shared
RRD directory. After the first worker renames the directory, the second
worker sees the old path missing and calls:

```c
rrdcache_drop_host(safeold, 0);
```

This frees its distinct pending updates instead of writing them under the
new hostname.

I reproduced this with two live workers. Worker one flushed and renamed.
Worker two logged:

```text
renamehost: .../testhost already moved, discarding its cached updates
```

The second timestamp was absent from the resulting RRD. Caching is enabled
by default.

Move or re-key the second worker's pending entries to the new hostname, or
flush them into the renamed files. Add a two-worker regression case.

### Failed filesystem renames still update locator state

`xymond/xymond_rrd.c:522-524`

`rename()` remains unchecked and `locator_rename_host()` runs
unconditionally.

An occupied non-empty destination reproduces the failure: both old and new
directories remain, but locator mode is still instructed to move the host.
The duplicate-worker `ENOENT` case needs special handling, but it does not
justify treating `ENOTEMPTY`, permission errors, and other failures as
successful renames.

Distinguish a directory already moved by the peer worker from a genuine
rename failure. Update locator state only for a successful or verified
already-completed move.

### The test does not detect restoration of background deletion

`tests/server/xymond-rrd-drophost.sh:45-56`

The central assertion sends a status followed by a drop, then checks the
small directory after the worker exits. It sends no message while deletion
is active and does not force deletion to remain in progress.

I changed:

```c
dropdirectory(hostdir, 0);
```

back to:

```c
dropdirectory(hostdir, 1);
```

All 20 test runs still passed. The test therefore does not guard the main
race this PR claims to fix.

Use a sufficiently large tree and a following update, or another
deterministic synchronization mechanism, so restoring background deletion
reliably fails the test.

## Validation

- Linux suite: 55 passed, 3 expected skips, 0 failed.
- Focused test passes on FreeBSD 14.3, NetBSD 11.0, and OpenBSD 7.9.
- `shellcheck -x` and `git diff --check` pass.

bonomani:
All three fixed in `6eb57777c`. Thanks — the third one is the important one and I had not caught it.

### The test did not guard the race

Reproduced your check: with `dropdirectory(hostdir, 1)` restored, **10 runs, 10 passes**. The dropped directory holds a single file, so the forked child finishes long before the assertion runs, and nothing is sent while the deletion is in flight. The test asserted the end state, which is the same either way.

It now drops a host with ~50k files. Synchronously the directory is gone the moment the worker returns; forked, the child is still emptying it. Verified in both directions rather than assumed:

- barrier reverted: **10 runs, 10 failures** — `the drop was still in progress when the worker exited (.../rrd3/testhost/1.rrd)`
- barrier in place: 5 runs, 5 passes

It is timing-based, so it is sized with margin rather than tuned: removal takes the order of 100ms against the microseconds the parent needs to finish the drop and exit, and a slower machine widens that gap rather than narrowing it. If you would rather have a deterministic synchronisation point than a margin, say so — I could not find one that did not require the worker to cooperate with the test.

### The second worker's cached updates

You are right, and the comment I wrote there offered a false choice — "discard, or an rrdupdate failure and an error line" — when there is a third option. Those readings are the second worker's own, from a different channel, not duplicates of the first worker's. They are keyed on the old path but the files are now under the new one, so they can simply be written there.

`rrdcache_rename_host()` does that, sharing the walk with `rrdcache_drop_host()` rather than duplicating it.

### The unchecked rename

Fixed, with the distinction you asked for. `ENOENT` with the destination present stays a success — that is the peer worker having got there first, and the end state is the one we wanted, so the locator should still be told. Everything else (`ENOTEMPTY`, `EACCES`, …) is reported and leaves locator state alone.

Suite 58 passed, 0 skipped, 0 failed; shellcheck clean on the test.

One aside, since it cost me two debugging rounds and is the same class of bug: my first version of the new assertion used `find … | head -1`, which SIGPIPEs `find` under `set -o pipefail`, fails the assignment, and kills the script with `set -e` and no output at all. It is `sed -n 1p` now — the same reason your `find`/`wc` change in #341 was right.

SoundGoof (approved):
lgtm

PR #265, opened 2026-07-20
showgraph: pass exec-title arguments as literal shell words

bonomani:
## What

`showgraph.cgi` supports dynamic graph titles via a graphs.cfg
`TITLE exec:<command>` directive: the command runs and its first output
line becomes the graph title. The command line is assembled for `popen()`
from the admin's command plus several request- and filename-derived
values — the graph's `displayname` (the `disp=` CGI parameter), the
service name, the legend, and the matched RRD filenames.

Those values were interpolated without quoting, so shell-active
characters in them were interpreted by the shell rather than passed
through as data.

## Change

Every value appended after the admin's (trusted) command is now
single-quote-escaped, so each reaches the exec-title script as one
literal argument regardless of its contents. The command from graphs.cfg
still runs verbatim, so legitimate exec titles behave exactly as before.

## Test

Adds `tests/web/showgraph-exec-title.sh`, which drives the real CGI with
values containing `$(...)`, double quotes and backticks and asserts each
is delivered to the script verbatim with no shell evaluation. Full suite
green on this branch.

### DIFF
--- a/tests/web/showgraph-exec-title.sh
+++ b/tests/web/showgraph-exec-title.sh
@@ -1,0 +1,106 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/web/showgraph-exec-title.sh
+#
+# A graphs.cfg "TITLE exec:<cmd>" runs <cmd> with the graph's displayname,
+# service and matched RRD filenames as arguments, and uses its first line
+# of output as the graph title. The displayname comes from the "disp="
+# CGI parameter, so those arguments must be passed to the script as
+# literal words, never interpreted by the shell. This test drives the
+# real showgraph CGI with values containing shell-active characters and
+# asserts they reach the script verbatim, with no shell evaluation.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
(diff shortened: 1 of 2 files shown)


SoundGoof:
The implementation looks correct, but could you add a regression case containing an embedded single quote to showgraph-exec-title.sh?

The current test covers $(), double quotes, and backticks, while single quotes exercise the dedicated '\'' branch in shellquote().

I tested a value such as O'Brien;touch .../SIDEEFFECT locally and it passes, but that behavior should be preserved by the committed test suite.

bonomani:
Added — the committed test now has a single-quote case.

One note on the value: I built the pre-fix `showgraph.c` and ran `O'Brien;touch .../SIDEEFFECT` through it, and it passes there too — inside the old `"%s"` interpolation both the quote and the `;` are inert. So that value would have been green before and after the fix.

The committed case is `a'$(touch .../SIDEEFFECT)'b` instead. It covers what you asked for and a bit more: it fails against the pre-fix code (the command substitution runs), and it also fails if `shellquote()` wraps in single quotes without rewriting an embedded one as `'\''` — I verified both by deleting that branch and re-running. The three existing assertions also moved to `grep -qxF --`, since the values are data rather than patterns.

SoundGoof (approved):
lgtm

PR #270, opened 2026-07-23
xymond_history: do not write a closing hist record with an empty color

### LINKED ISSUE #269: xymond_history writes corrupt hist records (missing color) when the hist file tail is unparseable
## Symptom

Reported by several users after migrating from Big Brother / upgrading (4.3.28 → 4.3.30): the history CGI shows incomplete or no entries, and the `hist/` files contain records **missing the color field**:

```
Sat Apr  4 09:27:22 2020 1586017642 188529        <- corrupt (no color)
Sat Apr  4 09:27:22 2020 red 1586017642 188529    <- expected format
```

Thread: https://sourceforge.net/p/xymon/discussion/435278/thread/3479b2301a/

## Root cause

`xymond_history` writes the corrupt line itself. In the `save_statusevents` block (`xymond/xymond_history.c`):

1. The hist-file convention is that the **last line is an "open" record** (`timestamp color start-epoch`, no duration, no trailing newline). On the next color change, the writer scans the last 512 bytes for that open record (`sscanf(l+24, " %s %d %d") == 2` + valid color) so it can re-print it with the final duration.
2. If the scan finds nothing (`pos == -1`) — file left over from bbd whose tail is not in open-record form, an empty file created by a partial copy, or any unparseable tail — the fallback sets `lastchg = st.st_mtime` and leaves `oldcol` as the **empty string** it was initialized to.
3. The closing record is then written anyway:

```c
if (logexists) {
        ...
        fprintf(statuslogfd, "%s %s %d %d\n",
                timestamp, oldcol /* == "" */, (int)lastchg, (int)(tstamp - lastchg));
}
```

→ exactly the corrupt line above: empty color (double space), start = file mtime, duration = now − mtime.

The reader side (`lib/availability.c:get_historyline()`) requires a valid color and **silently skips** such lines (`dbgprintf` only), so the affected period disappears from the history page — or, when records end up merged onto an unterminated line, whole stretches of history vanish. The events page still works because it uses a different write path (hostlog/allevents), which matches the reports in the thread.

## Reproduction

```sh
mkdir -p /tmp/ht/hist /tmp/ht/logs /tmp/ht/histlogs
: > /tmp/ht/hist/skeletor,feld,me.foo                  # existing but empty hist file
touch -d '2020-04-04 09:27:22' /tmp/ht/hist/skeletor,feld,me.foo
printf '@@stachg#1|1586206051.00000|127.0.0.1|xymond|skeletor.feld.me|foo|1586206351|red|green|1586205951|0||0|0|\nbody\n@@\n' > /tmp/ht/msg
env XYMONSERVERLOGS=/tmp/ht/logs XYMONHISTLOGS=/tmp/ht/histlogs \
    ./xymond/xymond_history --histdir=/tmp/ht/hist < /tmp/ht/msg
cat -A /tmp/ht/hist/skeletor,feld,me.foo
```

Result (current main):

```
Sat Apr  4 09:27:22 2020  1585985242 220809$          <- empty color, mtime-based, bogus duration
Mon Apr  6 22:47:31 2020 red 1586206051
```

## Workaround (existing installations)

1. Find corrupt records:
   ```sh
   awk '$6 !~ /^(green|yellow|red|blue|purple|clear)$/' $XYMONVAR/hist/HOST.SERVICE
   ```
2. Repair them — insert the missing color (the color of the period that ended there = the previous record's color; the fabricated records have no matching `histlogs/` entry) and split any glued records. One-liner, run **with xymond_history stopped**; every file gets a `.bak` backup:
   ```sh
   cd $XYMONVAR/hist && for f in *.*; do cp -p "$f" "$f.bak" && sed -E 's/([0-9])((Sun|Mon|Tue|Wed|Thu|Fri|Sat) [A-Z][a-z]{2} )/\1\n\2/g' "$f.bak" | awk '{n=split($0,f," "); if(n>=6 && f[6]~/^(green|yellow|red|blue|purple|clear)$/){prev=f[6]; print} else if((n==6||n==7) && f[6]~/^[0-9]+$/){printf "%s %s %s%s\n", substr($0,1,24), (prev!=""?prev:"clear"), f[6], (n==7?" "f[7]:"")} else print}' > "$f"; done
   ```
   The `sed` splits glued records (digit running straight into a weekday name); the `awk` rebuilds any record whose 6th field is numeric (= color missing) as `<original 24-char timestamp> <color> <epoch> [<duration>]`, preserving the fixed timestamp column that the readers parse by offset. The trailing newline it leaves on the open last record is harmless (the writer's tail scan parses it identically and fully overwrites it on the next status change).
3. On an unpatched server, a file whose last record is *completed* (e.g. straight from a BB migration) re-triggers the corruption on the next status change — re-run the repair afterwards, or apply the fix below first.

## Fix

Only write the closing record when the color of the period being closed is actually known:

```c
if (logexists && *oldcol) {
```

With the fix, the unparseable-tail case simply appends a fresh open record; migrated/partial files stay intact. PR follows.

## Note on the devel branch

The rewritten block on `devel` has the same flaw plus additional ones (the new-file guard was dropped, so every freshly created hist file starts with an empty-color record; `write(fd, newrec, strlen(newrec)+1)` embeds a NUL byte in the file; the `pread` buffer is used unterminated). That deserves a separate issue once this one is settled.

bonomani:
Fixes #269

## Problem

When a status change arrives and the tail of the existing hist file contains no *valid open record* (empty file from a partial copy, bbd/BB-migrated file, torn last line, or any garbage tail), the fallback path in the `save_statusevents` block writes corrupt records. Three variants, all from the same scan:

1. **Empty color** — scan matched nothing, `oldcol` still `""`:
   ```
   Sat Apr  4 09:27:22 2020  1585985242 220809
   ```
2. **Garbage color / negative duration** — `sscanf` filled `oldcol` as a side effect of a line it *rejected*:
   ```
   Thu Jul 23 07:49:06 2026 :31 1784785746 -198576095
   ```
3. **Glued records** — with an unterminated last line, the appended record starts mid-line, is unreadable, and re-triggers the failure (and further gluing) on every subsequent status change.

`lib/availability.c:get_historyline()` silently skips all of these, so the history CGI shows holes — or nothing. Full analysis in #269.

## Fix (three parts, same code block)

1. Only write the closing record when the color of the period being closed is known: `if (logexists && *oldcol)`.
2. Only copy the scanned color into `oldcol` when the record fully validates (scan into a temp buffer, bounded with `%99s` so an overlong garbage token cannot overflow it) — kills the garbage-color/negative-duration variant.
3. When falling back to append-at-EOF, terminate an unterminated last line first, so the appended record starts on a line of its own and the file re-syncs.

**Intentional behavior change:** for a bbd-migrated file (tail = completed, newline-terminated record), the old code fabricated a closing record from side-effect state (last record's color spread across an mtime-based timespan). Now nothing is fabricated: the migrated history is left untouched and the new open record starts cleanly. The gap between migration and first Xymon status is simply unrecorded — which is what every reader already saw, since the fabricated record was frequently malformed.

## How to test (before/after)

Build `xymond/xymond_history`, then, with two status-change messages:

```sh
mkdir -p /tmp/ht/hist /tmp/ht/logs /tmp/ht/histlogs
printf '@@stachg#1|1586206051.00000|127.0.0.1|xymond|junk.host|disk|1586206351|red|green|1586205951|0||0|0|\nbody\n@@\n@@stachg#2|1586209651.00000|127.0.0.1|xymond|junk.host|disk|1586209951|yellow|red|1586206051|0||0|0|\nbody\n@@\n' > /tmp/ht/msg2x
RUN='env XYMONSERVERLOGS=/tmp/ht/logs XYMONHISTLOGS=/tmp/ht/histlogs ./xymond/xymond_history --histdir=/tmp/ht/hist'
```

**Case A — unterminated garbage tail (+2 changes):**
```sh
printf 'xx yy zz' > /tmp/ht/hist/junk,host.disk; touch -d '2020-04-04 09:27:22' /tmp/ht/hist/junk,host.disk
$RUN < /tmp/ht/msg2x; cat -A /tmp/ht/hist/junk,host.disk
```
Before: `xx yy zzMon Apr  6 ... red 1586206051Thu Jul 23 ... :31 1784785746 -198576095` — glued + garbage record, repeats forever.
After:
```
xx yy zz
Mon Apr  6 22:47:31 2020 red 1586206051 3600
Mon Apr  6 23:47:31 2020 yellow 1586209651
```

**Case B — existing but empty hist file (+1 change):**
Before: `Sat Apr  4 09:27:22 2020  1585985242 220809` (empty color) + open record. After: the open record only.

**Case C — regression, normal file ending with a proper open record (+1 change):**
Output byte-identical before and after (record closed with final duration, new open record appended).

**Case D — bbd-style file, completed newline-terminated tail (+2 changes):**
```sh
printf 'Wed Apr  1 08:00:00 2020 green 1585724400 261622\n' > /tmp/ht/hist/bb,host.disk
```
Before: fabricated `... green <mtime> <gap>` record inserted. After:
```
Wed Apr  1 08:00:00 2020 green 1585724400 261622
Mon Apr  6 22:47:31 2020 red 1586206051 3600
Mon Apr  6 23:47:31 2020 yellow 1586209651
```
(migrated history untouched, subsequent operation normal).

All four cases were run against both binaries; outputs above are actual results.

### DIFF
--- a/tests/server/history-tail-resync.sh
+++ b/tests/server/history-tail-resync.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/server/history-tail-resync.sh
+#
+# Guard for the hist-file tail recovery in xymond_history
+# (xymon-monitoring/xymon#269).
+#
+# When the tail of a hist file holds a valid open record followed by a
+# malformed suffix (torn write, partial copy), xymond_history seeks back to
+# the open record and rewrites it closed. The rewrite must also TRUNCATE the
+# file after the newly written open record: without that, the stale suffix
+# survives past the (shorter) rewrite, ends up glued to the new open record,
+# and is carried forward onto every subsequent status change - the file never
+# re-synchronizes.
+#
+# Drives the real worker over a hist file seeded with an open record plus a
+# 300-byte garbage suffix (inside the 512-byte tail scan window) and two
+# consecutive status changes, then asserts the suffix is gone and every
+# record is well-formed. Skips when the worker binary has not been built.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+ROOT=$(find_root)
+BIN="$ROOT/xymond/xymond_history"
+
+[ -x "$BIN" ] || skip "xymond/xymond_history not built"
+
+WORK=$(mktempdir)
+mkdir -p "$WORK/hist" "$WORK/logs" "$WORK/histlogs"
+HISTFILE="$WORK/hist/junk,host.disk"
+
+# Valid open record, then a malformed suffix shorter than the scan window.
+printf 'Mon Apr  6 22:37:31 2020 green 1586205451\n' > "$HISTFILE"
+head -c 300 /dev/zero | tr '\0' 'G' >> "$HISTFILE"
+
+# Two consecutive status changes: green->red, then red->yellow.
+printf '@@stachg#1|1586206051.00000|127.0.0.1|xymond|junk.host|disk|1586206351|red|green|1586205451|0||0|0|\nbody\n@@\n@@stachg#2|1586209651.00000|127.0.0.1|xymond|junk.host|disk|1586209951|yellow|red|1586206051|0||0|0|\nbody\n@@\n' \
+	> "$WORK/msg2x"
+
+env XYMONSERVERLOGS="$WORK/logs" XYMONHISTLOGS="$WORK/histlogs" \
+	"$BIN" --histdir="$WORK/hist" < "$WORK/msg2x" > /dev/null 2>&1
+
+result=$(cat "$HISTFILE")
+
+# The stale suffix must not survive - not glued to a record, not anywhere.
+assert_not_contains "G" "$result" \
+	"stale tail bytes survived the seek-back rewrite (#269): file did not re-sync"
+
+# Exactly the three expected records: two closed (with the durations fixed by
+# the message timestamps, so timezone-independent) and the final open record.
+# The open record must be the last thing in the file - nothing after it.
+assert_match " green 1586205451 600
+.* red 1586206051 3600
+.* yellow 1586209651$" "$result" \
+	"hist file does not hold the two closed records plus a clean final open record (#269)"
+
+pass "xymond_history truncates a stale tail after seek-back rewrite (#269)"
--- a/xymond/xymond_history.c
+++ b/xymond/xymond_history.c
@@ -299,6 +299,7 @@ int main(int argc, char *argv[])
 					 */
 					off_t pos = -1;
 					char l[1024];
+					char scancol[100];
 					int gotit;
 
 					MEMDEFINE(l);
@@ -323,14 +324,15 @@ int main(int argc, char *argv[])
 						if (fgets(l, sizeof(l)-1, statuslogfd)) {
 							/* Sun Oct 10 06:49:42 2004 red   1097383782 602 */
 
-							if ((strlen(l) > 24) && 
-							    (sscanf(l+24, " %s %d %d", oldcol, &lastchg_i, &dur_i) == 2) &&
-							    (parse_color(oldcol) != -1)) {
-								/* 
+							if ((strlen(l) > 24) &&
+							    (sscanf(l+24, " %s %d %d", scancol, &lastchg_i, &dur_i) == 2) &&
+							    (parse_color(scancol) != -1)) {
+								/*
 								 * Record the start location of the line
 								 */
 								pos = tmppos;
 								lastchg = lastchg_i;
+								strncpy(oldcol, scancol, sizeof(oldcol));
 							}
 						}
 						else {
@@ -339,13 +341,27 @@ int main(int argc, char *argv[])
 					}
 
 					if (pos == -1) {
-						/* 
+						/*
 						 * Couldnt find anything in the log.
 						 * Take lastchg from the timestamp of the logfile,
 						 * and just append the data.
 						 */
 						lastchg = st.st_mtime;
 						fseeko(statuslogfd, 0, SEEK_END);
+						if (st.st_size > 0) {
+							char lastbyte;
+
+							/*
+							 * If the last line is unterminated garbage, terminate it
+							 * so the record we append below starts on a line of its own.
+							 */
+							fseeko(statuslogfd, -1, SEEK_END);
+							if ((fread(&lastbyte, 1, 1, statuslogfd) == 1) && (lastbyte != '\n')) {
+								fseeko(statuslogfd, 0, SEEK_END);
+								fputc('\n', statuslogfd);
+							}
+							fseeko(statuslogfd, 0, SEEK_END);
+						}
 					}
 					else {
 						/*
--- a/xymond/xymond_history.c
+++ b/xymond/xymond_history.c
@@ -387,7 +387,7 @@ int main(int argc, char *argv[])
 				}
 
 				if (statuslogfd) {
-					if (logexists) {
+					if (logexists && *oldcol) {
 						struct tm oldtm;
 
 						/* Re-print the old record, now with the final duration */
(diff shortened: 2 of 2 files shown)


SoundGoof:
## Review: Changes requested

### Finding

**[P1] Bound the color scan to the destination buffer**

`xymond/xymond_history.c:328`

`scancol` is 100 bytes, but `%s` is unbounded while `l` can contain approximately 1,022 bytes. A malformed history tail containing 100 or more non-whitespace characters therefore overflows the stack buffer in the corruption-recovery path this PR is hardening.

I reproduced the overwrite using the same buffer sizes: a marker immediately following `scancol` changed from `0x12345678` to `0x41414141`.

Limit the conversion:

    sscanf(l+24, " %99s %d %d", scancol, &lastchg_i, &dur_i)

This also guarantees that `scancol` is terminated before it is copied to `oldcol`.

### Other observations

The remaining changes appear correct:

- Rejected records no longer leak partially parsed colors into `oldcol`.
- Empty and completed-history tails no longer produce fabricated closing records.
- Unterminated tails are separated from newly appended records.
- Valid open records retain the previous closing behavior.
- Later changes on `main` do not modify this file.

### Validation

- All four CI jobs pass.
- `git diff --check` passes.
- The PR is mergeable without conflicts.
- The overflow was reproduced with a compiler-built marker test.

bonomani:
Fixed in afb733b: the tail-scan conversion is now bounded (`%99s` into `scancol[100]`), so an overlong garbage token can no longer overflow the buffer, and `scancol` is guaranteed NUL-terminated before the copy to `oldcol`. The branch is also squashed to a single commit.

SoundGoof:
## Findings

### [P2] Truncate stale bytes after rewriting the recovered open record

[`xymond/xymond_history.c:366-371`](https://github.com/xymon-monitoring/xymon/blob/afb733bd89355e19cc8410c440166ab4d5be0a34/xymond/xymond_history.c#L366-L371)

When the scan finds a valid open record followed by a malformed tail, `pos` points to the open record. The code seeks there and writes the closed record plus the new open record, but never truncates the old suffix.

For example, start with:

```text
Mon Apr  6 22:37:31 2020 green 1586205451
<300 bytes of garbage>
```

After one patched update, the file contains:

```text
Mon Apr  6 22:37:31 2020 green 1586205451 600
Mon Apr  6 22:47:31 2020 red 1586206051<remaining garbage>
```

A second update moves the same garbage onto the new yellow open record. The file therefore never re-synchronizes, despite malformed-tail recovery being part of this change.

Truncate the file at the end of the newly written open record. Add a real-worker regression case with a valid open record followed by a malformed suffix shorter than the 512-byte scan window, and assert that no suffix remains after consecutive updates.

## Validation

- The commit cherry-picks cleanly onto current `main` (`eea531bd`).
- Empty-file, garbage-only, completed-record, and normal-open cases behave as described.
- Unpatched main reproduces the empty-color, glued-record, and fabricated-duration corruption.
- Full suite: 30 passed, 3 unrelated build-dependent skips, 0 failed.
- Patch whitespace check passed.

bonomani:
Fixed in 1278423b6 (branch re-squashed to a single commit) — both parts of the finding addressed, CI green.

**Truncation**: after writing the new open record, the file is flushed and truncated at the current position. This drops the stale suffix when the seek-back rewrite lands inside the file; in the append-at-EOF and fresh-file paths the position is already EOF, so it's a no-op there.

**Regression test**: `tests/server/history-tail-resync.sh` drives the real worker over a valid open record followed by a 300-byte malformed suffix (inside the 512-byte scan window) and two consecutive status changes, asserting no suffix survives and that exactly the two closed records plus a clean final open record remain (assertions key on the epoch/duration fields, so they're timezone-independent). It fails against the unfixed binary — reproducing exactly the carry-forward you described, garbage glued to the final open record — and passes against the fixed one. Skips with exit 77 when the worker isn't built, per suite convention.

Also re-ran the PR's original cases A–D against the fixed binary: outputs are byte-identical to what the PR description documents.

SoundGoof (approved):
## Review

No blocking issues found. The history-tail recovery behaves correctly across the tested malformed, empty, migrated, and oversized-tail cases.

The unchecked write-error path is not a practical regression for this change and does not warrant blocking it.

## Validation

- Rebases cleanly onto current `main`.
- Server build succeeded.
- Full suite: 35 passed, 3 skipped, 0 failed.
- Focused history-tail test passed.
- All GitHub checks are green.

PR #272, opened 2026-07-25
docs: add HTTP content check examples

### LINKED ISSUE #231: hosts.cfg.5 - needs http cont examples
HTTP response content checks are missing even a single example showing the correct regex syntax

SoundGoof:
## Summary

Adds practical HTTP response-content check examples to `hosts.cfg.5`.

- Shows a complete `cont;URL;REGEX` tag using `[[:space:]]`.
- Clarifies that the regular expression follows the second semicolon directly and does not use surrounding slash delimiters.
- Demonstrates assigning a content check to a custom status column.
- Adds an advanced POSIX extended-regex example using grouping and alternation.

Closes #231.

## Testing

- Rendered `common/hosts.cfg.5` successfully with `groff`.
- Built and ran the real `xymonnet` against local HTTP fixtures.
- Confirmed all three documented examples produced green `Content OK` statuses.
- Confirmed the advanced `status=(ok|ready)` expression rejects `status=degraded` with a red `Content match failed` status.
- Verified the patch with `git diff --check`.

### DIFF
--- a/common/hosts.cfg.5
+++ b/common/hosts.cfg.5
@@ -1179,11 +1179,13 @@ For example, the complete content-check tag is:
 .br
 The regular expression follows the second semicolon directly and is
 written without surrounding slash delimiters.
-To report the result in a custom column and require an exact response:
+To report the result in a custom column and match "OK", optionally
+followed by a line ending:
 .br
-   cont=health;https://example.com/health;^OK$
+   cont=health;https://example.com/health;^OK[[:cntrl:]]*$
 .br
-This reports in the "health" column and matches only the response "OK".
+This reports in the "health" column and rejects additional printable text.
+For a byte-exact response check, use the message-digest form described above.
 For a more advanced match, grouping and alternation can accept multiple values:
 .br
    cont=api;https://example.com/api;status=(ok|ready)
--- a/common/hosts.cfg.5
+++ b/common/hosts.cfg.5
@@ -1173,6 +1173,23 @@ Note that this may depend on your particular implementation
 of the regex functions found in your C library. Thanks to 
 Charles Goyard for this tip.
 
+For example, the complete content-check tag is:
+.br
+   cont;https://example.com/;All[[:space:]]is[[:space:]]OK
+.br
+The regular expression follows the second semicolon directly and is
+written without surrounding slash delimiters.
+To report the result in a custom column and require an exact response:
+.br
+   cont=health;https://example.com/health;^OK$
+.br
+This reports in the "health" column and matches only the response "OK".
+For a more advanced match, grouping and alternation can accept multiple values:
+.br
+   cont=api;https://example.com/api;status=(ok|ready)
+.br
+This matches either "status=ok" or "status=ready" in the response.
+
 Note: If you are migrating from the "cont2.sh" script,
 you must change the '_' used as wildcards by cont2.sh 
 into '.' which is the regular-expression wildcard character.


bonomani:
Three of the four examples check out against the code. One does not work with the engine as it currently stands:

    cont=health;https://example.com/health;^OK$

Content regexes are compiled with `REG_EXTENDED|REG_NOSUB` (`xymonnet/httptest.c:475`). The body is accumulated into a single buffer as it arrives (`httptest.c:142-154`) and matched once (`httpresult.c:471`). With no `REG_NEWLINE`, `^` and `$` anchor to the ends of that whole buffer rather than to lines, so any trailing newline defeats the match.

Ran the real `xymonnet` against a local HTTP fixture:

| response body | `^OK$` | `^OK[[:cntrl:]]*$` |
|---|---|---|
| `OK\n` | red — Content match failed | green — Content OK |
| `OK\r\n` | red — Content match failed | green — Content OK |
| `OK` (no newline) | green — Content OK | green — Content OK |

Since `echo OK > health` — and every editor that adds a final newline — produces `OK\n`, the example as written fails for the common case.

Suggested replacement:

    cont=health;https://example.com/health;^OK[[:cntrl:]]*$

`[[:cntrl:]]` covers `\n`, `\r` and `\t` but not the space character, so this stays true to "exact response": it still rejects `OKAY`, `NOT OK` and a trailing `OK `. (`[[:space:]]*` also fixes the newline problem but accepts `OK `, which reads oddly next to the word "exact". A literal `\n` is not an option — whitespace terminates a hosts.cfg tag, and backslash-n is read as the letter `n`.)

It may be worth a word on which flavour of "exact" is meant, since the anchored regex forms are all exact-modulo-whitespace. The digest form already documented above at line 1140 — `cont;URL;#md5:...` via `xymondigest(1)` — is the byte-exact one.

To be clear, this is a documentation fix, not an engine bug. Setting `REG_NEWLINE` would also stop `.` and `[^...]` from crossing newlines, which would break existing multi-line `cont` patterns and silently change verdicts in both directions.

The rest is accurate: the regex really is everything after the second semicolon (`lib/url.c:560` splits at the first `;` after the URL), and `status=(ok|ready)` works unescaped because of `REG_EXTENDED`.

bonomani (approved):
Approving — all four examples now match what the engine does.

Re-checked the revised `^OK[[:cntrl:]]*$` against the real `xymonnet` and under `REG_EXTENDED|REG_NOSUB`, the flags `httptest.c:475` compiles with:

| response body | result |
|---|---|
| `OK` | match |
| `OK\n` | match |
| `OK\r\n` | match |
| `OK ` (trailing space) | no match |
| `OKAY` / `NOT OK` | no match |

Each part of it is load-bearing, for anyone tempted to shorten it later: `?` instead of `*` drops `OK\r\n` (two control characters); `.?` instead of the class lets `OKx` through; without `^` it accepts `NOT OK`; without `$` it accepts `OKAY`. `\n` is not an option — POSIX ERE reads it as the letter `n`.

The prose change is the part that matters most. "Rejects additional printable text" is exactly the guarantee this expression gives, where "matches only the response OK" promised an exactness it does not have, and pointing at the digest form for byte-exact checks gives the reader the right tool for the stricter case.

The other three examples check out too: the regex really is everything after the second semicolon (`lib/url.c:560` splits at the first `;` after the URL), `status=(ok|ready)` works unescaped thanks to `REG_EXTENDED`, and `[[:space:]]` is needed because whitespace would otherwise terminate the hosts.cfg tag.

`groff -man -z` clean.

PR #274, opened 2026-07-25
fix: initialize alpns in httptest.c SSL options to avoid uninitialized read

SoundGoof:
## Summary

Fixes a segfault in `xymonnet` on HTTPS checks that set a custom cipher list, SSL version, or client certificate (e.g. `https://CERT:client.pem@host/path`).

## Root cause

`add_http_test()` in `xymonnet/httptest.c` `malloc()`s a per-request `ssloptions_t` but never initializes the `alpns` field (added by 75cfe50b for ALPN support). `setup_ssl()` in `contest.c` reads that garbage value as a fallback ALPN string and `strdup()`s it, segfaulting on invalid memory.

Pre-existing on `main`, not introduced by this branch — just never hit until a client-cert HTTPS test exercised the path.

## Fix

Initialize `sslopt->alpns = NULL;` alongside the other fields.

## Verification

Reproduced under gdb (`item->ssloptions->alpns` held `0x1`); confirmed the crash is gone after the fix.

### DIFF
--- a/xymonnet/httptest.c
+++ b/xymonnet/httptest.c
@@ -675,11 +675,10 @@ void add_http_test(testitem_t *t)
 
 	/* Pickup any SSL options the user wants */
 	if (sslopt_ciphers || (sslopt_version != SSLVERSION_DEFAULT) || sslopt_clientcert){
-		sslopt = (ssloptions_t *) malloc(sizeof(ssloptions_t));
+		sslopt = (ssloptions_t *) calloc(1, sizeof(ssloptions_t));
 		sslopt->cipherlist = sslopt_ciphers;
 		sslopt->sslversion = sslopt_version;
 		sslopt->clientcert = sslopt_clientcert;
-		sslopt->alpns = NULL;
 	}
 
 	/* Add to TCP test queue */
--- a/xymonnet/httptest.c
+++ b/xymonnet/httptest.c
@@ -679,6 +679,7 @@ void add_http_test(testitem_t *t)
 		sslopt->cipherlist = sslopt_ciphers;
 		sslopt->sslversion = sslopt_version;
 		sslopt->clientcert = sslopt_clientcert;
+		sslopt->alpns = NULL;
 	}
 
 	/* Add to TCP test queue */


bonomani (dismissed):
Superseded — see the comment below for the current position (short version: the diagnosis holds, and `calloc()` is the form I would take rather than the explicit `= NULL`). This review is left dismissed; its original text filed that as an optional nit, which undersold it.

bonomani:
I dismissed my own approval — not because the fix is wrong, but because I filed the `calloc()` remark as an optional nit when it is actually the better change, and that is worth settling before this lands rather than after.

The diagnosis is confirmed. `ssloptions_t` has exactly four members, `malloc()` at `httptest.c:678` is followed by three assignments, and `alpns` was the one left holding whatever the allocator returned. `contest.c:575-576` then tests it and passes it to `strdup()`, so a non-NULL garbage pointer is dereferenced as a string. It is also the **only** allocation site for that struct in the tree, so there is no second instance hiding elsewhere.

What changed my mind on the form:

```diff
-		sslopt = (ssloptions_t *) malloc(sizeof(ssloptions_t));
+		sslopt = (ssloptions_t *) calloc(1, sizeof(ssloptions_t));
```

1. **It fixes the class, not the instance.** This bug exists because `alpns` was added to the struct by 75cfe50b6 and this allocation was not updated to match. `alpns = NULL` fixes today's field; a zeroing allocator fixes the next one too. Given the struct has already grown once, it will likely grow again.

2. **Zero happens to be the correct default for the other members**, which was not obvious until checked: `SSLVERSION_DEFAULT` is `0` (`contest.h:60`). All three are reassigned immediately, so nothing changes today — but if a future edit stops assigning `sslversion`, `calloc()` leaves the right value where `malloc()` leaves noise.

3. Smaller diff, and with a single allocation site the change is trivially auditable.

So my recommendation is `calloc()`. The explicit `= NULL` is not wrong — it is correct and complete for the bug as reported — but it is the weaker of the two for the reasons above, and I would rather say that plainly than leave it as a footnote to an approval.

bonomani (approved):
lgtm

PR #275, opened 2026-07-26
fix: check SSL_CTX_set_cipher_list() return value in setup_ssl

SoundGoof:
## Summary

Fixes a silent SSL misconfiguration in `xymonnet`: forcing a cipher-strength restriction (e.g. the `m`/`httpsm` scheme-suffix dialect from hosts.cfg(5)) can silently have no effect instead of restricting the connection or failing.

## Root cause

`setup_ssl()` in `xymonnet/contest.c` calls `SSL_CTX_set_cipher_list()` but never checks its return value:

```c
if (item->ssloptions->cipherlist) 
    SSL_CTX_set_cipher_list(item->sslctx, item->ssloptions->cipherlist);
```

On failure, OpenSSL leaves the SSL_CTX's previous cipher list untouched rather than clearing it — so the connection silently proceeds without the requested restriction, indistinguishable from success.

Confirmed on OpenSSL 3.0.13: `openssl ciphers MEDIUM` errors with "no cipher match" (that cipher class has been removed entirely), so any test using `httpsm` was actually connecting with the default cipher list, not a medium-strength one.

## Fix

Check the return value and treat failure the same as the existing SSL-context-creation error a few lines above (report via `errprintf`, mark the test as an SSL error) instead of silently continuing.

## Verification

Ran the full loopback test suite in a container: before the fix, a test forcing `MEDIUM` ciphers reported green (false success); after the fix, it correctly reports red ("SSL error"), while `HIGH` (which does match ciphers) still reports green.

### DIFF
--- a/RELEASENOTES
+++ b/RELEASENOTES
@@ -1,5 +1,5 @@
           <<<<<<<<<<<<<<<<<<<<<<<<>>>>>>>>>>>>>>>>>>>>>>>>
-           * * *   Release notes for Xymon 4.3.30   * * *
+           * * *   Release notes for Xymon 4.3.31   * * *
           <<<<<<<<<<<<<<<<<<<<<<<<>>>>>>>>>>>>>>>>>>>>>>>>
 
 
@@ -8,6 +8,16 @@ changes you should be aware of when upgrading.
 
 For a full list of changes and enhancements, please see the 
 Changes file.
+
+
+Changes for 4.3.31
+==================
+
+xymonnet tests with a configured cipher-list restriction now report an SSL error
+when OpenSSL rejects the restriction, instead of silently using the default
+cipher list. Such tests may therefore turn red after upgrading Xymon, or after
+an OpenSSL upgrade invalidates a previously accepted restriction. xymonnet tests
+without a cipher-list restriction are unaffected.
 
 
 Changes for 4.3.30
--- a/xymonnet/contest.c
+++ b/xymonnet/contest.c
@@ -564,8 +564,16 @@ static void setup_ssl(tcptest_t *item)
 		SSL_CTX_set_quiet_shutdown(item->sslctx, 1);
 
 		/* Limit set of ciphers, if user wants to */
-		if (item->ssloptions->cipherlist) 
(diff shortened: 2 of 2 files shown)


bonomani (dismissed):
Superseded — see the approval below.

bonomani (approved):
Approving. No regression: the patch only adds a check on a path that previously did nothing, and behaviour with a valid `cipherlist` is unchanged.

Verified rather than assumed — measured against OpenSSL 3.0.2:

```
  HIGH:!aNULL                     ret=1
  ECDHE-RSA-AES128-GCM-SHA256     ret=1
  TLS_AES_128_GCM_SHA256          ret=0   no cipher match
  CE-NEST-PAS-UNE-SUITE           ret=0   no cipher match
  ""                              ret=0   no cipher match
```

A rejected list also populates the OpenSSL error queue, so `ERR_error_string(ERR_get_error(), ...)` yields a real reason rather than `error:00000000` — the error path as written works.

Two notes, neither blocking.

**Worth a line in the release notes.** This turns a silent pass into a red test, and the trigger is not only a typo. `TLS_AES_128_GCM_SHA256` above is a perfectly legitimate suite name — TLS 1.3 suites are configured through `SSL_CTX_set_ciphersuites()`, not this call. Anyone who put one in `cipherlist` has a setting that quietly did nothing until now and goes red the moment this lands. That is the correct outcome, and it is also the one that will generate the support question. Worth noting that the diagnosis is not buried: the shipped `tasks.cfg` runs `xymonnet --report`, so the message lands on xymonnet's own column and turns it yellow, cipher string and OpenSSL reason included.

**A nit for whenever you are next in this function:** the new path returns without `SSL_CTX_free(item->sslctx)`, and `socket_shutdown()` only frees the context while `item->sslrunning` is set — which this path clears first. In practice it costs a few KB in a process that `xymonlaunch` re-runs every five minutes, so it is hygiene rather than a problem. Two existing paths in the same function do the same thing (the client-certificate load failure and the key mismatch), while `SSL_new` at `:646` and `SSL_set_fd` at `:683` free correctly.

PR #276, opened 2026-07-29
fix: keep hosts.cfg noflap= lists intact across status updates

SoundGoof:
## Summary

`noflap=test1,test2,...` in `hosts.cfg` silently stopped suppressing flap detection for every test except the first one evaluated. The bare `noflap` flag form was never affected.

## Root cause

`xmh_item()` for `XMH_NOFLAP` returns a pointer **into the host record's own tag buffer**, not a copy. `isset_noflap()` handed it straight to `strtok()`, which overwrote the list's first comma with a `NUL` — truncating the stored tag permanently. Since `isset_noflap()` runs once per incoming status message, the first evaluation destroyed the list for every later one:

```
before any call    raw tags: [conn] [noflap=web,cpu,disk]
isset_noflap(web)  = 1
after first call   raw tags: [conn] [noflap=web]        <-- truncated in place
isset_noflap(cpu)  = 0
isset_noflap(disk) = 0
```

Whichever test reported first kept suppression; the rest quietly lost it until the next `hosts.cfg` reload.

## Fix

Tokenize a copy (`strdup` / `xfree`) instead of the live buffer. Matching semantics, return values and the debug log are unchanged.

This is the established pattern — four of the six places that tokenize an `xmh_item()` result already copy first: `convertnk.c` (`XMH_NK`), `xymond_client.c` and `xymongen/loaddata.c` (`XMH_NOCOLUMNS`), `lib/webaccess.c` (`XMH_ALLPAGEPATHS`). That's what makes this an oversight rather than a design choice.

## Testing

`tests/server/xymond-noflap-list.sh`, in two halves since `isset_noflap()` is `static` and xymond has no library form to link against:

1. a source assertion that `xymond.c` still copies before tokenizing — the actual guard on the real file, matching the existing style of `digest-md5hash.sh`;
2. a C harness with an identical copy of the function, driven against real host records through the real `lib/loadhosts.c`.

Verified failing before the change (`exit=1`) and passing after (`exit=0`). Full suite: 18 passed, 1 skipped (RRD not built), 0 failed.

Deliberate in the test: the list assertions share one host and run in sequence, because the defect only appeared on the *second* and later evaluations — a fresh host per test name would have passed against the bug.

## Scope / notes

- The `strdup()` NULL check deviates slightly from the precedent sites. `XYMON_MEMORY_WRAPPERS` is `#undef`'d, so this is plain `strdup`, and on failure `strtok(NULL, ...)` would resume stale tokenizer state and return a wrong answer rather than crash.

### DIFF
--- a/tests/server/xymond-noflap-list-harness.c
+++ b/tests/server/xymond-noflap-list-harness.c
@@ -12,14 +12,10 @@
  * stored tag permanently -- so every later status update saw only the first
  * test.
  *
- * Why this mirrors the function instead of calling it: isset_noflap() is
- * static in xymond/xymond.c and xymond has no library form, so it cannot be
- * linked here. The copy below is kept identical to the real one, and the
- * companion shell test asserts at the source level that xymond.c still
- * copies before tokenizing -- that source assertion, not this file, is what
- * stops the two drifting apart. What this file adds is proof that the
- * copy-first logic actually produces the right answers on real host records
- * loaded through the real lib/loadhosts.c.
+ * The companion shell test prepends xymond/xymond.c's real static
+ * isset_noflap() implementation when it builds the harness. This file adds
+ * only the assertions that drive it against real host records loaded through
+ * the real lib/loadhosts.c.
  *
  * The list assertions deliberately share one host and run in sequence: the
  * old defect only appeared on the SECOND and later evaluations, because the
@@ -27,12 +23,6 @@
  * where isset_noflap() runs once per incoming status message. A test using
  * a fresh host per test name would have passed even before the fix.
  */
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
-
-#include "libxymon.h"
-
 static int failures = 0;
 
 static void expect(const char *label, int cond, const char *detail)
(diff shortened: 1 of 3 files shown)


bonomani:
Fix is correct. Two things in the test.

**1. Drop `-lssl -lcrypto`.** The harness links without them, and they turn `--no-ssl` builds and macOS/Homebrew (keg-only `openssl@3`, missing the `-L` from `configure.server:513`) into `fail` rather than `skip`. Pattern from `tests/web/showgraph-exec-title.sh:28`:

```diff
+ssllibs=$(sed -n 's/^SSLLIBS *= *//p' "$ROOT/Makefile")
+
 "$CC" -I"$ROOT/include" -I"$ROOT/lib" -o "$work/harness" \
 	"$here/xymond-noflap-list-harness.c" "$ROOT/lib/libxymoncomm.a" \
-	-lssl -lcrypto 2>"$work/cc.log" \
+	$ssllibs 2>"$work/cc.log" \
 	|| { cat "$work/cc.log" >&2; fail "harness does not compile"; }
```

**2. Neither half of the test runs the real function.** The shell half searches the text of `xymond.c` for `strdup(`; the C half tests a hand-written copy of `isset_noflap()`. If the real function is edited later and the bug comes back, the copy does not change and the assertions still pass. The text search is meant to catch that drift, but renaming `dstr` to `p` satisfies `assert_not_contains 'strtok(dstr'` with the bug back — and `assert_contains 'strdup('` would reject a correct fix using `strtok_r` on a stack buffer.

The test already awk-extracts the function body; compiling that instead removes both problems, and the copy with them:

```sh
{
	printf '#include <stdio.h>\n#include <stdlib.h>\n#include <string.h>\n'
	printf '#include "libxymon.h"\n'
	awk '/^static int isset_noflap/,/^}/' "$XYMOND_C"
	cat "$here/xymond-noflap-list-harness.c"
} > "$work/harness.c"
```

Then delete the copied function from the harness (keep `expect()` and `main()`), compile `"$work/harness.c"`, and drop both source assertions. The extracted body needs only `xmh_item`, `XMH_NOFLAP`, `strdup`, `xfree` and `dbgprintf`, all in `libxymon.h`.

bonomani (approved):
lgtm

PR #277, opened 2026-07-29
fix: register XMH_FLAG_MULTIHOMED under its documented name

SoundGoof:
## Summary

`xmh_item_name[XMH_FLAG_MULTIHOMED]` was registered as `"XMH_MULTIHOMED"`, missing the `FLAG_` that the enum identifier and `xymon-xmh(5)` both carry. One-word fix in `lib/loadhosts.c`, plus a test.

That string is public: clients filter/select `xymondboard`/`hostinfo` fields by it, xymond emits it on the wire (`loadhosts_net.c` parses it back), and web templates resolve it via `xmh_item_byname()`.

## Consequences

- **The documented name didn't resolve.** `xymon-xmh(5)` documents `XMH_FLAG_MULTIHOMED` and never mentions `XMH_MULTIHOMED`, so a field selector, filter or template written from the manpage was silently rejected.
- **MULTIHOMED wasn't treated as a flag.** `xmh_item_isflag[]` is derived by prefix-matching `"XMH_FLAG_"` on the registered name, so `xmh_item()` returned `""` instead of the canonical key every other flag yields.

It was the only entry where enum and registered names disagreed — 21 vs 20; now 21/21.

The one in-tree consumer (`xymond.c`, multiple-source-IP warning) tests `== NULL`, and `""` was already non-NULL, so its behaviour is unchanged.

## History

A typo at birth, never since altered. Both halves landed the same day in the 4.3.0 merge:

- `e3f90f0e` (2011-03-08) added `xymon-xmh.5` documenting `XMH_FLAG_MULTIHOMED`
- `8b01109b` (2011-03-08) added `xmh_item_name[...] = "XMH_MULTIHOMED"`

`git blame` shows line 176 still on that original commit, and a pickaxe over all history finds only those two commits ever touching the string. So the code has disagreed with its own manpage for ~15 years, and no commit ever deliberately chose the short name — there's no historical intent to preserve.

It also means the documented name has never worked in any release, so `"XMH_MULTIHOMED"` was only ever discoverable by reading the source. 

## Testing

`tests/server/xmh-item-names.sh`, two halves:

1. **Source check** that each `xmh_item_name[XMH_X]` registers `"XMH_X"` — this is what catches the typo class, since a misnamed entry is self-consistent at runtime and no runtime assertion can see it.
2. **C harness** on the real library: documented name resolves, MULTIHOMED yields its key like the `dialup` control, present/absent still read non-NULL/NULL, and all names round-trip (catching duplicates).

Verified against `origin/main`: unfixed `exit=1`, fixed `exit=0`. Full build clean; suite 19 passed, 0 failed.

## Upgrade note

xymond now emits `XMH_FLAG_MULTIHOMED` in `hostinfo`. A mixed-version pair silently drops that one field (receivers ignore unresolvable `XMH_` keys); nothing else is affected, and nothing in-tree reads it over the network.

### DIFF
--- a/lib/loadhosts.c
+++ b/lib/loadhosts.c
@@ -173,7 +173,7 @@ static void xmh_item_list_setup(void)
 	xmh_item_key[XMH_NOFLAP]               = "NOFLAP";
 	xmh_item_name[XMH_NOFLAP]              = "XMH_NOFLAP";
 	xmh_item_key[XMH_FLAG_MULTIHOMED]      = "MULTIHOMED";
-	xmh_item_name[XMH_FLAG_MULTIHOMED]     = "XMH_MULTIHOMED";
+	xmh_item_name[XMH_FLAG_MULTIHOMED]     = "XMH_FLAG_MULTIHOMED";
 	xmh_item_key[XMH_FLAG_HTTP_HEADER_MATCH]             = "headermatch";
 	xmh_item_name[XMH_FLAG_HTTP_HEADER_MATCH]            = "XMH_FLAG_HTTP_HEADER_MATCH";
 	xmh_item_key[XMH_FLAG_SNI]             = "sni";			// Enable SNI (Server name Indication) for TLS requests
--- a/tests/server/xmh-item-names-harness.c
+++ b/tests/server/xmh-item-names-harness.c
@@ -1,0 +1,111 @@
+/* Regression test for the XMH_ item-name table in lib/loadhosts.c.
+ *
+ * xmh_item_name[] holds the public name of each host attribute. Those names
+ * are not internal labels - they are the field names clients use in
+ * xymondboard/hostinfo requests (xymond/xymond.c), the names xymond emits in
+ * hostinfo responses and lib/loadhosts_net.c parses back, and what web
+ * templates resolve via xmh_item_byname() (lib/headfoot.c). They are
+ * documented in xymon-xmh(5).
+ *
+ * XMH_FLAG_MULTIHOMED was registered as "XMH_MULTIHOMED", missing the FLAG_
+ * that its enum name and xymon-xmh(5) both carry. Two consequences:
+ *   - the documented name XMH_FLAG_MULTIHOMED did not resolve, so a
+ *     board/hostinfo field selector or template reference using it was
+ *     silently rejected;
+ *   - xmh_item_isflag[] is derived by prefix-matching "XMH_FLAG_" against
+ *     the registered name, so MULTIHOMED was never treated as a flag and
+ *     xmh_item() returned the empty remainder after the key instead of the
+ *     canonical key string that every other flag yields.
+ *
+ * Everything here drives the real lib/loadhosts.c through its public
+ * interface - no mirrored logic.
+ */
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "libxymon.h"
+
+static int failures = 0;
+
+static void expect(const char *label, int cond, const char *detail)
+{
+	if (!cond) {
+		fprintf(stderr, "%s: %s\n", label, detail);
+		failures++;
+	}
+}
+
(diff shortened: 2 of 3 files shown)


bonomani:
Fix is correct. Two things.

**1. Drop `-lssl -lcrypto`.** The harness links without them, and they turn `--no-ssl` builds and macOS/Homebrew (keg-only `openssl@3`, missing the `-L` from `configure.server:513`) into `fail` rather than `skip`. Pattern from `tests/web/showgraph-exec-title.sh:28`:

```diff
+ssllibs=$(sed -n 's/^SSLLIBS *= *//p' "$ROOT/Makefile")
+
 "$CC" -I"$ROOT/include" -I"$ROOT/lib" -o "$work/harness" \
 	"$here/xmh-item-names-harness.c" "$ROOT/lib/libxymoncomm.a" \
-	-lssl -lcrypto 2>"$work/cc.log" \
+	$ssllibs 2>"$work/cc.log" \
 	|| { cat "$work/cc.log" >&2; fail "harness does not compile"; }
```

**2. The source check misses the typo shape closest to this bug.** Its grep only matches assignments whose registered string already starts with `XMH_`, so anything that loses the prefix is never compared:

```
"XMH_FLAG_DIALUPX"  -> caught
"DIALUP"            -> not caught
"xmh_flag_dialup"   -> not caught
```

Match any string, then compare:

```diff
-mismatched=$(grep -o 'xmh_item_name\[XMH_[A-Z0-9_]*\][[:space:]]*=[[:space:]]*"XMH_[A-Z0-9_]*"' "$LOADHOSTS_C" \
-	| sed -E 's/xmh_item_name\[(XMH_[A-Z0-9_]*)\][[:space:]]*=[[:space:]]*"(XMH_[A-Z0-9_]*)"/\1 \2/' \
+mismatched=$(grep -o 'xmh_item_name\[XMH_[A-Z0-9_]*\][[:space:]]*=[[:space:]]*"[^"]*"' "$LOADHOSTS_C" \
+	| sed -E 's/xmh_item_name\[(XMH_[A-Z0-9_]*)\][[:space:]]*=[[:space:]]*"([^"]*)"/\1 \2/' \
 	| awk '$1 != $2 { printf "  %s registered as \"%s\"\n", $1, $2 }')
```

Catches all three variants, no false positives across the 72 entries, still catches `MULTIHOMED` on the unpatched file.

Nothing needed from you on `Changes` or `RELEASENOTES`: release documentation is prepared on the `Changes` branch by a maintainer, and the entries for this are already in place. (Disregard my earlier ask for one here.)

bonomani (approved):
Both review points are addressed, and I re-measured rather than taking them on trust.

**Link flags.** `$ssllibs` now comes from the build's own Makefile, so a `--no-ssl` build or a macOS/Homebrew tree no longer fails where it should skip.

**Source check.** The pattern accepts any registered string instead of only ones already starting with `XMH_`. Re-run against the three typo shapes: `"XMH_FLAG_DIALUPX"`, `"DIALUP"` and `"xmh_flag_dialup"` are all caught now — the last two were not. Clean on this branch's `loadhosts.c` across all 72 entries, and still catches `MULTIHOMED` on unpatched `main`.

**The fix itself**, probed against the real library before and after: `xmh_key_idx("XMH_FLAG_MULTIHOMED")` goes from `XMH_LAST` to resolving, and `xmh_item()` on a tagged host goes from `""` to `"MULTIHOMED"`, matching `dialup` and the other nineteen flags. An untagged host still reads NULL, so the one in-tree consumer (`xymond.c:1641`, testing `== NULL`) is unaffected.

Also worth recording: `xmh_item_isflag[]` has exactly one read site (`loadhosts.c:246`), so flipping this entry to flag semantics changes nothing beyond the returned string.

`Changes` and `RELEASENOTES` entries are already staged on the `Changes` branch — nothing needed here.

Issue #278, opened 2026-07-30
Reserved hosts.cfg tags leak into the info page's "Other tags" row — xmh_item_idx() misses 10 keys and is case-sensitive


SoundGoof:
## `xmh_item_idx()` fails to recognize reserved hosts.cfg tags — two causes

`xmh_item_idx()` (`lib/loadhosts.c`) answers *"is this raw hosts.cfg tag a recognized reserved attribute?"* for two callers: `web/svcstatus-info.c` (to avoid echoing a tag that already has its own field) and `xymonnet/xymonnet.c:492` (to decide a tag is *not* a network test). It answers wrongly in two independent ways.

### 1. Ten keys never match, in any case

The scan stops at the first NULL slot in `xmh_item_key[]`, index 49 (`XMH_IP`):

```c
while (xmh_item_key[i] && strncmp(xmh_item_key[i], value, strlen(xmh_item_key[i]))) i++;
```

That was deliberate: `enum xmh_item_t` mixes tag-keyed items with computed attributes (`XMH_IP`, `XMH_HOSTNAME`, `XMH_PAGEPATH`, `XMH_RAW`, …) that have no key, and keyed items were kept contiguous at the front — so the first NULL meant "no more tag keys". `lib/loadhosts.c:217` guards it:

```c
i = 0; while (xmh_item_key[i]) i++;
if (i != XMH_IP) errprintf("ERROR: Setup failure in xmh_item_key position %d\n", i);
```

Ten tag keys were later appended *past* that boundary, interleaved with the computed ones:

| idx | key | | idx | key |
| --- | --- | --- | --- | --- |
| 51 | `DOC:` | | 64 | `OS:` |
| 52 | `NOPROP:` | | 65 | `NOCOLUMNS:` |
| 61 | `ACCEPTONLY:` | | 67 | `NOTBEFORE:` |
| 63 | `CLASS:` | | 68 | `NOTAFTER:` |
| 69 | `COMPACT:` | | 70 | `INTERFACES:` |

The self-check still passes, because it only asserts *where the first NULL sits* (still 49), not the property it exists to protect — that no keys live above it. `git blame` dates the table to `8b01109b` (2011-03-08) and a pickaxe finds no commit ever altering these entries.

`NOPROP:` is the legacy alias for `NOPROPYELLOW:` (`xymongen/loadlayout.c:636`), so the oldest configs are the most exposed.

### 2. Case rules disagree with `xmh_find_item()`

[...]

PR #279, opened 2026-07-30
docs: match the tag case in hosts.cfg(5) to the keys loadhosts stores (#278)

SoundGoof:
Partially addresses #278.

## Summary

`hosts.cfg(5)` documented four tags in lower case while `lib/loadhosts.c` stores their keys in upper case: `noclear`, `pulldata`, `noflap`, `multihomed`. They were the only tags where the manual and the key table disagreed.

## Why the case matters

`xmh_find_item()` matches a tag case-insensitively, so either spelling still sets the attribute. But `xmh_item_idx()` matches **case-sensitively**, and that is what `web/svcstatus-info.c` uses to decide a tag is already a recognized attribute, and `xymonnet/xymonnet.c` to decide a tag is *not* a network test.

Written in the documented case, these four worked yet were reported as unrecognized — they appeared in the info page's "Other tags:" row despite having been applied, and xymonnet accepted them as test specs. **Following the manual was what triggered the mismatch**, which is the user-facing half of #278's bug 2.

## Scope — docs only

This fixes the documentation side and nothing else. It is not a fix for #278:

- **Bug 2 (case sensitivity)** stays open. Whether `xmh_item_idx()` should match case-insensitively — as `lib/matching.c`'s `namematch()` was changed to do in #182, for the same class of inconsistency — is a decision about the code. Aligning the manual removes the trap for anyone following it, but two hosts spelling the same tag differently still render differently.
- **Bug 1 (the ten keys past the cutoff)** is untouched and unaffected by case, so nothing here helps it.

The two are not mutually exclusive: if `xmh_item_idx()` later goes case-insensitive, this change stays correct and the new test keeps the manual honest regardless.

Tags with no manpage entry at all (`holidays=`, `nobb2`, `NOPROP:`, `NOTBEFORE:`, `NOTAFTER:`, `ACCEPTONLY:`) are a separate gap and untouched. `OS:` legitimately has none — xymond sets it from client data.

## Changes

**`7296efb7`** — 12 lines in `common/hosts.cfg.5`: the `.IP` entries plus every prose reference (`noclear` alone appears 8 times and has two `.IP` entries, in *GENERAL PER-HOST OPTIONS* and *NETWORK TEST SETTINGS* — deliberate, they cross-reference each other), and the `.default.` allowed-tag list. Also one `tasks.cfg.DIST` comment naming `PULLDATA`, the only shipped example using the old spelling.

Plus `tests/packaging/hosts-cfg-tag-case.sh`, which compares every `.IP <tag>` entry against `xmh_item_key[]` — so a *new* tag documented in the wrong case is caught too, not just these four. No build required; both inputs are source files.

**`6b4ea6d0`** — closes an unbalanced quote on the second `dns=` `.IP` entry, the only one in the file. Cosmetic: groff tolerated it and the rendered output is byte-identical. Kept the quote rather than dropping it (the tag has no spaces, so quotes are optional) to match the stacked `"dns=hostname"` entry directly above it, which shares its description.

## Verification

The test against main's manpage:

```
hosts.cfg(5) says "multihomed", lib/loadhosts.c stores "MULTIHOMED"
hosts.cfg(5) says "noclear",    lib/loadhosts.c stores "NOCLEAR"
hosts.cfg(5) says "noflap",     lib/loadhosts.c stores "NOFLAP"
hosts.cfg(5) says "pulldata",   lib/loadhosts.c stores "PULLDATA"
```

Fails on main, passes here; a re-scan confirms zero mismatches remain. `groff -mandoc -z` output is identical to main (two pre-existing warnings, at lines unrelated to this change), and the four entries render as expected. Suite: 16 passed, 3 skipped (unbuilt tree), 0 failed.

### DIFF
--- a/common/hosts.cfg.5
+++ b/common/hosts.cfg.5
@@ -80,7 +80,7 @@ the file or directory is not present on a system.
 
 
 .SH GENERAL PER-HOST OPTIONS
-.IP noclear
+.IP NOCLEAR
 Controls whether stale status messages go purple or clear when
 a host is down. Normally, when a host is down the client statuses
 ("cpu", "disk", "memory" etc) will stop updating - this would usually
@@ -88,9 +88,9 @@ make them go "purple" which can trigger alerts. To avoid that, Xymon
 checks if the "conn" test has failed, and if that is true then the
 other tests will go "clear" instead of purple so you only get alerts
 for the "conn" test. If you do want the stale statuses to go purple,
-you can use the "noclear" tag to override this behaviour.
+you can use the "NOCLEAR" tag to override this behaviour.
 
-Note that "noclear" also affects the behaviour of network tests;
+Note that "NOCLEAR" also affects the behaviour of network tests;
 see below.
 
 .IP prefer
@@ -107,7 +107,7 @@ Note: This only applies to hosts that are defined multiple
 times in the hosts.cfg file, although it will not hurt to add it
 on other hosts as well.
 
-.IP multihomed
+.IP MULTIHOMED
 Tell Xymon that data from the host can arrive from multiple IP-addresses.
 By default, Xymon will warn if it sees data for one host coming from
 different IP-addresses, because this usually indicates a mis-configuration
@@ -586,7 +586,7 @@ E.g. "WARNSTOPS:5" turns the report red once a 6th outage occurs in
 the reporting interval.
 Default: -1 (disabled - no limit on the number of down-periods).
 
-.IP "noflap[=test1,test2,...]"
+.IP "NOFLAP[=test1,test2,...]"
 Disable flap detection for this host, or for specific tests on this
 host. Flap detection is globally controlled by options given to 
 xymond on the command line, but, if enabled, it can be disabled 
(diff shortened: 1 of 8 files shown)


bonomani:
The research here is sound and the facts check out — I ran the test's own logic against unpatched `main` and it finds exactly these four, no more. `xmh_item_idx()` really is a case-sensitive `strncmp`, and its two callers are the ones you name.

The blocker is upstream of the code: **this PR picks a direction we have not decided on yet, so there is nothing for it to be worth until we do.** That decision is ours to make, not yours, and it has not been made — #278 poses it and nobody has answered.

## The decision

The manual and the code disagree about the case of four tags.

**A — move the manual to the code** (this PR): document them in upper case.
**B — move the code to the manual** (#278's own fix 2): `xmh_item_idx()` compares with `strncasecmp`, matching what `xmh_find_item()` has always done, and the case stops mattering.

They are not equivalent:

- **A only helps configurations written after it.** Every existing `hosts.cfg` spelling `noclear` in lower case — all of them, after twenty years of the manual saying so — keeps showing the tag under "Other tags:" unless an admin edits the file. **B fixes every installation at once**, and covers all 49 reachable keys rather than four.
- **A is bigger than this PR.** The tags are documented outside `hosts.cfg(5)` too: `xymond/xymonfetch.8` (2, `pulldata`, with a worked example) and `xymond/xymond.8` (1, `noflap`, in a cross-reference pointing at `hosts.cfg(5)`). Landing only part of it trades a code/doc inconsistency for a doc/doc one. **B needs no documentation change at all.**

## Why this is on hold rather than in review

Until that is settled, merging this buys little. Writing `noclear` keeps working either way; the only difference is a cosmetic row on the info page, and only for configurations rewritten afterwards.

And the test does more than wait: it asserts that the documented case equals the stored case, which writes direction A into the tree as an invariant. It does not break under B — but choosing B later would then cost two edits instead of one: revert the manpage, and delete the test. A test that must be deleted to make a legitimate change is a lock, and the direction it locks is the undecided one.

## So: nothing needed from you for now

Please don't rework anything on my account. The order is decision first, then this PR:

- **if A is chosen**, this lands close to as-is, plus those two other man pages — and there is a way to keep the test without the lock (have it read the comparison `xmh_item_idx()` actually performs, rather than hardcoding a case policy; I checked it holds under both directions and still fails on today's `main`). I will post the detail if we go that way.
- **if B is chosen**, this closes as unnecessary — the manual is already correct under it.

**In short: this PR cannot bring value before we know what we want.** Not because the work is wrong — it is careful and it checks out — but because it answers a question we have not asked ourselves yet. I will come back here once we have.

PR #280, opened 2026-07-30
docs: document the holidays=, NOTBEFORE: and NOTAFTER: hosts.cfg tags

SoundGoof:
Documents three `hosts.cfg` tags that work but had no entry in `hosts.cfg(5)`.

`holidays=` was the most conspicuous gap — `xymonserver.cfg(5)` and `xymon-xmh(5)` both refer to *"the holidays tag for a host in the hosts.cfg file"*, a tag `hosts.cfg(5)` never described. `xymon-xmh(5)` likewise documents `XMH_NOTBEFORE`.

## Verified

**`holidays=SET`** selects a named `[section]` of holidays.cfg. Consumers, all via `xmh_item(hinfo, XMH_HOLIDAYS)`: alert rules (`do_alert.c`, `loadalerts.c`), analysis.cfg thresholds (`client_config.c`), `DOWNTIME` windows (`xymond.c`), and the `NKTIME` critical window (`loadlayout.c`). Holidays declared before any named section are merged into every set at load (`lib/holidays.c`), so the entry says they always apply too. The fallback chain — tag, then the `HOLIDAYS` setting, then common holidays only — matches `xymonserver.cfg(5)`'s existing wording.

**`NOTBEFORE:` / `NOTAFTER:`** limit when a host definition is active. Documented as exactly 12 digits in local time because that is what `timestr2timet()` accepts; any other length is logged and ignored, leaving the bound at its default.

### DIFF
--- a/common/hosts.cfg.5
+++ b/common/hosts.cfg.5
@@ -148,7 +148,10 @@ The holiday set matters wherever a time-specification is matched against
 the current time, such as alert rules, analysis.cfg thresholds and
 \fBDOWNTIME\fR windows, because on a holiday the day is treated as a
 different weekday. Which one is set by HOLIDAYLIKEWEEKDAY in
-holidays.cfg; the default is Sunday.
+holidays.cfg. Xymon's default holidays.cfg sets HOLIDAYLIKEWEEKDAY=0,
+so holidays are treated as Sundays. If the setting is omitted or set
+to -1, the actual weekday is used and holidays have no effect on time
+specifications.
 
 If this tag is not present, the set named by the HOLIDAYS setting in
 .I xymonserver.cfg(5)
--- a/common/hosts.cfg.5
+++ b/common/hosts.cfg.5
@@ -138,6 +138,36 @@ effect immediately, without any delay.
 .IP delayyellow=STATUSCOLUMN:DELAY[,STATUSCOLUMN:DELAY...]
 Same as \fBdelayred\fR, but defers the change to a yellow status.
 
+.IP holidays=SET
+Selects which set of holidays applies to this host. Holiday sets are the
+named sections of the holidays.cfg file, so \fBholidays=us\fR selects the
+"[us]" set. Holidays listed before any named section are common to all
+sets and always apply as well.
+
+The holiday set matters wherever a time-specification is matched against
+the current time, such as alert rules, analysis.cfg thresholds and
+\fBDOWNTIME\fR windows, because on a holiday the day is treated as a
+different weekday. Which one is set by HOLIDAYLIKEWEEKDAY in
+holidays.cfg; the default is Sunday.
+
+If this tag is not present, the set named by the HOLIDAYS setting in
+.I xymonserver.cfg(5)
+is used. If that is unset too, only the common holidays apply.
+
+.IP NOTBEFORE:YYYYMMDDHHMM
+.IP NOTAFTER:YYYYMMDDHHMM
+Limit the period in which this host definition is active, so that a host
+can be scheduled in or out of monitoring without editing hosts.cfg.
+Outside the period the host is treated as if it were not listed in
+hosts.cfg: it is left off the generated web pages, and xymond treats
+status messages about it as coming from an unknown host.
+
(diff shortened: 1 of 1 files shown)


bonomani:
Good addition — `holidays=` in particular was referenced by two other man pages without ever being described here. I traced each claim rather than taking them on trust: the named sections, the fallback chain (tag → `HOLIDAYS` setting → common holidays), the consumers via `xmh_item(hinfo, XMH_HOLIDAYS)`, the 12-digit validation at `lib/timefunc.c:539`, and either tag working alone (`notbefore` defaults to 0, `notafter` to `INT_MAX`). All correct.

The tag cases match the stored keys too — `holidays=` lower, `NOTBEFORE:`/`NOTAFTER:` upper — so this stays neutral whichever way #279 goes.

## One sentence to fix

> Which one is set by HOLIDAYLIKEWEEKDAY in holidays.cfg; **the default is Sunday**.

The default in the code is `-1`, not `0`:

```c
/* lib/holidays.c:28  */  static int holidays_like_weekday = -1;
/* lib/holidays.c:549 */  if (holidays_like_weekday == -1) return t->tm_wday;
```

At `-1`, holidays have no effect at all — the real weekday is returned, which is the opposite of what the paragraph describes. Sunday is only what the shipped `holidays.cfg` sets (`HOLIDAYLIKEWEEKDAY=0`); an admin whose file lacks the line gets no holiday handling. Worth saying as "the shipped holidays.cfg sets 0 (Sunday); without the setting, holidays are not applied at all", since that distinction is exactly what a reader of this paragraph needs.

## The tags you are documenting are broken — #298

Checking the `NOTBEFORE:` claims turned up two defects in `timestr2timet()`, whose only callers are these two tags. Filed as #298; summarised here because it bears on this text.

`struct tm` is never initialised and `tm_sec` is never assigned, so `mktime()` normalises whatever is on the stack. Converting `202001011200`:

```
expected   2020-01-01 12:00:00
actual     1995-12-20 17:57:17     (and it moves as the stack changes)
```

So a host carrying `NOTBEFORE:` is either permanently visible or permanently invisible, unpredictably. The function also writes NULs into the caller's string — which `lib/loadhosts.c:339` takes straight from `xmh_item()`, i.e. the host record's own buffer, the same aliasing as #276 — leaving `XMH_RAW : conn|NOTBEFORE:2020|NOTAFTER:2030` in the stored record.

None of this is a reason to hold the PR. Documenting the intended behaviour is what made the gap visible, and the text describes what the tags are meant to do, which is what #298 should restore. It might be worth a sentence pointing at #298 until it is fixed, so a reader who tries `NOTBEFORE:` and sees nothing sensible knows why — your call.

bonomani (approved):
Verified each claim against the source rather than the prose:

| claim | checked |
|---|---|
| `holidays=SET` selects a named `[section]` | `xmh_item(XMH_HOLIDAYS)`, consumers as listed |
| holidays before any section are common to all sets | `lib/holidays.c:488` `xtreeFind(holidays, "")`, merged into every set at `:497` |
| default `HOLIDAYLIKEWEEKDAY=0`, so Sunday | `xymond/etcfiles/holidays.cfg:9` |
| omitted or `-1` → the real weekday | `holidays.c:28` initialiser, `:549` `return t->tm_wday` |
| outside the window, left off the web pages | `loadhosts.c:460` `hostinfo()` returns NULL |
| outside the window, an unknown host to xymond | `loadhosts.c:414` `knownhost()` returns NULL |
| exactly 12 digits, other lengths logged and ignored | `timefunc.c:540`, bound reverts to its default in `loadhosts.c:341/345` |

All hold. `holidays=` was a real gap — `xymonserver.cfg(5)` and `xymon-xmh(5)` both pointed at a tag `hosts.cfg(5)` never described.

One follow-up, not a change request: the three new labels use the bare `.IP holidays=SET` form, while #303 converts every other label in this file to `.IP "\fBholidays\fR=\fISET\fR"` — bold for what is typed, italic for what is replaced. The two merge without a conflict, so nothing would flag it. I will restyle these three in #303 rather than send this back for three lines of font.

Issue #283, opened 2026-08-05
xymond_hostdata: channel hostname used as a path component without confinement (reachable with --ghosts=allow)


bonomani:
## Summary

`xymond_hostdata` turns the hostname carried in a channel message straight into a filesystem path component, with no confinement on the write path and an insufficient one (`basename()`) on the delete and rename paths. With `--ghosts=allow`, that hostname is whatever the reporting client called itself.

**Not reachable in a default install** — the default is `--ghosts=log` (`xymond.c:256`), and the shipped `tasks.cfg.DIST` passes no `--ghosts` at all. This needs an admin who has deliberately enabled `allow`.

## The three call sites

| message | what it does with the hostname | filtering before this report |
|---|---|---|
| `@@clichg` | `mkdir "$CLIENTLOGS/<host>"`, then writes `"<host>/<tstamp>"` | **none** |
| `@@drophost` | `dropdirectory("$CLIENTLOGS/<host>")` — **recursive delete** | `basename()` |
| `@@renamehost` | `rename("$CLIENTLOGS/<old>", "$CLIENTLOGS/<new>")` | `basename()` |

`basename("..")` returns `".."`, so the two that filter still permit one level of traversal.

## Why the name is untrusted under `--ghosts=allow`

```c
/* xymond.c:4550-4553 */
hname = knownhost(hostname, hostip, ghosthandling);
if (hname == NULL) {
	hname = log_ghost(hostname, sender, msg->buf);
}
```

- `knownhost()` returns an unknown name **verbatim** in that mode (`lib/loadhosts.c:409`), and returns before `log_ghost()` runs.
- `log_ghost()` is what holds the character whitelist rejecting `/` (`xymond.c:1007`).

In the other three modes unknown hosts are dropped (`loadhosts.c:415`), so the name reaching the worker is always a canonical `hosts.cfg` entry.

**Correction to an earlier revision of this issue:** it described `allow` as "bypassing the only validation on this path". That is wrong, and worth stating plainly because it points at the wrong fix. `log_ghost()` returns `result`, which is only ever assigned in the `GH_MATCH` branch (`xymond.c:1031`) — so in `GH_LOG` and `GH_IGNORE` the function returns NULL whether the whitelist fired or not, and the message is dropped either way. What the whitelist actually prevents is a junk name entering `rbghosts` and showing up in the "Ghost Clients" report. It is display hygiene, not path safety. There is no security control that `allow` disables; there was never one on this path.

[...]

PR #284, opened 2026-08-05
tests: path-traversal guards for xymond_hostdata and svcstatus.cgi (#283)

bonomani:
Regression guard for #283: the `xymond_hostdata` **delete and rename** paths.

`@@drophost` recursively deletes `$CLIENTLOGS/<host>` and `@@renamehost` renames it, and with `--ghosts=allow` the name is whatever the client sent — so an unconfined `".."` takes out or moves all of `$XYMONVAR`. `confine_name()` stops that; nothing drove it.

Scope is what `main` doesn't already pin: `hostdata-hostname-sanitize.sh` (#289) covers the `@@clichg` host and test fields, so this covers the delete and the rename — including the rename **destination** — plus the `"."` and `"/"` host spellings.

Assertions compare the whole scratch tree before and after, **by path and by content**: an escape can land beside `$XYMONVAR` (`"../.."`) or inside the client-log root (`"."`), and one that rewrites an existing file in place moves no path at all.

**Every case is measured, not assumed.** The list is built from a vacuity matrix against a guard-less worker — 15 of 15 bite. That matters because the kernel rejects several plausible spellings on its own: `rename(2)` gives `EINVAL` for a source that is an ancestor of the destination and `ENOTEMPTY` for an existing non-empty destination, and `drophost` of a missing path removes nothing. So hostile sources name things that exist beside the client-log root, hostile destinations name things that don't exist yet, and `seed()` plants a decoy so `"../realhost"` has a target. The same reasoning excludes the degenerate spellings in the clichg *test* field — they resolve onto a directory, so `open()` fails either way; that field is pinned by the sanitize test.

**Mutation-tested** against all five `confine_name()` sites. Removing any of the four this file owns makes it fail; the fifth (clichg test name) is caught by `hostdata-hostname-sanitize.sh`.

Built through `setup_xymond_hostdata` / `run_xymond_hostdata_keepvar`, so `require_gnu_make`, `$XYMON_MAKE` and `xymon_cflags` apply.

No CGI test here: `main`'s `tests/web/svcstatus-client-traversal.sh` (`9e8f1806c`, refs #146) already covers that path and asserts refusal rather than absence of a canary, which is strictly stronger.

Suite: 58 passed, 0 skipped, 0 failed.

### DIFF
--- a/tests/xymond/hostdata-path-traversal.sh
+++ b/tests/xymond/hostdata-path-traversal.sh
@@ -1,0 +1,200 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/xymond/hostdata-path-traversal.sh
+#
+# Regression guard: a host name from a channel message must not let
+# xymond_hostdata escape the client-log directory through the two paths that
+# do not merely write a file. @@drophost recursively deletes
+# "$CLIENTLOGS/<host>", and @@renamehost renames it - and with
+# --ghosts=allow the name is whatever the client sent, so an unconfined ".."
+# takes out or moves the whole of XYMONVAR. The confinement is
+# confine_name() in the worker.
+#
+# Scope: hostdata-hostname-sanitize.sh already covers the @@clichg save path
+# for a hostile host name and a hostile test name. This one covers what that
+# leaves: the delete and the rename, the rename *destination* as well as its
+# source, and the degenerate "." and "/" spellings.
+#
+# Every assertion is "the tree did not change", compared over the whole
+# scratch directory: an escape can land beside XYMONVAR ("../..") or inside
+# the client-log root ("."), and a whole-tree comparison never has to predict
+# which. The comparison covers both the path set and the contents of every
+# file under var, because a traversal that overwrites an existing file in
+# place - "$CLIENTLOGS/../CANARY.txt" - moves no path at all.
+#
+# Every case below is chosen so that it *can* fail: each was measured against
+# a worker with confine_name() removed and confirmed to change the tree. A
+# hostile spelling that the kernel rejects on its own (see the renamehost
+# notes) proves nothing about the guard and is not listed.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+# shellcheck source=tests/lib/xymond-hostdata.sh
+. "$(dirname "$0")/../lib/xymond-hostdata.sh"
+
+# dropdirectory() forks (lib/files.c), so the delete outlives the worker and
+# settle() has to wait for the child. There is no PID to wait on - the shell
+# never forked it - so the wait matches on the command line, which needs
+# pgrep. Declare it: without this the wait below would degrade to nothing and
+# the assertions would race the delete.
+command -v pgrep >/dev/null 2>&1 \
+	|| skip "pgrep not available (needed to wait for the backgrounded drophost child)"
+
+work=$(mktempdir)
+
+setup_xymond_hostdata "$work"
+
+seed() {  # a pristine XYMONVAR with canaries inside and outside the client-log dir
+	rm -rf "${work:?}/var"
+	mkdir -p "$work/var/hostdata/realhost" "$work/var/rrd" "$work/var/realhost"
+	printf 'CANARY-OUTSIDE-CLIENTLOGS\n' >"$work/var/CANARY.txt"
+	printf 'existing client data\n'      >"$work/var/hostdata/realhost/1785830400"
+	printf 'rrd payload\n'               >"$work/var/rrd/some.rrd"
+	# A decoy sharing the real host's name one level up, so "../realhost"
+	# names something that exists. Without it that spelling resolves to a
+	# missing directory, dropdirectory() removes nothing, and the case holds
+	# whether or not the name was confined.
+	printf 'decoy above the client-log root\n' >"$work/var/realhost/decoy"
+	: >"$work/worker.log"
+}
+
+# Paths *and* contents. `find | sort` alone cannot see an escape that opens an
+# existing file and rewrites it. worker.log lives above var and is appended to
+# by every feed(), so it is outside the content half by construction.
+# `|| true`: find exits 1 if an entry disappears under it, and pipefail would
+# turn that into a bare errexit abort with no verdict line.
+snapshot() {
+	{ find "$work" 2>/dev/null || true; } | sort
+	{ find "$work/var" -type f -exec cksum {} + 2>/dev/null || true; } | sort
+}
+
+feed() {  # feed <message-first-line> [body]
+	local rc
+	set +e
+	printf '%s\n%s\n@@\n' "$1" "${2:-payload}" \
+		| run_xymond_hostdata_keepvar "$work" >/dev/null 2>>"$work/worker.log"
+	rc=$?
+	set -e
+	# The worker exits 0 when its input pipe closes. Anything else is a crash
+	# or an early abort, and must not be mistaken for "the guard held": a
+	# worker that died on this message would satisfy every "nothing happened"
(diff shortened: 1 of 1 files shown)


SoundGoof:
## Findings

### Hostdata assertions miss several guarded traversal paths

In
[hostdata-path-traversal.sh:123-127](https://github.com/xymon-monitoring/xymon/blob/55ff463f9d4a8445ff3e3114751cd310a2627acd/tests/xymond/hostdata-path-traversal.sh#L123-L127),
the search only examines immediate children of `$work/var`.

It therefore misses the locations produced without confinement by:

- `"../.."`: `$work/9999999999`
- `"."` and `"/"`: `$work/var/hostdata/9999999999`

Mutation testing confirmed the problem. After bypassing `@@clichg`
hostname confinement and restricting the loop to those three advertised
inputs, the harness still reported `OK`.

The test also never supplies a hostile test name to `@@clichg` or a hostile
new hostname to `@@renamehost`. The unmodified harness still reported `OK`
after both corresponding production guards were removed.

Assert each expected escape location, or compare a complete tree snapshot.
Also add hostile values for the test-name and rename-destination fields.

### Both new tests fail before their assertions on BSD

Both scripts invoke literal `make` and compile without the configured include
flags:

- [hostdata-path-traversal.sh:20-47](https://github.com/xymon-monitoring/xymon/blob/55ff463f9d4a8445ff3e3114751cd310a2627acd/tests/xymond/hostdata-path-traversal.sh#L20-L47)
- [svcstatus-path-traversal.sh:20-48](https://github.com/xymon-monitoring/xymon/blob/55ff463f9d4a8445ff3e3114751cd310a2627acd/tests/web/svcstatus-path-traversal.sh#L20-L48)

Reproduced on OpenBSD 7.9, FreeBSD 14.3, and NetBSD 11.0:

1. BSD `make` rejects the generated GNU Makefiles.
2. With `make` temporarily redirected to `gmake`, compilation still fails
   because `pcre2.h` is outside the default include path.

Use the current shared builders in `tests/lib/build-worker.sh` and
`tests/lib/svcstatus-cgi.sh`. They already use `require_gnu_make`,
`$XYMON_MAKE`, `xymon_cflags`, and configured linker flags.

## Validation

- `git diff --check` passes.
- Linux full suite: 56 passed, 0 skipped, 0 failed.
- Both BSD failures reproduced on all three available BSD systems.
- Both false-negative findings were confirmed with isolated mutation tests.

bonomani:
Both findings confirmed and fixed — though I departed from the suggested remedy on one of them, so let me say where and why.

### The BSD finding — fixed as suggested

Both scripts now build through the shared helpers (`setup_xymond_hostdata` → `build_xymond_worker`), so `require_gnu_make`, `$XYMON_MAKE` and `xymon_cflags` apply. No literal `make`, no hand-rolled include flags.

### The CGI test is dropped, not ported

Before porting it I checked what `main` already covers, and `tests/web/svcstatus-client-traversal.sh` (`9e8f1806c`, refs #146) turned out to be a strict superset of it: the same hostile `CLIENT` list, the same `CANARY_ABOVE`/`CANARY_IN` canaries, the same legitimate and IP-spelling checks — and it asserts the CGI **refuses** (exit 1) rather than only that no canary leaked, which is the stronger form of exactly the false-negative concern you raised. It also carries normalized-host cases (`../realhost`, `/realhost`, ...) and a hostile `TIMEBUF` that mine never had.

Porting mine would have left two tests doing one job with the weaker one newer, so `svcstatus-path-traversal.sh` is deleted. If you would rather have it repaired and kept, say so and I will.

### The hostdata assertions — fixed, and your method found two more

You were right that the `-maxdepth 1` scan could not see `"../.."` (beside `$XYMONVAR`) or `"."` / `"/"` (inside the client-log root). Replaced with a before/after snapshot of the whole scratch tree, which cannot miss a location because it never has to predict one.

Scope also narrowed: `hostdata-hostname-sanitize.sh` from #289 already pins the `@@clichg` host **and** test fields, so this file now covers what that leaves — `@@drophost`, `@@renamehost`, and the rename **destination** you noted was never exercised.

Then I ran your mutation method across all five `confine_name()` call sites, and it caught two cases of mine that proved nothing:

- the hostile rename **sources** (`".."`, `"."`, `"/"`, `"../.."`) are all ancestors of the destination, and `rename(2)` refuses to move a directory into itself, so they held with the guard removed. Replaced with `"../rrd"` and `"../CANARY.txt"`, siblings of the client-log root, which really do move.
- the degenerate spellings in the clichg **test** field resolve the save onto an existing directory, so `open()` fails either way. That loop is gone, with a comment recording why instead of a vacuous check.

Final state — removing any of the four guards this file owns makes it fail:

| guard removed | result |
| --- | --- |
| `@@clichg` hostname | caught |
| `@@drophost` hostname | caught |
| `@@renamehost` source | caught |
| `@@renamehost` destination | caught |
| `@@clichg` test name | caught by `hostdata-hostname-sanitize.sh` (mutated it to confirm) |

Now one commit, one file, 172 lines. Suite 56 passed / 0 skipped / 0 failed, `shellcheck -x` clean. Thanks — the mutation testing is what turned this from plausible into checked.

SoundGoof (dismissed):
## Findings

No blocking findings.

Non-blocking: the title still mentions `svcstatus.cgi`, but the force-updated
PR now contains only the `xymond_hostdata` test.

The previous findings are resolved:

- whole-tree snapshots detect escapes above and below `XYMONVAR`
- effective rename sources exercise source confinement
- hostile rename destinations are covered
- existing `@@clichg` coverage is reused without duplication
- the shared portable worker builder replaces local build logic

## Validation

- `git diff --check` passes.
- Linux full suite: 58 passed, 0 skipped, 0 failed.
- Changed test passes on OpenBSD 7.9, FreeBSD 14.3, and NetBSD 11.0.

Independent mutation testing confirmed detection of removed confinement for:

- `@@drophost` hostname
- `@@renamehost` source
- `@@renamehost` destination
- `@@clichg` hostname

The existing `hostdata-hostname-sanitize.sh` also detects removal of the
delegated `@@clichg` test-name guard.

bonomani:
Both findings confirmed and fixed, though I departed from the suggested remedy on one.

**BSD:** fixed as suggested — both scripts now build through the shared helpers, so `require_gnu_make`, `$XYMON_MAKE` and `xymon_cflags` apply.

**The CGI test is dropped, not ported.** `main`'s `tests/web/svcstatus-client-traversal.sh` (`9e8f1806c`) is a strict superset: same hostile `CLIENT` list, same canaries, same IP spellings — and it asserts the CGI *refuses* rather than only that no canary leaked, which is the stronger form of your concern. Porting mine would leave two tests doing one job with the weaker one newer. Say the word if you would rather have it repaired and kept.

**Hostdata assertions:** you were right that `-maxdepth 1` could not see `"../.."` or `"."`. Replaced with a whole-tree before/after snapshot. Scope also narrowed, since #289 already pins the clichg fields.

Then your mutation method caught two cases of mine that proved nothing: the hostile rename *sources* were all ancestors of the destination, so `rename(2)` refused them regardless of the guard; and the degenerate spellings in the clichg *test* field resolve onto an existing directory, so `open()` fails either way. Replaced and removed respectively.

15 of 15 hostile cases now bite against a guard-less worker, and all four guards this file owns are caught by mutation (the fifth by the sanitize test). One commit, one file. Suite 58/0/0.

SoundGoof (approved):
## Review

No blocking findings.

Non-blocking: the title still mentions `svcstatus.cgi`, although this PR
only adds the `xymond_hostdata` test.

Validation:

- Linux suite: 55 passed, 3 expected skips, 0 failed.
- Focused test passes on FreeBSD 14.3, NetBSD 11.0, and OpenBSD 7.9.
- Four isolated mutations confirm detection of removed confinement for:
  - `@@clichg` hostname
  - `@@drophost` hostname
  - `@@renamehost` source
  - `@@renamehost` destination
- Delegated clichg test-name and CGI traversal tests pass.
- `shellcheck -x` and `git diff --check` pass.

PR #286, opened 2026-08-05
xymond_hostdata: don't drop client data during the first hour of uptime (#285)

### LINKED ISSUE #285: xymond_hostdata: client data silently dropped for the first hour of uptime (zero-initialised timestamps vs monotonic clock)
## Summary

`xymond_hostdata` silently discards **all** client data while the machine's uptime is below `--recent-period` (default 3600s). After a reboot of the Xymon server, no hostdata is saved for the first hour, and nothing is logged to say so.

This is **not** a case for changing the clock — see "Why `gettimer()` is not the bug" below.

## Mechanism

`xymond_hostdata.c` throttles how often a host's data is saved:

```c
time_t now = gettimer();
...
for (i=0, recentcount=0; ((i < 12) && (itm->tstamp[i] > (now - recentperiod))); i++) recentcount++;
if (!logdirfull && (recentcount < maxrecentcount)) { /* ...save... */ }
```

A host seen for the first time gets a freshly `calloc`'d record, so `tstamp[]` is all zeroes:

```c
itm = (savetimes_t *)calloc(1, sizeof(savetimes_t));
```

`gettimer()` returns `CLOCK_MONOTONIC` (`lib/timing.c:44`) — seconds since boot. So when uptime < `recentperiod`, the bound `now - recentperiod` is **negative**, `0 > negative` holds for all twelve slots, `recentcount` reaches 12, and `recentcount < maxrecentcount` (5) is false. The save is skipped, and that branch prints nothing.

## Reproduction

Workers read messages on stdin, so the binary can be driven directly. `--recent-period` stands in for a short uptime (the comparison only cares about `now - recentperiod`):

| invocation | result |
|---|---|
| `--recent-period=60` (default window, host up 9.7h) | file written |
| `--recent-period=99999` (window exceeds uptime) | **nothing written, no message** |
| `--recent-period=0` | file written |

Same message in all three cases:

```
@@clichg#1|1|10.0.0.99|realhost|SAVED|linux
payload
@@
```

On a freshly booted host the default 3600s window produces the middle row.

## Why `gettimer()` is not the bug

#29 dispositions TBT patch `231 xymon.hist_timer.patch` as **drop — monotonic→wall-clock regression; PR #174 keeps `gettimer()`**, and rejects `62 xymon.proxynow.patch` partly because `getcurrenttime()` "would be the wrong timebase vs main's `gettimer()`". That stance is right, and this report does not argue against it: monotonic is the correct clock for measuring an interval, which is exactly what this throttle does.

The defect is the **initialiser**. `0` means "never" on a wall clock (1970) but is a valid, recent instant on a monotonic one (boot). Please do not fix this by swapping the clock — that is the regression #29 already rejected.

## Suggested fix

Floor the comparison bound so "never saved" cannot read as "saved recently":

```c
time_t cutoff = (now > recentperiod) ? (now - recentperiod) : 0;
...
for (i=0, recentcount=0; ((i < 12) && (itm->tstamp[i] > cutoff)); i++) recentcount++;
```

Note a sentinel initialiser does *not* work: setting `tstamp[]` to `-1` still satisfies `-1 > (now - recentperiod)` when uptime is low, and no fixed sentinel is safe because `--recent-period` is admin-settable. The bound has to be floored.

Throttling behaviour is unchanged once uptime exceeds the window, and remains correct below it: real saves record `now > 0`, so they still count against the limit.

## Impact

Data loss, silent, on every restart — bounded to the first `--recent-period` of uptime (1 hour by default). No security impact. Found while writing a regression test for #283: the test passed locally and failed on CI runners, which boot fresh.

bonomani:
Fixes #285.

## What

`xymond_hostdata` silently discards **all** client data while the machine's uptime is below `--recent-period` (default 3600s; the flag itself takes **minutes**, so this is `--recent-period=60`). After a reboot of the Xymon server, nothing is stored for the first hour, and nothing is logged to say so.

The save-throttle counts how many of a host's twelve recorded save times fall inside the window:

```c
for (i=0; (i < 12) && (itm->tstamp[i] > (now - recentperiod)); i++) recentcount++;
if (!logdirfull && (recentcount < maxrecentcount)) { /* ...save... */ }
```

A host seen for the first time gets a `calloc`'ed record, so those slots are zero — and `gettimer()` is `CLOCK_MONOTONIC` (`lib/timing.c:44`), so zero is not "long ago", it is the moment this machine booted. Below the window the bound goes negative, `0 > negative` holds for all twelve slots, `recentcount` reaches 12, and the save is skipped in a branch that prints nothing.

## Change

Floor the cutoff at zero.

A sentinel initialiser does **not** work, which is worth stating because it is the obvious first idea: `-1` still satisfies `-1 > (now - recentperiod)` when uptime is low, and no fixed value is safe because `--recent-period` is admin-settable. The bound is what has to be floored.

## This does not touch `gettimer()`

Deliberately. Monotonic is the correct clock for measuring an interval, which is exactly what this throttle does. #29 already dispositioned the monotonic→wall-clock swap (TBT patch `231 xymon.hist_timer.patch`) as **"drop — monotonic→wall-clock regression; PR #174 keeps `gettimer()`"**, and rejected `62 xymon.proxynow.patch` partly on the same grounds. That stance is right. The defect here is the initialiser, not the timebase.

## Evidence

`tests/xymond/hostdata-recent-window.sh` drives the real binary — workers read messages on stdin, so no daemon is needed. The test cannot reboot the host, but the comparison only depends on `now - recentperiod`, so a window larger than the current uptime reproduces the condition exactly.

| window | before | after |
|---|---|---|
| larger than uptime (stands in for a fresh boot) | 🔴 nothing written, nothing logged | data saved |
| 60 min, host up 9.7h | saved | saved |
| 0 | saved | saved |

The test fails against the code this fixes, not merely passes against the fix.

## Observed while verifying, and left alone

The throttle caps nothing either way: seven messages fed to one process write seven files both before and after this change. The shift loop

```c
for (i = 10; (i > 0); i--) itm->tstamp[i+1] = itm->tstamp[i];
```

stops at `i=1`, so `tstamp[1]` is never written and stays zero forever; the counting loop halts at the first non-recent slot, so `recentcount` never exceeds 1 and `--recent-count` has no effect. That is a separate defect with a different fix, and widening this PR to cover it would make both harder to review. Flagged here so the numbers above are not mistaken for a regression introduced by this change.

## Note on overlap

Independent of #284, which touches the same function for an unrelated reason (path confinement). Both apply to `main`; whichever lands second will need a trivial rebase — the hunks are adjacent but do not overlap.

### DIFF
--- a/tests/xymond/hostdata-recent-window.sh
+++ b/tests/xymond/hostdata-recent-window.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/xymond/hostdata-recent-window.sh
+#
+# Regression guard: xymond_hostdata must save client data even when the
+# machine's uptime is below --recent-period.
+#
+# A first-seen host has calloc'ed (zero) save timestamps, and gettimer() is
+# CLOCK_MONOTONIC (seconds since boot), so while uptime is below the window
+# "now - recentperiod" is negative, every empty slot counts as a recent
+# save, and the data is dropped unlogged.
+#
+# The test cannot reboot the host, but the comparison only depends on
+# "now - recentperiod", so a window larger than the current uptime
+# reproduces the condition exactly.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+ROOT=$(find_root)
+
+CC=${CC:-cc}
+command -v "$CC" >/dev/null 2>&1 || skip "no C compiler available (CC=$CC)"
+command -v make >/dev/null 2>&1 || skip "make not available"
+[ -f "$ROOT/include/config.h" ] || skip "tree not configured (no include/config.h)"
+[ -f "$ROOT/Makefile" ] || skip "tree not configured (no Makefile)"
+
+ssllibs=$(sed -n 's/^SSLLIBS *= *//p' "$ROOT/Makefile")
+pcre_libs=${PCRELIBS:-}
+if [ -z "$pcre_libs" ] && command -v pkg-config >/dev/null 2>&1; then
+	pcre_libs=$(pkg-config --libs libpcre2-8 2>/dev/null || true)
+fi
+[ -n "$pcre_libs" ] || pcre_libs="-lpcre2-8"
+
+work=$(mktemp -d "${TMPDIR:-/tmp}/xymon-hostdata-window.XXXXXX")
+trap 'rm -rf "$work"' EXIT HUP INT TERM
+
+make -C "$ROOT/lib" libxymon.a libxymoncomm.a libxymontime.a >"$work/libbuild.log" 2>&1 \
+	|| { cat "$work/libbuild.log" >&2; fail "cannot build the xymon libraries"; }
+
+# HOSTDATAOBJS = xymond_hostdata.o xymond_worker.o (xymond/Makefile).
+# Archives listed twice rather than --start-group, which is GNU ld only.
+"$CC" -I"$ROOT/include" -I"$ROOT/lib" -I"$ROOT/xymond" -o "$work/xymond_hostdata" \
+	"$ROOT/xymond/xymond_hostdata.c" "$ROOT/xymond/xymond_worker.c" \
+	"$ROOT/lib/libxymon.a" "$ROOT/lib/libxymoncomm.a" "$ROOT/lib/libxymontime.a" \
+	"$ROOT/lib/libxymon.a" \
+	$pcre_libs $ssllibs 2>"$work/cc.log" \
+	|| { cat "$work/cc.log" >&2; fail "xymond_hostdata does not build -- cannot verify the save window"; }
+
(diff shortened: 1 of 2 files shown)


SoundGoof:
## Finding

**[P2] Use a window that exceeds realistic server uptime**

`tests/xymond/hostdata-recent-window.sh:72-74` uses `99999` minutes to simulate uptime below `--recent-period`, but that is only 69.44 days. On a server up longer than that, the original buggy code has a positive cutoff and this regression test passes without reproducing the defect.

Use a larger non-overflowing value such as `10000000` minutes. That represents roughly 19 years while `60 * 10000000` still fits the signed `int` arithmetic used by option parsing.

## Validation

The production cutoff fix is correct, all four CI checks pass

bonomani:
Fixed in a7e9896: the stand-in window is now 10000000 minutes (~19 years), so the before-case reproduces the negative cutoff regardless of host uptime, and 60*10000000 still fits the option-parsing int. Test re-run locally, all three cases pass. Still a single commit.

SoundGoof (approved):
lgtm

PR #287, opened 2026-08-05
xymond_rrd: confine channel hostnames to one path component (#290)

### LINKED ISSUE #290: xymond_rrd: channel hostname used as a path component without confinement -- @@drophost deletes $RRDDIR recursively

bonomani:
Fixes #290. Same defect class as #284 (#283), in a different worker.

> **Stacked on #147** — uses the `safe_basename()` helper that PR adds, so #147 goes first. Until it merges, its commits show in this diff; the change under review is the last commit only.

## What

`xymond_rrd` turns the hostname from a channel message into a path component in two handlers:

| handler | effect | filtering before |
|---|---|---|
| `@@drophost` (`:418`) | `dropdirectory("$RRDDIR/<host>", 1)` — **recursive delete** | `basename()` |
| `@@renamehost` (`:440-441`) | `rename("$RRDDIR/<old>", "$RRDDIR/<new>")` | **none** |

`basename("..")` returns `".."`, so the one that filters confines nothing.

## Measured

Feeding the unpatched worker `@@drophost#1|…|..`:

```
Starting to remove directory $RRDDIR/..
Recurse into $RRDDIR/../rrd
Removing file $RRDDIR/../rrd/realhost/tcp.rrd
Removing directory $RRDDIR/../rrd/realhost
Removing directory $RRDDIR/../rrd
Removing directory $RRDDIR/..
```

The entire RRD tree is destroyed, `$RRDDIR` included. That is the historical data.

| case | before | after |
|---|---|---|
| `@@drophost` host `..`, `../..`, `/`, `.` | 🔴 `$RRDDIR` and its contents deleted | dropped |
| `@@renamehost` old `..` / new `../escaped` | 🔴 directory created outside `$RRDDIR` | dropped |
| legitimate `drophost` / `renamehost` | works | works |

## Why the fix belongs in the worker

`xymond_rrd.c:243` calls `net_worker_run()`. In a distributed setup the worker listens on its own ip:port and takes messages from remote xymond instances, so its trust boundary is not "our own xymond" — no entry-side check in xymond could cover that input. #283 originally suggested hardening xymond's ghost handling instead; that suggestion has been withdrawn there, for this reason and because the check in question (`log_ghost()`'s whitelist) was never a path-safety control in the first place.

## A note on the test

`tests/xymond/rrd-path-traversal.sh` asserts on `$RRDDIR` and on a sibling *directory*, not on a sibling file. `dropdirectory()` unlinks entries while `readdir()` is still walking them, which is undefined for entries not yet returned — a sibling file survives by luck often enough that an assertion resting on it passes for the wrong reason. It did exactly that during development, which is why the assertions are written this way.

Verified to fail against the code this replaces, and against a variant regressing only the `renamehost` handler, so each assertion is shown to catch its own defect.

## Related

- #284 — same fix for `xymond_hostdata`; independent, no file overlap
- #221 — hardens `xymond_filestore`'s notes path with `basename()`, which does not stop `..`; worth revisiting once `safe_basename()` lands
- `xymond_filestore`'s `@@status`→HTML (`:331`) and `@@notes` (`:359`) paths remain unconfined; not in this PR

### DIFF
--- a/tests/xymond/rrd-path-traversal.sh
+++ b/tests/xymond/rrd-path-traversal.sh
@@ -1,0 +1,150 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/xymond/rrd-path-traversal.sh
+#
+# Regression guard: xymond_rrd must not let a hostname taken from a channel
+# message escape the RRD directory.
+#
+#   @@drophost   -> dropdirectory("$RRDDIR/<host>", 1)   <- recursive delete
+#   @@renamehost -> rename("$RRDDIR/<old>", "$RRDDIR/<new>")
+#
+# drophost filtered with basename(), which returns ".." for ".." and so
+# confines nothing; renamehost did not filter at all. Since the delete is
+# recursive, an unconfined ".." takes out the parent of the RRD directory --
+# the historical data.
+#
+# Workers read their messages from stdin (xymond_worker.c) and are also
+# reachable over the network in a distributed setup (net_worker_run(),
+# xymond_rrd.c), so the component has to be confined here rather than
+# upstream in xymond.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+ROOT=$(find_root)
+
+CC=${CC:-cc}
+command -v "$CC" >/dev/null 2>&1 || skip "no C compiler available (CC=$CC)"
+command -v make >/dev/null 2>&1 || skip "make not available"
+[ -f "$ROOT/include/config.h" ] || skip "tree not configured (no include/config.h)"
+[ -f "$ROOT/Makefile" ] || skip "tree not configured (no Makefile)"
+
+rrddef=$(sed -n 's/^RRDDEF *= *//p' "$ROOT/Makefile")
+rrdlibs=$(sed -n 's/^RRDLIBS *= *//p' "$ROOT/Makefile")
+[ -n "$rrdlibs" ] || skip "tree built without RRD support (no RRDLIBS)"
+ssllibs=$(sed -n 's/^SSLLIBS *= *//p' "$ROOT/Makefile")
+pcre_libs=${PCRELIBS:-}
+if [ -z "$pcre_libs" ] && command -v pkg-config >/dev/null 2>&1; then
+	pcre_libs=$(pkg-config --libs libpcre2-8 2>/dev/null || true)
+fi
+[ -n "$pcre_libs" ] || pcre_libs="-lpcre2-8"
+
+work=$(mktemp -d "${TMPDIR:-/tmp}/xymon-rrd-traversal.XXXXXX")
+trap 'rm -rf "$work"' EXIT HUP INT TERM
+
+make -C "$ROOT/lib" libxymon.a libxymoncomm.a libxymontime.a >"$work/libbuild.log" 2>&1 \
+	|| { cat "$work/libbuild.log" >&2; fail "cannot build the xymon libraries"; }
+
+# RRDOBJS = xymond_rrd.o xymond_worker.o do_rrd.o client_config.o (xymond/Makefile).
+# Archives listed twice rather than --start-group, which is GNU ld only.
+"$CC" -I"$ROOT/include" -I"$ROOT/lib" -I"$ROOT/xymond" $rrddef -o "$work/xymond_rrd" \
+	"$ROOT/xymond/xymond_rrd.c" "$ROOT/xymond/xymond_worker.c" \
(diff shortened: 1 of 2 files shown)


bonomani:
Superseded by #264, which landed as `9ae7b5a8e`. Closing rather than rebasing.

`xymond/xymond_rrd.c` on `main` already confines both the drophost and the
renamehost names with `safe_basename()`, and does more than this PR does:

| | this PR | `main` today |
|---|---|---|
| confinement | `safe_basename()` | `safe_basename()` |
| path truncation | unchecked `snprintf()` | checked, refuses and reports |
| pending updates | left in the cache | `rrdcache_drop_host()` before the delete |
| deletion | `dropdirectory(hostdir, 1)` — forked | synchronous, so it cannot race a message posted after the drop |
| rename result | unchecked | checked, with an ENOENT/stat fallback |

Merging this branch would replace the checked, unraceable version with the
unchecked, forked one, so the code half is a regression rather than a fix.

The test half is covered too. `tests/server/xymond-rrd-drophost.sh` on `main`
exercises the same hostile shapes and carries the same guard against a
vacuous pass.

I checked whether the shapes unique to this branch were worth grafting onto
that test, and they are not — every one of them is redundant against
`safe_basename()` (lib/cgi.c:36), which rejects on exactly four conditions:
empty, contains `/`, `.`, and `..`.

- `../..` contains `/`, so it takes the same branch as the existing
  `foo/testhost`.
- `..|..` short-circuits on the first name, exactly like the existing
  `..|newhost`.
- an empty name and `.` both make `safe_basename()` return `""`, so the
  caller's `if (!*safehost)` cannot tell them apart.

All four rejection branches are already reached by the cases `main` has.

Thanks for the original report — the traversal it describes is real, and it
is fixed; it just got fixed on the other branch first.

PR #289, opened 2026-08-05
xymond_hostdata: make the save throttle actually throttle (#288)

### LINKED ISSUE #288: xymond_hostdata: the save throttle never triggers -- --recent-count has no effect
## Summary

`xymond_hostdata`'s save-throttle never triggers: `--recent-count` has no effect, and a host reporting continuously has every message written to disk. Measured — eight messages in one process with `--recent-count=5` write eight files.

## Mechanism

The limit is kept as twelve save timestamps per host, shifted down on each save:

```c
for (i = 10; (i > 0); i--) itm->tstamp[i+1] = itm->tstamp[i];
itm->tstamp[0] = now;
```

The loop stops at `i = 1`, so it writes `tstamp[11]`…`tstamp[2]` and never `tstamp[1]`. Since the record comes from `calloc()`, `tstamp[1]` stays zero for the life of that host's entry.

The counting loop halts at the first slot that is not recent:

```c
for (i=0, recentcount=0; ((i < 12) && (itm->tstamp[i] > cutoff)); i++) recentcount++;
```

`tstamp[0]` is the last save (recent), `tstamp[1]` is zero (not recent) — so it stops at `i = 1` and `recentcount` is 1 whatever the history. `recentcount < maxrecentcount` is then true for any `--recent-count` above 1, and nothing is ever dropped.

## Reproduction

Workers read messages on stdin, and the save times are per-process state, so the burst goes through one process:

```console
$ for k in $(seq 1 8); do
    printf '@@clichg#%d|1|10.0.0.99|realhost|F%d|linux\npayload\n@@\n' "$k" "$k"
  done | ./xymond_hostdata --env=xymonserver.cfg --logdir=$V/hostdata \
           --minimum-free=0 --recent-period=1 --recent-count=5
$ ls $V/hostdata/realhost | wc -l
8                                  # expected 5
```

With the loop shifting down to slot 0, the same burst writes 5.

## Impact

No security impact — the throttle exists to bound disk churn, and its failure means more writes than intended, not fewer. A host whose client data changes on every poll has every version stored, which is exactly the load `--recent-count` was added to cap. Worth noting the shipped `tasks.cfg` does not pass `--recent-count`, so every installation runs the default of 5 and none of them get it.

## Fix

```c
for (i = 10; (i >= 0); i--) itm->tstamp[i+1] = itm->tstamp[i];
```

## Relationship to #285

Different defect in the same function, and both had to be fixed before the throttle behaves: #285 is the cutoff going negative below one hour of uptime, which makes *every* slot count as recent and drops everything. This one is the opposite failure. Neither masks the other — with only #285 fixed the throttle still never triggers, which is how this was found.

bonomani:
Fixes #288.

> **Stacked on #286** — the last commit is the change under review. The defects are independent, but without #286's floored cutoff a machine whose uptime is below `--recent-period` drops every save, so the throttle can't be tested deterministically.

## What

`xymond_hostdata`'s save throttle never triggers: the twelve-slot timestamp shift stops at `i = 1`, so `tstamp[1]` keeps its `calloc`'ed zero, the recent-save count halts there, and `--recent-count` has no effect — a host whose data changes every poll writes every message to disk. The shipped `tasks.cfg` runs the default of 5, so every installation is affected. (No security impact: throttle failure means more writes, not fewer.)

## Fix

- **Ring instead of shifted array** — the per-host history holds the `--recent-count` most recent save times, so any limit works (the fixed twelve silently disabled values above 12) and the check is one comparison of the oldest slot against the cutoff. A slot is consumed only by a *completed* save, so a failed write never burns budget; instead a persistent write failure `chkfreespace()` can't see (read-only mount, permissions) backs off *that host alone* for 5 minutes, retrying and logging once per window without stopping saves for the rest of the fleet.
- **Confined `clichg` paths** (review) — raw host/test names let a `/` or a bare `..` escape the CLIENTLOGS tree. Names now go through `safe_basename()`, which *rejects* `/`-bearing and degenerate `.`/`..`/`""` names (rejected → message dropped). The throttle tree is keyed on the sanitized name, so prefix variants of one host can't each claim a budget. `drophost`/`renamehost` share the same guard and free the host's throttle entry (otherwise leaked).
- **Plain-integer option parsing** (review) — `--recent-count`, `--recent-period`, `--minimum-free` use `parse_int_opt()`, a shared `strtol` helper in `lib/misc.c`. A trailing non-digit is rejected, so `--minimum-free=10%` falls back to the default rather than becoming 10; out-of-range integers clamp, matching `atoi`-era negatives so upgrades don't change deployed setups. `xymond_history`'s `--minimum-free` had the same `atoi` bug and shares the helper and `DEFAULT_MINLOGSPACE`.

Measured: eight messages with `--recent-count=5` wrote 8 files before, 5 after.

## Testing

All tests drive the real built worker on stdin (shared recipe in `tests/lib/build-worker.sh`: absent toolchain/tree skips, a build failure in a configured tree **fails**):

- `hostdata-save-throttle.sh` — the limit follows `--recent-count` (incl. 12/13), `0` disables saving, `--recent-period=0` disables the throttle, failed saves don't burn slots. Fails unfixed: *8 messages with --recent-count=5 wrote 8 files, expected 5*.
- `hostdata-hostname-sanitize.sh` — `/`-bearing and bare `..` host/test names are rejected and touch nothing in or above the tree.
- `hostdata-minfree-validate.sh` — non-integer values (incl. a `%` suffix) fall back, negatives clamp, out-of-range caps.
- `hostdata-rename-throttle-reset.sh` — renamehost clears the old name's throttle budget (renamehost, not drophost, whose background-fork rmdir would race an immediate re-save).
- `hostdata-write-error-backoff.sh` — a persistent write failure on one host trips a single 5-minute backoff and does not suppress a second, healthy host.

### DIFF
--- a/docs/manpages/man8/xymond_history.8.html
+++ b/docs/manpages/man8/xymond_history.8.html
@@ -52,7 +52,10 @@ Section: Maintenance Commands (8)<br/>Updated: Version 4.3.30: 4 Sep 2019<br/><a
   <dt id="minimum-free"><a class="permalink" href="#minimum-free"><b>--minimum-free</b>=<i>N</i></a></dt>
   <dd>Sets the minimum percentage of free filesystem space on the $XYMONHISTLOGS
       directory. If there is less than N% free space, xymond_history will not
-      save the detailed status-logs. Default: 5
+      save the detailed status-logs. N is a percentage between 0 and 100; 0
+      disables the free-space check. The value must be a plain integer; anything
+      else (including a &quot;%&quot; suffix) falls back to the default.
+      Out-of-range values are clamped. Default: 5
     <p class="Pp"></p>
   </dd>
   <dt id="pidfile"><a class="permalink" href="#pidfile"><b>--pidfile</b>=<i>FILENAME</i></a></dt>
--- a/docs/manpages/man8/xymond_hostdata.8.html
+++ b/docs/manpages/man8/xymond_hostdata.8.html
@@ -43,10 +43,29 @@ Section: Maintenance Commands (8)<br/>Updated: Version 4.3.30: 4 Sep 2019<br/><a
 <p>&#160;</p>
 <h1 class="Sh" id="OPTIONS"><a class="permalink" href="#OPTIONS">OPTIONS</a></h1>
 <dl class="Bl-tag">
+  <dt id="recent-period"><a class="permalink" href="#recent-period"><b>--recent-period</b>=<i>N</i></a></dt>
+  <dd>Sets the length of the save-throttling window, in <b>minutes</b>. Client
+      data for a host is saved at most --recent-count times within this window;
+      further messages are dropped. A value of 0 disables the throttle, so all
+      client data is saved. The value must be a plain integer number of minutes;
+      anything else falls back to the default. Negative values are clamped to 0.
+      Default: 60
+    <p class="Pp"></p>
+  </dd>
+  <dt id="recent-count"><a class="permalink" href="#recent-count"><b>--recent-count</b>=<i>N</i></a></dt>
+  <dd>Sets how many times a host's client data may be saved within the
+      --recent-period window. A value of 0 disables saving entirely. The value
+      must be a plain integer; anything else falls back to the default. Negative
+      values are clamped to 0 and values above 10000 to 10000. Default: 5
+    <p class="Pp"></p>
+  </dd>
   <dt id="minimum-free"><a class="permalink" href="#minimum-free"><b>--minimum-free</b>=<i>N</i></a></dt>
   <dd>Sets the minimum percentage of free filesystem space on the $CLIENTLOGS
       directory. If there is less than N% free space, xymond_hostdata will not
-      save the data. Default: 5
+      save the data. N is a percentage between 0 and 100; 0 disables the
+      free-space check. The value must be a plain integer; anything else
+      (including a &quot;%&quot; suffix) falls back to the default. Out-of-range
+      values are clamped. Default: 5
     <p class="Pp"></p>
   </dd>
   <dt id="debug"><a class="permalink" href="#debug"><b>--debug</b></a></dt>
--- a/lib/misc.c
+++ b/lib/misc.c
@@ -747,3 +747,46 @@ int chkfreespace(char *path, int minblks, int mininodes)
 	return 1;
 }
 
+/*
+ * Parse the value of a numeric "--option=N" argument. The value must be a
+ * plain integer: strtol (not atoi) so overflow is not undefined behaviour,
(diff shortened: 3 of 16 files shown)


SoundGoof:
## Review

### Finding: Guard and document `--recent-count`

The fixed timestamp history stores only 12 entries, but `--recent-count` accepts arbitrary values through `atoi()`. With `--recent-count=13`, `recentcount` can never reach the configured limit, so throttling remains disabled and every message is saved.

Restrict the input to `1`–`12`, reject invalid values, and document the supported range in `xymond_hostdata.8`. Add boundary coverage for `12` and `13`.

bonomani:
Addressed (folded into the commit): the per-host history is sized from `--recent-count`, so any limit works and no 1–12 range needs documenting. Options are strtol-parsed — non-integer values fall back to the defaults, out-of-range values are clamped, `--recent-count=0` keeps its historical never-save meaning — and documented in `xymond_hostdata.8`. Boundary tests: 12 and 13 hold exactly. Restacked onto #286's head.

SoundGoof:
### Validate `--minimum-free` with `parse_int_opt()`

The new parser validates `--recent-period` and `--recent-count`, but the adjacent `--minimum-free` option still uses unchecked `atoi()`.

Current behavior:

- `--minimum-free=junk` becomes `0`, silently disabling free-space protection.
- Negative values also disable protection.
- Values above `100` prevent hostdata saves.

### Suggested changes

1. Parse the option using the new helper:

   ```c
   minlogspace = parse_int_opt(argv[argi], 0, 100, 5);

bonomani:
Fixed (folded into the commit): `--minimum-free` is parsed with `parse_int_opt()` — a non-integer (including a `%` suffix) falls back to the default instead of silently becoming 0, negatives clamp to 0, values above 100 cap; range documented in `xymond_hostdata.8`. `tests/xymond/hostdata-minfree-validate.sh` asserts the parser's verdicts (disk fullness can't be arranged in a test); verified failing with the old `atoi()` and passing after.

SoundGoof (dismissed):
lgtm

SoundGoof (approved):
## Review: 

No findings.

### Changes since last review

Compared previous reviewed head `302a8325` with current head `06201a07` 

The PR patch is functionally unchanged. It was rebased onto newer `main`; the only patch-level difference is that two upstream `sprintf` calls had become `snprintf` before this PR replaces them with checked `join_path()` calls.

### Verification

- Clean merge against current `main`
- Server and web build succeeded
- Full suite: 53 passed, 0 skipped, 0 failed
- All six hostdata regression tests passed
- Editor diagnostics: no errors
- `git diff --check`: clean
- GitHub CI: 5/5 checks green

PR #295, opened 2026-08-05
xymond: match noflap= entries by name, not by prefix

### LINKED ISSUE #293: xymond: noflap= list matches test names by prefix, not by name
`noflap=test1,test2,...` in `hosts.cfg` matches a test name against the list by prefix, not by equality, so a listed entry also suppresses flap detection for every test whose name is a prefix of it.

`xymond/xymond.c`, `isset_noflap()`:

```c
keylen = strlen(testname);
tok = strtok(dstr, ",");
while (tok && (strncmp(testname, tok, keylen) != 0)) tok = strtok(NULL, ",");
```

`keylen` is the length of the *incoming* test name, so `strncmp("ssh", "ssh2", 3)` is 0 and `ssh` matches the entry `ssh2`. The reverse does not happen — `strncmp("ssh2", "ssh", 4)` compares four bytes and differs — so the effect is one-directional: **a listed entry silently covers its own prefixes.**

Measured against `noflap=web,cpu,disk`:

```
cp        -> suppressed (should not be)
we        -> suppressed (should not be)
webserver -> not suppressed (correct)
```

This is not hypothetical. Among the 44 services shipped in `xymonnet/protocols.cfg`, 18 pairs collide:

```
noflap=imaps    also suppresses  imap
noflap=imap2    also suppresses  imap
noflap=pop3     also suppresses  pop
noflap=pop3s    also suppresses  pop, pop3
noflap=ssh1     also suppresses  ssh
noflap=ssh2     also suppresses  ssh
noflap=ftps     also suppresses  ftp
noflap=ldaps    also suppresses  ldap
noflap=smtps    also suppresses  smtp
noflap=nntps    also suppresses  nntp
noflap=telnets  also suppresses  telnet
noflap=amqps    also suppresses  amqp
```

So a host configured `noflap=imaps` also stops flap detection on its `imap` test, with nothing in the configuration to suggest it.

`hosts.cfg(5)` documents the tag as `noflap[=test1,test2,...]` — a list of test names, with no mention of prefix matching.

The fix is `strcmp` instead of `strncmp`, which drops `keylen` as well. It is a behaviour change: hosts relying, knowingly or not, on the prefix effect will start seeing flap detection on the shorter test name again. That is the documented behaviour, but it is worth a `Changes` entry.

Pre-existing; unrelated to the list-truncation bug in #276, which affects the same function.

bonomani:
Fixes #293.

`isset_noflap()` compared a test name against each `noflap=` entry with `strncmp()` over `strlen(testname)`, so a listed entry also covered every test whose name is a prefix of it:

```c
-	keylen = strlen(testname);
-	while (tok && (strncmp(testname, tok, keylen) != 0)) tok = strtok(NULL, ",");
+	while (tok && (strcmp(testname, tok) != 0)) tok = strtok(NULL, ",");
```

The effect was one-directional and silent. Among the 44 services in `xymonnet/protocols.cfg`, 18 pairs collide — `noflap=imaps` also stopped flap detection on `imap`, `noflap=ssh2` on `ssh`, `noflap=pop3s` on both `pop` and `pop3`, and so on. `hosts.cfg(5)` documents the tag as a list of test names, with no mention of prefix matching.

## Behaviour change

A host whose configuration relied on the prefix effect, knowingly or not, will see flap detection return on the shorter test name. That is the documented behaviour, so this is a fix rather than a regression, but it is a visible change and gets a `Changes` entry.

## Test

`tests/server/xymond-noflap-prefix.sh` extracts `isset_noflap()` from `xymond/xymond.c` at run time and compiles it, so the test drives the shipped function rather than a copy that could drift from it.

Verified both directions — the same harness against the unpatched function:

```
noflap=imaps must not suppress the imap test: got 1, want 0
noflap=ssh2 must not suppress the ssh test: got 1, want 0
FAILED
```

and against the patched one:

```
ALL OK
```

The suite includes the anti-vacuity cases, so the assertions cannot be satisfied by simply never suppressing anything: `noflap=imaps` must still suppress `imaps`, and a bare `noflap` must still suppress everything.

## Note on #276

That pull request changes the same function, for the unrelated bug where `strtok()` truncates the stored list at its first comma. The two touch adjacent lines and will conflict textually; whichever lands second needs a trivial rebase.

This test is written to be independent of that bug: every assertion uses its own host with a single-entry list, so nothing here depends on which order they merge in.

### DIFF
--- a/tests/server/xymond-noflap-prefix.sh
+++ b/tests/server/xymond-noflap-prefix.sh
@@ -1,0 +1,119 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/server/xymond-noflap-prefix.sh
+#
+# Regression guard for issue #293: the hosts.cfg "noflap=test1,test2,..." list
+# matched test names by prefix rather than by name, so a listed entry silently
+# covered every test whose name is a prefix of it. Among the 44 services in
+# xymonnet/protocols.cfg, 18 pairs collide this way -- "noflap=imaps" also
+# silenced imap, "noflap=ssh2" also silenced ssh, and so on.
+#
+# isset_noflap() is static in xymond/xymond.c and xymond has no library form,
+# so the function is extracted from the real source at run time and compiled
+# here. That way the test drives the shipped code rather than a copy of it,
+# and cannot drift from it.
+#
+# Each assertion gets its own host, deliberately: isset_noflap() tokenises the
+# pointer xmh_item() returns, so a list with a comma is truncated in the host
+# record by the first evaluation (fixed separately in #276). Single-entry
+# lists and one host per query keep this test independent of that.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+ROOT=$(find_root)
+
+XYMOND_C="$ROOT/xymond/xymond.c"
+[ -f "$XYMOND_C" ] || skip "xymond/xymond.c not present in this checkout"
+
+CC=${CC:-cc}
+command -v "$CC" >/dev/null 2>&1 || skip "no C compiler available (CC=$CC)"
+
+[ -f "$ROOT/include/config.h" ] && [ -f "$ROOT/lib/libxymoncomm.a" ] \
+	|| skip "tree not built (run make first; the post-build CI suite covers this)"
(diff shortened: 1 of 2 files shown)


SoundGoof:
this needs a rebase

bonomani:
Rebased

SoundGoof (approved):
lgtm

PR #296, opened 2026-08-05
lib/loadhosts: drop the duplicate XMH_CLIENTALIAS name registration

### LINKED ISSUE #294: lib/loadhosts: XMH_CLIENTALIAS name is registered twice
`xmh_item_list_setup()` in `lib/loadhosts.c` registers the name of `XMH_CLIENTALIAS` twice.

`lib/loadhosts.c:87-88`, with its key, among the tag-backed items:

```c
xmh_item_key[XMH_CLIENTALIAS]          = "CLIENT:";
xmh_item_name[XMH_CLIENTALIAS]         = "XMH_CLIENTALIAS";
```

and again at `lib/loadhosts.c:203`, in the block that follows:

```c
xmh_item_name[XMH_IP]                  = "XMH_IP";
xmh_item_name[XMH_CLIENTALIAS]         = "XMH_CLIENTALIAS";
xmh_item_name[XMH_HOSTNAME]            = "XMH_HOSTNAME";
```

That second block is for items that have a **name but no key** — values derived at runtime rather than parsed from a `hosts.cfg` tag. The self-check right after it depends on exactly that split:

```c
i = 0; while (xmh_item_key[i]) i++;
if (i != XMH_IP) errprintf("ERROR: Setup failure in xmh_item_key position %d\n", i);
```

`XMH_CLIENTALIAS` does have a key (`CLIENT:`), so it belongs in the first block only. The line at :203 is a stray.

No consequence today: both assignments store the same string, so the second is a no-op, and it does not disturb the key-contiguity check, which reads only `xmh_item_key[]`. Removing it is a cleanup, not a fix.

Worth noting for whoever picks this up: a runtime round-trip check over the name table cannot see this — the entry resolves back to its own slot perfectly well. Only reading the source, or a duplicate-assignment check, finds it.

bonomani:
Fixes #294.

`xmh_item_list_setup()` registered the name of `XMH_CLIENTALIAS` twice: at `lib/loadhosts.c:88` alongside its `CLIENT:` key, and again at `:203` inside the block for items that have a **name but no key** — values derived at run time rather than parsed from a `hosts.cfg` tag.

`XMH_CLIENTALIAS` has a key, so the second line was a stray. The split between the two blocks is not cosmetic; the self-check immediately below depends on it:

```c
i = 0; while (xmh_item_key[i]) i++;
if (i != XMH_IP) errprintf("ERROR: Setup failure in xmh_item_key position %d\n", i);
```

## No functional change

Both lines stored the same string, so the second was a no-op, and it never touched `xmh_item_key[]`, which is what that check reads.

Verified rather than assumed: I dumped the full name table and the `xmh_key_idx()` resolution for all 71 entries from a build before and after the change. The two dumps are identical, and the contiguity check stays silent.

```
=== diff of the tables ===
IDENTIQUES (71 entrées)
```

No test is added. There is no behaviour to pin — the change removes a dead assignment — and a runtime round-trip check cannot see this class of duplicate anyway, since the entry resolves back to its own slot either way. Only reading the source finds it.

### DIFF
--- a/lib/loadhosts.c
+++ b/lib/loadhosts.c
@@ -200,7 +200,6 @@ static void xmh_item_list_setup(void)
 	xmh_item_name[XMH_ACCEPT_ONLY]         = "XMH_ACCEPT_ONLY";
 
 	xmh_item_name[XMH_IP]                  = "XMH_IP";
-	xmh_item_name[XMH_CLIENTALIAS]         = "XMH_CLIENTALIAS";
 	xmh_item_name[XMH_HOSTNAME]            = "XMH_HOSTNAME";
 	xmh_item_name[XMH_PAGENAME]            = "XMH_PAGENAME";
 	xmh_item_name[XMH_PAGEPATH]            = "XMH_PAGEPATH";


SoundGoof (approved):
lgtm

PR #297, opened 2026-08-06
ci: update actions/checkout to v7

SoundGoof:
## Summary

Update `actions/checkout` from v6 to v7 in all GitHub Actions workflows:

- Build workflow
- Shell regression test workflow
- Release tarball workflow

`softprops/action-gh-release@v3` is already on its latest major release and remains unchanged.

## Motivation

`actions/checkout` v7 is the current stable major release. It moves the action to the supported Node.js 24 runtime and includes current dependency, compatibility, and security updates.

## Security implications

Checkout v7 includes additional safeguards around unsafe pull-request checkouts and incorporates the latest security hardening from the action maintainers.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,7 +34,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v6
+        uses: actions/checkout@v7
 
       - name: Install build dependencies
         run: |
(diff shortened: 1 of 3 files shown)


bonomani (approved):
Verified before approving.

`actions/checkout` v7 exists (v7.0.1, 20 July) and its breaking change is narrow: it blocks checking out fork PRs for `pull_request_target` and `workflow_run`. Neither trigger is used here — `build.yml` and `tests.yml` run on `push`/`pull_request`, and `workflow-release.yaml` on tag pushes — so nothing in this repo is affected.

These are also the only three `actions/checkout@` references in the tree, so the update is complete. And because they are `pull_request` events, the green checks ran the workflow files from this branch, exercising v7 rather than v6.

PR #299, opened 2026-08-06
lib/timefunc: make timestr2timet() correct and non-destructive

### LINKED ISSUE #298: NOTBEFORE:/NOTAFTER: do not work - timestr2timet() reads an uninitialised struct tm and corrupts the host record
`timestr2timet()` in `lib/timefunc.c:536` converts the `YYYYMMDDHHMM` timestamps of the `NOTBEFORE:` and `NOTAFTER:` hosts.cfg tags. Those two tags are its only callers (`lib/loadhosts.c:339` and `:343`), so both defects below are confined to them — but for those tags the feature does not work.

```c
time_t timestr2timet(char *s)
{
	/* Convert a string "YYYYMMDDHHMM" to time_t value */
	struct tm tm;

	if (strlen(s) != 12) {
		errprintf("Invalid timestring: '%s'\n", s);
		return -1;
	}

	tm.tm_min = atoi(s+10); *(s+10) = '\0';
	tm.tm_hour = atoi(s+8); *(s+8) = '\0';
	tm.tm_mday = atoi(s+6); *(s+6) = '\0';
	tm.tm_mon = atoi(s+4) - 1; *(s+4) = '\0';
	tm.tm_year = atoi(s) - 1900; *(s+4) = '\0';
	tm.tm_isdst = -1;
	return mktime(&tm);
}
```

## 1. `struct tm` is never initialised, so the result is arbitrary

Six fields are assigned: `tm_min`, `tm_hour`, `tm_mday`, `tm_mon`, `tm_year`, `tm_isdst`. **`tm_sec` is not.** `mktime()` reads it and normalises out-of-range values, so whatever is on the stack shifts the result without bound — not by seconds, by years.

Measured, converting `202001011200`:

```
expected   2020-01-01 12:00:00
actual     1995-12-20 17:57:17
```

and the value moves as the stack changes. In one run I got 1955-04-22 with a clean stack and 2046-09-03 after dirtying it, from the same input string.

The practical effect on a host carrying `NOTBEFORE:`: `lib/loadhosts.c:414` and `:460` hide the host when `notbefore > now`, so depending on stack residue the host is either permanently visible or permanently invisible — and it can flip between builds. A tag whose documented purpose is to schedule a host in and out of monitoring instead does something unpredictable.

`memset(&tm, 0, sizeof(tm))` before the assignments is enough. Verified against the same input: the result becomes `2020-01-01 12:00:00` and is stable across runs.

## 2. It writes into the caller's string, corrupting the host record

Each `*(s+N) = '\0'` truncates the string in place. `lib/loadhosts.c:339` passes it the pointer `xmh_item()` returned, which points **into the host record's own tag buffer** — the same aliasing that caused #276.

Loading a host defined as:

```
127.0.0.1 nb.example.com # conn NOTBEFORE:202001011200 NOTAFTER:203001011200
```

leaves the stored record as:

```
XMH_RAW : conn|NOTBEFORE:2020|NOTAFTER:2030
```

The parse itself survives, because each field is read before the NUL that follows it is written. What is destroyed is the stored tag, so everything that reports raw tags shows a mutilated value: `XMH_RAW`, `hostinfo` responses, the info page's "Other tags:" row, `xymongrep`.

Copying before tokenising fixes it, as in `xymond_client.c`'s `want_msgtype()`, or the function could parse without writing at all — `sscanf(s, "%4d%2d%2d%2d%2d", ...)` needs no truncation.

## Note

The trailing `*(s+4) = '\0'` on the `tm_year` line is a duplicate of the one on the line above it; it writes a NUL that is already there.

Both defects predate any recent change. Documenting these two tags in #280 is what led me to check them.

bonomani:
Fixes #298.

`NOTBEFORE:` and `NOTAFTER:` did not work. `timestr2timet()` — whose only callers are those two tags (`lib/loadhosts.c:339` and `:343`) — had two defects.

## 1. The converted time was arbitrary

`struct tm` was never initialised, and `tm_sec` was never assigned. `mktime()` reads it and normalises out-of-range values, so whatever was on the stack moved the result without bound. Converting `202001011200`:

```
expected   2020-01-01 12:00:00
measured   1995-12-20 17:57:17     (1955-04-22 with a clean stack, 2046-09-03 after dirtying it)
```

`lib/loadhosts.c:414` and `:460` hide a host when `notbefore > now`, so a host carrying the tag was either permanently visible or permanently invisible, decided by stack residue rather than by its timestamp.

## 2. It destroyed the tag it parsed

Each `*(s+N) = '\0'` truncated the string in place, and `loadhosts.c` passes the pointer `xmh_item()` returned — which points into the host record's own tag buffer. The same aliasing that caused the noflap bug in #276.

```
127.0.0.1 nb.example.com # conn NOTBEFORE:202001011200 NOTAFTER:203001011200

XMH_RAW : conn|NOTBEFORE:2020|NOTAFTER:2030
```

The parse survived, because each field was read before the NUL after it was written. What was destroyed was the stored tag, so everything reporting raw tags showed the mutilated value: `XMH_RAW`, `hostinfo` responses, the info page's "Other tags:" row, `xymongrep`.

## 3. It accepted what is not a date

Both raised by SoundGoof in review.

`sscanf()` does not enforce the documented format: `%d` skips leading whitespace and accepts a sign, so twelve characters that are not twelve digits still convert five fields.

```
"2020 1011200"  ->  2020-10-11 20:00     the space shifts every field
"-20200101120"  ->  year -202
"20200101120+"  ->  the trailing sign is ignored
```

`strlen()` and `strspn()` answer different questions - how long, and made of what - so both are needed to say *exactly twelve digits*, which is what `hosts.cfg(5)` documents.

`mktime()` then normalises a date that does not exist instead of refusing it:

```
202013011200  ->  2021-01-01     month 13
202002311200  ->  2020-03-02     31 February
202102291200  ->  2021-03-01     29 February outside a leap year
202001012500  ->  2020-01-02     hour 25
```

A `NOTAFTER:` of `202002311200` would not have been rejected - the host would have disappeared two days after the date its administrator wrote. Same class of defect as the two above: a wrong value accepted in silence.

`mktime()` rewrites the struct while normalising, so comparing the fields back against what was parsed separates the two cases. 29 February 2020 exists and still converts; 2021 does not and no longer does.

## The fix

`sscanf()` settles both at once — it reads the fields without writing to the source — and an explicit `memset()` leaves nothing for `mktime()` to normalise.

The duplicated `*(s+4) = '\0'` on the old `tm_year` line goes with it.

## Test

`tests/server/notbefore-notafter.sh` drives the real library through its public interface. Against the unpatched code:

```
timestr2timet does not write into the string it was given: ...
a host inside its window is visible: ...
loading a host leaves its NOTBEFORE/NOTAFTER tags intact: (no record)
FAILED
```

and with the fix applied, `ALL OK`.

Worth stating: only the record assertion fails *deterministically* against the old code. The window assertions depend on what was on the stack, which is the bug — so the record one is what anchors the test, and the header says so.

The fixture uses 1990 and 2090 so the test does not rot, and includes an untagged host as a control, so the two "must be hidden" assertions cannot be satisfied by hiding everything.

Five further assertions cover the validation: a signed timestamp, a spaced one, an impossible month, an impossible day, and a real leap day as the control that stops the other four from being satisfied by refusing everything. Unlike the window assertions, all four failure cases fail deterministically against the code before the fix.

### DIFF
--- a/lib/timefunc.c
+++ b/lib/timefunc.c
@@ -537,19 +537,45 @@ time_t timestr2timet(char *s)
 {
 	/* Convert a string "YYYYMMDDHHMM" to time_t value */
 	struct tm tm;
+	int year, mon, day, hour, min;
+	time_t result;
 
-	if (strlen(s) != 12) {
+	/* sscanf() does not enforce the format: "%d" skips whitespace and takes a
+	 * sign. strlen() for the length, strspn() for the digits. */
+	if ((strlen(s) != 12) || (strspn(s, "0123456789") != 12)) {
 		errprintf("Invalid timestring: '%s'\n", s);
 		return -1;
 	}
 
-	tm.tm_min = atoi(s+10); *(s+10) = '\0';
-	tm.tm_hour = atoi(s+8); *(s+8) = '\0';
-	tm.tm_mday = atoi(s+6); *(s+6) = '\0';
-	tm.tm_mon = atoi(s+4) - 1; *(s+4) = '\0';
-	tm.tm_year = atoi(s) - 1900; *(s+4) = '\0';
+	/* sscanf(), not atoi() over truncated copies: the callers pass the pointer
+	 * xmh_item() returned, which aliases the host record's own tag buffer. */
+	if (sscanf(s, "%4d%2d%2d%2d%2d", &year, &mon, &day, &hour, &min) != 5) {
+		errprintf("Invalid timestring: '%s'\n", s);
+		return -1;
+	}
+
+	/* mktime() reads every field, and normalizes an uninitialized tm_sec into
+	 * the result. */
+	memset(&tm, 0, sizeof(tm));
+	tm.tm_year = year - 1900;
+	tm.tm_mon = mon - 1;
+	tm.tm_mday = day;
+	tm.tm_hour = hour;
+	tm.tm_min = min;
 	tm.tm_isdst = -1;
-	return mktime(&tm);
+
+	result = mktime(&tm);
+	if (result == -1) return -1;
+
+	/* mktime() turns 31 February into 2 March rather than failing, rewriting
+	 * the struct as it goes - compare the fields back to catch it. */
+	if ((tm.tm_year != year - 1900) || (tm.tm_mon != mon - 1) ||
+	    (tm.tm_mday != day) || (tm.tm_hour != hour) || (tm.tm_min != min)) {
+		errprintf("Invalid timestring: '%s'\n", s);
+		return -1;
+	}
+
+	return result;
 }
 
 
--- a/tests/server/notbefore-notafter.sh
+++ b/tests/server/notbefore-notafter.sh
@@ -1,0 +1,173 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/server/notbefore-notafter.sh
+#
+# Regression guard for issue #298: the NOTBEFORE:/NOTAFTER: hosts.cfg tags did
+# not work. timestr2timet() converts their YYYYMMDDHHMM timestamps and had two
+# defects, both fixed together:
+#
+#   1. struct tm was never initialized. tm_sec was left holding whatever was on
+#      the stack, and mktime() normalizes it, so the converted time moved by an
+#      unbounded amount -- "202001011200" came out as 1995-12-20 17:57:17 in one
+#      measurement and 2046-09-03 in another. A host carrying NOTBEFORE: was
+#      therefore either permanently visible or permanently invisible, depending
+#      on stack residue.
+#
+#   2. It truncated the string in place with *(s+N) = '\0'. lib/loadhosts.c
+#      hands it the pointer xmh_item() returned, which points into the host
+#      record's own tag buffer, so the stored tag was left as "NOTBEFORE:2020" --
+#      the same aliasing that caused the noflap bug in #276. Everything that
+#      reports raw tags showed the mutilated value: XMH_RAW, hostinfo responses,
+#      the info page's "Other tags:" row, xymongrep.
+#
+# The window assertions below exercise defect 1 and the record assertion
+# exercises defect 2. Only the record assertion fails deterministically against
+# the old code -- the window ones depend on what was on the stack -- so that is
+# the one that anchors this test.
+#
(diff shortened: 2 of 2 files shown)


SoundGoof:
## Proposed changes

`sscanf()` does not enforce the documented `YYYYMMDDHHMM` format. `%d` accepts signs and whitespace, so malformed timestamps can still produce five successful conversions:

```text
2020+1011200 -> 2020-01-01 12:00
2020 1011200 -> 2020-10-11 20:00
```

Validate that the input contains exactly 12 digits before parsing:

```c
if ((strlen(s) != 12) || (strspn(s, "0123456789") != 12)) {
    errprintf("Invalid timestring: '%s'\n", s);
    return -1;
}
```

The checks serve different purposes: `strlen()` validates the complete length, while `strspn()` verifies that all 12 characters are digits.

`mktime()` also normalizes invalid dates:

```text
202013011200 -> 2021-01-01 12:00
202002311200 -> 2020-03-02 12:00
```

Compare the resulting `struct tm` fields with the parsed values after `mktime()` so invalid calendar dates return `-1`.

bonomani:
Both applied in be75afe0. Confirmed each one before changing anything — the second I had missed.

**Twelve digits.** Your form, with the two checks kept for the two questions they answer:

```c
if ((strlen(s) != 12) || (strspn(s, "0123456789") != 12)) {
```

**Impossible dates.** `mktime()` rewrites the struct while normalising, so the fields are compared back against what was parsed, and a mismatch returns -1.

Measured over the vectors, before → after:

| input | was | now |
|---|---|---|
| `2020 1011200` | 2020-10-11 20:00 | refused |
| `-20200101120` | year -202 | refused |
| `202013011200` | 2021-01-01 | refused |
| `202002311200` | 2020-03-02 | refused |
| `202102291200` | 2021-03-01 | refused |
| `202002291200` | 2020-02-29 | 2020-02-29 |

The last row is the one that matters: a real leap day still converts, so this refuses impossible dates rather than merely refusing more.

The test gains five assertions, that leap day among them as the control — without it the other four could be satisfied by refusing everything. All four failure cases fail deterministically against the previous commit, which the window assertions do not.

The PR description also claimed the fix "rejects non-numeric input that atoi() silently read as zero". It did not. Removed, and the two defects you found are now described there.

SoundGoof (approved):
lgtm

PR #301, opened 2026-08-06
xymongen: copy the COMPACT: tag before tokenising it, and guard the class

### LINKED ISSUE #300: xymongen: COMPACT: tag tokenised in place, corrupting the host record
`generate_compactitems()` in `xymongen/loaddata.c:419` tokenises the `COMPACT:` tag in place, through the pointer `xmh_item()` returned:

```c
compacted = xmh_item(xmh, XMH_COMPACT);
if (!compacted) continue;

tok1 = strtok_r(compacted, ",", &savep1);
while (tok1) {
        itm->compactname = strdup(strtok_r(tok1, "=", &savep2));
        members = strtok_r(NULL, "\n", &savep2);
```

That pointer points into the host record's own tag buffer, so both `strtok_r()` calls overwrite separators with NULs and leave the stored tag truncated. It is the same aliasing that caused #276 (noflap) and #298 (NOTBEFORE:/NOTAFTER:).

## No demonstrated symptom, and that is worth stating precisely

I looked for one rather than assuming it. `XMH_COMPACT` has exactly two readers:

- `web/svcstatus.c:189` — a CGI, a separate process with its own `load_hostnames()`, so it never sees xymongen's damage;
- this loop.

Within xymongen the ordering saves it. `xymongen.c` runs `load_layout()` — which walks raw tags at `xymongen/loadlayout.c:654` — **before** `load_state()`, and `generate_compactitems()` is called from `load_state()`. So nothing reads raw tags after the corruption, and the process exits shortly afterwards.

So this produces no wrong output today. What makes it worth closing is that the guarantee is accidental: it rests on the order of two calls in `main()` and on nothing re-reading the tag later. Reorder the passes, add a second consumer, or emit host records at the end of a run, and the damage becomes visible — with nothing in the code to say why.

## Fix

Copy before tokenising, as `xymond/xymond_client.c`'s `want_msgtype()`, `xymond/convertnk.c` and `loaddata.c:114` itself already do:

```c
compacted = xmh_item(xmh, XMH_COMPACT);
if (!compacted) continue;
compacted = strdup(compacted);
...
xfree(compacted);
```

Note that `loaddata.c:114`, twenty lines above in the same file, already uses exactly that pattern for `XMH_NOCOLUMNS` — so this is an inconsistency within one file rather than a missing convention.

Found while auditing every caller that mutates a pointer from the loadhosts API. It is the last such site with no issue of its own; the others are #276 (fixed), #298 (fix in #299), and the xymonnet testspec split described in #278.

bonomani:
Fixes #300. One commit: the fix and the guard for its class.

## The fix

`generate_compactitems()` handed the pointer `xmh_item()` returned straight to `strtok_r()`, truncating the stored `COMPACT:` tag in place. Copy first (`xstrdup`, so allocation failure aborts loudly), free after. No wrong output today, only because `load_layout()` walks raw tags before `load_state()` reaches this function; the guarantee rests on the order of two calls in `main()`, which is what makes it worth closing.

Review added two fixes in the same file: the compact list is now scoped and freed per host (every host used to inherit its predecessors' COMPACT groups and emit their columns), and `unwantedcolumn()` no longer leaks its copy of `XMH_NOCOLUMNS`. Two related bugs in `svcstatus.c`'s COMPACT parsing are split into #306.

## The guard

The same aliasing caused #276, #298 and #300. `tests/packaging/loadhosts-aliasing.sh` hunts the pattern, not the instances: one awk pass per file strips comments and strings, tracks brace depth (first `#if` branch taken), and scans every identifier assigned from the `xmh_item()` family — any shape, spacing, cast, chain or ternary — to the end of its function for a standard-library call that writes through it; `strsep`'s `&v` form and mutators handed the API result directly are matched too. A reassignment ends the scan (comparisons and refetches cannot mask it), except a refetch from the four accessors exactly. Source-only, tracked files only (grep fallback for tarballs); well under a second on gawk, mawk or busybox awk.

A fixture self-test runs first; its expected report is generated from markers, so a scanner regression fails with a one-line diff. With the `xstrdup` removed the scan flags `loaddata.c`; as-is the tree passes.

Blind spots, documented in the header: derived pointers (#278's shape), struct-member targets, assignments split across lines, aliases returned through out-parameters (#306's shape), and non-standard-library mutators (#298's kind — fixed at the function instead, #299). The complete fix is const-qualifying the accessors (194 assignments across 30 files); this narrows the class for a fraction of that.

### DIFF
--- a/tests/packaging/loadhosts-aliasing.sh
+++ b/tests/packaging/loadhosts-aliasing.sh
@@ -1,0 +1,415 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/packaging/loadhosts-aliasing.sh
+#
+# xmh_item(), xmh_item_walk(), xmh_item_byname() and xmh_item_multi() return a
+# pointer INTO the host record, not a copy: a caller that writes through it
+# corrupts the stored configuration for every later reader. That caused #276
+# and #300; the fix is always the same line -- copy before tokenising. This
+# test hunts the pattern, so the class cannot regrow quietly.
+#
+# One awk pass per file: comments and string literals are stripped, brace
+# depth is tracked (first #if branch taken), and every assignment from the
+# API -- any shape, spacing, cast, chain or ternary -- opens a scan segment
+# running to the end of its function. Inside a segment, a standard-library
+# call writing through the variable is reported (strsep via its &v form, a
+# mutator handed the API result directly too). A reassignment ends the
+# segment -- comparisons and refetches cannot mask it -- except a refetch
+# from the API, which keeps the alias. A fixture self-test runs first; its
+# expected report is generated from the fixture's "HIT: <mutator> <var>"
+# markers, so a scanner regression fails with a one-line diff.
+#
+# The mutator list is standard-library only: a project function that mutates
+# its argument is a bug in that function (#298, fixed by #299), not an entry
+# here.
+#
+# KNOWN BLIND SPOTS, on purpose: pointers derived from the alias (#278's
+# strchr shape); struct-member targets; assignments split across lines;
+# aliases handed back through out-parameters (svcstatus.c's loadhostdata()).
+# The complete fix is const-qualifying the accessors -- 194 assignments in
+# 30 files; this narrows the class for a fraction of that.
+#
+# Source-only: no build required.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+ROOT=$(find_root)
+[ -f "$ROOT/lib/loadhosts.c" ] || skip "lib/loadhosts.c not present in this checkout"
+command -v awk >/dev/null 2>&1 || skip "awk not available"
+
+work=$(mktempdir)
+
+# Write through their first argument; read-only friends are absent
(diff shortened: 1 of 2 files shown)


SoundGoof:
## Proposed changes

### Guard misses declaration assignments

The variable extractor only recognizes assignments beginning directly with the variable name:

```c
tag = xmh_item(...);
```

It misses the common declaration form:

```c
char *tag = xmh_item(...);
strtok(tag, ",");
```

This means the aliasing class can return undetected despite the guard’s claim that it checks every variable assigned from the loadhosts API.

Extract the identifier immediately before `=` instead of assuming it begins the line, and add a self-test covering declaration assignments.

bonomani:
Confirmed — and it was worse: the line-anchored extractor silently dropped 31 call sites, though none hid a live offender. Fixed as you proposed; the self-test now generates its expected report from markers in a fixture covering every shape it must catch and every correct shape it must leave alone.

Continued review hardened the scanner into a single comment-stripping, brace-depth-tracking awk pass per file, and turned up neighbouring bugs along the way: the compact list was never reset between hosts and `unwantedcolumn()` leaked its copy (both fixed here), plus two COMPACT-parsing bugs in `svcstatus.c` (split into #306). Everything is now squashed into one commit; over to you.

SoundGoof (approved):
lgtm

PR #305, opened 2026-08-07
xymond: save next client log on request

SoundGoof:
## Summary

This PR should wait for #286 and #289 to merge

Add the administrative command:

`hostdatasave HOSTNAME [MINUTES]`

The command arms a one-shot request that sends the host's next client report through the existing `CLICHG` pipeline for storage by `xymond_hostdata`.

It also provides the synchronous response protocol needed for a future button on the host info page.

## Motivation

This provides an authorized remote trigger for capturing fresh client reports around maintenance.

Operators can capture a report before maintenance, and compare the resulting snapshots to verify that the client server returned to the same state it was in before maintenance.

## Behavior

- Defaults to a 30-minute request lifetime.
- Accepts lifetimes from 1 to 1440 minutes.
- Supports configured hosts before their first client report.
- Consumes the request after one client report.
- Cancels pending requests when a host is dropped or renamed, or when `xymond` restarts.
- Requires in-memory client logging and an active `CLICHG` reader.
- Returns an immediate `OK` or `ERROR` response.
- Marks explicitly requested `CLICHG` messages as forced.
- Forced saves bypass the recent-save quota without consuming quota history.
- Existing free-space protection still applies.

An accepted request confirms that it was armed. It does not guarantee a disk write if the worker subsequently encounters insufficient free space or an I/O error.

## Responses

Successful requests return:

`OK: next client report for HOSTNAME armed for MINUTES minutes`

Requests return a descriptive error for:

- Missing or invalid arguments.
- Lifetimes outside the supported range.
- Unknown hosts.
- Disabled in-memory client logging.
- No active hostdata worker.

## Info Page Integration

The synchronous response protocol prepares this command for a future button on the host info page.

Existing Xymon CGI actions communicate with `xymond` through the internal `sendmessage()` API. They construct normal Xymon protocol commands directly instead of invoking the `xymon` executable.

A future secure CGI action can follow the same pattern:

1. Receive `HOSTNAME` and optional `MINUTES` from a POST form.
2. Apply the existing web authentication and host-access checks.
3. Call `sendmessage()` with `hostdatasave HOSTNAME [MINUTES]`.
4. Capture and display the returned `OK` or `ERROR` response.

No additional daemon protocol or command-line process is required.

## Authorization

The command uses the existing `adminsenders` authorization check.

## Testing

- Built the affected `common` and `xymond` components.
- Added an end-to-end test using the real daemon, channel, worker, and command client.
- Verified success and rejection responses.
- Verified malformed requests receive immediate responses.
- Verified one-shot request consumption.
- Verified forced saves bypass `--recent-count=0`.
- Verified ordinary saves remain subject to the quota.
- Verified the saved payload remains intact.
- Full regression suite: 18 passed, 6 environment-dependent skips, 0 failures.

### DIFF
--- a/common/xymon.1
+++ b/common/xymon.1
@@ -404,8 +404,10 @@ hostdata worker for storage. The request expires after 30 minutes unless the
 optional MINUTES value specifies another lifetime between 1 and 1440 minutes.
 It is held in memory and consumed by the next client message only; dropping or
 renaming the host, or restarting xymond, cancels it. Cached client messages must
-be enabled, and the hostdata worker must be running. The hostdata worker's
-normal recent-save limits apply.
+be enabled, and the hostdata worker must be running. The command reports whether
+the request was armed or rejected. Explicitly requested saves bypass the
+hostdata worker's recent-save limit, but its free-space check still applies;
+therefore an armed response does not guarantee that the report reaches disk.
 
 .IP "\fBping\fR"
 Attempts to contact the Xymon server. If successful, the Xymon server version ID
--- a/common/xymon.1
+++ b/common/xymon.1
@@ -405,9 +405,11 @@ hostdata. The request expires after 30 minutes, or after MINUTES when specified
 canceled when the host is dropped or renamed, or when xymond restarts. Cached
 client messages and the
 hostdata worker must be enabled. Requested saves bypass the recent-save limit,
-but still require sufficient free space. The response reports whether the
-request was armed, not whether the hostdata was written to disk. Hostdata is the
-client data shown as the clientlog service in the web interface.
+but still require sufficient free space. Requests are rejected when no CLICHG
+channel reader is connected. A connected reader does not prove that the
+hostdata worker is running. The response reports whether the request was armed,
+not whether the hostdata was written to disk. Hostdata is the client data shown
+as the clientlog service in the web interface.
 
 .IP "\fBping\fR"
 Attempts to contact the Xymon server. If successful, the Xymon server version ID
--- a/common/xymon.1
+++ b/common/xymon.1
@@ -398,6 +398,15 @@ normal client data.
 Retrieves the current raw client message last sent by HOSTNAME. The optional
 "section" filter is used to select specific sections of the client data.
 
+.IP "\fBhostdatasave HOSTNAME [MINUTES]\fR"
+Requests that the next client message received from HOSTNAME be sent to the
+hostdata worker for storage. The request expires after 30 minutes unless the
+optional MINUTES value specifies another lifetime between 1 and 1440 minutes.
+It is held in memory and consumed by the next client message only; dropping or
+renaming the host, or restarting xymond, cancels it. Cached client messages must
+be enabled, and the hostdata worker must be running. The hostdata worker's
+normal recent-save limits apply.
+
 .IP "\fBping\fR"
 Attempts to contact the Xymon server. If successful, the Xymon server version ID
 is reported.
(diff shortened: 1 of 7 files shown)


bonomani:
Heads-up on an interaction with #289, plus what stock Xymon already covers here.

**Throttle interaction with #289.** Until now the worker's save throttle never actually fired (the off-by-one #289 fixes), so a forced save always reached disk. Once #289 merges the limit is real: if the host already had its `--recent-count` saves in the current window — likely for exactly the busy hosts you'd snapshot around maintenance — `xymond_hostdata` drops the forced message **silently**, and the armed request is still consumed. The operator gets no file and no error. The man-page line "the hostdata worker's normal recent-save limits apply" is accurate but hides that consequence.

Suggestion: append a marker field to the CLICHG message the forced path emits (extra trailing field — existing workers ignore it), and let the worker bypass the throttle for marked messages. Combined with a reply to the sender (armed / bad TTL / unknown host / clientlog disabled), all the command's failure modes become visible. Happy to help with the worker side; on top of #289 it's a few lines.

**What stock Xymon can already do.** For calibration: the snapshot itself needs no daemon change. `xymondboard fields=clntstamp` exposes the client-data timestamp (poll it to catch the *next* report — it's also exactly the filename `xymond_hostdata` uses), and `clientlog HOST` returns the same bytes the worker stores. Tested end-to-end against a live `xymond` from this tree:

```sh
#!/bin/sh
# hostdatasnap HOST [TIMEOUT_MINUTES] -- wait for HOST's next client report and
# save it to $CLIENTLOGS/HOST/<timestamp>, exactly as xymond_hostdata would.
# Run on the Xymon server (e.g. under xymoncmd, so XYMSRV/CLIENTLOGS are set).
HOST=${1:?usage: hostdatasnap HOST [TIMEOUT_MINUTES]}
SRV=${XYMSRV:-127.0.0.1}
DEADLINE=$(( $(date +%s) + ${2:-30} * 60 ))

stamp() { xymon "$SRV" "xymondboard host=^$HOST\$ fields=clntstamp" | head -1; }

armed=$(stamp)
[ -n "$armed" ] || { echo "unknown host: $HOST" >&2; exit 2; }

while cur=$(stamp); [ "$cur" = "$armed" ] || [ "$cur" = "0" ]; do
	[ "$(date +%s)" -lt "$DEADLINE" ] || { echo "timeout: no new report from $HOST" >&2; exit 1; }
	sleep 30
done

mkdir -p "$CLIENTLOGS/$HOST"
xymon "$SRV" "clientlog $HOST" > "$CLIENTLOGS/$HOST/$cur"
echo "saved $CLIENTLOGS/$HOST/$cur"
```

Same one-shot semantics, same default 30-minute lifetime, survives a `xymond` restart (it just keeps polling), can't be throttled, and every outcome is visible (exit 0 saved / 1 timeout / 2 unknown host). Not an argument against the PR — it's what defines its value: what `hostdatasave` adds over this is *remote, adminsenders-authorized triggering*. The marker + sender-reply above are what would make the in-daemon command strictly better than the script instead of quieter than it.

**On the test:** `hostdata-save-next.sh` asserts source-code strings rather than behavior, so it stays green if the logic breaks and goes red on harmless refactors. The pattern from #289 (`tests/lib/build-worker.sh`: build the real binary, drive it, assert on output) may be adaptable, though I realize `xymond` is much heavier to stand up than a channel worker.

bonomani (approved):
Approved. I reviewed the hostdatasave command path, the forced CLICHG message handling, and the hostdata worker quota bypass. Built the PR locally and verified the focused tests plus the full regression suite.\n\nLocal verification: ./tests/xymond/hostdata-forced-save.sh passed; ./tests/server/hostdata-save-next.sh passed; ./tests/testsuite passed with 54 passed, 1 skipped, 0 failed; git diff --check passed.

Issue #187, opened 2026-06-20
xymond/combostatus.c: stack buffer overflow via unbounded sprintf into errtext[1024]


bonomani:
### Summary

`evaluate()` in `xymond/combostatus.c` formats an error message into a fixed
1 KB stack buffer with an unbounded `sprintf()`, while the formatted expression
can be up to `MAX_LINE_LEN` (16 KB). A sufficiently long combo expression that
triggers a `compute()` error overflows the stack buffer.

### Details

```c
static long evaluate(char *symbolicexpr, char **resultexpr, value_t **valuelist, char **errbuf)
{
        char expr[MAX_LINE_LEN];     /* 16384 bytes (lib/stackio.h) */
        ...
        char errtext[1024];
        ...
        if (error) {
                sprintf(errtext, "compute(%s) returned error %d\n", expr, error);   /* <-- overflow */
                ...
        }
}
```

`expr` is built from `symbolicexpr` (the combo test expression read from
`combo.cfg`) and may be far longer than `errtext`'s 1024 bytes. When `compute()`
returns an error, `sprintf()` writes `"compute(" + expr + ") returned error N\n"`
into the 1 KB buffer, overflowing the stack for any erroring expression longer
than ~990 characters.

This is a pre-existing defect, independent of any current network-test work; it
was noticed during an unrelated clean build.

### Impact

Stack buffer overflow driven by `combo.cfg` content. Configuration is normally
admin-controlled, so this is primarily a robustness/correctness bug rather than
a remotely exploitable one, but a long combo expression that fails to compute
corrupts the stack and can crash `xymond` (the combostatus module).

### Suggested fix

Use `snprintf(errtext, sizeof(errtext), ...)`.

The other `errtext` buffer in `getvalue()` (written by `getxymondvalue()` via a
pointer) was checked and is **not** affected: its format substitutes only a
fixed string plus a small `%d`, so the write is bounded well under 1 KB.

### Location

`xymond/combostatus.c` — `sprintf(errtext, ...)` in `evaluate()` (around line 355).

PR #307, opened 2026-08-08
tests: guard seven under-tested behaviours (config, matching, CGI, crash/overflow fixes)

bonomani:
A regression audit of recently merged behavioural changes against `tests/`
surfaced user-visible features and fixes that shipped without a guard. This
adds one focused test per item (one commit each), each pinning the actual edge
case rather than a smoke path, and each verified against a pre-fix build/source
so the assertion provably catches the regression. All run under
`./tests/testsuite` and skip cleanly on an unbuilt tree.

### Behavioural (compile/run and check output)

**1. FILE `IFEXIST` alias for `OPTIONAL` (#161)** —
`tests/server/analysis-file-ifexist.sh` drives `xymond_client --dump-config`
over a FILE rule written three ways; asserts `IFEXIST` folds onto the same
`OPTIONAL` flag, `OPTIONAL` is unchanged, and a bare FILE rule is *not*
`OPTIONAL` (so the alias assertion can't pass vacuously).

**2. analysis.cfg first-match for DS rules (#32)** —
`check_rrdds_thresholds()` has no CLI, so `analysis-ds-firstmatch.sh` compiles a
harness against the real `xymond/client_config.c` and counts `modify` lines per
RRD key. Three scenarios pin both sides: a non-triggering specific rule shadows
a later same-colour rule (0, was 1); yellow + red on one dataset both still fire
(2, shadow stays per-colour); two same-colour rules resolve to the first (1, was
2).

**3. Shipped `.d` drop-in directories (#222)** —
`config-include-dir.sh` asserts every server/client `.DIST` still declares its
`optional directory @XYMONHOME@/etc/<name>.d`, then drives the real stackio
reader built `STANDALONE` to show a present dir merges its fragments, a missing
`optional directory` is silent, and a missing non-optional `directory` still
warns (proving the silence is the `optional` keyword).

**4. svcstatus crash on missing `SCRIPT_NAME` (f01d07dfa)** —
`tests/web/svcstatus-scriptname.sh` runs the built `svcstatus.cgi` with
`SCRIPT_NAME` unset and asserts it is not killed by a signal. `parse_query()`
touches `SCRIPT_NAME` before any host lookup, so no server is needed; the
unguarded `strlen(NULL)` segfaults (139) while the fix exits 1.

**5. namematch() case-insensitive name lists (#182)** —
`tests/server/namematch-case.sh` links libxymon and calls `namematch()` directly
on the plain-list branch #182 switched from `strcmp()` to `strcasecmp()`: both
case directions, a mixed-case token in a comma list, negation, and a genuine
non-match (so it can't pass by matching everything).

### Hybrid source-bind + compiled property demo (static targets, following `digest-md5hash.sh`)

**6. .netrc password off-by-one (#226)** —
`tests/network/netrc-password.sh` binds to `url.c` (corrected `login_len + 1`
present, the one-short `login_len` gone) and compiles a faithful copy of the
pack step showing the fixed size keeps the whole `login:password` while the
short size drops the last character. `load_netrc()` is static.

**7. combostatus evaluate() stack overflow (#187)** —
`tests/server/combostatus-overflow.sh` binds to `combostatus.c` (bounded
`snprintf` present, unbounded `sprintf` into `errtext[1024]` gone) and compiles
a faithful copy showing a 16 KB expression is bounded into the 1 KB buffer with
a trailing canary intact. `evaluate()` is static; the unbounded form is not
executed — it is the overflow under test.

### DIFF
--- a/tests/network/netrc-password.sh
+++ b/tests/network/netrc-password.sh
@@ -1,0 +1,95 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/network/netrc-password.sh
+#
+# Regression guard for the off-by-one in load_netrc() that dropped the last
+# character of every .netrc password (commit b4c4775dc).
+#
+# load_netrc() packs "login:password" into item->auth. It allocated the right
+# size -- malloc(login_len + 1), where login_len = strlen(login) + strlen(pw)
+# + 1 for the ':' -- but passed login_len (one byte short) as snprintf()'s
+# size, so snprintf's own NUL landed on the password's last character and
+# silently truncated it. The fix passes login_len + 1.
+#
+# load_netrc() is static and pulls the whole URL machinery in, so -- like
+# tests/server/digest-md5hash.sh -- this (1) binds to the real source: the
+# corrected size argument must be present and the short one gone, and (2)
+# compiles a faithful copy of the pack step and shows the fixed size yields the
+# whole password while the buggy size truncates it. Skips if the file or a C
+# compiler is absent.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+ROOT=$(find_root)
+SRC="$ROOT/lib/url.c"
+CC=${CC:-cc}
+
+[ -f "$SRC" ] || skip "lib/url.c absent"
+src=$(cat "$SRC")
+
+# (1) Bind to the real code: fixed size argument present, buggy one gone. The
+# fixed line is `snprintf(item->auth, login_len + 1, ...)`; the bug was
+# `snprintf(item->auth, login_len, ...)`. The " + 1," / "," suffixes keep the
+# two apart so the fixed form cannot satisfy the buggy pattern.
+assert_contains "snprintf(item->auth, login_len + 1," "$src" \
+	"url.c load_netrc() lost the corrected snprintf size (login_len + 1) (#226 / b4c4775dc)"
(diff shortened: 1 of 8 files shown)


SoundGoof:
## Findings

### Medium: netrc regression guard misses equivalent truncation

`tests/network/netrc-password.sh:37` only verifies the `snprintf(..., login_len + 1, ...)` argument. Its behavioral probe is a handwritten copy with its own correct `login_len` calculation.

Changing production from:

`strlen(login) + strlen(password) + 1`

to:

`strlen(login) + strlen(password)`

recreates the one-character truncation, but the test still exits 0. Exercise the real packing code or also bind the guard to the production length calculation.

### Medium: drop-in guard accepts missing or commented declarations

`tests/server/config-include-dir.sh:41-43` skips each missing `.DIST` file and checks the directive as an unanchored substring.

I verified that both deleting `alerts.cfg.DIST` and commenting its declaration out still produce a passing test. Since the test claims to guard every shipped declaration, expected files should be required and directives should match an active complete line.

## Validation

- Exact PR head built successfully.
- Full suite: 30 passed, 0 skipped, 0 failed.
- GitHub CI is green.
- ShellCheck found no substantive errors.

bonomani:
Both valid — thanks, fixed in b67fb04aa.

- netrc: now exercises the real code — parse_url() runs load_netrc() and I check the packed url.auth, so dropping the +1 from the length calc (not just the snprintf size) fails. Kept a build-independent source check as a first layer too.
- drop-in guard: requires every expected .DIST whose directory is present and matches an active whole-line declaration, so a deleted file or a commented-out directive now fails.

SoundGoof (approved):
lgtm

PR #308, opened 2026-08-08
xymond_history: validate command-line options

SoundGoof:
## Summary

Harden `xymond_history` command-line option handling.

## Changes

- Parse `--minimum-free` with `strtol()`:
  - accept `0..100`
  - use the default of `5` for invalid or negative values
  - cap values above `100`
  - emit warnings for fallback and capping
- Replace the fixed `PATH_MAX` PID-file buffer and unchecked `strcpy()` with dynamic storage, matching `xymond`.
- Reject empty `--histdir=`, `--histlogdir=`, and `--pidfile=` values.
- Bound construction of the `allevents` path.
- Report and ignore unknown options, matching `xymond`.
- Document the validation behavior under each affected option.

## Impact

Previously:

- `--minimum-free=junk` or a negative value silently became `0`, disabling free-space protection.
- Values above `100` prevented historical status logs from being saved.
- A long `--pidfile` value could overflow the fixed buffer.
- Unknown options were silently ignored.

The default configuration and valid option values retain their existing behavior.

## Compatibility

This can merge independently of #219. When #219 is rebased, the default PID directory can be changed from `XYMONSERVERLOGS` to `XYMONRUNDIR` while retaining dynamic PID-path storage.

## Testing

`tests/server/xymond-history-options.sh` drives the real built worker and covers:

- junk, negative, excessive, and overflowing `--minimum-free` values
- valid boundaries `0` and `100`
- unknown options and exact `--debug` matching
- empty history, history-log, and PID-file values
- a 5000-character PID-file value

Validation:

- Focused test passes.
- Full suite: 22 passed, 2 RRD-dependent tests skipped.
- The regression test fails against unpatched `main`.

### DIFF
--- a/docs/manpages/man8/xymond_history.8.html
+++ b/docs/manpages/man8/xymond_history.8.html
@@ -41,34 +41,38 @@ Section: Maintenance Commands (8)<br/>Updated: Version 4.3.30: 4 Sep 2019<br/><a
 <dl class="Bl-tag">
   <dt id="histdir"><a class="permalink" href="#histdir"><b>--histdir</b>=<i>DIRECTORY</i></a></dt>
   <dd>The directory for the history files. If not specified, the directory given
-      by the XYMONHISTDIR environment is used.
+      by the XYMONHISTDIR environment is used. An empty directory name produces
+      an error and prevents xymond_history from starting.
     <p class="Pp"></p>
   </dd>
   <dt id="histlogdir"><a class="permalink" href="#histlogdir"><b>--histlogdir</b>=<i>DIRECTORY</i></a></dt>
   <dd>The directory for the historical status-logs. If not specified, the
-      directory given by the XYMONHISTLOGS environment is used.
+      directory given by the XYMONHISTLOGS environment is used. An empty
+      directory name produces an error and prevents xymond_history from
+      starting.
     <p class="Pp"></p>
   </dd>
   <dt id="minimum-free"><a class="permalink" href="#minimum-free"><b>--minimum-free</b>=<i>N</i></a></dt>
   <dd>Sets the minimum percentage of free filesystem space on the $XYMONHISTLOGS
       directory. If there is less than N% free space, xymond_history will not
       save the detailed status-logs. N is a percentage between 0 and 100; 0
-      disables the free-space check. The value must be a plain integer; anything
-      else (including a &quot;%&quot; suffix) falls back to the default.
-      Out-of-range values are clamped. Default: 5
+      disables the free-space check. The value must be a plain integer; invalid
+      or negative values produce a warning and use the default of 5. Values
+      above 100 produce a warning and are capped at 100.
     <p class="Pp"></p>
   </dd>
   <dt id="pidfile"><a class="permalink" href="#pidfile"><b>--pidfile</b>=<i>FILENAME</i></a></dt>
   <dd>xymond_history writes the process-ID it is running with to this file. This
       is for use in automated startup scripts. The default file is
-      $XYMONSERVERLOGS/xymond_history.pid.
+      $XYMONSERVERLOGS/xymond_history.pid. An empty filename produces an error
+      and prevents xymond_history from starting.
     <p class="Pp"></p>
   </dd>
   <dt id="debug"><a class="permalink" href="#debug"><b>--debug</b></a></dt>
-  <dd>Enable debugging output.
-    <p class="Pp"></p>
-  </dd>
+  <dd>Enable debugging output.</dd>
 </dl>
+<p class="Pp">Unknown options produce a warning and are ignored.</p>
+<p class="Pp"></p>
 </section>
 <section class="Sh">
 <p>&#160;</p>
--- a/docs/manpages/man8/xymond_history.8.html
+++ b/docs/manpages/man8/xymond_history.8.html
@@ -41,35 +41,35 @@ Section: Maintenance Commands (8)<br/>Updated: Version 4.3.30: 4 Sep 2019<br/><a
 <dl class="Bl-tag">
   <dt id="histdir"><a class="permalink" href="#histdir"><b>--histdir</b>=<i>DIRECTORY</i></a></dt>
   <dd>The directory for the history files. If not specified, the directory given
-      by the XYMONHISTDIR environment is used. An empty directory name produces
-      an error and prevents xymond_history from starting.
+      by the XYMONHISTDIR environment is used.
     <p class="Pp"></p>
   </dd>
   <dt id="histlogdir"><a class="permalink" href="#histlogdir"><b>--histlogdir</b>=<i>DIRECTORY</i></a></dt>
   <dd>The directory for the historical status-logs. If not specified, the
-      directory given by the XYMONHISTLOGS environment is used. An empty
-      directory name produces an error and prevents xymond_history from
-      starting.
+      directory given by the XYMONHISTLOGS environment is used.
     <p class="Pp"></p>
   </dd>
   <dt id="minimum-free"><a class="permalink" href="#minimum-free"><b>--minimum-free</b>=<i>N</i></a></dt>
   <dd>Sets the minimum percentage of free filesystem space on the $XYMONHISTLOGS
       directory. If there is less than N% free space, xymond_history will not
       save the detailed status-logs. N is a percentage between 0 and 100; 0
-      disables the free-space check. The value must be a plain integer; invalid
-      or negative values produce a warning and use the default of 5. Values
-      above 100 produce a warning and are capped at 100.
+      disables the free-space check. The value must be a plain integer. Invalid
+      values produce a warning and use the default of 5. Negative values produce
+      a warning and are clamped to 0, preserving their historical meaning of
+      disabling the check. Values above 100 produce a warning and are capped at
+      100.
     <p class="Pp"></p>
   </dd>
   <dt id="pidfile"><a class="permalink" href="#pidfile"><b>--pidfile</b>=<i>FILENAME</i></a></dt>
   <dd>xymond_history writes the process-ID it is running with to this file. This
       is for use in automated startup scripts. The default file is
-      $XYMONSERVERLOGS/xymond_history.pid. An empty filename produces an error
-      and prevents xymond_history from starting.
+      $XYMONSERVERLOGS/xymond_history.pid. If the file cannot be opened or
+      written, a warning identifies the path and history processing continues.
     <p class="Pp"></p>
   </dd>
-  <dt id="debug"><a class="permalink" href="#debug"><b>--debug</b></a></dt>
-  <dd>Enable debugging output.</dd>
+  <dt id="debug"><a class="permalink" href="#debug"><b>--debug</b>[=VALUE]</a></dt>
+  <dd>Enable debugging output. Suffixed forms such as <b>--debug=1</b> are
+      accepted for compatibility.</dd>
 </dl>
 <p class="Pp">Unknown options produce a warning and are ignored.</p>
 <p class="Pp"></p>
--- a/tests/server/xymond-history-options.sh
+++ b/tests/server/xymond-history-options.sh
@@ -8,34 +8,53 @@ set -euo pipefail
 require_bin XYMOND_HISTORY xymond/xymond_history
 
 work=$(mktempdir)
-mkdir -p "$work/hist"
+mkdir -p "$work/hist" "$work/histlogs"
+input=/dev/null
+allhist=TRUE
+hosthist=FALSE
+statuslogs=FALSE
+pidarg="--pidfile=$work/xymond_history.pid"
+
+printf '%s\n' \
+	'@@stachg#1|1586206051.00000|127.0.0.1|xymond|test.host|disk|1586206351|red|green|1586205451|0||0|0|' \
+	'body' '@@' > "$work/message"
 
 run_history() {
 	set +e
 	output=$(env \
 		XYMONSERVERLOGS="$work" \
-		XYMONALLHISTLOG=TRUE \
-		XYMONHOSTHISTLOG=FALSE \
-		SAVESTATUSLOG=FALSE \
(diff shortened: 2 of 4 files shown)


bonomani:
## Review

Ran the built binary rather than reading only the diff. One crash reproduces; the rest are behaviour changes on upgrade paths, plus some gaps in the new test.

### Blocking: shutdown crashes when XYMONALLHISTLOG=FALSE

`fclose(alleventsfd)` at the end of `main()` is unguarded, and with `save_allevents == 0` the handle is never assigned (stays NULL from line 93):

```
$ XYMONSERVERLOGS=$W XYMONALLHISTLOG=FALSE XYMONHOSTHISTLOG=FALSE SAVESTATUSLOG=FALSE \
    xymond/xymond_history --histdir=$W/hist --pidfile=$W/p.pid </dev/null
$ echo $?
134
```

`strace` shows `SIGSEGV si_addr=NULL` immediately after the pid file is written, converted to SIGABRT by xymon's segv handler.

This lands on the PR because the new `unlink(pidfile); xfree(pidfile);` sit on the next two lines and never run — so the shutdown that was meant to clean up the pid file instead dumps core and leaves it behind (`p.pid` still present afterwards). The same function already knows the guard is needed at line 247: `if (rotatefiles && alleventsfd)`.

Pre-existing, but the PR makes it matter and is one line from fixing it.

### Behaviour changes that bite on upgrade

**Negative `--minimum-free` reverses meaning.** `parse_int_opt()` clamped `-1` to `0`, and `chkfreespace(path, 0, 0)` returns 0 unconditionally (`lib/misc.c:745`) — so a negative value *disabled* the check. `parse_minimum_free()` now returns the default `5` instead, and on a histlogs filesystem below 5% free the worker sets `logdirfull` and stops writing detailed status logs (line 446). An admin who deliberately disabled the check loses detail logging after an upgrade, with one startup warning. It also makes the two workers disagree about the same option: `xymond_hostdata.c:213` still uses `parse_int_opt` and still clamps `-1` to `0`.

The summary describes this as fixing "silently became 0, disabling free-space protection" — worth confirming that was intended as a behaviour change rather than a bug fix, because `lib/misc.c:756-760` says the clamping exists specifically to keep deployed setups working.

**Empty `--histlogdir=` is now fatal.** With the default `SAVESTATUSLOG=TRUE`, `--histlogdir=$UNSET_VAR` used to cost only the detail-log writes; hist files, allevents and per-host event logs kept working. Now it prints "No history-log directory given, aborting" and returns 1, and xymond_channel re-forks per batch, so the whole history subsystem stops. Same shape for empty `--pidfile=`: a wrapper passing `--pidfile=$PIDFILE` with the variable unset used to run fine and record history; it now exits 1 before reading the channel.

**`--debug=1` no longer enables debug.** `argnmatch()` prefix-matched anything starting with `--debug`; the exact `strcmp()` sends `--debug=1` to the new unknown-option branch, so it prints "Unknown option ... - ignored" and runs with debug off. Silent for whoever is mid-diagnosis.

### PID-file handling

- The new `return 1` for an over-long allevents path fires *after* the pid file was written (lines 199-203), so it exits leaving a stale pid file with a dead PID. A later `xymon stop` signals whatever now owns it. The pre-existing `return 1` at line 218 has the same shape — moving the pid-file write after validation fixes both.
- When `fopen` of the pid file fails, the new else-branch warns and continues, but `unlink(pidfile)` at exit still runs and succeeds if the *directory* is writable — deleting a pid file this process never owned.
- `malloc()` at line 162 and `strdup()` at line 128 are unchecked; the fixed `pidfn[]` buffer they replace had no failure mode. `sprintf` writes to the result on the next line.
- "Cannot open PID file: %s" omits the path. All workers on a channel share `history.log`, so the operator can't tell which path failed — exactly the case the new 5000-character test exercises.

### The bounds check covers one site

`snprintf` for `histdir + "/allevents"` is bounded, but ~20 sibling `sprintf()` calls that build `PATH_MAX` buffers are not — `sprintf(statuslogfn, "%s/%s.%s", histdir, hostnamecommas, testname)` at lines 330/714/829, `sprintf(fname, "%s/%s/%s", histlogdir, hostdash, testname)` at 456, and more. A histdir of `PATH_MAX-15` passes the new check and then overflows on the first status change. Those are the sites taking host- and test-name components; the guarded one takes a constant suffix.

Also: the empty-string `histlogdir` guards don't cover NULL. With `SAVESTATUSLOG=FALSE` and no `--histlogdir` — the configuration the new test uses — `histlogdir` stays NULL and `chkfreespace(NULL, ...)` runs every 300s, so `lib/misc.c:737` does `errprintf("Cannot stat filesystem %s", NULL)`.

### Test coverage

The new file drives the real binary, which is right, but its fixed environment means most of the changed code never executes:

- `SAVESTATUSLOG=FALSE` for every case, so `--histlogdir=` hits the `!save_histlogs` branch. With `SAVESTATUSLOG=TRUE` the same argument takes the *other* new branch ("No history-log directory given, aborting") — the one that runs in production, asserted by nothing.
- every case feeds `</dev/null`, so `get_xymond_message()` returns NULL and the loop exits before line 265, the only place `minlogspace` is used. A parser returning 100 for `--minimum-free=0` would still pass.
- `run_history` always passes `--pidfile=`, so the new default-path block (xgetenv + malloc + sprintf) has no coverage.
- only `--debugging` is asserted as unknown; nothing asserts plain `--debug` still works, which is the regression the `argnmatch`→`strcmp` switch actually risks. Changing the branch to `--debugx` keeps the suite green.

### Note for #219

The pid-file rewrite here is the code #219 will rebase onto, and two findings sit in it: the unchecked allocations, and `unlink()` removing a file that was never opened. The second interacts with moving the default directory from `XYMONSERVERLOGS` to `XYMONRUNDIR`, since the unlink target moves with it. Cheaper to settle here than after the rebase.

SoundGoof:
Thanks for the thorough runtime review. I addressed all findings in
`ba5604a4`.

- Guarded `fclose(alleventsfd)`, so `XYMONALLHISTLOG=FALSE` shuts down
  cleanly and removes its PID file.
- Preserved upgrade behavior for negative `--minimum-free`, empty path
  options, and `--debug=1`.
- Moved PID creation after startup validation.
- Track PID-file ownership, preventing removal of files the process did
  not create.
- Use `xmalloc()`/`xstrdup()` and include the failing PID path in errors.
- Detect PID write and `fclose()` failures.
- Bounded all fixed-buffer history path construction. An overlong path
  now skips only the affected output.
- Avoid free-space checks when detailed history logging is disabled or
  no directory is configured.
- Expanded the test to use real channel messages and cover the parser at
  runtime, default PID handling, debug compatibility, empty paths,
  startup failures, shutdown cleanup, write failures, and oversized
  message-derived paths.

I also repeated the syscall-level checks with `strace`, including your
`XYMONALLHISTLOG=FALSE` reproduction. It now exits `0`, unlinks the PID
file, and receives no `SIGSEGV` or `SIGABRT`. Additional traces confirmed
PID ownership, `/dev/full` write failure handling, empty histlog
compatibility, and independent event writes after rejecting an overlong
status path.

Validation passes on Linux, FreeBSD 14.3, and NetBSD 11.0. All current
GitHub Actions checks are green.

bonomani:
Second review pass over `ba5604a4f`, run as a multi-agent sweep. Everything below I re-checked against the source myself; where the sweep overstated something I say so.

### New in this revision

**`unlink()` on the PID-write failure path deletes a file the worker never created** (`xymond_history.c:218`). The `pidfile_created` flag correctly fixed the exit-time unlink from the last round, but the new write-failure branch is unconditional:

```c
errprintf("Cannot write PID file '%s': %s\n", pidfile, strerror(savederrno));
unlink(pidfile);
```

`fopen("/dev/full","w")` succeeds, the buffered `fprintf` succeeds, `fclose()` fails with ENOSPC → `unlink("/dev/full")`. The new test drives exactly that (`--pidfile=/dev/full`, line 143), so **running the suite as root removes the device node**. It also deletes an operator's existing pidfile target whenever that filesystem is full — and it contradicts the contract the neighbouring assertion states ("worker removed a pidfile it never opened"). The open-failure branch deliberately doesn't unlink; this one should match.

### Crash / cost, adjacent to the diff

- **`fprintf(alleventsfd, ...)` at :619 is unguarded** while `fclose` at :861 is. `alleventsfd` starts NULL (:91) and the post-SIGHUP reopen (:250-251) only `errprintf`s on failure, leaving it NULL with `save_allevents` still 1 — next `@@stachg` segfaults. trimhistory sends that SIGHUP.
- **`rotatefiles` is never cleared.** Only three references exist: declaration (:38), `= 1` in the handler (:51), test at :248. After the first rotation every message pays an `fclose`+`fopen`+`setvbuf` forever, which also multiplies the exposure above.

### Still open from the previous round

- `argnmatch(argv[argi], "--debug")` at :128 still prefix-matches, so `--debugx` enables debug instead of reaching the new unknown-option warning at :132.
- Empty `--histdir=` / `--histlogdir=` are still accepted (only NULL is rejected at :147), so paths root at `/`. The test now *blesses* this — "empty history-log directory must not prevent startup", and `--histdir=` asserted status 0.
- Path components from the channel are still unconfined. `xymond_hostdata.c` uses `confine_name()`/`safe_basename()` for the same job; this PR rewrites all these call sites and is the natural place.
- `format_path()` has no `__attribute__((format(printf,3,4)))`, so converting ~20 `sprintf()` sites silently drops `-Wformat` checking at exactly the sites building paths from message-supplied names.
- The truncation message carries no host/test context and repeats per message.

### Test

- **Sleepless spin** at :127 — `for _ in {1..1000}` with no `sleep` is ~2 ms total; the worker was measured hitting it at iteration 159. On a loaded runner or ASAN build it exhausts and the job goes red spuriously.
- **`/dev/full` is Linux-only**, so that case fails rather than skips on the BSDs and macOS.
- **The truncation case runs with `statuslogs=FALSE`** (set at :106, never re-enabled), so `save_histlogs` is off and the histlog `format_path` chain — 3 call sites, the only one that `mkdir`s — never executes.
- `assert_not_contains "is invalid"` at :70 is vacuous. The string exists in `common/xymonlaunch.c`, but `xymond_history` never emits it, so the `--minimum-free=0/100` boundary cases are unguarded.
- `--minimum-free=200` clamps to 100 rather than falling back to the default, i.e. "never save detail logs".
- Startup aborts on a too-long `$histdir/allevents` even when `XYMONALLHISTLOG=FALSE` and the file is never opened.

### Two the sweep raised that I don't think are defects

- It flagged the `goto statuspathdone` as splitting one event across inconsistent stores. Your test asserts precisely that behaviour is wanted (`"rejected status path suppressed the all-events record"`), so it's a deliberate design choice — worth confirming it's intended, not fixing.
- It called `"is invalid"` absent from the codebase; it's present, just in a different binary. The assertion is still dead for this test.

bonomani:
Re-checked `ba5604a4` by running it, not by reading the list. Every finding I raised holds:

| | measured |
|---|---|
| `XYMONALLHISTLOG=FALSE` shutdown | exit 0, PID file removed, no signal |
| PID file created after startup validation | `--histdir=/nonexistent` → exit 1, **no** PID file left |
| foreign PID file on a refused startup | survives, contents intact |
| PID path that cannot be opened | reported, nothing unlinked, directory intact |
| PID write failure (`/dev/full`) | `Cannot write PID file '/dev/full': No space left on device` |
| negative `--minimum-free`, empty paths, `--debug=1` | all accepted, exit 0 |
| overlong message-derived path | rejected and logged; the next host's history is written normally, in either order |
| `SAVESTATUSLOG=FALSE` | 0 `statfs` calls |
| your own tests + full suite on this branch | 58 passed, 0 skipped, 0 failed |

Two notes, neither blocking:

**The overlong-path case took me a wrong turn worth recording.** My first runs showed the following host writing nothing, which looked like the rejection poisoning the rest of the stream. It was my fixture: every message reused `@@stachg#1`, so everything after the first was dropped for that reason. With distinct sequence numbers the behaviour is what you describe. A control with two ordinary hosts is what settled it.

**`--histlogdir=` is treated as a configured directory**, so `chkfreespace("")` still runs — one `statfs("")` returning `ENOENT` per 300s window. Following that thread: history paths are then built relative to nothing and the daemon does `mkdir("/a_host")`. It fails with `EACCES` only because I ran unprivileged; under root it would create a directory per host at the filesystem root. That is identical on `main`, so it is not yours — you preserved it on purpose, which is the right call for an option-validation change. Filed separately as #369.

Ready from my side. The branch is 30 commits behind `main`; a rebase before merge would also let it run against the current suite (74 tests rather than 58). Worth knowing that #174 touches `xymond_history.c` too, so whichever lands second will need the other's changes taken into account.

bonomani:
Conflicts with #400, #401 and #402 in `xymond/xymond_history.c`.

All four apply cleanly to `main` on their own, so nothing here is broken. #400 #401 #402 are one series — each carries the one before it, so they go in that order — and whichever side merges second needs a rebase before it can go in, which can cost it a fresh review.

The overlap is not small on either side: this changes 104 lines over 44 hunks of that file, #402 changes 172 over 30. Cheapest sequence is the series first, then this rebased once onto the result, rather than the series rebasing three times over a file this has reorganised.

bonomani:
The compatibility note here says:

> This can merge independently of #219.

That is true of each against `main` separately, but not of the two together. Merging both, in either order, conflicts:

```
$ git checkout -B t origin/main && git merge pr219 && git merge pr308
CONFLICT (content): xymond/xymond_history.c
CONFLICT (content): xymond/xymond_history.8
CONFLICT (content): docs/manpages/man8/xymond_history.8.html
```

Both rewrite the same pidfile handling in `xymond_history` — #219 bounds the `strcpy()` on the path the shipped `tasks.cfg` actually passes, this PR validates the options around it. Whichever lands second needs a real resolution in `xymond_history.c`, not a trivial one.

Nothing to change here; the note just reads as a stronger guarantee than it is. #219 has just been rebased onto current `main`, so the conflict is against its current head.

(Noting it because it is the kind of thing #501 exists to catch automatically, and that is still open.)

PR #312, opened 2026-08-08
docs: per-OS parity for the client filesystem filter (cfg.DIST + cfg.5)

bonomani:
## Problem

`xymonclient.cfg.DIST` and `xymonclient.cfg.5` are both shipped identically to every client OS, so OS-specific behavior of the `XYMONCLIENT_FS_*` filter (the only per-OS-variant client config surface) must be scoped *inside* them. Coverage was uneven: **Linux** documented in both, **macOS** only in cfg.DIST, **FreeBSD** (#175) in neither — and cfg.5 described the variables as Linux settings outright, so a FreeBSD/macOS admin reading the manual got another OS's defaults. (Same gap PR #177 just closed for NetBSD, per review there.)

## Change

The split of labor is now explicit: **cfg.5 is the canonical, complete contract; cfg.DIST carries only per-OS defaults stubs** (DRY - the DIST comments no longer duplicate the manual).

- **cfg.DIST**: the verbose per-OS blocks are replaced by one short "Filesystem reporting" section with a defaults stub per OS - `<OS> (xymonclient-<os>.sh)`, that OS's three defaults, and a single hang-warning pointer at the manual. FreeBSD (previously undocumented) gains its stub; the macOS example that showed `EXCLUDE_TYPES="lifs"` although the script default is empty is corrected to `""`.
- **cfg.5**: the three entries get an OS-neutral definition followed by explicit "On Linux:" / "On FreeBSD:" / "On macOS:" paragraphs (defaults, mechanism, macOS attribute matching, the MNT_LOCAL note for DF_LOCAL_ONLY). Details that previously lived only in cfg.DIST comments (e.g. the Linux /run-tmpfs-is-filtered-server-side note) moved here, so nothing is lost. An unlisted OS means the variable does nothing there; #177's NetBSD paragraph slots in as one more.
- **tests/packaging/fs-filter-docs-parity.sh** pins the rule mechanically: every `client/xymonclient-*.sh` consuming `XYMONCLIENT_FS_` must have its defaults stub in cfg.DIST and its OS mention in cfg.5. It fails against unpatched main (the macOS/FreeBSD gaps) and passes here, so the next OS port cannot ship undocumented.

## Coordination

Overlaps with #177 in the same cfg.5 entries (both restructure them); whichever merges second resolves a mechanical conflict by keeping all per-OS paragraphs. All facts cross-checked against the shipped scripts (`xymonclient-freebsd.sh`, `xymonclient-darwin.sh`).

## Validation

- `groff -man -ww` clean on cfg.5
- New test: passes with the docs, fails against unpatched main (negative control)
- Full suite: 21 passed, 13 build-dependent skips, 0 failed

### DIFF
--- a/client/xymonclient.cfg.DIST
+++ b/client/xymonclient.cfg.DIST
@@ -18,120 +18,42 @@ XYMONCLIENTLOGS="$XYMONHOME/logs"     # Where we store the client logfiles
 LOGFETCHOPTS=""
 
 
-# Filesystem reporting (Linux client) -- used by xymonclient-linux.sh when
-# collecting df and inode data. By default it excludes the nodev pseudo
-# filesystems (except rootfs) and the always-full iso9660/squashfs images,
-# reports the real local filesystems including zfs/virtiofs/tmpfs, and passes
-# -l to df. The noisy tmpfs mounts (/run, /run/lock, /run/user/N) are filtered
-# server-side by the CLASS=linux block in analysis.cfg, not here.
+# Filesystem reporting -- the XYMONCLIENT_FS_* variables tune which mounts the
+# client's disk and inode reports include. The full contract - per-OS
+# mechanisms, matching rules, examples - lives in xymonclient.cfg(5); the
+# blocks below list only each OS's defaults. Unlisted OSes ignore the
+# variables. WARNING: XYMONCLIENT_FS_DF_LOCAL_ONLY=no can hang df on a stale
+# remote mount and turn the whole host purple - see the manual first.
 #
-# XYMONCLIENT_FS_INCLUDE_TYPES: whitespace-separated nodev FS types to surface
-#    (they would otherwise be dropped). Defaults to "zfs virtiofs tmpfs" -- the
-#    real local filesystems that merely happen to be nodev. Examples:
-#      ""                    -- restore the historical "exclude every nodev
-#                               type" behaviour (drops zfs/virtiofs/tmpfs too).
-#      "zfs virtiofs tmpfs nfs nfs4 ceph"
-#                            -- also include remote filesystems (requires
-#                               XYMONCLIENT_FS_DF_LOCAL_ONLY=no, since
-#                               df -l hides remote mounts).
-#
-# XYMONCLIENT_FS_EXCLUDE_TYPES: whitespace-separated FS types to also exclude,
-#    on top of the nodev default. Defaults to "iso9660 squashfs fuse.snapfuse"
-#    -- read-only images always reported 100% full by design (snaps mount as
-#    squashfs, or as fuse.snapfuse where snapd falls back to FUSE). Matching is
-#    on the exact df type token, and nodev types (overlay, bare fuse, ...) are
-#    already excluded by default, so other effective entries name a type that is
-#    NOT a nodev default -- a device-backed type, or a specific FUSE subtype the
-#    bare nodev "fuse" default cannot match. Examples:
-#      ""                    -- monitor iso9660/squashfs/snaps too.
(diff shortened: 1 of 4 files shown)


SoundGoof:
this needs a rebase

SoundGoof:
## Findings

### Current `main` makes the new parity test fail

PR #176 added `XYMONCLIENT_FS_*` support to `xymonclient-openbsd.sh` after this branch diverged, but this PR has no OpenBSD defaults stub or manual section.

The PR merges textually, but running `fs-filter-docs-parity.sh` against the resulting merge fails:

> xymonclient.cfg.DIST lacks the "OpenBSD (xymonclient-openbsd.sh)" defaults stub

Rebase onto current `main`, add OpenBSD coverage to `xymonclient.cfg.DIST` and `xymonclient.cfg.5`, and regenerate the HTML manpage.

### “root data volume is always kept” contradicts documented behavior

`client/xymonclient.cfg.DIST` says the macOS root data volume is “always kept,” and the include section in `xymonclient.cfg.5` says it is “always reported.”

However, the exclusion section correctly states that explicit exclusions override this exemption. The Darwin regression test confirms `XYMONCLIENT_FS_EXCLUDE_TYPES=apfs` removes the root data volume.

Describe it as exempt from the default `nobrowse`/`read-only` filters rather than always reported.

## Validation

- Branch CI: all five checks pass.
- Branch test suite: 23 passed, 23 skipped, 0 failed.
- Synthetic merge with current `main`: clean Git merge, but the new parity test fails due to missing OpenBSD documentation.

bonomani:
Both findings addressed. Branch rebased onto current `main` (`8f25b4c90`), still a single commit, now `9f00b3911`.

**OpenBSD coverage.** Added the missing half of #176: a defaults stub in `xymonclient.cfg.DIST` (`kernfs procfs cd9660`, `df -l`, tmpfs kept in both reports) and `On OpenBSD:` paragraphs in both `cfg.5` entries. The parity test now reports all 5 participating client OSes, and it still fails against unpatched `main` — which has no OpenBSD stub, and whose `cfg.5` mentions no OS at all. HTML manpage regenerated; a full-tree `build/makehtml.sh` touches only that one page.

**"root data" wording.** Corrected in both files: the volume is exempt from the default `nobrowse`/`read-only` drops, not always reported, and an explicit `XYMONCLIENT_FS_EXCLUDE_TYPES` entry (for instance `apfs`) still removes it — which is exactly what `tests/client/fs-filter-darwin.sh:115` pins. I dropped a first draft that also credited the local-only filter with removing it: the data volume is flagged `local`, so that filter never drops it.

Two things the rebase pulled in beyond the review:

- NetBSD support landed via #313 (#177 was closed). Its text sat as one undifferentiated `.IP` block at the tail of `EXCLUDE_TYPES`, covering all three variables at once; it is now folded into the same `On <OS>:` shape as the rest, and both entries use `cfg.DIST`'s OS order (Linux, FreeBSD, NetBSD, OpenBSD, macOS).
- The manual rendered the exclude flag as `-t no<list>` while the scripts emit it concatenated (`-tnokernfs,procfs,cd9660`). Now written `-tno<list>`, with "one comma-joined argument" spelled out at the first occurrence.

Every per-OS default in the new text was read off the client scripts and cross-checked against their tests. One caveat worth stating plainly: the BSD tests are argv mocks — `fs-filter-openbsd.sh` says so in its own header — so the statements about OpenBSD/NetBSD `df -i` behaviour rest on the scripts' comments, not on a run against real hardware.

Validation: 5/5 CI checks green on `9f00b3911`; locally `groff -man -ww` clean, testsuite 26 passed / 28 build-dependent skips / 0 failed.

SoundGoof (approved):
lgtm

PR #313, opened 2026-08-08
Fix/netbsd inode report

### LINKED ISSUE #311: NetBSD client: add inode filesystem reporting
# NetBSD client: add inode filesystem reporting

## Problem

The NetBSD client emits only a `[df]` section. It does not emit an `[inode]` section, so Xymon cannot monitor inode exhaustion on NetBSD filesystems.

This differs from the FreeBSD client, which provides end-to-end inode monitoring:

- `client/xymonclient-freebsd.sh` emits `df -i` output under `[inode]`
- `xymond/client/freebsd.c` reads the section
- The server calls `unix_inode_report()`
- `tests/client/fs-filter-freebsd.sh` covers filtering and failure handling

The gap was noted in #170 and PR #177, but no dedicated issue currently tracks the required two-sided implementation.

### Client

Update `client/xymonclient-netbsd.sh` to:

- Emit an `[inode]` section using NetBSD `df -i`
- Apply the same filesystem include/exclude and local-only settings as `[df]`
- Preserve NetBSD-compatible output for the server parser
- Emit a headerless failure marker when `df` exits non-zero without output, preventing a false-green inode status

NetBSD `df(1)` supports `-i`, `-l`, and `-t no<csv>`.

Unlike FreeBSD, NetBSD `tmpfs` inode counts are memory-derived and potentially meaningful. Do not copy FreeBSD’s unconditional `tmpfs` exclusion without NetBSD-specific evidence.

### Server

Update `xymond/client/netbsd.c` to:

- Read the `[inode]` section
- Pass it to `unix_inode_report()` using the NetBSD column names
- Preserve existing behavior when older clients do not send the section

### Documentation

Document that the NetBSD client reports both disk capacity and inode usage.

Describe how these settings apply to both reports:

- `XYMONCLIENT_FS_INCLUDE_TYPES`
- `XYMONCLIENT_FS_EXCLUDE_TYPES`
- `XYMONCLIENT_FS_DF_LOCAL_ONLY`

### Tests

Extend the NetBSD filesystem-filter test to cover:

- `[inode]` section emission
- Correct NetBSD `df -i` arguments
- Include/exclude precedence
- Local-only behavior
- Meaningful NetBSD `tmpfs` handling
- A failed inode probe producing a yellow failure marker
- Compatibility when the `[inode]` section is absent

Add a server-side test proving that a NetBSD inode report is parsed and can produce green, yellow, and red statuses.

## Acceptance criteria

- NetBSD clients report inode usage end to end
- Existing NetBSD disk reporting remains unchanged
- Older NetBSD clients without `[inode]` remain compatible
- Filtering is consistent between disk and inode reports
- Failure cannot silently produce a green inode status
- Behavior is verified against real NetBSD `df -i` output

Related: #170, #177

SoundGoof:
## Summary

Add complete NetBSD inode filesystem reporting and configurable filesystem filtering.

Fixes #311.

Supersedes #177 and extends its NetBSD filesystem-filtering approach to inode reporting. Thanks to @bonomani for the original filtering implementation.

## Changes

- Add normalized `[inode]` output to the NetBSD client.
- Process NetBSD inode reports through `xymond_client`.
- Generate per-mount inode RRDs using the existing disk RRD infrastructure.
- Apply shared filesystem filtering to disk and inode collection.
- Exclude `kernfs`, `procfs`, `cd9660`, `null`, and `ptyfs` by default.
- Keep NetBSD `tmpfs` reporting enabled.
- Add configurable include, exclude, and local-only filesystem settings.
- Report failed `df` collection instead of producing a false-green status.
- Document the NetBSD filesystem settings.
- Add client filtering and inode RRD regression tests.

## Validation

- Full build and regression suite: 35 passed, 0 skipped, 0 failed.
- Native full-client builds tested on:
  - NetBSD 9.4
  - NetBSD 10.1
  - NetBSD 11.0
- Verified FFS, vnode-backed FFS, tmpfs, null, procfs, kernfs, and ptyfs behavior.
- Verified disk, inode, memory, network, and vmstat processing.
- Verified green disk and inode statuses on all three releases.
- Verified per-mount disk and inode RRD creation and stored values.
- Verified include/exclude precedence and default filtering.


bonomani:
Thanks for taking this over the line. The native NetBSD 9.4/10.1/11.0 verification is the missing piece #177 could not provide, and this should supersede #177 once a few consistency fixes are made.

I checked the new NetBSD inode path against the existing FreeBSD, Linux, and macOS clients. Four small fixes would keep the behavior aligned:

### 1. Drop inode rows with no usable inode accounting

`client/xymonclient-netbsd.sh` currently normalizes every `df -i` row. Rows where the inode total is zero or otherwise not usable become `ifree=0` rows, which `unix_inode_report()` can turn red for absolute `INODE ... U` thresholds, and `do_disk_rrd` will also graph as flat-zero inode data. The other clients already filter this class of inode row before sending it.

Suggested NetBSD formatter:

```awk
NR == 1 { print "Filesystem itotal iused ifree %iused Mounted on"; next }
($6 + $7) <= 0 { next }
{ printf "%s %d %d %d %s %s\n", $1, $6+$7, $6, $7, $8, $9 }
```

Please add a zero/dash inode row to `tests/client/fs-filter-netbsd.sh` and assert it is dropped.

### 2. Exclude ZFS from the inode report only

NetBSD ZFS has no fixed inode pool. Its `df -i` values are derived from available objects/space, so a nearly full pool can make every dataset look near the default inode panic threshold even though it cannot run out of inodes in the traditional sense. This matches the FreeBSD `zfs` and macOS `apfs` inode-only exclusions.

Suggested call:

```sh
if emit_df inode Inode -i zfs; then
```

Keep `tmpfs` reported; NetBSD tmpfs inode availability is memory/node-accounting derived and can be a real signal.

### 3. Restore the Linux historical-behavior note

The manpage rewrite removes the Linux note that `XYMONCLIENT_FS_INCLUDE_TYPES=""` restores the historical behavior of dropping every `nodev` type. The Linux client still supports that, so please re-add the sentence to the Linux part of `common/xymonclient.cfg.5`.

### 4. Call out the new `ptyfs` default exclusion in release notes

Excluding `ptyfs` is consistent with the other clients: it is NetBSD's `/dev/pts` pseudo-terminal filesystem, not a meaningful disk/inode capacity signal. The old NetBSD client did not type-exclude it, though, so keeping the new default is upgrade-visible.

Please add an upgrade/release note, for example:

> The NetBSD client now excludes `ptyfs` (`/dev/pts`) from disk and inode reporting by default. Set `XYMONCLIENT_FS_INCLUDE_TYPES="ptyfs"` to restore the old reportable behavior.

Not blockers for this PR: `getdata("msgsstr")` in `handle_netbsd_client()` looks like a pre-existing `getdata("msgs")` typo, and the broader shared filter cleanups can be handled separately.

bonomani (approved):

PR #314, opened 2026-08-09
lib/loadhosts: warn about web-unreachable canonical hostnames (#309)

### LINKED ISSUE #309: Define and enforce a canonical-hostname character set (CGI ⇄ hosts.cfg parser)
Split out of the review discussion on #147 (thanks @SoundGoof).

### Problem
The CGI programs and the `hosts.cfg` parser disagree on what a canonical hostname may contain. The `hosts.cfg` parser accepts names with bytes outside a strict ASCII set (e.g. `ä`, `ö`, `å`); such a host loads and is monitored fine, but with #147 the CGI programs confine host parameters to an ASCII name charset (`[A-Za-z0-9:,._-]`) and refuse anything outside it. So the two sides now disagree: a host can load yet be unreachable via `svcstatus.sh` / `history.sh` / `reportlog.sh`. Whichever way we lean, the accepted set should be agreed on in one place and documented, rather than emergent.

### Context
#147 hardens the CGI path-component confinement (traversal + protocol/shell injection) and, for hostnames, confines to the ASCII set above — deliberately matching the pre-existing `reportlog`/`history` behaviour and refusing (never normalizing) the rest. That is a security fix; the character-set **policy** is a separate, compatibility-affecting decision, which is why it is broken out here.

### To decide
- Pick the canonical-hostname character set. "Legal in a URL" (RFC 3986 unreserved/sub-delims) is a natural ASCII basis; note a decoded UTF-8 name falls outside it, so this is effectively an ASCII-canonical-name policy with Unicode available via the `NAME:` display tag.
- If we keep the restricted set, enforce it in **one** place of record — the `hosts.cfg` loader — so an illegal canonical name fails at config load, not later in a CGI. The CGI confinement (already in #147) then never disagrees with it.
- Document the accepted set in `hosts.cfg(5)` and note the compatibility change in `RELEASENOTES`.

### Non-goal
Not the path-traversal / injection confinement (done in #147); this is about the *positive* definition of a legal canonical hostname and where it is enforced. Related but distinct: #283 (hostname confinement in `xymond_hostdata`).

bonomani:
Closes #309.

## Problem
The `hosts.cfg` parser accepts any non-whitespace token as a canonical hostname (field 2), while the web CGIs and the xymond ghost-name guard confine host names to a safe set. Such a host loads and is monitored, yet its `svcstatus`/`history`/`reportlog` pages are unreachable — with nothing said.

## What this does
`xymond` warns once per config (re)load about a canonical name the web cannot serve — from the daemon, **not** the shared loader (which runs in every short-lived CGI process and would flood the error log). The host is **never dropped**: dropping it removes it from the roster (out of alerting) and lets `trimhistory(1)` delete its history as orphaned. Non-ASCII/decorated spellings belong in the `NAME:` display tag.

The character set is one `XYMON_HOSTNAME_CHARS` (`lib/cgi.h`), shared by `cgi.c`'s `CGI_NAMECHARS` and the ghost guard; a `hostname_web_safe()` helper backs both. A literal comma is also flagged — it stays in the set for the CGI `192,168,1,1` spelling, but `uncommafy()` collapses `foo,bar`→`foo.bar`.

Documented in `hosts.cfg(5)` (HTML regenerated); tested via the built `xymongrep`, which loads every host and stays silent. `RELEASENOTES`/`Changes` text is staged in the comment below.

### DIFF
--- a/common/hosts.cfg.5
+++ b/common/hosts.cfg.5
@@ -27,6 +27,24 @@ The IP-address and hostname are mandatory; all of the tags are optional.
 Listing a host with only IP-address and hostname will cause a network
 test to be executed for the host - the connectivity test is enabled
 by default, but no other tests.
+.sp
+The hostname is the host's canonical name: it keys the status database and
+appears in web-page URLs and RRD file paths. It is reachable through the web
+interface only when built from the characters
+.B A-Z a-z 0-9
+and
+.BR : . _ -
+(letters, digits, colon, period, underscore and hyphen). A host whose name
+contains any other character - other punctuation, a literal comma (which the
+web URLs use to encode a period, so it cannot appear in a name), or a
+non-ASCII byte - is still monitored, but
+.I xymond(8)
+writes a warning to the Xymon log when it loads the configuration and its
+status pages are not reachable through the web interface. To show a non-ASCII
+or otherwise decorated name in the web interface, give the host an ASCII
+canonical name and carry the display spelling in the
+.B NAME:
+tag (see below).
 
 The optional tags are then used to define which tests are
 relevant for the host, and also to set e.g. the time-interval used
--- a/docs/manpages/man5/hosts.cfg.5.html
+++ b/docs/manpages/man5/hosts.cfg.5.html
@@ -52,6 +52,18 @@ Section: File Formats (5)<br/>Updated: Version 4.3.30: 4 Sep 2019<br/><a href="#
     optional. Listing a host with only IP-address and hostname will cause a
     network test to be executed for the host - the connectivity test is enabled
     by default, but no other tests.</p>
+<p class="Pp">The hostname is the host's canonical name: it keys the status
+    database and appears in web-page URLs and RRD file paths. It is reachable
+    through the web interface only when built from the characters <b>A-Z a-z
+    0-9</b> and <b>:</b>.<b>_</b>- (letters, digits, colon, period, underscore
+    and hyphen). A host whose name contains any other character - other
+    punctuation, a literal comma (which the web URLs use to encode a period, so
+    it cannot appear in a name), or a non-ASCII byte - is still monitored, but
+    <i><a href="../man8/xymond.8.html">xymond</a>(8)</i> writes a warning to the Xymon log when it loads the
+    configuration and its status pages are not reachable through the web
+    interface. To show a non-ASCII or otherwise decorated name in the web
+    interface, give the host an ASCII canonical name and carry the display
+    spelling in the <b>NAME:</b> tag (see below).</p>
 <p class="Pp">The optional tags are then used to define which tests are relevant
     for the host, and also to set e.g. the time-interval used for availability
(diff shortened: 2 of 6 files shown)


bonomani:
Release-notes text to stage on the `Changes` branch at merge:

**RELEASENOTES (4.3.31):**
> A canonical hostname (field 2 of hosts.cfg) with a character the web interface cannot serve now makes xymond log a warning when it loads the configuration. The host is still monitored; only its svcstatus/history/reportlog pages are unreachable. Give it an ASCII canonical name and put the original spelling in the NAME: display tag. The accepted set (letters, digits and ":._-") is shared by the CGIs and the xymond ghost-name check.

**Changes:**
> * Warn at config load about canonical hostnames the web interface cannot serve (#309)

SoundGoof:
## Findings

### Emit startup warnings after log redirection

[xymond/xymond.c](xymond/xymond.c#L5553-L5554) emits hostname warnings before `--log` redirects stderr at [xymond/xymond.c](xymond/xymond.c#L5673-L5676).

Runtime testing confirmed that startup warnings appear only on the launching terminal. After a changed `hosts.cfg` reload, the same warnings appear in the configured log.

Move the initial warning scan after log setup so startup and reload behavior match.

### Do not classify literal commas as web-safe

[lib/cgi.h](lib/cgi.h#L25-L33) includes comma in `XYMON_HOSTNAME_CHARS`, so canonical hostname `foo,bar` produces no warning.

History URLs use commas to encode periods: `commafy()` leaves an existing comma unchanged, while `uncommafy()` converts every comma to a period. Consequently, a generated history link for `foo,bar` resolves as `foo.bar`.

Exclude literal commas from the web-safe canonical set or introduce unambiguous encoding before documenting them as reachable.

## Validation

- Rebases cleanly onto current `main`.
- Server build succeeded.
- Full suite: 44 passed, 0 failed, 0 skipped.
- GitHub checks are green.
- Runtime testing confirmed the startup/reload logging difference.

bonomani:
@SoundGoof both fixed, thanks:

- **Startup vs reload logging:** the startup scan now runs *after* stderr is redirected to `--log`, so startup and reload warnings land in the same place.
- **Literal comma:** now treated as unreachable and warned about. It stays in `XYMON_HOSTNAME_CHARS` (the CGIs need it for the `192,168,1,1` IP spelling), but `warn_unreachable_hostnames()` and `hosts.cfg(5)` now exclude it, since `uncommafy()` collapses a configured `foo,bar` to `foo.bar`. The warning text is generalized to "characters the web interface cannot serve."

Also rebased onto current `main` (the #147 base commit is merged, so it drops) and regenerated `hosts.cfg.5.html` for the new manual-page gate.

SoundGoof (approved):
## Review

No blocking issues found. Startup and reload warnings are logged consistently, literal commas are handled explicitly, and invalid canonical names remain loaded.

Also, räksmörgås is underrated.

## Validation

- Build succeeded.
- Tests: 44 passed, 0 failed.
- Generated `hosts.cfg.5.html` matches its source exactly.
- Rebases cleanly onto current `main`.

Approved.

PR #321, opened 2026-08-10
tests: decide on ASAN by running a probe, not by compiling one

bonomani:
Five tests prefer a sanitized build and fall back to a plain one, and each picked between them by asking whether `-fsanitize=address,undefined` **compiles**. That is not the question that matters.

The RPM families ship the sanitizer runtime in its own package (`libasan`, `libubsan`). Where the package is absent — Amazon Linux 2023 — the compile and the link both succeed, and only the finished binary fails, at startup:

```
error while loading shared libraries: libasan.so.6: cannot open shared object file
```

exiting 127. The fallback never fired, because nothing had failed yet.

The three tests that run a harness reported a broken parser. The two that run a CGI saw exit 127 and reported a crash, having sent the loader's message to `/dev/null`:

```
FAIL: history.cgi exited 127 (crash?) on QUERY_STRING='HISTFILE=..'
FAIL: svcstatus exited 127 (crash?) on QUERY_STRING=HOST=realhost&SERVICE=web.grp&TIMEBUF=20260101
```

All five blamed the code under test for a missing package.

## Change

`asan_usable()` in `tests/lib/assert.sh`: build a probe, **run** it, cache the verdict. Running is the point — a successful link says nothing about whether the runtime is installed. The five callers gate on it, so a host without the runtime takes the plain build they already had.

The pattern is not new here. `xtree-destroy.sh` and `xtree-iterate-deleted.sh` have always executed their probe:

```sh
"$CC" -fsanitize=address -o "$work/probe" "$work/probe.c" 2>/dev/null \
	&& "$work/probe" 2>/dev/null \
	|| skip "cc does not support -fsanitize=address"
```

which is exactly why those two came through Amazon Linux 2023 green while the other five did not. This puts the check in one place rather than adding a sixth copy, and leaves those two untouched.

Standalone copies of the three harness tests — the branch taken when `assert.sh` is not readable — define `asan_usable` as false and build plain.

## Verification

A `cc` shim reproduces the condition: it strips `-fsanitize` and links a stub library reachable via `-L` but not by the loader, so the sanitized link succeeds and the binary dies at exec with 127. It reproduces both CGI failure lines above verbatim.

| | before | after |
|---|---|---|
| suite under shim | **5 failed** | **0 failed** — 50 passed, 3 skipped |
| suite unshimmed (real ASAN) | 52 / 1 / 0 | 52 / 1 / 0 |

Under the shim the two xtree tests skip via their own probe, as they should.

The non-regression check matters most: unshimmed, `asan_usable` returns true and the harness still links `libasan.so.6` and `libubsan.so.1`. This does not quietly disable the sanitizer where it works.

`shellcheck` output is unchanged before and after on every touched file.

## Note

Installing `libasan`/`libubsan` on the RPM lanes would give those distros real sanitizer coverage instead of a graceful downgrade. That is a separate CI change; this fix is needed regardless, since any host can lack the runtime.

### DIFF
--- a/tests/lib/assert.sh
+++ b/tests/lib/assert.sh
@@ -142,6 +142,43 @@ require_cc() {
 	command -v "$CC" >/dev/null 2>&1 || skip "no C compiler available (CC=$CC)"
 }
 
+# asan_usable -- true when $CC can build a sanitized binary AND the result
+# runs. Compiling is not the question: distros that ship the sanitizer
+# runtime in its own package (Amazon Linux 2023 and the other RPM families,
+# where it is libasan/libubsan) link -fsanitize=address happily, and only
+# the finished binary fails, at startup, with
+#
+#     error while loading shared libraries: libasan.so.6: ...
+#
+# and exit 127. A caller that gated on the compile alone then reads that as
+# a crash in the code under test -- the symptom this helper exists to stop.
+# So the probe is executed, not just built.
+#
+# Callers use it to pick between a sanitized build and a plain one, or to
+# skip outright when the sanitizer is the point of the test. The verdict
+# cannot change inside one test process, so it is computed once.
+#
+# The probe directory is removed here rather than left to the EXIT trap
+# registered at the top of this file: several callers install a trap of
+# their own for their work dir, which replaces that one, and the probe
+# would then survive the run. A helper cannot assume the shared cleanup is
+# still armed by the time it is called, so it cleans up after itself.
+asan_usable() {
+	if [ -z "${__xymon_tests_asan_usable:-}" ]; then
+		local probe
+		__xymon_tests_asan_usable=no
+		probe=$(mktempdir)
+		printf 'int main(void){return 0;}\n' >"$probe/probe.c"
+		if "${CC:-cc}" -fsanitize=address,undefined -o "$probe/probe" \
+					"$probe/probe.c" 2>/dev/null \
+				&& "$probe/probe" >/dev/null 2>&1; then
+			__xymon_tests_asan_usable=yes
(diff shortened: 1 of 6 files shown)


SoundGoof:
## Review

### Finding

#### Low: Clean up probe directories when callers replace the EXIT trap

`asan_usable()` creates its probe with `mktempdir()`, relying on
`assert.sh`'s EXIT trap for cleanup.

The three harness tests replace that trap before calling the helper:

- `tests/network/ntp-probe.sh:29-36`
- `tests/network/udp-query.sh:28-40`
- `tests/rrd/ntp-offset-parse.sh:28-35`

Consequently, every invocation leaves an
`xymon-test.<pid>.*` directory containing `probe.c` behind.

I reproduced this on the PR head for all three tests. Each exited
successfully and leaked one probe directory.

Remove the probe explicitly inside `asan_usable()`, or preserve the shared
cleanup mechanism in these callers.

### Verification

CI is green and the PR merges cleanly into current `main`.

The three focused harness tests pass. Shellcheck found no new substantive
diagnostics beyond the cleanup issue above.

bonomani:
Confirmed and fixed — thanks.

Reproduced on a built tree: the three trap-replacing tests leaked one `xymon-test.<pid>.*` each, the two CGI callers leaked none. Leftover dirs 3 → 0.

Fixed inside `asan_usable()` rather than in the callers: a helper cannot assume the shared EXIT trap is still armed when it runs.

Squashed into the single commit. Full suite 52 passed / 1 skipped / 0 failed.

SoundGoof (approved):
lgtm

PR #324, opened 2026-08-11
Add the ping commando to the msgcache so we can query the version (and state) of the daemon

StefCoene:
This is needed for the latest version of the Windows Powershell client so it can test if a new client version or config file is working.

### DIFF
--- a/client/msgcache.c
+++ b/client/msgcache.c
@@ -48,6 +48,7 @@ char *client_response = NULL;		/* The latest response to a "client" message */
 char *logfile = NULL;
 int maxage = 600;			/* Maximum time we will cache messages */
 sender_t *serverlist = NULL;		/* Who is allowed to grab our messages */
+time_t lastpull = 0;			/* When a server last collected from us */
 
 typedef struct conn_t {
 	time_t tstamp;
@@ -125,6 +126,32 @@ void grabdata(conn_t *conn)
 	 * that we'll return the next time a client sends us the "client" message.
 	 */
 
+	if (strncmp(STRBUF(conn->msgbuf), "ping", 4) == 0) {
+		/*
+		 * Answered here and now, before the queueing below: a ping is not
+		 * client data, so it must not go on the outbound queue, and it must
+		 * not travel through client_response - that global holds the config
+		 * the server pushed, and overwriting it would hand the next client a
+		 * version string where its configuration belongs.
+		 *
+		 * What we can honestly report is ourselves and the age of the last
+		 * pull. We never connect to the Xymon server; it connects to us. So
+		 * "the server is up" is not ours to say, and naming xymond here would
+		 * let a caller read "safe to upgrade" off a msgcache whose server has
+		 * been gone for a week - while maxage quietly discards its messages.
+		 * lastpull is -1 until a server has collected at least once.
+		 */
+		char id[128];
+
+		snprintf(id, sizeof(id), "msgcache %s\nlastpull %ld\n", VERSION,
+			 (long)(lastpull ? (getcurrenttime(NULL) - lastpull) : -1));
+		clearstrbuffer(conn->msgbuf);
+		addtobuffer(conn->msgbuf, id);
+		conn->ctype = C_CLIENT_OTHER;
+		conn->action = C_WRITING;
+		return;
+	}
+
 	if (strncmp(STRBUF(conn->msgbuf), "pullclient", 10) == 0) {
 		char *clientcfg;
 		int idnum;
@@ -153,6 +180,7 @@ void grabdata(conn_t *conn)
 
 		conn->ctype = C_SERVER;
 		conn->action = C_WRITING;
+		lastpull = getcurrenttime(NULL);
 
 		/* Save any client config sent to us */
 		clientcfg = strchr(STRBUF(conn->msgbuf), '\n');
--- a/tests/client/msgcache-ping.sh
+++ b/tests/client/msgcache-ping.sh
@@ -1,0 +1,73 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/client/msgcache-ping.sh
+#
+# msgcache answers "ping" locally, and says two things: what it is, and how
+# long ago a Xymon server last collected from it.
+#
+# Both matter because msgcache never connects to the server -- the server
+# connects to it with "pullclient". "The server is up" is therefore not
+# msgcache's to say, and a caller that read it as such could upgrade a client
+# against a server that has been gone for a week, while maxage quietly
+# discarded its messages. The age of the last pull is the honest signal.
+#
+# The other half is what a ping must NOT disturb: the config the server pushed
+# is held in one global, and answering through it would hand the next client a
+# version string where its configuration belongs.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+require_bin MSGCACHE client/msgcache
+command -v nc > /dev/null 2>&1 || skip "no nc to speak to msgcache with"
+
+work=$(mktempdir)
+
+free_port() {
+	local p tries=0
+	while [ "$tries" -lt 50 ]; do
+		p=$(( 20000 + (RANDOM % 20000) ))
+		nc -z 127.0.0.1 "$p" 2>/dev/null || { printf '%s' "$p"; return 0; }
+		tries=$((tries + 1))
+	done
+	return 1
+}
+
+PORT=$(free_port) || skip "no free port for msgcache"
+"$MSGCACHE" --listen=127.0.0.1:"$PORT" --server=127.0.0.1 --no-daemon > "$work/log" 2>&1 &
+MC=$!
+register_cleanup "kill $MC 2>/dev/null || true"
+
+i=0
+while [ "$i" -lt 50 ]; do
+	nc -z 127.0.0.1 "$PORT" 2>/dev/null && break
+	kill -0 "$MC" 2>/dev/null || { cat "$work/log" >&2; fail "msgcache exited during startup"; }
+	sleep 0.1
+	i=$((i + 1))
+done
+
+say() { printf '%s\n' "$1" | nc -q1 127.0.0.1 "$PORT" 2>/dev/null; }
+
+# --- before any server has collected -----------------------------------------
+out=$(say 'ping')
+assert_contains "msgcache" "$out" "the ping names msgcache, not the daemon it never talks to"
+assert_contains "lastpull -1" "$out" "and reports no pull at all until a server has collected"
+
+# --- a server collects, pushing a client config ------------------------------
+say 'pullclient 1
+CONFIG-FROM-SERVER' > /dev/null
+
+out=$(say 'ping')
+assert_not_contains "lastpull -1" "$out" "after a pull, the age must be a real one"
+assert_contains "lastpull " "$out" "and still be reported"
+
+# --- and the ping did not eat the config -------------------------------------
+# The reply travels through its own buffer, not through the global holding the
+# pushed config: sharing it would give this client a version string instead.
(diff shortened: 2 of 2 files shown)


bonomani:
Hi Stef,

I have to admit I did not fully understand the goal. The title says "query the version
(and state) of the daemon", the body says "test if a new client version or config file is
working" — those read as two different things to me, and which one it is changes what the
implementation should look like. Could you say a bit more about what the PowerShell client
needs to decide?

I built the patch to try it out, and ran into two things I would like your opinion on.

First, `client_response` is the global holding the client configuration the server pushed
to us via `pullclient`, so `client_response = strdup(id)` replaces it. In my test a
`client` message returned the config before the ping and `xymond 4.3.30` after it — am I
right that a ping would then override the config for every client until the next
`pullclient` refills it? (The `pullclient` branch also does
`if (client_response) xfree(client_response);` first, so we may be leaking the old buffer
as well.)

Second, with `ctype = C_CLIENT_CLIENT` the ping seems to fall into the `!= C_SERVER`
branch below and end up on the outbound queue — I saw it come back in the next
`pullclient`, so I think it gets forwarded to the server by xymonfetch.

If that reading is right, both would go away by handling ping early, before the queueing
code:

```c
if (strncmp(STRBUF(conn->msgbuf), "ping", 4) == 0) {
	/* Liveness probe answered locally - not forwarded to the Xymon server */
	clearstrbuffer(conn->msgbuf);
	addtobuffer(conn->msgbuf, "msgcache " VERSION "\n");
	conn->ctype = C_CLIENT_OTHER;
	conn->action = C_WRITING;
	return;
}
```

I used `msgcache` rather than `xymond` in the reply, since msgcache answers this itself
and it never reaches the server — answering `xymond` could make a caller believe it talked
to the Xymon server. (xymonproxy is a different case: it forwards ping on and relays the
real answer.)

And one idea, depending on your answer above: if what you need is "is my link to the
server working", ping alone cannot show that, since msgcache replies whether the server is
up or not. But it does know when xymonfetch last collected from it, so it could add that:

```
msgcache 4.3.30
lastpull 42
queued 3
```

`lastpull` = seconds since the last `pullclient`. One `time_t` in the existing
`pullclient` branch, no new command, nothing to change server-side. Not implemented, just
an idea — would that be useful for you?

StefCoene:
> Hi Stef,
> 
> I have to admit I did not fully understand the goal. The title says "query the version (and state) of the daemon", the body says "test if a new client version or config file is working" — those read as two different things to me, and which one it is changes what the implementation should look like. Could you say a bit more about what the PowerShell client needs to decide?
> 
> I built the patch to try it out, and ran into two things I would like your opinion on.
> 
> First, `client_response` is the global holding the client configuration the server pushed to us via `pullclient`, so `client_response = strdup(id)` replaces it. In my test a `client` message returned the config before the ping and `xymond 4.3.30` after it — am I right that a ping would then override the config for every client until the next `pullclient` refills it? (The `pullclient` branch also does `if (client_response) xfree(client_response);` first, so we may be leaking the old buffer as well.)
> 
> Second, with `ctype = C_CLIENT_CLIENT` the ping seems to fall into the `!= C_SERVER` branch below and end up on the outbound queue — I saw it come back in the next `pullclient`, so I think it gets forwarded to the server by xymonfetch.
> 
> If that reading is right, both would go away by handling ping early, before the queueing code:
> 
> ```c
> if (strncmp(STRBUF(conn->msgbuf), "ping", 4) == 0) {
> 	/* Liveness probe answered locally - not forwarded to the Xymon server */
> 	clearstrbuffer(conn->msgbuf);
> 	addtobuffer(conn->msgbuf, "msgcache " VERSION "\n");
> 	conn->ctype = C_CLIENT_OTHER;
> 	conn->action = C_WRITING;
> 	return;
> }
> ```
> 
> I used `msgcache` rather than `xymond` in the reply, since msgcache answers this itself and it never reaches the server — answering `xymond` could make a caller believe it talked to the Xymon server. (xymonproxy is a different case: it forwards ping on and relays the real answer.)
> 
> And one idea, depending on your answer above: if what you need is "is my link to the server working", ping alone cannot show that, since msgcache replies whether the server is up or not. But it does know when xymonfetch last collected from it, so it could add that:
> 
> ```
> msgcache 4.3.30
> lastpull 42
> queued 3
> ```
> 
> `lastpull` = seconds since the last `pullclient`. One `time_t` in the existing `pullclient` branch, no new command, nothing to change server-side. Not implemented, just an idea — would that be useful for you?

Hi,

The Windows PowerShell client has a test mechanism that sends a 'ping' statement to the configured Xymon server to test a new client and/or a new configuration file.

Xymond answers with its version, but msgcache does not.

So, clients connecting to a msgcache daemon (we use that in our setup) will not update to a new version of the PowerShell client and/or a new configuration file.

I cobbled this code together myself (that was before the AI era) and it works for me.
Also, I'm not a C programmer at all, I can understand the code and copy or tweak it, but that's about it. :)
So it could be that while it works for me, it's actually totally wrong under the hood...


Stef

bonomani:
Thanks Stef — that's the piece I was missing, and the gap is real: msgcache replies nothing today, so your clients never get past the self-test.

Your goal is "is it safe to apply this new client/config", and that's what shapes the answer. msgcache never connects to the Xymon server — the server connects to *it* and drains the queue with `pullclient`. So a ping msgcache answers can only say "msgcache is up", and the version it reports is msgcache's own build, not the server's. If it replied `xymond 4.3.30`, your client would read that as "server is fine, safe to upgrade" with the server down for a week, or on a different version. That's the one failure you'd never notice.

What msgcache *does* know is when the server last collected from it, which is exactly the signal you need:

```c
/* near the other globals */
static time_t lastpull = 0;

/* in the pullclient branch, next to "conn->ctype = C_SERVER;" */
lastpull = getcurrenttime(NULL);

/* first test in grabdata(), before the pullclient one */
if (strncmp(STRBUF(conn->msgbuf), "ping", 4) == 0) {
	char id[128];

	/* Answered locally - never forwarded to the Xymon server. */
	snprintf(id, sizeof(id), "msgcache %s\nlastpull %ld\n", VERSION,
		 (long)(lastpull ? (getcurrenttime(NULL) - lastpull) : -1));
	clearstrbuffer(conn->msgbuf);
	addtobuffer(conn->msgbuf, id);
	conn->ctype = C_CLIENT_OTHER;
	conn->action = C_WRITING;
	return;
}
```

`lastpull` is seconds since the server last pulled, `-1` if it never has. Your client can then say: got an answer *and* `lastpull` is within the polling interval → the whole path is alive, go ahead; otherwise don't. That's the check you actually want, and one `time_t` buys it.

Returning early also fixes two things in the current patch, which share a cause. `C_CLIENT_CLIENT` can only reply with `client_response` — the global holding the config the server pushed — so the version string had to be written there, which means one ping replaces the config for every client until the next `pullclient`. And `C_CLIENT_CLIENT` falls past the `!= C_SERVER` test, so the ping goes on the outbound queue and gets forwarded to the server. Answering before the queueing code avoids both.

I built and ran it here:

```
$ printf 'ping\n' | nc 127.0.0.1 1984
msgcache 4.3.30
lastpull -1                     <- server has never pulled

(after a pullclient that also pushes a config)
$ printf 'ping\n' | nc ...
msgcache 4.3.30
lastpull 2

$ printf 'client foo.linux\ndata\n' | nc ...
CONFIG-LINE-FOR-CLIENTS         <- config survives the ping

$ printf 'pullclient 1\n' | nc ...
client foo.linux ...            <- the two client messages, no ping
```

Could you try it against your PowerShell client and confirm it actually solves your problem? It needs a change on your side — the reply says `msgcache`, not `xymond`, and there's the extra line. Nothing breaks in the meantime, since msgcache answers nothing at all today. But if reworking the client script across your fleet is a bigger problem than I'm assuming, say so and we'll think again — you know that cost and I don't.

StefCoene:
Hi,

I tested it and it works.


Stef

bonomani:
Which of the two did you test — the branch as it stands, or the `lastpull` version I sketched? The branch has not moved since, so I assume the former, but I would rather ask than guess.

If it was the branch: it does answer, and that part works. The two problems I raised are ones a single ping does not show.

- The reply goes through `client_response`, the global holding the config the server pushed. After a ping, the **next** client to send `client` gets `xymond 4.3.x` where its config should be, until the next `pullclient` refreshes it. Visible only from a second client, after the ping.
- `C_CLIENT_CLIENT` falls past the `!= C_SERVER` test, so the ping is also queued and forwarded to the server. Visible only server-side.

Worth checking on your setup: send the ping, then have a second client send `client`, and see whether it still gets its configuration.

The other point is not about the implementation. msgcache never connects to the Xymon server — the server connects to *it* with `pullclient` — so `xymond <version>` reports msgcache's own build under the server's name. A client reading it would take "server is fine, safe to upgrade" from a msgcache whose server has been down for a week. The one thing msgcache does know is when it was last pulled, which is the signal your question actually needs: got an answer *and* `lastpull` within the polling interval means the whole path is alive.

Happy to push the `lastpull` variant to a branch you can test directly if that is easier than transcribing it.

bonomani:
Pushed it so you can build it rather than transcribe it: `bonomani:feat/msgcache-ping-lastpull` (one commit on top of current `main`, no PR — it is for comparison, not competition).

```
git fetch https://github.com/bonomani/xymon.git feat/msgcache-ping-lastpull && git checkout FETCH_HEAD
./configure --client && make
```

What it answers, driven against a running msgcache here:

```
ping (no server has pulled yet)   →  msgcache 4.3.30
                                     lastpull -1
... server does "pullclient 1" + pushes a config ...
ping                              →  msgcache 4.3.30
                                     lastpull 1
client host.linux linux           →  CONFIG-FROM-SERVER
```

The last line is the difference that does not show in a single ping: the config the server pushed survives, because the reply goes into its own buffer instead of `client_response`. On the branch as it stands, that client gets `xymond 4.3.30` where its configuration should be.

`lastpull` is seconds since a server last collected, `-1` if none ever has. Your client can then say: got an answer *and* `lastpull` under the polling interval → the whole path is alive, apply the update. It is honest about its limits — it says a server came, not that xymond stored anything downstream.

There is a test with it (`tests/client/msgcache-ping.sh`) covering both halves, including the config surviving the ping.

StefCoene:
> Pushed it so you can build it rather than transcribe it: `bonomani:feat/msgcache-ping-lastpull` (one commit on top of current `main`, no PR — it is for comparison, not competition).
HI,

I tested your patch.


Stef

bonomani:
So just update you pr when you are want it to be merged!
Bruno

StefCoene:
> So just update you pr when you are want it to be merged! Bruno

Hi,

I've updated the branch with your implementation


Stef

bonomani:
@StefCoene Sorry for the back-and-forth — you'd already pulled my implementation into this PR as I'd asked, and then I ran a much deeper analysis on it (a few review passes), which turned up some refinements. Rather than keep pushing changes onto your branch, I put the polished version up as **#399** (single commit on current main).

So the easiest path for now is probably to **approve #399** — it supersedes this one, and we can close #324 once you've confirmed it works in your setup (`./configure --client && make`; test is `tests/client/msgcache-ping.sh`).

What the deeper pass changed, so it's no black box — and so next time you can take it the same distance yourself:
- `lastpull` is stamped when a pull *completes*, not when it's requested, so it can't report a fresh collection off a pull whose write failed;
- only an exact `ping` is answered — `pingpong…`-style client data is still queued and forwarded to the server;
- a backward clock step clamps the age to 0 instead of faking the "never pulled" −1;
- the test drives the wire with the in-tree `xymon` CLI instead of `nc`, so it stops failing on the BSD/macOS lanes.

The core is still exactly what you had once my earlier notes were in — answer early in `grabdata()`, into `conn->msgbuf`, `C_CLIENT_OTHER`, before the queueing, so it never touches `client_response` or the outbound queue. Thanks for starting this off.

StefCoene:
Superseded by #399 — closing this one in its favour.

PR #325, opened 2026-08-11
tests: read inode RRD through librrd

### LINKED ISSUE #322: tests/rrd/inode-netbsd.sh: skips for a CLI it does not need, and is named after the wrong axis
Two separable problems in `tests/rrd/inode-netbsd.sh`. Neither is a bug in
xymon itself; both make the test cover less than it appears to.

## 1. It skips for the rrdtool CLI, which the build does not require

The test's only use of the CLI is one line:

```sh
lastupdate=$(rrdtool lastupdate "$rrd")
```

and its precondition is `command -v rrdtool || skip "rrdtool not found"`.

But `xymond_rrd` is linked against librrd already:

```
$ ldd xymond/xymond_rrd | grep rrd
    librrd.so.8 => /lib/x86_64-linux-gnu/librrd.so.8
```

`librrd-dev` is a build dependency of `--server`; the `rrdtool` **CLI** is a
separate package that nothing in the build needs. So the test skips on every
machine that can build and run the very binary it is testing, unless one extra
package happens to be installed. On a local server build here it is the only
test in the whole suite that skips:

```
passed:  52   skipped: 1   failed:  0
  ----- tests/rrd/inode-netbsd.sh -----
  SKIP: rrdtool not found
```

`.github/workflows/build.yml` installs `rrdtool` solely for this one test, and
says so in a comment.

**Suggested fix:** read the RRD back through librrd instead of the CLI. The
suite already compiles C harnesses for other tests (`tests/rrd/ntp-offset-parse-harness.c`,
`tests/server/xmh-item-names-harness.c`, ...), so a small harness calling
`rrd_lastupdate_r()` or `rrd_info_r()` fits the existing pattern and adds no
host requirement beyond what building the server already needs. CI could then
drop `rrdtool` from the apt list.

## 2. The name says "netbsd", but the code switches on disk-report type

`xymond/rrd/do_disk.c` has no OS-specific branches at all. It picks one of six
column layouts from substrings in the message (do_disk.c:58-63):

```c
else if (strstr(msg, "DASD"))            dsystype = DT_AS400;
else if (strstr(msg, "NetWare Volumes")) dsystype = DT_NETWARE;
else if (strstr(msg, "NetAPP"))          dsystype = DT_NETAPP;
else if (strstr(msg, "Summary"))         dsystype = DT_BBWIN;
else if (strstr(msg, "Filesystem"))      dsystype = DT_NT;
else                                     dsystype = DT_UNIX;
```

The test's sample header is `Filesystem itotal iused ifree %iused Mounted on`,
so it is classified **DT_NT**, and then flips per line to **DT_UNIX** at
do_disk.c:115 because `columns[0]` contains a `/` and `columns[5]` is
non-empty. The DT_UNIX branch then takes `pused` from `columns[4]` (`15%`) and
`aused` from `columns[2]` (`41020`) - which is exactly what the test asserts.

That is a genuinely subtle path and worth pinning. But the filename describes
the OS that produced the sample, not the axis the code branches on, with two
consequences:

- It implies four missing siblings (`inode-linux`, `inode-freebsd`, ...). They
  would be **redundant** - every unix `df -i` report takes the same
  DT_NT-then-flip-to-DT_UNIX path.
- It hides five layouts that really are untested: `DT_AS400`, `DT_NT` proper
  (no flip), `DT_BBWIN`, `DT_NETAPP`, `DT_NETWARE`.

**Suggested fix:** name it for the path it covers rather than the sample's
origin - something like `inode-unix-columns.sh` - keeping the NetBSD report as
a fixture. A fully generic name (`inode-rrd.sh`) would be worse than the
current one: it would suggest the test covers the inode pipeline in general
when it covers one sixth of it.

## Not urgent, and independent

The test passes wherever the CLI is present, so nothing is silently broken.
Filing separately rather than folding it into the variant-testing work in #167,
since it touches neither.

SoundGoof:
# tests: read inode RRD through librrd

## Summary

This follows up on the inode RRD regression test added in #313 and
addresses the test-maintenance issues reported in #322.

- replace the `rrdtool lastupdate` CLI call with a small librrd harness
- rename `inode-netbsd.sh` to `inode-unix-columns.sh`
- describe the parser layout under test instead of the fixture’s original OS
- compile the harness with the configured RRD include, library, linker,
  and runtime-path flags

The effective Make-variable probe follows the test harness pattern
introduced in #289. This is necessary because platform-specific linker
settings come from Makefiles included by the generated top-level
Makefile and cannot be reliably extracted from that file with `sed`.

This removes the test-only `rrdtool` package dependency while preserving
coverage of the shared UNIX inode-column parsing path.

## Testing

- focused inode RRD test
- deliberately failing `rrdtool` function, confirming the CLI is unused
- custom librrd runtime-path flag expansion
- ShellCheck
- full test suite: 50 passed, 3 skipped, 0 failed

Closes #322.


bonomani (approved):
Approved. The PR is narrow, CI is green, and I reproduced the focused test plus the full regression suite locally.\n\nOne non-blocking note: the new harness uses rrd_lastupdate_r(), which is present in RRDtool 1.4.x but not in older 1.2/1.3 headers that the historical install docs still mention. If that old baseline is still intended, a follow-up compile probe/fallback would avoid surprising downstream test failures.

Issue #328, opened 2026-08-11
tests: `-I$ROOT/lib` shadows the macOS SDK's Availability.h


bonomani:
Sixteen C harnesses in the suite compile with `-I"$ROOT/lib"`. On macOS, whose
filesystem is case-insensitive by default, that directory shadows the SDK's own
`Availability.h` with xymon's `lib/availability.h`, and every subsequent system
prototype becomes a parse error. All sixteen fail; the macOS lanes report
`passed: 27  failed: 26`.

## Symptom

macOS 15, Xcode 26.6:

```
/Users/runner/work/xymon/xymon/lib/Availability.h:53:53: error: unknown type name 'time_t'
   53 | extern int history_color(FILE *fd, time_t snapshot, time_t *starttime, char **histlogname);
      |                                                     ^
In file included from xymond/client_config.c:18:
In file included from .../MacOSX.sdk/usr/include/stdio.h:61:
.../MacOSX.sdk/usr/include/sys/stdio.h:48:56: error: expected function body after function declarator
   48 | int     renameat(int, const char *, int, const char *) __OSX_AVAILABLE_STARTING(__MAC_10_10, __IPHONE_8_0);
      |                                                        ^
fatal error: too many errors emitted, stopping now
```

The first line names it outright. Clang prints the path as spelled in the
`#include`, `Availability.h`, but line 53 is `history_color()` — that is
xymon's `lib/availability.h`, reached through a case-insensitive match. Apple's
`stdio.h` chain includes `<Availability.h>` expecting the SDK's, gets ours,
and so `__OSX_AVAILABLE_STARTING` and `__API_AVAILABLE` are never defined. Every
declaration carrying one then fails to parse. The cascade is loud but entirely
downstream of that single wrong file.

## Why the real build is immune

The build never passes `-I…/lib`. `build/Makefile.rules:11` adds exactly one
include directory:

```make
CFLAGS += -I$(BUILDTOPDIR)/include
```

which is why the macOS lanes build fine and only the test stage falls over. The
harnesses added a second directory the build does not use.

`-I` is also stronger than these harnesses need: it makes the directory
eligible for `#include <...>`, when everything the harnesses actually want from
`lib/` is reached by `#include "..."`.

## Affected

```
tests/lib/svcstatus-cgi.sh
tests/network/netrc-password.sh
tests/server/analysis-ds-firstmatch.sh
tests/server/config-include-dir.sh
tests/server/namematch-case.sh
[...]

Issue #327, opened 2026-08-11
tests: C harnesses link PCRE2 but never resolve its include directory


bonomani:
Two C harnesses in the suite link against PCRE2 but compile without a PCRE
include directory. They work on Linux by accident — `pcre2.h` happens to sit in
`/usr/include` there. On any platform that installs it under a prefix, the
harness does not compile and the test hard-fails.

## Symptom

FreeBSD 15.0, where `pcre2.h` lives in `/usr/local/include`:

```
In file included from /tmp/xymon-test.7165.Fvp6No/harness.c:3:
In file included from include/libxymon.h:81:
include/../lib/loadalerts.h:20:10: fatal error: 'pcre2.h' file not found
   20 | #include <pcre2.h>
      |          ^~~~~~~~~
1 error generated.
FAIL: netrc harness does not compile
```

Nothing about PCRE is under test here. `netrc-password.sh` drives `parse_url()`
in `lib/url.c`; PCRE arrives only because `libxymon.h:81` pulls in
`lib/loadalerts.h`, which includes `<pcre2.h>`. The harness needs the header to
exist, not to use it.

Affected: `tests/network/netrc-password.sh`, `tests/server/config-include-dir.sh`.

## Where

`tests/network/netrc-password.sh` already does half the job — it resolves the
PCRE **libraries** through pkg-config, with an env override and a fallback:

```sh
pcre_libs=${PCRELIBS:-}
if [ -z "$pcre_libs" ] && command -v pkg-config >/dev/null 2>&1; then
	pcre_libs=$(pkg-config --libs libpcre2-8 2>/dev/null || true)
fi
[ -n "$pcre_libs" ] || pcre_libs="-lpcre2-8"
```

and then compiles with no corresponding include path:

```sh
"$CC" -I"$ROOT/include" -I"$ROOT/lib" -o "$work/harness" \
	"$work/harness.c" "$SRC" "$ROOT/lib/libxymoncomm.a" $ssllibs $pcre_libs
```

`tests/server/config-include-dir.sh:91` has the same shape.

The asymmetry is the bug: link flags are discovered, compile flags are assumed.
`pkg-config --cflags libpcre2-8` is the same call, one word apart.

## Suggested fix

Derive `pcre_cflags` next to `pcre_libs`, from the same three sources in the
same order — `$PCREINCDIR`/`$PCRECFLAGS` if exported, then `pkg-config --cflags
libpcre2-8`, then nothing (which is correct on a system where the header is in
[...]

PR #329, opened 2026-08-11
tests: run the tree's Makefiles with GNU make, not whatever is named make

### LINKED ISSUE #326: tests: harnesses call `make`, which is not GNU make on the BSDs
The regression suite invokes `make`. On the BSDs `make` is BSD make, and every
Makefile in this tree is GNU make, so 22 of the 24 tests that build anything
fail there — not because the code under test is broken, but because the harness
picked the wrong tool.

## Symptom

```
make: /home/runner/work/xymon/xymon/lib/Makefile:10: Invalid line "ifeq ($(LOCALCLIENT),yes)", expanded to "ifeq "
make: /home/runner/work/xymon/xymon/lib/Makefile:12: Invalid line "endif"
make: Fatal errors encountered -- cannot continue
make: stopped making "libxymon.a libxymoncomm.a" in /home/runner/work/xymon/xymon/lib
FAIL: cannot build libxymon.a libxymoncomm.a
```

FreeBSD 15.0, server variant: `passed: 28  skipped: 1  failed: 24`. Twenty-two
of those 24 are this one cause; the other two are #327 and #328.

The same failure appears on every FreeBSD, NetBSD and OpenBSD lane. Every Linux
lane passes, which is the whole tell: the split is exactly "is `/usr/bin/make`
GNU make?".

## Why the build works and only the tests fail

The build already knows. `ci/bootstrap-install.sh` selects gmake for the BSDs:

```sh
prepare_os() {
  ...
  if [ "${BUILD_TOOL}" = "make" ]; then
    MAKE_BIN="gmake"
  fi
```

and the build log shows it working — `gmake[1]: Leaving directory
'/home/runner/work/xymon/xymon/web'`. The test stage runs next, in the same
job, and calls plain `make`.

## Where

`tests/lib/assert.sh`, the shared helper both halves of the contract live in:

```sh
require_c_buildenv() {
	require_cc
	command -v make >/dev/null 2>&1 || skip "make not available"      # :189
	[ -f "$1/include/config.h" ] || skip "tree not configured (no include/config.h)"
}

build_xymon_libs() {
	local root=$1 log=$2
	shift 2
	make -C "$root/lib" "$@" >"$log" 2>&1 \                            # :198
		|| { cat "$log" >&2; fail "cannot build $*"; }
}
```

The probe is the more insidious half: on a BSD host with no gmake installed it
finds BSD make, reports the build environment as present, and the test proceeds
to a hard `fail` — exactly the outcome `require_c_buildenv` exists to prevent.

Direct callers outside the helper:

- `tests/lib/build-worker.sh:40` — `make -s -C "$root" -f Makefile -f - __testlibs`
- `tests/lib/build-worker.sh:59` — `make -C "$root/lib" libxymon.a libxymoncomm.a libxymontime.a`
- `tests/rrd/inode-unix-columns.sh:19` — `make -s -C "$ROOT" -f Makefile -f - inode-test-flags`
- `tests/server/analysis-ds-firstmatch.sh:58`, `namematch-case.sh:31`,
  `notbefore-notafter.sh:49`, `xymond-noflap-list.sh:56`,
  `xymond-noflap-prefix.sh:40` — `make -C "$ROOT/lib" libxymoncomm.a`
- `tests/web/showgraph-exec-title.sh:39`, `showgraph-single-service.sh:44` — same

One more, already correct in shape but never supplied: `tests/buildsystem/docs-templating.sh`
reads `MAKE="${MAKE:-make}"`, so it honours an override — nothing in the tree
sets one, and it fails the same way on its own temp Makefile:

```
make: /tmp/xymon-test.5498.1ajFjU/Makefile:27: Invalid line "ifndef PKGBUILD"
```

## Suggested fix

A `xymon_make()` in `tests/lib/assert.sh` that resolves GNU make once — prefer
`$MAKE` if exported, then `gmake`, then `make` if it identifies as GNU make
(`make --version` printing "GNU Make"), and skip otherwise with a message that
names the real condition rather than "make not available". `require_c_buildenv`
probes through the same resolver, so the probe and the invocation can no longer
disagree. The eleven direct call sites route through it.

The skip case matters: on a BSD host without gmake, "GNU make not found (the
Makefiles are GNU make; install gmake)" is a legitimate missing host dependency
under the tests/README.md contract, and skipping is correct. What is not
correct is today's silent selection of a tool that cannot parse the file.

`main`'s two suite-running workflows are both `runs-on: ubuntu-latest`. The
defect is latent on main, not absent — a developer running `./tests/testsuite`
on FreeBSD from a plain main checkout hits all 22 today.

bonomani:
Closes #326.

Every Makefile in the tree is GNU make. The suite ran them with plain `make`,
which on the BSDs is BSD make and stops at `lib/Makefile`'s first conditional,
so 22 tests hard-failed on trees that build fine. The build itself already
picks gmake there; only the test stage did not.

`require_gnu_make()` resolves it once into `$XYMON_MAKE` — `$MAKE` if exported,
then `gmake`, then `make`, each having to identify itself as GNU make. Eleven
call sites go through it.

The probe was the more insidious half of the old code. `command -v make` finds
BSD make, reports the build environment as present, and leaves the test to
hard-fail — the exact outcome `require_c_buildenv` exists to prevent. Now a
host with no GNU make skips, like a missing compiler; what it can no longer do
is select a tool that cannot parse the file.

## Verified

Linux server build, three runs, no change from the `main` baseline of
`53/0/0`:

```
passed:  53 skipped: 0 failed:  0 (run 1)
passed:  53 skipped: 0 failed:  0 (run 2)
passed:  53 skipped: 0 failed:  0 (run 3)
```

The three runs are not decoration. The first draft used `make --version | head
-1 | grep -q`, and under the `set -o pipefail` every test runs with, `head`
exiting early can SIGPIPE the writer and fail the pipeline for a candidate that
answered correctly — it skipped a different test on each run. Matching the
banner through a `case` removed the pipeline and the flakiness with it.

**A BSD-like `make` first on `PATH`, gmake present** — a stub reporting `bmake
(NetBSD make) 20200710` and refusing to parse conditionals, exactly what the
FreeBSD lanes have:

```
passed:  52 skipped: 1 failed:  0
```

All 22 previously-failing tests pass by resolving gmake. The one skip is
`configure-no-rrd.sh`, which runs the real `./configure --server`; configure
does its own GNU-make detection and aborts before reaching the RRD probe, which
is that test's own documented skip path, not a regression.

**No GNU make anywhere** (both `make` and `gmake` stubbed non-GNU):

```
tests/server/namematch-case.sh          rc=77  SKIP: GNU make not found (the tree's Makefiles are GNU make -- install gmake)
tests/buildsystem/docs-templating.sh    rc=77  SKIP: GNU make not found ...
tests/xymond/hostdata-save-throttle.sh  rc=77  SKIP: GNU make not found ...
tests/rrd/inode-unix-columns.sh         rc=77  SKIP: GNU make not found ...
```

Skip, not fail, with a message naming the fix.

**`MAKE=gmake` override** — honoured (`rc=0`). `docs-templating.sh` already read
`$MAKE`; it keeps that and gains the probe.

## What this does not prove

The BSD behaviour above is a simulation on Linux, not a BSD run. `main`'s CI is
`ubuntu-latest` only, so nothing here can execute on a platform where
`/usr/bin/make` is BSD make. Real confirmation needs a run of the
multi-platform lane set after this merges.

## Merge order

Touches files also touched by the PRs for #327 and #328 — `tests/lib/assert.sh`,
`tests/lib/build-worker.sh` and seven of the inline-`make` tests. Whichever
merges second needs a rebase; the edits are on adjacent lines, not the same
ones, except in `build-worker.sh` where all three change the compile/build
lines.

### DIFF
--- a/tests/buildsystem/docs-templating.sh
+++ b/tests/buildsystem/docs-templating.sh
@@ -19,9 +19,8 @@ set -euo pipefail
 
 ROOT=$(find_root)
 DOCS="$ROOT/docs"
-MAKE="${MAKE:-make}"
 
-command -v "$MAKE" >/dev/null 2>&1 || skip "no make"
+require_gnu_make
 for f in Makefile install.html.DIST xymon-apacheconf.txt.DIST; do
 	[ -f "$DOCS/$f" ] || skip "docs/$f absent -- predates #90"
 done
@@ -30,7 +29,7 @@ work=$(mktempdir)
 cp "$DOCS/Makefile" "$DOCS/install.html.DIST" \
    "$DOCS/xymon-apacheconf.txt.DIST" "$work/"
 
-"$MAKE" -C "$work" install.html xymon-apacheconf.txt \
+"$XYMON_MAKE" -C "$work" install.html xymon-apacheconf.txt \
 	INSTALLETCDIR=/etc/xymon XYMONTOPDIR=/opt/xymon XYMONUSER=xymonuser \
 	INSTALLWWWDIR=/var/www/xymon CGIDIR=/cgi SECURECGIDIR=/secure-cgi \
 	>"$work/make.log" 2>&1 \
--- a/tests/lib/assert.sh
+++ b/tests/lib/assert.sh
@@ -179,14 +179,50 @@ asan_usable() {
 	[ "$__xymon_tests_asan_usable" = yes ]
 }
 
-# require_c_buildenv ROOT -- skip unless a C compiler, make, and a configured
-# tree (include/config.h) are present. The shared baseline for every test
-# that compiles a harness against the in-tree libraries, so a new shared
+# require_gnu_make -- resolve GNU make into $XYMON_MAKE, or skip.
+#
+# Every Makefile in the tree is GNU make: lib/Makefile opens with an `ifeq`,
+# build/Makefile.rules uses pattern rules. On the BSDs /usr/bin/make is BSD
+# make, which stops at the first conditional --
+#
+#     make: lib/Makefile:10: Invalid line "ifeq ($(LOCALCLIENT),yes)"
+#
+# -- so a test that ran `make` there hard-failed on a tree that builds
+# perfectly well with gmake, which is what the build itself uses.
+#
+# Resolution order: $MAKE if the caller exported one (autopkgtest, a distro
+# build), then gmake, then make. Each candidate must identify itself as GNU
+# make; a candidate that exists but is not GNU make is passed over rather
+# than trusted, because "a program named make is on PATH" is the assumption
+# that produced the failure above.
+#
+# No GNU make at all is a missing host dependency, so it skips -- same
+# contract as a missing compiler (tests/README.md). Resolved once: PATH
+# cannot change inside one test process.
+require_gnu_make() {
+	[ -n "${XYMON_MAKE:-}" ] && return 0
+	local candidate
+	for candidate in ${MAKE:-} gmake make; do
+		command -v "$candidate" >/dev/null 2>&1 || continue
+		# Matched on the whole banner rather than `--version | head -1`:
+		# every test runs under `set -o pipefail`, and head exiting after
+		# the first line can SIGPIPE the writer, failing the pipeline for
+		# a candidate that answered correctly.
+		case $("$candidate" --version 2>/dev/null) in
+			"GNU Make"*) XYMON_MAKE=$candidate; return 0 ;;
+		esac
+	done
+	skip "GNU make not found (the tree's Makefiles are GNU make -- install gmake)"
+}
+
+# require_c_buildenv ROOT -- skip unless a C compiler, GNU make, and a
+# configured tree (include/config.h) are present. The shared baseline for every
+# test that compiles a harness against the in-tree libraries, so a new shared
 # skip condition lands in one place; a test may still add stricter
 # conditions of its own (e.g. "tree already built") next to its call.
 require_c_buildenv() {
 	require_cc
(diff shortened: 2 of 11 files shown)


SoundGoof (approved):
## Review

No blocking findings.

The resolver correctly rejects BSD make and selects GNU Make from
`$MAKE`, `gmake`, or `make`. All migrated call sites use the resolved
executable.

Validation:

- Linux suite passes.
- NetBSD: 50 passed, 3 skipped, 0 failed.
- FreeBSD: 52 passed, 1 skipped, 0 failed.
- OpenBSD: 50 passed, 3 skipped, 0 failed.
- Full builds passed with RRDtool, OpenSSL, OpenLDAP, system c-ares,
  and Net-SNMP enabled.
- `git diff --check` passes.
- No remaining plain `make -C` invocation exists in the changed files.

Residual test limitations:

- OpenBSD's base compiler does not support AddressSanitizer, so the
  two `xtree` ASan tests were not exercised there. They are covered
  on platforms with ASan support.
- Two unrelated fixtures are not natively portable to OpenBSD:
  `fs-filter-linux.sh` assumes `printf` is a shell builtin in its
  restricted `PATH`, and `history-tail-resync.sh` uses GNU-only
  `head -c`. Both passed with fixture-only compatibility adjustments
  outside the reviewed PR tree.
- `configure-no-rrd.sh` intentionally removes GNU Make and exits 77.
  This is expected fixture behavior, not missing coverage.

These limitations are unrelated to PR #329 and do not indicate a
defect in its GNU Make resolution.

Approved from review and validation.

PR #330, opened 2026-08-11
tests: resolve the PCRE include directory, not just its libraries

### LINKED ISSUE #327: tests: C harnesses link PCRE2 but never resolve its include directory

bonomani:
Closes #327.

Harnesses that include `libxymon.h` need `<pcre2.h>` whether or not they use
PCRE — `libxymon.h` pulls in `lib/loadalerts.h`, which includes it. They
resolved the PCRE *libraries* through an env override and pkg-config, then
compiled with no matching include path. That works on Linux only because
`/usr/include/pcre2.h` is already on the default search path; where PCRE lives
under a prefix it does not:

```
lib/loadalerts.h:20:10: fatal error: 'pcre2.h' file not found
FAIL: netrc harness does not compile
```

on every FreeBSD lane, in tests that have nothing to do with PCRE.

`pcre_cflags()` resolves it in the order the link flags already use, most
specific first: `$PCRECFLAGS`, then the configured tree's `PCREINCDIR` — by
definition what the library being linked was built against, and it already
carries its `-I` — then `pkg-config --cflags`, then nothing, which is correct
where the header is already found.

**Three call sites, not the two the issue lists.** `tests/lib/build-worker.sh`
compiles xymond sources through the same include chain and would have hit this
next, once #326 lets the BSD lanes get far enough to compile at all.

## Verified

Linux server build: `passed: 53  skipped: 0  failed: 0`, unchanged from the
`main` baseline. Here `pcre_cflags` resolves to the empty string — the header is
in `/usr/include` — so the no-op path is the one CI exercises today.

Each source in turn:

```
default:      []
env override: [-I/opt/pcre/include]        # PCRECFLAGS
makefile:     [-I/from/the/configured/tree] # PCREINCDIR appended to a scratch Makefile
```

That the flag reaches the compiler, and takes precedence, is shown with a decoy
header — a directory holding a `pcre2.h` that is nothing but `#error decoy
pcre2.h was used`, pointed at with `PCRECFLAGS`:

```
$ PCRECFLAGS=-I.../decoy ./tests/network/netrc-password.sh
.../decoy/pcre2.h:1:2: error: #error decoy pcre2.h was used
```

Same for `config-include-dir.sh` (4 hits) and, through `build-worker.sh`,
`hostdata-save-throttle.sh` (8 hits). A prefixed `pcre2.h` is now found for the
same reason the decoy is.

## What this does not prove

The failing platform is not reachable from `main`'s CI, which is
`ubuntu-latest` only. The decoy stands in for `/usr/local/include/pcre2.h`;
real confirmation needs a FreeBSD run after this merges.

## Merge order

`tests/lib/assert.sh`, `tests/lib/build-worker.sh`,
`tests/network/netrc-password.sh` and `tests/server/config-include-dir.sh` are
also touched by the PRs for #326 and #328. Whichever merges second needs a
rebase.

### DIFF
--- a/tests/lib/assert.sh
+++ b/tests/lib/assert.sh
@@ -236,6 +236,57 @@ build_xymon_libs() {
 		|| { cat "$log" >&2; fail "cannot build $*"; }
 }
 
+# pcre_cflags ROOT -- print the compile flags a harness needs to reach
+# <pcre2.h>, or nothing when the header is already on the default search
+# path.
+#
+# Every harness that includes libxymon.h needs this whether or not it uses
+# PCRE: libxymon.h pulls in lib/loadalerts.h, which includes <pcre2.h>. The
+# harnesses resolved the PCRE *libraries* and assumed the *header*, which
+# holds on Linux (/usr/include/pcre2.h) and fails wherever PCRE lives under
+# a prefix -- on FreeBSD, /usr/local/include:
+#
+#     lib/loadalerts.h:20:10: fatal error: 'pcre2.h' file not found
+#
+# Resolution mirrors the PCRELIBS order already used for the link, most
+# specific first: an explicit override, then the configured tree (whose
+# PCREINCDIR is by definition what the library being linked was built
+# against, and already carries its -I), then pkg-config, then nothing.
+pcre_cflags() {
+	local root=$1 flags=${PCRECFLAGS:-}
+	[ -n "$flags" ] || [ ! -f "$root/Makefile" ] \
+		|| flags=$(sed -n 's/^PCREINCDIR *= *//p' "$root/Makefile")
+	if [ -z "$flags" ] && command -v pkg-config >/dev/null 2>&1; then
+		flags=$(pkg-config --cflags libpcre2-8 2>/dev/null || true)
+	fi
+	printf '%s' "$flags"
+}
+
+# xymon_cflags ROOT -- the include flags a harness compiled against the
+# in-tree libraries needs: the tree's own headers, and wherever <pcre2.h>
+# lives. Callers add whatever else they need after it (the harness's own
+# -iquote ROOT/web or ROOT/xymond, -DSTANDALONE, the RRD defines).
+#
+# The tree's directories go on the *quoted* path and the PCRE directory on
+# the angle-bracket one, because that is how each is included: every in-tree
+# header is reached through #include "...", while loadalerts.h asks for
+# <pcre2.h>. Putting the tree on the angle-bracket path is what makes
+# lib/availability.h answer the macOS SDK's #include <Availability.h> on a
+# case-insensitive filesystem (#328).
+#
+# Bundled into one helper rather than offered as separate flags on purpose.
+# Neither mistake here was a wrong flag: each was a flag that every harness
+# had to remember for itself. The PCRE path had to be remembered fourteen
+# times and eleven compile lines did not remember it; -iquote has to be
+# remembered at forty-two flags across nineteen files, and the next harness
+# written will have to remember it again. A compile line can no longer say
+# where the tree's headers are without also saying how to reach them and
+# where pcre2.h is.
+xymon_cflags() {
+	local root=$1
+	printf '%s' "-iquote $root/include -iquote $root/lib $(pcre_cflags "$root")"
+}
+
 # ---- repo location -----------------------------------------------------------
 
 # find_root -- print the absolute path of the repo root, derived from the
--- a/tests/lib/build-worker.sh
+++ b/tests/lib/build-worker.sh
@@ -21,7 +21,7 @@
 __XYMON_TESTS_BUILD_WORKER_SOURCED=1
 
 build_xymond_worker() {
-	local outdir=$1 prog=$2 root cc treelibs pcre_libs src srcs mkfiles inc
+	local outdir=$1 prog=$2 root cc treelibs pcre_libs harness_cflags src srcs mkfiles inc
 	shift 2
 	root=$(find_root)
 	cc=${CC:-cc}
@@ -55,6 +55,7 @@ build_xymond_worker() {
 		pcre_libs=$(pkg-config --libs libpcre2-8 2>/dev/null || true)
 	fi
 	[ -n "$pcre_libs" ] || pcre_libs="-lpcre2-8"
+	harness_cflags=$(xymon_cflags "$root")
 
 	"$XYMON_MAKE" -C "$root/lib" libxymon.a libxymoncomm.a libxymontime.a >"$outdir/libbuild.log" 2>&1 \
 		|| { cat "$outdir/libbuild.log" >&2; fail "the xymon libraries do not build in this configured tree"; }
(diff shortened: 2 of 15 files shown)


SoundGoof:
## Review

### : Apply `pcre_cflags()` to all affected harnesses

The helper is correct, but only three compile paths use it.

Full builds succeeded on FreeBSD, NetBSD, and OpenBSD with PCRE2 in
their native prefixed locations. With ambient include-path variables
unset, the same fourteen tests fail on every BSD because their compiler
commands still cannot find `<pcre2.h>`.

Examples include:

- `tests/server/analysis-ds-firstmatch.sh:68`
- `tests/server/namematch-case.sh:84`
- `tests/server/notbefore-notafter.sh:165`
- `tests/server/xmh-item-names.sh:65`
- `tests/web/showgraph-exec-title.sh:42`
- `tests/web/showgraph-single-service.sh:47`
- The shared `tests/lib/svcstatus-cgi.sh` compile path

Apply `pcre_cflags()` to every affected harness compiler invocation,
including shared compile helpers, then run the complete suite with
`CPATH`, `C_INCLUDE_PATH`, and `CPLUS_INCLUDE_PATH` unset.

### Validation

The exact synthetic merge of current `main` and PR #330 was tested.

- All five existing GitHub checks pass.
- The synthetic merge is clean and preserves PR #329.
- FreeBSD: 38 passed, 1 skipped, 14 failed.
- NetBSD: 36 passed, 3 skipped, 14 failed.
- OpenBSD: 34 passed, 3 skipped, 16 failed.
- Fourteen identical PCRE include failures occur on every BSD.
- OpenBSD's other two failures are unrelated fixture portability issues.
- The three paths modified by this PR pass on all tested BSDs.

### Residual

Without `pkg-config`, existing harness library resolution falls back to
bare `-lpcre2-8` instead of configured `PCRELIBS`. Testing therefore
used each platform's library search path. This predates PR #330 but
remains a separate BSD harness limitation.

Changes requested because the PR only partially fixes the stated
prefixed-PCRE harness failure.

bonomani:
Thanks — rebased onto `main` with #331 merged, and extended.

Since your list says "examples", I derived the full set with `cc -H` instead of working from it: **eleven** sites needed the flag, not seven. The four extra are `xymond-noflap-list.sh`, `xymond-noflap-prefix.sh`, `html-log-custom-graphs.sh` and `history-reportlog-reject.sh`. If those were passing on your BSDs, my reasoning about them is wrong and I'd like to know.

Rather than adding the flag to eleven more compile lines, `xymon_cflags()` now emits the tree's include directories together with the PCRE one, and the fourteen harnesses take that instead of naming the directories themselves — so it can't be forgotten separately next time. That also folds in #331's `-iquote` for those sites; the files it touched that have no route to `<pcre2.h>` keep their own flags.

Checks: forcing `PCRECFLAGS` to a distinctive `-I`, 21 translation units reach `<pcre2.h>` and 0 lack it (reverting one harness makes the check report it, so the zero is real). Logging all 43 compiles, no in-tree directory appears on the angle-bracket path, so #328 stays fixed. Suite 53/0/0.

Linux can't reproduce the original failure — `/usr/include/pcre2.h` is on the default search path, so CI is green either way and unsetting `CPATH`/`C_INCLUDE_PATH`/`CPLUS_INCLUDE_PATH` changes nothing here. The BSD re-run is still the real check.

The `pkg-config`/`PCRELIBS` residual you flagged is untouched, as agreed.

SoundGoof (approved):
## Findings

No blocking findings.

The previous coverage issue is resolved. All test harnesses that reach
`pcre2.h` through `libxymon.h` now use `xymon_cflags()`.

Non-blocking: the PR description still says “Three call sites,” while the
refreshed patch now updates fourteen compilation paths. The description
should reflect the expanded scope.

## Validation

- Linux: 54 passed, 0 skipped, 0 failed.
- FreeBSD 14.3: 53 passed, 1 intentional skip, 0 failed.
- NetBSD 11.0: 51 passed, 3 capability skips, 0 failed.
- OpenBSD 7.9: every affected harness passed.
- No ambient `CPATH`, `C_INCLUDE_PATH`, or `CPLUS_INCLUDE_PATH` was used.
- `git diff --check` passes.

The two OpenBSD full-suite failures remain unrelated existing fixture
portability issues.

## Residual Risk

Library and runtime paths for prefixed PCRE installations remain a separate
problem. NetBSD still requires library-path support when `pkg-config` or an
embedded runtime path is unavailable. That does not affect this header-path
fix.

PR #331, opened 2026-08-11
tests: reach in-tree headers with -iquote, not -I

### LINKED ISSUE #328: tests: `-I$ROOT/lib` shadows the macOS SDK's Availability.h

bonomani:
Closes #328.

The harnesses put the tree's own source directories on the **angle-bracket**
include path. On macOS, whose filesystem is case-insensitive, `-I$ROOT/lib`
then answers the SDK's own `#include <Availability.h>` with xymon's
`lib/availability.h`. `__OSX_AVAILABLE_STARTING` and `__API_AVAILABLE` are
never defined, and every system prototype carrying one fails to parse:

```
lib/Availability.h:53:53: error: unknown type name 'time_t'
   53 | extern int history_color(FILE *fd, time_t snapshot, ...);
sys/stdio.h:48:56: error: expected function body after function declarator
```

The first line names it: clang prints the path as spelled in the `#include`,
but line 53 is `history_color()` — ours. All sixteen affected tests fail; the
macOS lanes report `27 passed, 26 failed`.

The real build is immune because it never passes the flag —
`build/Makefile.rules:11` adds `-I$(BUILDTOPDIR)/include` and nothing else.

`-iquote` grants exactly what the harnesses use. Every in-tree header they
reach is reached through `#include "..."`, and no source in the tree — product
or test — includes an in-tree header with angle brackets. So the SDK gets its
`Availability.h` back and the harnesses lose nothing.

Applied to all four in-tree directories (`include`, `lib`, `web`, `xymond`)
rather than `lib` alone: the collision is a property of putting the tree's
sources on the angle-bracket path at all, and `lib/availability.h` is only
where it landed first. 42 flags across 19 files.

## Verified

Linux server build: `passed: 53  skipped: 0  failed: 0`, unchanged from the
`main` baseline, and no in-tree `-I` left in `tests/`.

Linux is case-sensitive, so the collision itself cannot be reproduced here. What
can be isolated is the compiler semantics the fix turns on — a directory holding
an empty `availability.h`, reached both ways, on both compilers:

| | `#include <availability.h>` | `#include "availability.h"` |
| --- | --- | --- |
| `-I dir` | **found** — the macOS hazard | found |
| `-iquote dir` | **not found** | **found** |

gcc 12 and clang 14, identical results. The middle cell is the bug; the right
column is why nothing breaks.

Also checked before converting: nothing in the tree does `#include <libxymon.h>`
or the equivalent for any other in-tree header, and the only angle-bracket
includes in test harness sources are system headers
(`<netinet/in.h>`, `<arpa/inet.h>`).

## What this does not prove

That macOS goes green. `main`'s CI is `ubuntu-latest` only; real confirmation
needs a macOS run after this merges. One macOS failure is also **not** fixed
here and needs its own change: `tests/server/client-msgs-sections.sh:9` uses
`mapfile`, a bash 4 builtin, and macOS ships bash 3.2 (`rc=127`). Noted in #328.

## Merge order

Overlaps the PRs for #326 (nine files) and #327 (three files). Whichever merges
second needs a rebase; in `tests/lib/build-worker.sh` all three PRs touch the
compile line.

### DIFF
--- a/tests/lib/build-worker.sh
+++ b/tests/lib/build-worker.sh
@@ -63,7 +63,7 @@ build_xymond_worker() {
 	for src in "$@"; do srcs+=("$root/$src"); done
 
 	# Archives listed twice rather than --start-group, which is GNU ld only.
-	"$cc" -I"$root/include" -I"$root/lib" -I"$root/xymond" -o "$outdir/$prog" \
+	"$cc" -iquote "$root/include" -iquote "$root/lib" -iquote "$root/xymond" -o "$outdir/$prog" \
 		"${srcs[@]}" \
 		"$root/lib/libxymon.a" "$root/lib/libxymoncomm.a" "$root/lib/libxymontime.a" \
 		"$root/lib/libxymon.a" \
--- a/tests/lib/svcstatus-cgi.sh
+++ b/tests/lib/svcstatus-cgi.sh
@@ -116,7 +116,7 @@ svcstatus_build() {
 		return 0
 	fi
 
-	"$CC" "$@" -I"$ROOT/include" -I"$ROOT/lib" -I"$ROOT/web" -o "$work/svcstatus" \
+	"$CC" "$@" -iquote "$ROOT/include" -iquote "$ROOT/lib" -iquote "$ROOT/web" -o "$work/svcstatus" \
 		"$ROOT/web/svcstatus.c" "$ROOT/web/svcstatus-info.c" "$ROOT/web/svcstatus-trends.c" \
 		"$ROOT/lib/libxymon.a" "$ROOT/lib/libxymoncomm.a" "$ROOT/lib/libxymon.a" \
 		$pcre_libs $ssllibs $netlibs $librtdef 2>"$work/cc.log" || return 1
--- a/tests/network/netrc-password.sh
+++ b/tests/network/netrc-password.sh
@@ -86,7 +86,7 @@ EOF
 
 # lib/url.c goes before the archive so its fresh objects satisfy the netrc
 # symbols and the archive's (possibly stale) url.o is never pulled in.
-"$CC" -I"$ROOT/include" -I"$ROOT/lib" -o "$work/harness" \
+"$CC" -iquote "$ROOT/include" -iquote "$ROOT/lib" -o "$work/harness" \
 	"$work/harness.c" "$SRC" "$ROOT/lib/libxymoncomm.a" $ssllibs $pcre_libs 2>"$work/cc.log" \
 	|| { cat "$work/cc.log" >&2; fail "netrc harness does not compile"; }
 
--- a/tests/rrd/inode-unix-columns.sh
+++ b/tests/rrd/inode-unix-columns.sh
@@ -34,7 +34,7 @@ rrdlibs=${buildflags[4]#rrdlibs=}
 
 # Configured flags are deliberate word-split lists.
 # shellcheck disable=SC2086
-"$CC" $ldflags -I"$ROOT/include" $rrddef $rrdincdir \
+"$CC" $ldflags -iquote "$ROOT/include" $rrddef $rrdincdir \
 	-o "$work/rrd-lastupdate" $rpathopt \
 	"$ROOT/tests/rrd/inode-unix-columns-harness.c" $rrdlibs \
 	2>"$work/cc.log" || {
--- a/tests/server/analysis-ds-firstmatch.sh
+++ b/tests/server/analysis-ds-firstmatch.sh
@@ -65,7 +65,7 @@ if [ -z "$pcre_libs" ] && command -v pkg-config >/dev/null 2>&1; then
 fi
 [ -n "$pcre_libs" ] || pcre_libs="-lpcre2-8"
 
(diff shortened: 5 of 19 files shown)


SoundGoof (approved):
## Findings

No blocking findings.

The migration covers all test-harness source-tree include paths. Generated
shim directories correctly remain under `-I`.

No in-tree header is consumed through an angle-bracket include, so changing
these paths to `-iquote` does not remove required header resolution.

## Validation

- Linux: 53 passed, 0 skipped, 0 failed.
- FreeBSD 14.3: 52 passed, 1 intentional skip, 0 failed.
- NetBSD 11.0: 47 passed, 6 capability/build skips, 0 failed.
- OpenBSD 7.9: every affected runnable test passed.
- `git diff --check` passes.
- No stale in-tree `-I` paths remain.

The two OpenBSD full-suite failures are unrelated existing fixture issues:
the Linux filesystem fixture omits external `printf` from its restricted
PATH, and the history fixture uses GNU-only `head -c`.

## Residual Risk

The original case-insensitive macOS SDK collision was not reproduced on a
macOS host. The include-search invariant, compiler semantics, green CI, and
Linux/BSD regression runs support the fix.

PR #333, opened 2026-08-11
xymond: stop checkpointing a gap that can no longer bridge

### LINKED ISSUE #332: xymond: an overridden test keeps its no-data gap, and its checkpoint record, forever
### What happens

A no-data gap (`pregapcolor`/`gapstart`, added in #250) ends when the color xymond
records becomes the reporter's own again. While something *overrides* that color,
it never does — so the gap stays open indefinitely, and `save_checkpoint()` writes a
`.flapstate.` record for that log in every checkpoint from the first override onward.

The everyday trigger is not exotic: `xymond_client` emits a `modify` for every
matching RRDDS rule on every client message (`xymond/client_config.c`, the
`modify %s.%s %s rrdds` line). While one is in force, each genuine report is recorded
as the modifier's color, never the reporter's, so nothing can close the gap.

### Measured

Driving a real xymond — `status red`, then a `modify … yellow` refreshed alongside
three genuine `status red` reports, with `--checkpoint-interval=5`:

```
board=yellow, checkpoint gap: red gapstart=1786486814
```

The record is still there after the gap has aged past its bridge window, at which
point `holdtime_bridges()` would refuse it whatever color arrived next.

### Why it is small, and why it is still worth fixing

Nothing reads the gap except `holdtime_bridges()`, so an expired one cannot produce a
wrong `lastchange`: `gapstart` ages past `GAPBRIDGE_VALIDITIES * validity` and the
bridge is refused. The in-memory cost is zero — the fields live in an existing struct.

What it does cost is the checkpoint: one ~80-byte record per modified log, per save,
forever. On a server with RRDDS rules across a few thousand tests that is a few
hundred KB of state per checkpoint that can no longer affect any decision, and it
misleads anyone reading the file about what the daemon believes.

### Fix

Split the length half of `holdtime_bridges()` out as `gap_within_window()` and ask it
at save time as well, so a gap past its window is neither a reason to write the record
nor written into one flap state has already earned. Runtime behavior is unchanged.

Depends on #250 (the gap fields do not exist before it).

bonomani:
Closes #332.

**Stacked on #250** — the gap fields do not exist before it. Until #250 merges this
PR shows its commit too; that drops out of the diff once it lands.

A no-data gap ends when the color being recorded becomes the reporter's own again.
While something overrides that color it never does, so the gap stays open and its
`.flapstate.` record is written in every checkpoint from the first override onward.
`xymond_client` refreshes an RRDDS modifier every client cycle, so on a server using
DS rules that is every one of those logs, forever.

Past `GAPBRIDGE_VALIDITIES` report validities the gap decides nothing —
`holdtime_bridges()` refuses it whatever color arrives next. This splits that length
test out as `gap_within_window()` and asks it at save time as well.

Runtime behavior is unchanged: nothing else reads the gap, and one the saver drops is
one the loader could only restore to have refused. The single visible difference is
across a restart — validity is not checkpointed, so a restored log falls back to
`defaultvalidity`, and a gap long expired against its own `status+1` window would
otherwise come back bridgeable against a 30-minute one.

### Tests

`xymond-checkpoint-roundtrip.sh` gains both halves, driven without sending a status so
the gap is written back exactly as the daemon holds it: sixty seconds old it must
survive the round trip, a day old it must not. Reintroducing the defect fails it:

```
FAIL: a day-old gap was written back as red; nothing can bridge it any more,
      so it is state that never goes away
```

`xymond-holdtime-bridge.sh` extracts both halves of the rule now, so splitting the pair
without updating it fails to link rather than silently testing half the rule.

All 22 tests in `tests/server/` pass.

### DIFF
--- a/common/xymon.1
+++ b/common/xymon.1
@@ -206,7 +206,8 @@ Status color (green, yellow, red, blue, clear, purple)
 .IP "\fBtestflags\fR"
 For network tests, the flags indicating details about the test (used by xymongen).
 .IP "\fBlastchange\fR"
-Unix timestamp when the status color last changed.
+Unix timestamp when the current status color began, i.e. the hold-time of
+the color shown. See "Hold-time and no-data gaps" below.
 .IP "\fBlogtime\fR"
 Unix timestamp when the log message was received.
 .IP "\fBvalidtime\fR"
@@ -292,6 +293,21 @@ and with the following inequalities:  >= > <= < = !=
 
 These filters are: lastchange, logtime, validtime, acktime, disabletime
 
+Hold-time and no-data gaps
+
+\fBlastchange\fR is the hold-time of the color shown: when that color
+began, not when a message last moved it. If a test stops reporting and
+later resumes with the same color, the silence counts as hold-time and the
+value is unchanged across it - the status did not move while it was out of
+sight. A gap longer than four times the report validity is not bridged, and
+a resume with a different color starts a new hold-time. The gap is measured
+from when the status went stale, itself one report validity after the last
+report.
+
+This value can therefore move backwards, which matters for the
+\fBlastchange\fR filter above: a poller asking for changes since its last
+look will not see a bridged recovery.
+
 The response is one line for each status that matches the CRITERIA,
 or all statuses if no criteria is specified. The line is composed of
 a number of fields, separated by a pipe-sign. You can select which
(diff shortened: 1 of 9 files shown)


bonomani:
Folded into #250 — same unmerged commit introduces the gap state and this bug, and it could not merge before #250 anyway.

PR #334, opened 2026-08-12
xymond_channel: stop instead of spinning on a fatal semaphore error (TBT 177)

bonomani:
`xymond_channel` idles by blocking on `semop(GOCLIENT)`. When the channel's semaphore set goes away under it — xymond torn down, a stale set cleared by hand, `ipcrm` — that call stops blocking and starts failing immediately and *permanently* with `EIDRM`.

`main` handles that in the "Semaphore wait aborted" branch, which logs at debug level and `continue`s. The `continue` is what makes it expensive: it skips the rest of the loop body, including the `select()` in the peer-write section that is the cycle's only other blocking call. Nothing touches `running`, and the error is permanent, so the next iteration fails identically. An idle waiter becomes a busy-loop that runs until someone kills it.

### Measured, not assumed

Driving the real binary against a real channel and removing the semaphore set, before the fix:

| | state | CPU |
|---|---|---|
| blocked on `semop` | `S` | 0.0% |
| semaphore set removed | `R` | 322 ticks (3.2 s) per 3 s wall — a saturated core |

It never exited.

### The fix

Test the `semop` result where the error first surfaces, between the call and the `n == 0` branch. Anything that is not `EAGAIN` (the `IPC_NOWAIT` "nothing pending" answer) or `EINTR` (a signal) is permanent, so clear `running` and let the existing loop condition end the process through its normal shutdown path. No new exit route.

### Structure

Two commits, as with the other Terabithia ports (#200, #221, #225):

1. **`xymond_channel: exit when fatal semaphore errors are encountered`** — devel `7ffd904b103838a193d73e4fc08a3692ab7d63f7` (TBT 177) applied verbatim, authored by J.C. Cleaver with the original 2015-11-29 author date and a `cherry picked from` trailer.
2. **`xymond_channel: correct and cover the TBT 177 fatal-semop stop`** — our corrections and the test, so the correction diff reads against the port rather than against `main`.

The corrections in (2):

- `dbgprintf` → `errprintf`. As ported, a worker dying on a fatal IPC error leaves no trace unless someone happened to be running with `--debug`. This is a death notice, and it now goes where this file's other death notices go ("Child process %d died").
- Message reworded to match those, without the redundant program-name prefix `errprintf` does not need.
- The comment says which errnos are the normal ones and why the existing "Semaphore wait aborted" `continue` is the expensive path.
- Dropped the stray trailing whitespace and the extra space in the condition.

The resulting tree is byte-identical to the previous single-commit revision of this PR; only the history changed.

### Test

`tests/xymond/channel-fatal-semop.sh` drives the real `xymond_channel` against a real channel. The harness creates the channel's IPC the way xymond does (`setup_channel` with `CHAN_MASTER`; SysV IPC is kernel-persistent, so it can create and exit rather than live alongside the test). The test waits for `xymond_channel` to register itself by raising `CLIENTCOUNT` — without that gate it could break the channel before the process reached the loop under test and pass for the wrong reason — then `IPC_RMID`s the semaphore set and requires the process to exit.

- Passes in 0.17 s with the fix.
- Fails on the unfixed binary in ~12 s, reporting the burn it measured: `cpu ticks used: 1083` across the poll window.
- Exit status **and** the diagnostic are asserted, so a crash cannot pass as a clean stop.
- The channel is keyed off `ftok(XYMONHOME)`, so the test's private `XYMONHOME` keeps it clear of a real Xymon on the same host and of a parallel copy of itself. Cleanup removes both ids; the runs leave no IPC behind.

Suite: 54 passed, 0 skipped, 0 failed on a built server tree.

### Merge note

#172 / #173 insert a `hupchildren` block immediately **above** the same `semop` this touches. Different lines, shared context — adjacent hunks, so expect a trivial context resolution for whichever lands second rather than a surprise. No dependency either way.

Refs #29 (TBT `177` = `7ffd904b1`), #106

### DIFF
--- a/tests/xymond/channel-fatal-semop-harness.c
+++ b/tests/xymond/channel-fatal-semop-harness.c
@@ -1,0 +1,99 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later */
+/*
+ * tests/xymond/channel-fatal-semop-harness.c
+ *
+ * Stands in for xymond as the owner of a channel's SysV IPC, so a test can
+ * run the real xymond_channel against a real channel and then tear the
+ * semaphore set out from under it.
+ *
+ * SysV IPC is kernel-persistent: the ids stay valid after the creating
+ * process exits, so "create" can create and return rather than having to
+ * stay alive alongside the test.
+ *
+ *   create <channelname>    create the channel as CHAN_MASTER; print its
+ *                           shmid and semid on stdout as SHMID=/SEMID=
+ *   clients <semid>         print the CLIENTCOUNT semaphore.  A test must not
+ *                           break the channel before xymond_channel has
+ *                           attached to it, or the process exits for the
+ *                           wrong reason and the test passes vacuously;
+ *                           CLIENTCOUNT going to 1 is that readiness gate.
+ *   rmsem <semid>           IPC_RMID the semaphore set.  Every semop already
+ *                           blocked on it returns -1 with EIDRM, and every
+ *                           later one fails the same way: the permanent,
+ *                           fatal error the test is about.
+ *   rmall <shmid> <semid>   remove both, for cleanup.
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <errno.h>
+#include <sys/types.h>
+#include <sys/ipc.h>
+#include <sys/shm.h>
+#include <sys/sem.h>
+
+#include "libxymon.h"
+
+/* Same lookup xymond_channel does for --channel= (xymond_channel.c). */
+static int channel_id(char *name)
+{
+	int cnid;
+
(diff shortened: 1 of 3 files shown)


SoundGoof (approved):
## Findings

No blocking findings.

The error classification is correct. `EAGAIN` is expected with
`IPC_NOWAIT`, and `EINTR` is retryable. Other documented `semop()` errors
cannot be resolved by immediately retrying and would otherwise cause the
reported busy loop.

The test uses real SysV IPC and waits for `xymond_channel` to attach before
removing the semaphore set. This prevents a startup failure from producing
a false pass.

## Validation

- Linux full suite: 54 passed, 0 skipped, 0 failed.
- Linux real-IPC regression passed.
- FreeBSD 14.3 real-IPC regression passed.
- NetBSD 11.0 real-IPC regression passed.
- OpenBSD 7.9 real-IPC regression passed.
- `git diff --check` passes.

PR #335, opened 2026-08-12
tests: record why showgraph's exec-title injection checks are decidable

bonomani:
showgraph-exec-title.sh asserts four times that a file does *not* exist,
which is how it proves an injected touch never ran. An absence check is
only meaningful once whatever would create the file has finished, and
here that is not this shell's child: showgraph runs the title script
under popen(), and the wait happens in pclose() (web/showgraph.c).

Reading the title off the pipe would not be enough on its own -- that
proves the script wrote a line, not that it exited. So the four
assertions rest on that pclose(): in another file, incidental to why it
is there, and written down nowhere.

Remove it, or abandon the child on a timeout, and an injected touch can
land after the check. The assertions keep passing while the injection
actually works. A false pass in an injection guard is silent, which is
the direction nobody re-reads.

Comment only, no behaviour change. Test still passes.

### DIFF
--- a/tests/web/showgraph-exec-title.sh
+++ b/tests/web/showgraph-exec-title.sh
@@ -68,6 +68,14 @@ cat >>"$work/graphs.cfg" <<EOF
 	LINE2:p@RRDIDX@#@COLOR@:x
 EOF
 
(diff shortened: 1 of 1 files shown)


SoundGoof (approved):
lgtm

PR #336, opened 2026-08-12
tests: fix OpenBSD fixture portability

SoundGoof:
# tests: fix OpenBSD fixture portability

## Summary

- add external `printf` to the restricted filesystem-test `PATH`
- replace GNU-specific `head -c` with portable `dd`

## Problem

Two test fixtures failed on OpenBSD 7.9:

- `fs-filter-linux.sh` could not run `printf` because OpenBSD `/bin/sh`
  does not provide it as a builtin and the fixture's restricted `PATH`
  omitted `/usr/bin/printf`
- `history-tail-resync.sh` used the unsupported GNU `head -c` option

These failures were found while validating #331 but are unrelated to its
`-iquote` changes.

## Validation

| Platform | Passed | Skipped | Failed |
|---|---:|---:|---:|
| Linux | 54 | 0 | 0 |
| FreeBSD 14.3 | 53 | 1 | 0 |
| NetBSD 11.0 | 51 | 3 | 0 |
| OpenBSD 7.9 | 51 | 3 | 0 |

All skips are platform capability skips.

### DIFF
--- a/tests/client/fs-filter-linux.sh
+++ b/tests/client/fs-filter-linux.sh
@@ -314,6 +314,9 @@ for prog in sh bash awk sed; do
 		|| fail "inode-drop test needs a real '$prog' on PATH"
 	ln -s "$p" "$INODEDIR/$prog"
 done
+printf_path=$(PATH="$REALPATH" type -P printf) \
+	|| fail "inode-drop test needs a real 'printf' on PATH"
+ln -s "$printf_path" "$INODEDIR/printf"
 
 inode_out=$(
 	cd "$TMP"
--- a/tests/server/history-tail-resync.sh
+++ b/tests/server/history-tail-resync.sh
@@ -34,7 +34,7 @@ HISTFILE="$WORK/hist/junk,host.disk"
 
 # Valid open record, then a malformed suffix shorter than the scan window.
 printf 'Mon Apr  6 22:37:31 2020 green 1586205451\n' > "$HISTFILE"
-head -c 300 /dev/zero | tr '\0' 'G' >> "$HISTFILE"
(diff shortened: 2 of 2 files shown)


bonomani (approved):
Approved — thanks for doing the four-platform validation, that's the part I can't reproduce here.

The `head -c` → `dd` change is straightforwardly right; `-c` is a GNU extension and POSIX only requires `-n`. Using `type -P` rather than `command -v` for the printf lookup is also the right call, since `command -v` would return the builtin name and leave a broken symlink.

One note for whenever this area is next touched, not a reservation: the `printf` requirement comes from the fixture's own `$INODEDIR/df` stub rather than from the client code — the extracted section under test contains no `printf` — so switching that stub to `echo` (a builtin everywhere, and it drops the `%%` escaping) would fix OpenBSD without adding to the restricted PATH. I tried it against main and the test passes. Minor either way: `printf` is a builtin in dash and bash, so on Linux the client never consults PATH for it, and the allowlist is really there to catch non-builtins like `expr`.

SoundGoof:
Thanks for the suggestion. I tested replacing the fixture's `printf` calls
with `echo`, but the `printf` dependency does not come only from the fixture.

This test extracts the real `[df]` and `[inode]` implementation from
`client/xymonclient-linux.sh` and executes it with the restricted fixture
`PATH`. The extracted production `emit_df()` function contains:

    printf '%s\n' "$DFOUT" | sed ... | awk ...

On OpenBSD, `/bin/sh` does not provide `printf` as a builtin. Removing the
external `printf` symlink therefore causes this production pipeline to fail
with `printf: not found`, leaving the inode report empty.

Eliminating that dependency would require changing production code, which
would broaden this test-portability PR and introduce unnecessary behavioral
risk. I therefore prefer to retain the explicit external `printf` in the
fixture `PATH`.

Issue #310, opened 2026-08-08
Tests: use configured linker flags for libxymoncomm.a harnesses


SoundGoof:
# Tests: use configured linker flags for libxymoncomm.a harnesses

Several regression tests compile C harnesses against `lib/libxymoncomm.a` without using the archive's complete configured link dependencies.

The library declares:

```make
XYMONCOMMLIBS = $(XYMONCOMMLIB) $(ZLIBLIBS) $(SSLLIBS) $(NETLIBS) $(LIBRTDEF)
```

Affected tests:

- `tests/server/notbefore-notafter.sh`
- `tests/server/xmh-item-names.sh`
- `tests/server/xymond-noflap-list.sh`
- `tests/web/html-log-custom-graphs.sh`
- `tests/web/showgraph-exec-title.sh`
- `tests/web/showgraph-single-service.sh`
- `tests/server/xymond-noflap-prefix.sh` from PR #295

This works when selected archive objects do not require the omitted libraries, but it is not portable.

Examples:

- Linux uses `NETLIBS = -ltirpc`.
- Solaris uses `NETLIBS = -lresolv -lsocket -lnsl`.
- HP-UX uses `NETLIBS = -lnsl`.

Reading assignments from the generated Makefile with `sed` is insufficient because variables such as `NETLIBS` come from the included per-OS Makefile.

Add a shared test helper that asks `make` to expand:

```make
$(ZLIBLIBS) $(SSLLIBS) $(NETLIBS) $(LIBRTDEF)
```

Update every harness linking `libxymoncomm.a` to use that helper. Preserve test-specific dependencies such as `PCRELIBS` and `RRDLIBS`.

PR #289's `tests/lib/build-worker.sh` demonstrates the appropriate Make-based expansion pattern.

## Acceptance criteria

- All test harnesses use configured static-library dependencies.
- No test hardcodes `-lssl -lcrypto`.
- RRD and PCRE tests retain their additional configured dependencies.
- Tests work with SSL and non-SSL builds.
- The full regression suite passes.

bonomani:
Evidence that this is broader than the `libxymoncomm.a` dependency list, from a full BSD/macOS pipeline run on `cmake/bootstrap` ([run 31581958308](https://github.com/bonomani/xymon/actions/runs/31581958308)). Every Linux lane is green; 11 BSD lanes fail.

NetBSD 10.1 is the clearest case — 31 passed, **20 failed**, and 14 of those carry one of three signatures:

[...]

PR #340, opened 2026-08-12
build: fix malformed flag values in the FreeBSD and NetBSD makefiles

### LINKED ISSUE #337: Tests: RPATH values carry literal quotes, breaking any harness that passes RPATHOPT
The per-OS makefiles define `RPATH` with **literal double quotes inside the value**:

```make
# build/Makefile.FreeBSD:13
RPATH = "-Wl,--rpath,"
# build/Makefile.NetBSD:20
RPATH = "-Wl,--rpath,"
# build/Makefile.Linux
RPATH = -Wl,--rpath,        <- no quotes
```

`build/Makefile.rules:60` then builds `RPATHOPT := $(RPATH)$(shell echo $(RPATHVAL) | sed -e 's/ / $(RPATH)/g')`.

The product build is unaffected: make passes the recipe to a shell, which strips the quotes as shell syntax.

A test harness that reads the value out of make and word-splits it into `argv` does not get that. The quotes survive as literal characters:

```
clang: error: no such file or directory: '"-Wl,--rpath,"/usr/local/lib'   (FreeBSD)
clang: error: no such file or directory: '"-Wl,--rpath,"/usr/pkg/lib'     (NetBSD)
```

`tests/rrd/inode-unix-columns.sh` is the only harness that passes `RPATHOPT` today, and it fails on every FreeBSD and NetBSD lane for this reason — 4 occurrences on FreeBSD 14.3, 3 on NetBSD 10.1 in [run 31581958308](https://github.com/bonomani/xymon/actions/runs/31581958308). Linux is green because its `RPATH` has no quotes.

This matters beyond that one test: #310 proposes giving every harness the configured link flags. As soon as `RPATHOPT` is among them, this failure spreads from one test to all of them, so it is worth fixing first or alongside.

### Options

1. Drop the quotes from the `RPATH` values and let the rules file quote where needed. Cleanest, but touches every per-OS makefile and needs a check that nothing else depends on the current spelling.
2. Strip them where a harness consumes the value. Localised, but every future consumer has to remember.

Option 1 looks right — the quotes are not carrying meaning, they are an artifact of the value being written as if it were shell.

### Acceptance

- `tests/rrd/inode-unix-columns.sh` passes on FreeBSD and NetBSD.
- The product build's rpath handling is unchanged on all platforms (Linux, BSD, macOS, Solaris).

bonomani:
Two values in the per-OS makefiles are written as if they were shell
fragments rather than flag text. The product build survives both because
make hands its recipes to a shell; anything else that reads the values
does not.

RPATH carried literal double quotes on FreeBSD and NetBSD:

    build/Makefile.FreeBSD:13   RPATH = "-Wl,--rpath,"
    build/Makefile.NetBSD:20    RPATH = "-Wl,--rpath,"
    build/Makefile.Linux:18     RPATH = -Wl,--rpath,        <- and GNU, GNU_kFreeBSD

build/Makefile.rules:61 builds RPATHOPT from it. In a recipe the shell
strips the quotes, so the compiler sees the right thing. A test harness
that reads RPATHOPT out of make and word-splits it into argv does not
get that step, and the quotes arrive as literal characters:

    clang: error: no such file or directory: '"-Wl,--rpath,"/usr/local/lib'
    clang: error: no such file or directory: '"-Wl,--rpath,"/usr/pkg/lib'

tests/rrd/inode-unix-columns.sh is the only harness passing RPATHOPT
today and fails on every FreeBSD and NetBSD lane for this. It matters
more than that one test: #310 proposes giving every harness the
configured link flags, and RPATHOPT among them would spread this from
one failure to all of them.

Reproduced and checked both directions with the same make probe the test
uses:

    RPATH = "-Wl,--rpath,"      harness argv: ["-Wl,--rpath,"/usr/local/lib]
                                product:      [-Wl,--rpath,/usr/local/lib]
    RPATH = -Wl,--rpath,        harness argv: [-Wl,--rpath,/usr/local/lib]
                                product:      [-Wl,--rpath,/usr/local/lib]

So the harness is fixed and what the product build sees is unchanged,
which is the point: the quotes were never carrying meaning, and the
three other makefiles that define RPATH already omit them.

Second, NetBSD's CFLAGS had a stray comma after the library search path,
on all three variants (normal, non-pcre, and the commented debug one):

    -I${PKGDIR}/include -L${PKGDIR}/lib, -Wl,--rpath=${PKGDIR}/lib

That makes the search path "/usr/pkg/lib," -- a directory that does not
exist -- rather than /usr/pkg/lib. NetBSD lanes do fail with "ld: cannot
find -lpcre2-8"; I have not confirmed this comma is the whole cause,
since the product build there still links, so treat that as context
rather than a claim.

Neither file is read on Linux (the toplevel Makefile includes only
build/Makefile.Linux), so the local suite cannot exercise either change:
57 passed, 0 skipped, 0 failed, unchanged. The BSD lanes are the check.

Closes #337

### DIFF
--- a/build/Makefile.FreeBSD
+++ b/build/Makefile.FreeBSD
@@ -10,7 +10,7 @@ CC ?= cc
 CFLAGS = -g -O2 -Wall -Wno-unused -Wno-pointer-sign -D_REENTRANT -I/usr/local/include -L/usr/local/lib $(LFSDEF) $(OSDEF)
 
 
-RPATH = "-Wl,--rpath,"
+RPATH = -Wl,--rpath,
 
 # Compile flags for debugging
 # CFLAGS = -g -DDEBUG -Wall -D_REENTRANT -I/usr/local/include -L/usr/local/lib $(LFSDEF) $(OSDEF)
--- a/build/Makefile.NetBSD
+++ b/build/Makefile.NetBSD
@@ -12,16 +12,16 @@ CC= gcc
 GCCVER := $(shell gcc -dumpversion|cut -d. -f1)
 ifeq ($(GCCVER),4)
    CFLAGS = -g -O2 -Wall -Wno-unused -Wno-pointer-sign -D_REENTRANT $(LFSDEF) $(OSDEF) \
-       -I${PKGDIR}/include -L${PKGDIR}/lib, -Wl,--rpath=${PKGDIR}/lib
+       -I${PKGDIR}/include -L${PKGDIR}/lib -Wl,--rpath=${PKGDIR}/lib
 else
    CFLAGS = -g -O2 -Wall -Wno-unused -D_REENTRANT $(LFSDEF) $(OSDEF) \
-       -I${PKGDIR}/include -L${PKGDIR}/lib, -Wl,--rpath=${PKGDIR}/lib
+       -I${PKGDIR}/include -L${PKGDIR}/lib -Wl,--rpath=${PKGDIR}/lib
 endif
-RPATH = "-Wl,--rpath,"
+RPATH = -Wl,--rpath,
 
 # Compile flags for debugging
 # CFLAGS = -g -DDEBUG -Wall -D_REENTRANT $(LFSDEF) $(OSDEF) \
-       -I${PKGDIR}/include -L${PKGDIR}/lib, -Wl,--rpath=${PKGDIR}/lib
+       -I${PKGDIR}/include -L${PKGDIR}/lib -Wl,--rpath=${PKGDIR}/lib
 
 # Mail program: This must support "CMD -s SUBJECT ADDRESS" to send out a mail with a subject
 # Typically, this will be "mail" or "mailx"


SoundGoof (approved):
## Findings

No findings.

The corrected values reach both test harnesses and linked binaries.

For NetBSD, the generated compiler arguments now contain the library-search
argument `-L/usr/pkg/lib`. 

Before this change, the argument was `-L/usr/pkg/lib,`, with a comma included
in the directory name.

For FreeBSD and NetBSD, the generated `RPATH` and `RPATHOPT` values no longer
contain literal double-quote characters.

## Validation

- Exact merge against current `main` is conflict-free
- All five GitHub CI checks pass
- Linux suite: 57 passed, 0 skipped, 0 failed
- FreeBSD 14.3 build passes
- NetBSD 11.0 build passes
- The affected `inode-unix-columns.sh` test passes on both BSD systems
- FreeBSD `xymond_rrd` RUNPATH contains `/usr/local/lib`
- NetBSD `xymond_rrd` RPATH contains `/usr/pkg/lib`

PR #341, opened 2026-08-12
tests: make hostdata save test portable

SoundGoof:
## Summary

Make `tests/server/hostdata-save-next.sh` portable across supported BSD systems.

- Replace GNU-specific `find -quit` with portable first-result selection using `sed`.
- Replace padded BSD `wc -l` output with an exact numeric count using `awk`.

## Testing

The focused `hostdata-save-next.sh` runtime test passes on:

- Linux
- FreeBSD 14.3
- NetBSD 11.0
- OpenBSD 7.9

### DIFF
--- a/tests/server/hostdata-save-next.sh
+++ b/tests/server/hostdata-save-next.sh
@@ -71,7 +71,7 @@ assert_equal 'OK: next client report for testhost armed for 5 minutes' "$respons
 "$XYMON" "127.0.0.1:$port" $'client testhost.linux\n[test]\nforced payload'
 saved=''
 for attempt in {1..1000}; do
-	saved=$(find "$work/hostdata/testhost" -type f -print -quit 2>/dev/null || true)
+	saved=$(find "$work/hostdata/testhost" -type f -print 2>/dev/null | sed -n '1p' || true)
 	[ -n "$saved" ] && break
 done
 assert_file_exists "$saved" "the armed client report must reach the hostdata worker"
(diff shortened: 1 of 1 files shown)


bonomani (approved):
Both diagnoses check out, and I looked at the two things that usually go wrong with a portability fix like this.

**The sweep is complete.** `find -quit` appears only in this file, and the three other `wc -l` sites are already safe rather than overlooked: `hostdata-save-throttle.sh:43,132` feed `[ "$saved" -eq N ]`, and `-eq` tolerates the leading padding; `fs-filter-linux.sh:168` normalizes explicitly. So nothing is left half-converted — worth recording so nobody later "fixes" the `-eq` sites unnecessarily.

**`assert_equal` really is the trigger.** It is an exact string compare with no trimming (`[ "$want" != "$got" ]`), so BSD's padded `wc -l` output breaks that assertion specifically, while the arithmetic comparisons elsewhere shrug it off.

**`sed -n '1p'` is the right filter, and deliberately so.** It reads to EOF, so `find` is never SIGPIPE'd. Under this script's `set -o pipefail` an early-exiting filter (`head -1`, `sed -n '1q'`) would make the pipeline fail; it is currently masked by the `|| true`, but a later simplification that drops both would reintroduce a failure. Worth keeping as-is.

On `awk 'END { print NR }'` versus normalizing `wc -l` through `$(( ))` — I started out preferring the latter for consistency with the comment at `fs-filter-linux.sh:164-169`, then measured, and changed my mind:

| | empty | normal | unterminated last line | non-numeric input |
|---|---|---|---|---|
| `wc -l` | 0 | 3 | 2 | — |
| `awk 'END{print NR}'` | 0 | 3 | 3 | — |
| `$(( $(wc -l) ))` | 0 | 3 | 2 | 0, silently |

`$(( ))` normalization is a separable second step — a refactor can drop it, and the result then breaks only on BSD, which is this PR's own bug class. It also maps any non-numeric output to `0` with no error even under `set -e`, so a broken producer becomes a plausible count. The awk form has no half to lose. Your spelling is the better one; the awkward idiom is the existing site, not this change.

(The two forms are not interchangeable in general — they disagree when the final line has no trailing newline. Irrelevant for `find` output, but it means "swap `wc -l` for awk" is not a safe blanket refactor elsewhere.)

Nothing I would change. `-print` on the second `find` is redundant since it is the default, but it is harmless and arguably clearer next to the first one.

PR #342, opened 2026-08-12
docs: regenerate xymonserver.cfg.5.html after the instances-per-image rewording

bonomani:
4c1f842d9 (#262) reworded the GRAPHS entry in common/xymonserver.cfg.5 but
did not regenerate the committed HTML, so docs/manpages has been out of
step with its roff source since that merge.

CI catches this -- the "manual pages match their sources" job copies
docs/manpages aside, re-runs build/makehtml.sh and diffs -- and main has
been failing it since 4c1f842d9. This is that regeneration and nothing
else; the only change is the wording #262 already made in the source.

Verified with the job's own procedure: after this commit, regenerating
produces no diff.


SoundGoof (approved):
lgtm

PR #343, opened 2026-08-12
client/linux: bound a wedged remote df instead of losing the whole host (#316)

### LINKED ISSUE #316: client: non-blocking remote-df sentinel for hard-blocking mounts (DF_LOCAL_ONLY=no)
Follow-up to #49/#96, deferred out of #170 (which covered the per-OS port of the include/exclude/local-only filters).

## Problem

When `XYMONCLIENT_FS_DF_LOCAL_ONLY=no`, df stats remote mounts. A dead NFS/CIFS/Ceph server wedges df in an uninterruptible D-state wait: SIGKILL is ignored, so even a foreground `timeout df` blocks, and the whole client cycle hangs until the mount recovers or the host reboots. The host then goes purple — **all** metrics lost, not just disk. This is the documented hang warning in xymonclient.cfg(5); today the only mitigation is not enabling remote filesystems at all.

## Proposed design (prototyped on `feat/nfs-df-sentinel`, Linux client)

Split df collection into two invocations:

- **Local set** (`df -l`) runs synchronously and plainly (no timeout). It can never stat a remote server, so it cannot hard-block, and always stays fresh.
- **Remote set** (types listed in a hard-block type set, discovered from `/proc/mounts`) runs behind a non-blocking sentinel: df runs in the background and the client waits at most N seconds — a poll *budget*, not a kill timeout — before moving on. A wedged df is left running and detected next cycle via a stable pidfile (PID reuse guarded by `/proc/PID/comm`), so at most one df is ever outstanding per section; orphans never accumulate, and it auto-recovers when the server returns.

While unreachable, each remote mount is surfaced as a failed (100%) row rather than silently dropped (which reads as green), so a dead server turns that one filesystem red instead of purpling the whole host. If the probe dir is not writable, the sentinel fails safe by reporting the remote mounts unavailable — never falling back to a synchronous remote df that would reintroduce the hang it exists to prevent.

## Open design question

The prototype introduces two new variables (`XYMONCLIENT_FS_REMOTE_DF_TIMEOUT`, `XYMONCLIENT_FS_REMOTE_HARDBLOCK_TYPES`), where #170 originally sketched a single `XYMONCLIENT_FS_DF_TIMEOUT`. The PR should justify that split (budget vs kill-timeout semantics, and exact set membership rather than regex for types like `fuse.glusterfs`) against the project's current bar for new config surface.

## Plan

Rebase `feat/nfs-df-sentinel` once #312 (fs-filter docs parity) has merged — both rewrite the same `xymonclient.cfg.DIST` / `xymonclient.cfg.5` sections — then open the PR against this issue. The branch already carries tests (healthy path, wedge, PID-reuse guard, recovery, non-writable probe dir fail-safe) to be reconciled with the per-OS `fs-filter-*.sh` split that landed after it was written.

bonomani:
With `XYMONCLIENT_FS_DF_LOCAL_ONLY=no`, df stats remote mounts. A dead NFS/CIFS/Ceph server
leaves it in an uninterruptible D-state wait where SIGKILL is ignored, so a foreground
`timeout df` cannot help either: the client cycle never finishes, the host goes purple, and
every metric is lost -- not just disk -- until the mount recovers or the machine reboots.
Today the only mitigation is to not report remote filesystems at all.

Collection is split in two, on type rather than on locality. The plain df drops the
hard-blocking types with `-x` (`XYMONCLIENT_FS_REMOTE_HARDBLOCK_TYPES`, shipped with a
default so no site has to maintain one); everything else, local or remote, is still
reported, as `LOCAL_ONLY=no` asks. Those types -- found by reading `/proc/mounts`, which
never blocks -- run behind a sentinel: df goes to the background and the client waits at
most `XYMONCLIENT_FS_REMOTE_DF_BUDGET` seconds (default 30). That is a poll budget, not a
kill timeout: a wedged df cannot be killed, so it is left running and recognised next cycle
through a pidfile, and no second one is started. At most one df is ever outstanding per
report, and collection resumes by itself when the server answers. Meanwhile each mount of
the remote set is reported as a 100% row rather than dropped, because the server reads an
absent filesystem as green -- one filesystem red instead of the whole host purple.

Details that are easy to get wrong:

- **`-x` on the plain df, never `-l`.** Both filter the mount list before df `stat()`s
  anything; `-l` would also delete healthy remote filesystems from monitoring.
- **The exclusions stay in a subshell around that df.** `-x` applies to an explicitly named
  operand too: handed both, the guarded probe processes nothing, exits nonzero, and every
  healthy mount comes out unavailable and 100% full.
- **The pidfile is claimed with an O_EXCL create under `set -C`.** Test-then-create let two
  cycles both start a df with only the last pid recorded, so untracked probes -- the
  unkillable kind -- piled up every cycle. Winning also proves the pidfile writable before
  any df starts.
- **Writability is probed with `( : > file )`, never a bare `:`.** A redirection error on a
  POSIX special builtin is fatal to the shell rather than a non-zero status, so under dash
  -- `/bin/sh` on Debian and Ubuntu -- the guard's `return 124` never ran and run_df emitted
  nothing instead of the unavailable rows. Measured: dash 2, bash 124.
- **The remote df's header is stripped.** `unix_disk_report()` recognises one only on the
  first line (xymond_client.c); a second becomes a phantom "Mounted" filesystem that counts
  toward DISKCOUNT and reaches the RRD.
- **The probe dir must be local, decided from `/proc/mounts` without `stat()`ing it.** On
  the wedged mount even `test -w` blocks in D-state, inside the fail-safe meant to prevent
  that. Symlinks are not resolved for the same reason, and that is documented.
- **A probe that finished after the budget is discarded.** Its exit status went unseen, so a
  truncated file is indistinguishable from a complete one, and a partial df silently drops
  mounts -- which reads as green, the failure this exists to prevent.
- **PID reuse is checked against `/proc/PID/comm`**, and mount points are `\040`-decoded and
  iterated with `IFS=newline`, so a recycled pid does not strand a mount and a mount point
  containing a space stays one argument to df.

Known cost, deliberately accepted: the 100% row also reaches the disk RRD, so the graph
shows 100% used for the duration of the outage.

Tests. fs-sentinel-linux.sh drives a df stub and a `/proc/mounts` fixture: a healthy remote
type outside the hard-block list, the wedge (local set stays fresh, one outstanding df,
recovery), a type excluded by policy, an unrecordable pidfile, a probe dir on the wedged
mount, and the concurrent claim twice -- twenty parallel cycles, which land in the window
often but not always, and a deterministic setup of the state that window produces; only the
second reliably fails without the fix. The stub exits nonzero having processed nothing when
an exclusion leaves no operand, which is exactly what hid the exclusion defect.
fs-filter-linux.sh trades its ` -l ` assertion, which encoded the old mechanism, for the
property that matters: no hard-blocking type reaches the unguarded df.

Suite: 63 passed, 0 skipped, 0 failed. Mutations caught: reverting to `df -l`, dropping the
pidfile claim, handing the exclusions to the guarded probe. Not caught, and left as such:
restoring the fatal `:` idiom, because no case makes the probe file itself unwritable.

Closes #316

### DIFF
--- a/client/xymonclient-linux.sh
+++ b/client/xymonclient-linux.sh
@@ -113,11 +113,137 @@ case "$DFLOCALONLY" in
 		DFLOCALONLY=yes
 		;;
 esac
+# XYMONCLIENT_FS_REMOTE_DF_BUDGET: seconds to wait for the remote df probe
+# before giving up for this cycle (default 30, capped at 3600; non-numeric,
+# empty or zero -> 30). A poll BUDGET, not a kill timeout: a df wedged on a
+# dead server sits in uninterruptible D-state where even SIGKILL is ignored,
+# so it is left running and picked up on a later cycle rather than killed.
+DFBUDGET="${XYMONCLIENT_FS_REMOTE_DF_BUDGET:-30}"
+case "$DFBUDGET" in
+	*[!0-9]*|'') DFBUDGET=30 ;;
+	*[1-9]*) ;;
+	*) DFBUDGET=30 ;;
+esac
+[ "$DFBUDGET" -le 3600 ] 2>/dev/null || DFBUDGET=3600
+
+# XYMONCLIENT_FS_REMOTE_HARDBLOCK_TYPES: the df types whose stat() hard-blocks
+# when the server is unreachable. Shipped as a built-in list so no site has to
+# maintain one; assign the variable to override it. Only consulted with
+# DF_LOCAL_ONLY=no, since df -l keeps these mounts out entirely.
+: "${XYMONCLIENT_FS_REMOTE_HARDBLOCK_TYPES=nfs nfs4 cifs smb3 ceph glusterfs fuse.glusterfs lustre afs}"
+DFPROBEDIR="${XYMONTMP:-/tmp}"
+
+# probe_dir_is_local DIR : true when DIR is not itself on a hard-blocking
+# filesystem. Decided purely from /proc/mounts (reading it never blocks) by
+# longest mount-point prefix, and deliberately WITHOUT stat()ing DIR: if DIR
+# were on the wedged mount, even `test -w DIR` would block in D-state inside
+# the very fail-safe meant to prevent that. Symlinks are not resolved for the
+# same reason (readlink would stat), so keep XYMONTMP on a real local path.
+probe_dir_is_local()
+{
+	[ -r /proc/mounts ] || return 1
+	awk -v dir="$1" -v types="$XYMONCLIENT_FS_REMOTE_HARDBLOCK_TYPES" '
+		BEGIN { n = split(types, a, /[ \t]+/); for (i = 1; i <= n; i++) t[a[i]] = 1 }
+		{
+			mp = $2; gsub(/\\040/, " ", mp)
+			if (dir == mp || mp == "/" || substr(dir, 1, length(mp) + 1) == mp "/") {
+				if (length(mp) >= length(best)) { best = mp; besttype = $3 }
+			}
+		}
+		END { exit (besttype in t) ? 1 : 0 }
+	' /proc/mounts
+}
+
+# df_sentinel TAG ARGS... : df for the remote set with at most ONE outstanding
+# probe per TAG. Prints rows (header stripped); returns 0 with data, 124 when
+# unavailable. A wedged df cannot be killed, so it is left running and detected
+# on the next cycle through a pidfile -- orphans never accumulate, and it
+# recovers by itself once the server returns.
+df_sentinel()
+{
+	_tag="$1"; shift
+	_probe="$DFPROBEDIR/df-probe-$_tag"; _pidf="$_probe.pid"
+
+	probe_dir_is_local "$DFPROBEDIR" || return 124
+
+	# Claim the probe before starting it, atomically. Testing for the pidfile
+	# and creating it afterwards let two client cycles both find it absent and
+	# both start a df; only the last pid written was ever recorded, so the
+	# other probe was untracked from then on and more piled up every cycle -
+	# and these are the processes that cannot be killed. Under "set -C" the
+	# create is O_EXCL, so exactly one cycle wins, and winning doubles as the
+	# proof that the pidfile is writable before any df starts.
+	#
+	# A subshell, not a bare ":": ":" is a POSIX special builtin, and a
+	# redirection error on one is fatal to the shell rather than a non-zero
+	# status. Under dash - /bin/sh on Debian and Ubuntu - "return 124" would
+	# never run, the sentinel would die with status 2, and run_df would emit
+	# nothing for the remote set instead of the unavailable rows. It also keeps
+	# noclobber out of the caller.
+	#
+	# The claim names the claiming shell and is replaced by df's pid below, so
+	# a cycle killed in the gap leaves a pid that is simply dead and is
+	# collected like any finished probe, rather than a marker nothing clears.
+	if ! ( set -C; echo "claim:$$" > "$_pidf" ) 2>/dev/null; then
+		_old=$(cat "$_pidf" 2>/dev/null)
+		case "$_old" in
+			claim:*)
+				# Another cycle holds the claim and has not started its df yet.
+				_c=${_old#claim:}
+				[ -n "$_c" ] && kill -0 "$_c" 2>/dev/null && return 124
+				;;
+			?*)
+				# A df we recorded. Still wedged? comm is checked so a recycled
+				# PID does not look like one.
+				if kill -0 "$_old" 2>/dev/null && \
+				   [ "$(cat "/proc/$_old/comm" 2>/dev/null)" = df ]; then
+					return 124
+				fi
+				;;
+		esac
+		# It finished after we stopped waiting. Its output is not usable: we
+		# were not there to see the exit status, so a truncated file is
+		# indistinguishable from a complete one, and its age is unknown.
+		# Discard and re-probe -- fresh data next cycle beats stale or partial
+		# data now. One retry only: if another cycle claims it first, that
+		# cycle owns the probe and this one waits its turn.
+		rm -f "$_pidf" "$_probe"
+		( set -C; echo "claim:$$" > "$_pidf" ) 2>/dev/null || return 124
+	fi
+
+	# Never run a synchronous remote df here: a foreground df would reintroduce
+	# the hang this exists to prevent.
+	( : > "$_probe" ) 2>/dev/null || { rm -f "$_pidf"; return 124; }
+	df "$@" > "$_probe" 2>/dev/null &
+	_pid=$!
+	if ! echo "$_pid" > "$_pidf"; then
+		# The pre-flight passed and this still failed, so a df is running that
+		# the next cycle cannot recognise. Say so: an operator seeing repeated
+		# unavailable rows needs to know a df may be outstanding.
+		echo "xymonclient: cannot record remote df pid in $_pidf" >&2
+	fi
+	_n=0
+	while kill -0 "$_pid" 2>/dev/null; do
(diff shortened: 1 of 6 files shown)


SoundGoof:
## Review

###  `DF_LOCAL_ONLY=no` drops remote filesystem types outside the hard-block list

`run_df()` collects local filesystems with `df -l`, then adds only mount types
listed in `XYMONCLIENT_FS_REMOTE_HARDBLOCK_TYPES`.

This changes the existing and documented meaning of
`XYMONCLIENT_FS_DF_LOCAL_ONLY=no`, which reports remote filesystems as well.
Healthy remote types outside that list now disappear entirely.

This also makes an override destructive. I reproduced it with the PR's healthy
NFS fixture:

```sh
XYMONCLIENT_FS_REMOTE_HARDBLOCK_TYPES=cifs \
    ./tests/client/fs-sentinel-linux.sh
```

The NFS row disappears and the test fails:

```text
FAIL: the remote set must be reported
```

Keep reporting all remote filesystems when local-only is disabled. The
hard-block list should determine which mounts need the sentinel, not which
remote mounts remain visible.

### PID-file write failure leaves a wedged `df` untracked

`df_sentinel()` starts the background process before writing its PID:

```sh
df "$@" > "$_probe" 2>/dev/null &
_pid=$!
echo "$_pid" > "$_pidf"
```

The `echo` result is unchecked. If the probe output can be truncated but the
PID file cannot be created or written, the remote `df` is already running. If
it wedges, the next cycle has no sentinel PID and starts another one.

This defeats the at-most-one-process guarantee and can accumulate unkillable
D-state processes.

Verify PID-file usability before starting `df`, and handle the final PID write
failure explicitly. Add a regression case where the probe output is writable
but PID-file creation fails, asserting that no remote `df` starts.

bonomani:
Both fixed, and fixing the second turned up a third.

**Remote filesystems dropped.** Reproduced your case exactly. The first `df` now excludes the hard-blocking *types* with `-x` instead of keeping only locals with `-l`: `-x` filters the mount list before `df` stat()s anything, same as `-l`, so the call still cannot hard-block, but everything else — local or remote — is reported. Your framing was the right one; the list decides what needs the sentinel, not what is visible.

**PID file.** Pre-flighted before `df` starts, and the post-fork write is checked too.

**Third, found while doing that:** `:` is a POSIX special builtin, and a redirection error on one is fatal to the shell rather than a non-zero status. So the *existing* `: 2>/dev/null > "$_probe" || return 124` guard never returned 124 under dash — `/bin/sh` on Debian and Ubuntu:

```
dash : status=2      <- "|| return 124" never runs
bash : status=124
```

The sentinel died with status 2 and `run_df` emitted nothing for the remote set instead of the `unavailable:` rows. Both guards are now `( : > file )`.

**Tests.** Added the case you asked for (probe writable, pidfile not — asserts no `df` starts and the set is reported unavailable), plus a remote type outside the hard-block list that must still be reported. Both mutation-checked.

Two existing assertions changed, worth flagging since I changed tests to match my own fix: the `df` stub emitted the local row only for `-l`, which modelled the old design rather than `df`, and now models `-l`, `-x` and named mounts; and `fs-filter-linux.sh` asserted `" -l "` literally, and now asserts the property — no hard-blocking type reaches the unguarded `df`.

Not fixed: restoring the fatal `:` idiom on the probe guard is caught by nothing, because no case makes the probe itself unwritable.

SoundGoof:
## Review

### High: The guarded probe excludes every filesystem it must collect

`run_df()` adds every hard-blocking type to the argument list with `-x`,
then passes that same list to `df_sentinel()` alongside explicit mount
operands of those types.

GNU `df` applies exclusions to explicit operands. On Debian 13 with a real
ext4 filesystem:

```text
$ df -P /mnt/pr343-ext4
/dev/vdb ... /mnt/pr343-ext4

$ df -P -x ext4 /mnt/pr343-ext4
df: no file systems processed
```

The second command exits 1. The same happens to NFS, CIFS, Ceph, and other
mounts selected for guarded collection, so healthy mounts are rendered as
`unavailable:` and 100% used.

The test stub ignores `_ex` when named mounts are supplied. After changing
only that branch to honor `-x nfs4`, the healthy-path test failed:

```text
FAIL: the remote set must be reported
```

Keep the base arguments before adding the hard-blocking exclusions and use
those for the guarded probe. User-configured exclusions must still apply,
and excluded mount types should not be added to the remote operand list.

[Implementation](https://github.com/xymon-monitoring/xymon/blob/25a496c5073797c3b71b557e2409645d2b9db15d/client/xymonclient-linux.sh#L237-L276)

[Test stub](https://github.com/xymon-monitoring/xymon/blob/25a496c5073797c3b71b557e2409645d2b9db15d/tests/client/fs-sentinel-linux.sh#L54-L67)

### High: Concurrent client cycles can leave untracked probes

The pidfile protocol performs a non-atomic sequence:

1. Check the existing pidfile.
2. Truncate the probe and pidfile.
3. Start `df`.
4. Write the new PID.

Two client cycles can both pass the initial check and start remote probes.
The final pidfile write records only one process. If both become stuck in
uninterruptible sleep, the other process is permanently untracked and later
cycles can accumulate more.

I ran 40 concurrent cycles against the exact PR code on Debian 13:

```text
remote_calls=2
remote_pids=<7452 7457>
recorded_pid=7457
```

Use an atomic ownership operation around probe startup and add a concurrent
regression test. The current test covers only sequential cycles.

[Sentinel startup](https://github.com/xymon-monitoring/xymon/blob/25a496c5073797c3b71b557e2409645d2b9db15d/client/xymonclient-linux.sh#L169-L201)

[Sequential test](https://github.com/xymon-monitoring/xymon/blob/25a496c5073797c3b71b557e2409645d2b9db15d/tests/client/fs-sentinel-linux.sh#L113-L120)

### Validation

Exact head `25a496c5073797c3b71b557e2409645d2b9db15d` was configured
and built on Debian 13 with GNU coreutils 9.7.

The complete suite passed 58 tests with no skips or failures, demonstrating
that the existing tests do not detect either reproduced defect.

bonomani:
Both confirmed and fixed — thanks.

**Correction to this comment as first posted:** I had described the root cause as the mount type not being checked against the exclusion list. That was wrong — I had read a stale checkout that predated `25a496c50`, the head you reviewed. The actual cause is the one you gave: the hard-blocking `-x` flags are appended to `$@` for the plain df, and the remote mount points are then appended to that same `$@` before it reaches the sentinel. They now live in a subshell around the plain df, which gets its own copy of the positional parameters.

The exclusion check turned out to be needed as well, for the same reason one level down: a type the admin excluded has its `-x` in the *base* arguments, so naming its mount beside it is the same empty-and-nonzero df. Such a type is now simply not collected, guarded or not.

**Concurrent probes.** The claim is an `O_EXCL` create under `set -C`, before the probe starts — which also proves the pidfile writable, so it replaces the separate pre-flight. It names the claiming shell and is overwritten by df’s pid, so a cycle killed in the gap leaves a dead pid the existing staleness check collects, rather than a marker nothing could clear.

You were right that the stub could not see the first defect: it reported named operands without applying `-x` to them. Fixed, along with modelling that an exclusion leaving no operand makes df exit nonzero.

On the race, worth stating plainly: **twenty concurrent cycles catch it only sometimes** — I had one run detect it and one run miss it. So it is covered twice, and the reliable half is a deterministic case that constructs the state the window produces (a pidfile naming a live process that is not a df) rather than racing for it.

| Mutation | Caught by |
|---|---|
| hard-block `-x` shared with the probe | `the remote set must be reported` |
| exclusion check removed | `an excluded type was probed anyway and its healthy mount reported unavailable` |
| test-then-create claim | `a probe already claimed by a live cycle was started a second time` (deterministic) |

Full suite: 58 passed, 0 failed, 0 skipped.

SoundGoof (approved):
lgtm

PR #346, opened 2026-08-12
tests: give harnesses the configured link flags, not just libraries (#310)

### LINKED ISSUE #310: Tests: use configured linker flags for libxymoncomm.a harnesses

bonomani:
Completes #310. xymon_cflags() covers the compile side; nothing covered
the link side, and that is what the BSD lanes fail on.

Two failure modes, one cause. FreeBSD keeps the library search path
inside CFLAGS (build/Makefile.FreeBSD), which no harness passes, so the
link fails with "ld: cannot find -lpcre2-8" five times per lane. NetBSD
links but will not run: without RPATHOPT the loader cannot find
libpcre2-8.so.0, which is why six hostdata tests fail there and not on
FreeBSD.

xymon_ldflags() emits what the product link line gets: the search path
picked out of CFLAGS, RPATHOPT, and the libraries libxymoncomm.a needs
(XYMONCOMMLIBS). Asked of make rather than parsed out of the Makefile,
since three of those live in the per-OS include and the search path has
no variable of its own. An empty RPATHVAL leaves a bare "-Wl,--rpath,",
which is dropped.

Fourteen harnesses take it, replacing hand-rolled tails; build-worker.sh
drops its own make probe for the shared helper. Test-specific libraries
stay at the call sites, as #310 asked.

The svcstatus cache keyed only on the build arguments, so changing
SSLLIBS or the search path reused a stale binary and the test passed
without exercising the configuration it claims to. Pre-existing -- main
behaves the same -- but this PR widens it, so the key now hashes the
effective flags too.

Verified: 22 of 22 links against libxymoncomm.a now carry the configured
set, against 0 before; a broken SSLLIBS fails the test instead of
passing it, and the cache still reuses a good binary. Suite: 62 passed,
0 skipped, 0 failed.

Linux reproduces neither BSD failure, so those lanes are the check.

Closes #310

### DIFF
--- a/tests/lib/assert.sh
+++ b/tests/lib/assert.sh
@@ -287,6 +287,50 @@ xymon_cflags() {
 	printf '%s' "-iquote $root/include -iquote $root/lib $(pcre_cflags "$root")"
 }
 
+# xymon_ldflags ROOT -- the configured link flags for a harness built against
+# the in-tree libraries: the library search path, the runtime path, and the
+# libraries libxymoncomm.a itself needs (XYMONCOMMLIBS in xymond/Makefile).
+#
+# The counterpart to xymon_cflags(), and bundled for the same reason: the
+# product build gets all of this from $(CFLAGS) and $(RPATHOPT) on its link
+# line, while each harness spells its own and so far none carried any of it.
+# On the BSDs that costs a link (`ld: cannot find -lpcre2-8`, because
+# build/Makefile.FreeBSD keeps -L/usr/local/lib inside CFLAGS) or a run
+# (`Shared object "libpcre2-8.so.0" not found`, no rpath).
+#
+# Asked of make rather than sed'd out of the Makefile: NETLIBS, ZLIBLIBS and
+# LIBRTDEF live in the per-OS file the toplevel Makefile includes, and the
+# search path has no variable of its own at all - it is embedded in CFLAGS,
+# so the -L tokens are picked out of it.
+xymon_ldflags() {
+	local root=$1 probe out cflags rpath libs tok search=
+
+	[ -f "$root/Makefile" ] || return 0
+	require_gnu_make
+	# shellcheck disable=SC2016  # $(CFLAGS) etc. are make's to expand, not the shell's
+	probe='__xymon_ldflags:
+	@printf "%s\n" "$(CFLAGS)" "$(RPATHOPT)" "$(ZLIBLIBS) $(SSLLIBS) $(NETLIBS) $(LIBRTDEF)"
+'
+	out=$(printf '%s' "$probe" | "$XYMON_MAKE" -s -C "$root" -f Makefile -f - __xymon_ldflags 2>/dev/null) || return 0
+
+	cflags=$(printf '%s\n' "$out" | sed -n 1p)
+	rpath=$(printf '%s\n' "$out" | sed -n 2p)
+	libs=$(printf '%s\n' "$out" | sed -n 3p)
+
+	# Only the search path is wanted from CFLAGS; its -I/-D/-W belong to the
+	# compile side, which xymon_cflags() already covers.
+	for tok in $cflags; do
+		case $tok in -L*) search="$search $tok" ;; esac
+	done
+
+	# RPATHOPT is $(RPATH)$(RPATHVAL); with no RPATHVAL configured it is a bare
+	# "-Wl,--rpath," with nothing after the comma, so drop it rather than hand
+	# the linker a dangling flag.
+	case $rpath in *,) rpath= ;; esac
+
+	printf '%s' "$search $rpath $libs"
+}
+
 # require_shm_segments N -- skip unless one process may attach N SysV
 # shared-memory segments. xymond attaches one per channel, and macOS ships
 # kern.sysv.shmseg=8 against the nine channels xymond sets up, so a stock Mac
--- a/tests/lib/build-worker.sh
+++ b/tests/lib/build-worker.sh
@@ -21,7 +21,7 @@
 __XYMON_TESTS_BUILD_WORKER_SOURCED=1
 
 build_xymond_worker() {
-	local outdir=$1 prog=$2 root cc treelibs pcre_libs harness_cflags src srcs mkfiles inc
+	local outdir=$1 prog=$2 root cc pcre_libs harness_ldflags harness_cflags src srcs
 	shift 2
 	root=$(find_root)
 	cc=${CC:-cc}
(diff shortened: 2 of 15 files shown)


SoundGoof:
## Review

- **Medium:** Include effective linker flags in the svcstatus cache key.
  [tests/lib/svcstatus-cgi.sh#L104-L121](https://github.com/xymon-monitoring/xymon/blob/fd93353e68cc499c033a221355bb0e80e0aa89c1/tests/lib/svcstatus-cgi.sh#L104-L121)

  The cache key uses only the explicit compiler arguments, but the link now
  also depends on `pcre_libs` and `xymon_ldflags()`. Changes to configured
  search paths, RPATHs, or libraries can therefore reuse a stale binary and
  make a test pass without exercising the current configuration.

  I reproduced this by populating the cache and then adding a nonexistent
  library to `SSLLIBS`. The cached call returned success and retained the
  same binary timestamp. After removing the cached binary, the identical
  build failed as expected:

      /usr/bin/ld: cannot find -l__pr346_missing_ssl__

  Add the effective compile and link flags to `flagkey`, or include the
  configuration inputs such as `Makefile` in the freshness check.

### Validation

- Linux: 57 passed, 0 skipped, 0 failed
- FreeBSD 14.3: 57 passed, 0 skipped, 0 failed
- NetBSD 11.0: 55 passed, 2 expected ASAN skips, 0 failed
- OpenBSD 7.9: 55 passed, 2 expected ASAN skips, 0 failed

bonomani:
Thanks — and thank you especially for the BSD runs. I could not test any of this locally, so those numbers are the first real confirmation that the link and rpath halves were the whole story on FreeBSD and NetBSD.

Fixed in `9de6ce5fc`.

I reproduced your finding first, and it behaves exactly as you describe: populate the cache, add a nonexistent library to `SSLLIBS`, and the test still passes with the cached binary's mtime unchanged; remove the cached binary and the same build fails.

One thing worth recording: it is **pre-existing, not introduced by this PR**. `main` without these commits behaves identically, because the link already depended on `SSLLIBS`/`NETLIBS`/`LIBRTDEF` and none of those were in the key either. What this PR changes is the *exposure* — the link now also depends on the search path and the rpath, so there is more configuration that can drift behind the cache. That is why I have fixed it here rather than splitting it out.

The key now hashes the effective compile and link flags along with the arguments:

```sh
harness_cflags=$(xymon_cflags "$ROOT")
harness_ldflags=$(xymon_ldflags "$ROOT")
flagkey=$(printf '%s' "$* $harness_cflags $harness_ldflags $pcre_libs" | cksum | tr -cd '0-9')
```

Hashed rather than spelled out because the flags carry absolute paths, which do not belong in a filename.

I went with the key rather than adding `Makefile` to the freshness check, because the freshness check only catches an mtime change on a file. An environment override — `PCRELIBS=... ./tests/testsuite`, or a different `$CC` — changes the binary without touching any file, and the key covers that too.

Verified in three directions, since a cache fix can easily turn into "no cache at all":

- the broken `SSLLIBS` now **fails** the test instead of passing it
- restoring the configuration passes again
- a second test still **reuses** the cached binary rather than rebuilding, so the cache is still doing its job

Suite 57 passed, 0 skipped, 0 failed. No new shellcheck findings.

SoundGoof:
## Finding

### Medium: Changing `CC` still reuses a binary built by another compiler

[tests/lib/svcstatus-cgi.sh:112](https://github.com/xymon-monitoring/xymon/blob/9de6ce5fcb42673689e735733672ca1f0241c43f/tests/lib/svcstatus-cgi.sh#L112)

The repaired cache key includes the build arguments and configured compile,
link, and PCRE flags, but not the selected compiler.

I populated the cache with the normal compiler and reran the test with `CC`
pointing to a wrapper that always fails. The test passed without invoking that
wrapper and retained the cached binary's timestamp. After clearing the cache,
the same command invoked the wrapper and failed.

This can make a GCC-built binary satisfy a run intended to test Clang, or vice
versa. Include the effective `CC` executable in the cache key.

## Validation

All five GitHub checks pass.

The previous configured-flags finding is otherwise resolved: changes to SSL
libraries, search paths, rpath, compile flags, and PCRE flags now produce
different cache keys.

Tested platforms:

- Debian 13: 57 passed, 0 skipped, 0 failed.
- FreeBSD 14.3: 56 passed, 1 unrelated skip, 0 failed.
- NetBSD 11.0: 55 passed, 2 sanitizer-related skips, 0 failed.
- OpenBSD 7.9: 54 passed, 3 unrelated skips, 0 failed.

bonomani:
Fixed — the effective `CC` is in the key now.

Reproduced your experiment before and after, with `CC` pointed at a wrapper that always fails:

| Key | Result |
|---|---|
| without the compiler | `OK svcstatus-client-traversal.sh` — wrapper never invoked |
| with it | `wrapper refuses to compile` → `FAIL: svcstatus does not build` |

The resolved path goes in alongside the name, since `cc` is a different compiler on different hosts and PATH can put another one under the same name. It also joins the freshness set, so a compiler replaced in place invalidates the cache the same way a changed source does — otherwise a same-path upgrade would still be missed, which the name alone cannot catch.

On the conflict this PR was showing: it resolves itself. `upstream/main` has moved on today (#347, #225, #255) and this branch already contains it — GitHub now reports it mergeable, so there was nothing to rebase.

Full suite: 62 passed, 0 failed, 0 skipped.

bonomani:
Re-requesting review: the cache-key finding is fixed in `26ed000a4`, pushed after your pass, so this needs a fresh approval on that commit either way.

Worth taking before the others in the batch, if you have a choice of order: `xymon_ldflags()` here is the fix for the class you just reported on #204 (the harness linking without `RPATHOPT` and failing to load libpcre2 on NetBSD). The same gap is in the two harnesses I added today for #200 and #250 — once this lands, each of the three is a one-line change instead of three separate scrapes of the Makefile.

SoundGoof (approved):
No findings.

The centralized linker flags preserve library ordering and correctly propagate
configured search paths, runtime paths, and dependencies.

Validation:
- FreeBSD 14.3: 62 passed, 0 skipped, 0 failed.
- NetBSD 9.4: 62 passed, 0 skipped, 0 failed.
- NetBSD 11.0: 60 passed, 2 ASAN skips, 0 failed.
- NetBSD 11.0 with a PaX-aware ASAN compiler wrapper: both skipped tests pass.
- OpenBSD 7.9: 60 passed, 2 unsupported-ASAN skips, 0 failed.
- Linux: 62 passed, 0 skipped, 0 failed.
- All GitHub checks pass.
- Synthetic nonstandard `-L` and RPATH configuration produced the expected
  flags.
- `git diff --check` passed.

The BSD runs directly exercise the library search and runtime-path behavior
this change is intended to fix.

Issue #345, opened 2026-08-12
tests: xymond cannot attach its 9 channels on macOS (kern.sysv.shmseg=8)


bonomani:
`tests/server/hostdata-save-next.sh` (added by #305) fails on every macOS lane — 14, 15 and 26 — in [run 31625836308](https://github.com/bonomani/xymon/actions/runs/31625836308):

```
Loading hostnames
Setting up network listener on 127.0.0.1:37097
Setting up signal handlers
Setting up xymond channels
Could not attach shm Too many open files
Cannot setup user channel
FAIL: an invalid lifetime must be rejected immediately: '' does not contain 'ERROR: hostdatasave requires HOSTNAME'
```

macOS is otherwise green: 56 passed, 0 skipped, 1 failed.

## Cause

Despite the message, this is not a file-descriptor shortage. `xymond` attaches one SysV shared-memory segment per channel, and the failing one is `C_USER` — the **ninth**:

```c
/* lib/xymond_buffer.h:14 */
enum msgchannels_t { C_STATUS=1, C_STACHG, C_PAGE, C_DATA, C_NOTES,
                     C_ENADIS, C_CLIENT, C_CLICHG, C_USER,
                     C_FEEDBACK_QUEUE, C_LAST };
```

macOS ships `kern.sysv.shmseg=8` (segments attachable per process) by default, against 4096 on stock Linux. The first eight attach, the ninth fails, and `shmat` returns `EMFILE` — which `strerror()` renders as "Too many open files".

So a stock macOS cannot run `xymond` at all with the current channel count, and this is the first test that starts a real `xymond` rather than a stub — which is why it has not surfaced before.

## Scope

- Only `tests/server/hostdata-save-next.sh` starts a real `xymond` today, so it is the only test affected.
- Any future test that does the same will hit it.
- It says nothing about `xymond` in production on macOS, where an administrator can raise the limit — but it does mean the default install does not work there without tuning, which may itself be worth documenting.

## Options

1. **Raise the limit in the macOS lane** — `sysctl -w kern.sysv.shmseg=32` before the suite. Needs privileges on the runner, and only fixes CI, not a developer running the suite locally on a Mac.
2. **Skip with an explicit reason** when the limit is below the channel count, e.g. read `kern.sysv.shmseg` and skip if it is under `C_LAST`. Cheap and portable, but a silent-ish skip is how the `mapfile` breakage (#338) survived unnoticed for so long, so the skip message must name the sysctl and the value needed.
[...]

PR #347, opened 2026-08-12
tests: name the sysctl when xymond cannot attach its channels (#345)

bonomani:
xymond attaches one SysV shared-memory segment per channel and sets up
nine (C_STATUS through C_USER, xymond.c). macOS ships
kern.sysv.shmseg=8, so on a stock Mac the ninth attach fails and xymond
never starts:

    Setting up xymond channels
    Could not attach shm Too many open files
    Cannot setup user channel

tests/server/hostdata-save-next.sh is the only test that starts a real
xymond rather than a stub, so it is the only one affected, and it fails
on every macOS lane.

The message is what makes this worth a named check. shmat() reports the
limit as EMFILE, which strerror() renders "Too many open files" -- so the
error points at file descriptors and never mentions the sysctl actually
responsible. require_shm_segments() skips instead, naming the knob, the
current value, the value needed and the command to raise it.

The required count is read out of xymond.c rather than hardcoded, so
adding a channel cannot silently outgrow the check.

Linux has no comparable per-process cap and exposes no such sysctl, so
an absent knob is treated as "nothing to check" and the guard is inert
there: suite 57 passed, 0 skipped, 0 failed, unchanged. Verified in both
directions with a sysctl stub reporting 8, which produces

    SKIP: kern.sysv.shmseg is 8, need >= 9 (xymond attaches one shared-
    memory segment per channel) -- raise it with: sudo sysctl -w
    kern.sysv.shmseg=9

This is the developer-facing half of #345. The other half -- raising the
limit in the macOS lane so CI executes the test instead of skipping it --
belongs with the lane definition, which lives on cmake/bootstrap and not
on main.

Refs #345

### DIFF
--- a/tests/lib/assert.sh
+++ b/tests/lib/assert.sh
@@ -287,6 +287,33 @@ xymon_cflags() {
 	printf '%s' "-iquote $root/include -iquote $root/lib $(pcre_cflags "$root")"
 }
 
+# require_shm_segments N -- skip unless one process may attach N SysV
+# shared-memory segments. xymond attaches one per channel, and macOS ships
+# kern.sysv.shmseg=8 against the nine channels xymond sets up, so a stock Mac
+# cannot start xymond at all.
+#
+# Worth a named check rather than letting the run fail: shmat() reports the
+# limit as EMFILE, which strerror() renders "Too many open files", so the
+# error points at file descriptors and says nothing about the sysctl actually
+# responsible. The skip names the knob and the value so it is actionable.
+#
+# Linux has no comparable per-process cap (SHMSEG is effectively unlimited)
+# and exposes no such sysctl, so an absent knob means "no limit to check".
+require_shm_segments() {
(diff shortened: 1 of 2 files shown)


SoundGoof (approved):
## Findings

No findings.

## Validation

All five GitHub checks pass.

Xymond requires nine shared-memory segments.

Tested platforms:

- Debian 13: affected test passed; Linux has no comparable per-process segment
  limit. The full suite passed 54 tests with 3 unrelated skips.
- FreeBSD 14.3: affected test passed `kern.ipc.shmseg` is 128.
- NetBSD 11.0: affected test passed `kern.ipc.shmseg` is 128.
- OpenBSD 7.9: affected test passed `kern.shminfo.shmseg` is 128.

Stubbed sysctl tests confirmed that a limit of eight skips the test and reports
that at least nine segments are required.

macOS was not directly available; its default eight-segment behavior was
validated using a sysctl stub.

PR #348, opened 2026-08-12
xymonlaunch: refuse a tasklist config that was read short

bonomani:
`load_config()` clears every task's settings before re-reading the file, so the comparison afterwards can tell what changed — and nothing checked that it had read a whole file first. A config being rewritten in place is readable in that state: the re-read is driven by the file's timestamp, not by whoever is writing.

Acting on a short read costs two things:

- **a section read without its `CMD`** leaves a task with no command, which is fatal — the forked child hands the NULL to `expand_env()` and dies before `execvp()`, so it never reaches the point where a failure would be reported;
- **every section the read did not reach looks deleted**, so the tasks below the cut — `xymond` among them — are killed as if that had been asked for, with nothing in the log to say why.

A task with no command is the mark of a short read: no valid config drops a `CMD`, since a task is switched off with `DISABLED`. So when one turns up, put everything back and wait — the next pass re-reads 30 s later, unconditionally, rather than trusting a timestamp taken mid-write.

| mutation | result |
|---|---|
| no check at all | no report, task left with no command |
| restore only the command | `Task beat terminated by signal 15` — a task that lost its `ENVFILE` is restarted without it |
| open the file after clearing, not before | a config that cannot be opened leaves every task with a NULL command |
| restore before classifying | the task never restarts: `restore_task()` releases the copy that tells a pre-existing task from a new one, so the whole list was freed |
| leave the group list out of the rollback | *(not observable — see below)* |

The last two are @SoundGoof's, from two review rounds. The group one is real — the list is global, and a refused config that declared a group left the entry behind — but it has **no test**, and nothing outside the process can observe it: `dump_config()` prints only groups with `maxuse > 1`, and #376 means no group ever has one. It becomes testable when #376 is fixed, and the regression belongs there.

**Three limits, stated rather than papered over:**

- a cut landing **exactly on a section boundary** yields a well-formed prefix and is not detectable this way — so the "absent section" half of the test is asserted as a consequence of the first section being incomplete, not on its own;
- on the **first load** there is no previous state and nothing running, so refusing outright would start nothing at all; only the tasks with no command are dropped;
- `devel` carries the same `load_config()` and the same defect. No separate port is planned — noted so that whoever absorbs devel's xymonlaunch work does not bring the old version back with it.

`tests/server/xymonlaunch-partial-config.sh` drives a live xymonlaunch, forcing re-reads with SIGHUP, and each case above is verified by mutation. Suite on a built tree: 0 failed, 0 skipped.

### DIFF
--- a/common/xymonlaunch.c
+++ b/common/xymonlaunch.c
@@ -87,6 +87,48 @@ volatile int forcereload=0;
 # define xfreeassign(k,p) { if (k) { xfree(k); } k=p; }
 # define xfreedup(k,p) { if (k) { xfree(k); } k=strdup(p); }
 
+/* Release what a re-read parsed into a task, leaving its key and its place in
+   the list alone. */
+static void free_task_config(tasklist_t *t)
+{
+	xfreenull(t->cmd);
+	xfreenull(t->logfile);
+	xfreenull(t->envfile);
+	xfreenull(t->envarea);
+	xfreenull(t->onhostptn);
+	xfreenull(t->cronstr);
+	if (t->crondate) { crondatefree(t->crondate); t->crondate = NULL; }
+}
+
+/*
+ * Put a task back the way it was before the re-read cleared it. Nothing was
+ * lost on the way: the clearing loop NULLs the fields before parsing, so
+ * xfreedup() only ever replaced a NULL. Field by field rather than a memcpy,
+ * whose blanked next/copy pointers would cut the task out of the list.
+ */
+static void restore_task(tasklist_t *twalk)
+{
+	tasklist_t *saved = twalk->copy;
+
+	free_task_config(twalk);
+
+	twalk->disabled   = saved->disabled;
+	twalk->cmd        = saved->cmd;
+	twalk->interval   = saved->interval;
+	twalk->maxruntime = saved->maxruntime;
+	twalk->group      = saved->group;
+	twalk->logfile    = saved->logfile;
+	twalk->envfile    = saved->envfile;
+	twalk->envarea    = saved->envarea;
+	twalk->onhostptn  = saved->onhostptn;
+	twalk->cronstr    = saved->cronstr;
+	twalk->crondate   = saved->crondate;
+	twalk->depends    = saved->depends;
+
+	twalk->cfload = 0;
+	xfreenull(twalk->copy);
+}
+
 void load_config(char *conffn)
 {
 	static void *configfiles = NULL;
@@ -95,6 +137,11 @@ void load_config(char *conffn)
 	strbuffer_t *inbuf;
 	char *p;
 	char myhostname[256];
+	/* Nothing is running yet on the first load, so there is nothing to protect
+	   and an incomplete read is handled differently - see below. */
+	int firstload = (taskhead == NULL);
+	int incomplete = 0;
+	grouplist_t *groupmark;
 
 	/* First check if there were no modifications at all */
 	if (configfiles) {
@@ -109,15 +156,39 @@ void load_config(char *conffn)
 	}
 
 	errprintf("Loading tasklist configuration from %s\n", conffn);
+
 	if (gethostname(myhostname, sizeof(myhostname)) != 0) {
 		errprintf("Cannot get the local hostname, using 'localhost' (error: %s)\n", strerror(errno));
 		strcpy(myhostname, "localhost");
 	}
 
+	/*
+	 * Open the file before touching the task list. The clearing below throws
+	 * away every task's settings so that the re-read can detect what changed;
+	 * doing that first and then bailing out here left every task with a NULL
+	 * cmd and no way back, so each task crashed as soon as it was restarted.
+	 */
+	fd = stackfopen(conffn, "r", &configfiles);
+	if (fd == NULL) {
+		errprintf("Cannot open configuration file %s: %s\n", conffn, strerror(errno));
+		return;
+	}
+
 	/* The cfload flag: -1=delete task, 0=old task unchanged, 1=new/changed task */
+	/*
+	 * The group list is global and outlives the read, so the rollback has to
+	 * put it back too: a refused config that declared a group left it behind
+	 * with its own limit, and a later valid declaration of the same name
+	 * finds the existing entry instead of creating one (@SoundGoof).
+	 */
+	groupmark = grouphead;
+
 	for (twalk = taskhead; (twalk); twalk = twalk->next) {
 		twalk->cfload = -1;
-		twalk->group = NULL;
+		/* The group is cleared below, with the other settings and *after* the
+		   copy: cleared first, the copy held a NULL and a refused read gave
+		   every task back without its group - no limit applied to it, and no
+		   decrement when it exits. */
 		/* Create a copy, but retain the settings and pointers are the same */
 		twalk->copy = xmalloc(sizeof(tasklist_t));
 		memcpy(twalk->copy,twalk,sizeof(tasklist_t));
@@ -137,12 +208,6 @@ void load_config(char *conffn)
 		twalk->cronstr = NULL;
 		twalk->crondate = NULL;
 		twalk->depends = NULL;
-	}
-
-	fd = stackfopen(conffn, "r", &configfiles);
-	if (fd == NULL) {
-		errprintf("Cannot open configuration file %s: %s\n", conffn, strerror(errno));
-		return;
 	}
 
 	inbuf = newstrbuffer(0);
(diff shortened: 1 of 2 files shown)


SoundGoof:
## Finding

### High: Preserving only `CMD` can restart the task without required configuration

[common/xymonlaunch.c:363](https://github.com/xymon-monitoring/xymon/blob/bf75db143d01cbed66227505693773947dd6a35c/common/xymonlaunch.c#L363)

The missing-command guard restores only `CMD`. Other cleared fields remain
missing and are compared immediately afterward. In particular, losing
`ENVFILE` sets `reload`, kills the healthy process, and restarts its preserved
command without loading the required environment.

I reproduced this with a task whose command exits unless `ENVFILE` provides a
variable. After replacing its configuration with only `[beat]` and sending
SIGHUP, xymonlaunch logged that it was keeping the previous command, killed the
running task, and its replacement exited with status 42. Output stopped:

```text
Configuration error, no command for task beat - keeping the previous one
Task beat terminated by signal 15
Task beat terminated, status 42
```
This contradicts the intended guarantee that a partial read leaves the task
runnable. When a known task has no CMD, retain its complete previous
configuration or reject the incomplete reload, rather than restoring only the
command.

bonomani:
Confirmed and fixed — thanks, the finding was right and the first revision did not hold its own guarantee.

Reproduced your sequence as a mutation (restore only `CMD`, everything else as in this PR):

```
Configuration error, no command for task beat - keeping the previous one
Task beat terminated by signal 15
```

Rather than restoring every field for that one task, the reload is now refused whole and the previous configuration put back. Chasing your case turned up a second one it does not cover: sections the short read never reached look *deleted*, so they are killed — `xymond` included, if the cut lands before its section. The test now runs two tasks and truncates inside the first, so the second is absent entirely; restoring only the affected task leaves that one open.

Two limits I did not try to cover: on the first load there is no previous state, so only the tasks with no command are dropped (refusing outright would start nothing); and a cut exactly on a section boundary yields a well-formed prefix and stays undetectable.

Same code and same defect on `devel`.

bonomani:
Production evidence for the second path, from a Homebrew macOS install (`xymonlaunch.log`):

```
21:14:28.494276  Loading tasklist configuration from .../server/etc/tasks.cfg
21:14:28.494374  Cannot open configuration file .../server/etc/tasks.cfg: No such file or directory
21:14:28.528306  Task storeenadis terminated by signal 6
21:14:28.528354  Task storedata terminated by signal 6
                 ... 12 tasks, all within 34 ms
```

The re-read could not open the file — the keg was being swapped under `opt/` — and returned after the clearing loop had already run. Twelve tasks were then forked with no command and died on SIGABRT.

**Seven of those twelve were `DISABLED` in that config** (`distribute`, `xymonproxy`, `xymonfetch`, `storestatus`, `storedata`, `storeenadis`, `storenotes`). The clearing loop resets `disabled` as well as `cmd`, so a failed re-read does not only break the running tasks — it silently re-enables the ones the admin switched off, and starts them. I had not stated that in the description; opening the file before touching the task list covers it, since neither field is cleared unless the read can proceed.

SoundGoof:
## Finding

### Rollback removes restored tasks from the scheduler

`restore_task()` frees `twalk->copy` after restoring each existing task. The
following cleanup loops classify every task with `copy == NULL` as an entry
created during the incomplete read:

https://github.com/xymon-monitoring/xymon/blob/5d3d5c7b167d0f803076976010ed390cec0195a5/common/xymonlaunch.c#L428-L449

Because restoration has already set `copy` to `NULL`, existing restored tasks
satisfy the same condition as newly created tasks. The cleanup therefore
unlinks and frees the restored task list.

Already-running child processes continue independently, so the supplied
heartbeat test passes. However, `xymonlaunch` no longer supervises those
processes.

I reproduced this with two tasks in a group limited to one process. The first
task ran while the second remained queued. After a rejected partial reload, I
terminated the running task. The queued task never started, and debug output
showed empty scheduler scans after the rollback.

This failed consistently on:

| Platform | PR regression | Lifecycle probe |
|---|---:|---:|
| Debian 13 | Pass | Fail |
| NetBSD 9.4 | Pass | Fail |
| NetBSD 11.0 | Pass | Fail |
| OpenBSD 7.9 | Pass | Fail |
| FreeBSD 14.3 | Pass | Fail |

Preserve a separate distinction between pre-existing and newly created tasks
until cleanup is complete. Extend the regression test to terminate a supervised
task after rollback and verify that `xymonlaunch` restarts it or starts a queued
task.

bonomani:
Confirmed, reproduced here, and fixed. Your diagnosis is exact: `restore_task()` releases the copy that tells a pre-existing task from one the read created, so restoring first made every task look new and the unlinking pass freed the whole list.

**Fix is the ordering:** unlink first, restore second. No extra state is needed — the copy is already taken for every task that existed before the read (the clearing loop makes one unconditionally), so `copy == NULL` is a sound marker as long as nothing has consumed it yet.

**Reproduced** with your probe, on the unfixed build:

```
FAIL: a task did not restart after a refused read -- the rollback dropped it
from the task list, leaving its child running unsupervised
```

and it passes with the fix; restoring before classifying fails it again.

That probe is now a permanent case, placed where you suggested — after the refused read, kill the supervised task and require xymonlaunch to bring it back. It is the same shape the "could not open the config" case already used, for the same reason: a running child proves nothing about the list it came from, because it keeps running either way. That is also why the existing heartbeat assertions could not see this.

One knock-on: the following "no task was bounced by the good re-read" assertion now looks only at what that re-read logged, since the forced restart is a termination the test asked for.

Rebased on `2b2272eb9`, which also flattened the merge commit at the branch head — back to one commit, nothing lost (its second parent was `main`). Suite on a built tree: 74 passed, 0 skipped, 0 failed.

SoundGoof:
## Blocking

### A refused reload retains newly created group definitions

`common/xymonlaunch.c` adds new groups directly to the global `grouphead`
list. The incomplete-read rollback restores task fields but does not restore
this list.

I reproduced the resulting configuration poisoning:

1. Start with a valid supervised task.
2. Reload this incomplete configuration, which is correctly refused:

   ```text
   GROUP poison 1
   [keeper]
   ```

3. Load a valid configuration declaring `GROUP poison 3` and three tasks in
   that group.

Only one task starts. The group created by the refused configuration survives
with `maxuse=1`; the later declaration finds that existing group and does not
update its limit.

Repeated refused reloads with unique group names also retain every newly
allocated group.

Snapshot and restore the global group configuration alongside the task list,
or parse all configuration-owned state separately and commit it only after a
complete read. Add a regression covering a refused group declaration followed
by a valid declaration of the same group with a different limit.

## Validation

Reviewed commit `730d1002e52c36e1c9935aaed74d69965905bad4`.

- Submitted live-launcher regression: passed
- Full test suite: 57 passed, 17 skipped, 0 failed
- Test portability guard: passed
- `git diff --check`: passed
- Dedicated live group rollback reproducer: failed, 1 of 3 tasks started

bonomani:
Fixed — the group list is rolled back with the tasks now: a mark is taken before parsing, and the rollback unlinks down to it. Restored tasks point at entries below the mark, so nothing they hold is freed.

**But your reproduction does not show what it looks like it shows, and the difference matters.** I replayed it on the fixed build and still got one task out of three. Chasing that gave a second, older defect:

```c
groupname = p;
p += strcspn(p, " \t");
if (isdigit((int) *p)) maxuse = atoi(p); else maxuse = 1;
```

Nothing terminates the name at the separator, and the character `isdigit()` tests *is* that separator. Running exactly those lines over `GROUP g 3`:

```
groupname=[g 3]   character tested=[ ] (0x20)   maxuse=1
```

So `GROUP g 3` creates an entry literally named `"g 3"` with a limit of 1, and a task saying `GROUP g` does not match it — it gets a second entry, also 1. **Every group is a group of one, whatever the config says**, and that is true on `main` today, untouched by this PR: `xymonlaunch --config=… --dump` prints no `GROUP` line for a config that declares `GROUP g 3`. Filed as #376.

Your step 3 therefore starts one task with or without the rollback fix. The poisoned entry was real, but it was not what capped the group.

**The rollback fix carries no test, and I would rather say so than imply otherwise.** Nothing outside the process can observe the group list — `dump_config()` prints only groups with `maxuse > 1`, and #376 guarantees there are none. The case becomes observable once #376 is fixed, and the regression belongs there.

Rebased on `2b2272eb9`. Suite on a built tree: 74 passed, 0 skipped, 0 failed.

bonomani:
A second pass over this PR (Codex) raised four things. Three are limits this change already states — a cut landing exactly on a section boundary is undetectable, the first load drops tasks with no command rather than refusing, and the group rollback carries no test. The fourth is fair and now fixed, and it is about what a reader believes:

**The `[beat2]` case is not asserted independently, and cannot be.** The test truncates to `[beat]`, which loses that task's `CMD` — and *that* is what triggers the rollback; `[beat2]` surviving is a consequence of it, not a separate guarantee. Nothing in the test said so, so it read as "an absent section is always protected", which is exactly the belief the section-boundary limit contradicts. The comment now says which half does the work:

> What protects beat2 here is the *first* section being incomplete: a task with no command is the only mark of a short read this can see. So (B) is asserted as a consequence of (A), not on its own, and it cannot be otherwise.

No behaviour change; the assertions are unchanged. Suite on a built tree: 74 passed, 0 skipped, 0 failed.

bonomani:
Your rollback finding is fixed, and two later passes found more in the same rollback. Merged here.

**Yours.** The group list is global and survived a refused read; a mark is taken before parsing and the rollback unlinks down to it, restoring what was there before.

**But your reproduction does not show what it looks like.** Replayed on the fixed build it still starts one task of three, because of something older: `GROUP <name> <limit>` parses neither the name nor the limit. Running those two lines over `GROUP g 3` gives `groupname=[g 3]`, `maxuse=1` — nothing terminates the name at the separator, and the character `isdigit()` tests *is* that separator. **Every group is a group of one**, on `main` today. Filed as #376. Your step 3 therefore caps at one with or without the rollback fix.

**Two more, both in this PR's own work:**

- the rollback restored `group` as `NULL` — the clearing loop set it *before* the copy was taken, so every copy held a NULL and restored tasks came back without their group, with no limit applied and no decrement on exit. The clear belongs after the copy, where it already appears;
- both restart assertions could pass without a restart: they compared the beat file's line count across the kill, and a tick the dying child still had in flight satisfied that. They wait for the process to be gone, empty the file, and require it to fill again.

**No test for the group rollback, and I would rather say so:** nothing outside the process can observe the group list — `dump_config()` prints only groups with `maxuse > 1`, and #376 guarantees there are none. It becomes testable when #376 is fixed, and the regression belongs there.

SoundGoof (approved):
No findings.

The revised rollback preserves the complete task configuration, removes
objects created by the refused read before restoring snapshots, and retains
the original group objects referenced by restored tasks. This resolves the
previous rollback findings.

The focused live regression passed on a fresh out-of-tree build. It includes
a forced child restart that verifies tasks remain supervised after both a
short read and an open failure.

All five CI checks are green and `git diff --check` passes.

The known residual limitation is documented: a short read ending exactly on
a section boundary produces a valid-looking prefix and cannot be detected by
the missing-command check.

PR #349, opened 2026-08-12
showgraph: read an RRD's own update time, not its file's (supersedes TBT 215)

### LINKED ISSUE #377: showgraph: an RRD dropped for staleness is reported as a FNPATTERN problem, or not at all
`showgraph` drops an RRD whose file is more than a day old when `nostale` is set — the flag every generated graph link carries — and says nothing about it:

```c
if (ignorestalerrds && (stat(d->d_name, &st) == 0) && ((now - st.st_mtime) > 86400)) {
        continue;                       /* web/showgraph.c:1064 */
}
```

No counter is incremented, so the diagnostic below cannot mention it:

```c
if ((rrddbcount == 0) && svcrejects) {  /* web/showgraph.c:1115 */
        if (rrdparamisservice)
                errprintf("showgraph: no RRD file matched service '%s' - check that FNPATTERN group 1 captures the service component\n", service);
```

`svcrejects` counts only the service-pattern rejects at lines 1051 and 1057. Two outcomes, both wrong:

- every file dropped for staleness and nothing else rejected: `svcrejects` is 0, so the graph comes back empty with nothing logged anywhere;
- an unrelated sibling rejected by the service pattern: the message fires and blames the `FNPATTERN` capture of a configuration that was never at fault.

That second one is what sent the reporter of the macOS blank-graph case (#349) to check `graphs.cfg` instead of the file timestamps.

Counting the staleness drops separately and naming them in the message is the whole of it. Split out of #349, whose title promised this and whose diff does not contain it — the timestamp fix there stands on its own.

bonomani:
`nostale` — on every generated graph link — drops an RRD not updated for a day. That answer was
read off the file's mtime, which only tracked updates because `do_rrd.c` pushed it with a
Linux-only `utimes()`. On a Homebrew macOS server every mtime froze at creation time, and a day
later every `FNPATTERN` multi-graph (disk, tcp) rendered empty while the data kept arriving.
`showgraph` now asks the RRD itself (`rrd_last`, through the existing argv wrappers), and
`do_rrd.c` loses its stamp.

Deliberate semantics:
- Staleness means the last **successful** sample: an RRD whose every update fails leaves nostale
  views after a day. `xymond_rrd` logs each failed update.
- A file `rrd_last` cannot read is dropped with a log line naming it; one bad file no longer
  fails the whole multi-graph image.
- Data timestamps lagging wall clock by more than a day count as stale even while updates succeed.
- Every nostale request pays one `rrd_last` per pattern-matched file. Measured below; no mtime
  fast path is sound (the stale test case is a fresh file with 3-day-old data).

| measured, 200 disk RRDs | |
|---|---|
| `rrd_last()` warm / cold | 7.5 µs / 116 µs per file |
| filter on a cold 20-RRD render | 2.4 ms of 39.9 ms — 6% |
| removed from the write path | one `utimes()` per RRD per update |

Constraints:
- **This PR imposes no RRDtool version floor, and the build probes for none.** The staleness
  read does not depend on one: `rrd_last` is called with `./<name>` as its sole operand, which
  the legacy direct-`argv[1]` path (1.2.x) and the option-parsing path (>= 1.4) both accept. The
  wrappers reset `optind`/`opterr` before every call for the entry points that *do* parse
  options — create, update, fetch, graph — which is what a pre-1.7 RRDtool needs to re-scan and
  a no-op under 1.7's private optparse. Tested here against librrd 1.7.2.
- `rrd_last` honours `RRDCACHED_ADDRESS`: an rrdcached site needs the daemon reachable from the
  CGI. Unreachable drops every file, each with its log line. Deliberately not unset — `rrd_graph`
  wants the daemon too.
- Release note: the removed `utimes()` also kept mtimes fresh for out-of-tree consumers (rsync
  quick-check, mtime-keyed purge crons) on pre-3.5 Linux kernels — and permanently on macOS,
  where mmap writes never stamp the mtime.

Fixes #377 (second commit): an empty result now names the nostale drops — `no RRD for service
'x' - N matching file(s) dropped by the nostale filter` — instead of silence, or the FNPATTERN
blame that misdirected the original report to graphs.cfg.

`tests/web/showgraph-stale-filter.sh` drives the real CGI on RRDs whose file times contradict
their data:

| case | must | old code |
|---|---|---|
| written now, mtime backdated | be graphed | dropped — the macOS blank graph |
| last written 3 days ago, file just created | not be graphed | graphed |
| every match stale | named in the log, exact message | silence, or FNPATTERN blame |

Follow-up: lift the duplicated disk `@@status` fixture into tests/lib (shared with
xymond-rrd-drophost.sh).

### DIFF
--- a/build/test-rrd.c
+++ b/build/test-rrd.c
@@ -24,7 +24,6 @@ int main(void)
 	double ymin, ymax;
 
 	for (pcount = 0; (rrdargs[pcount]); pcount++);
-	rrd_clear_error();
 	result = xymon_rrd_graph(pcount, rrdargs, &calcpr, &xsize, &ysize, NULL, &ymin, &ymax);
 	(void)result;
 
--- a/lib/rrd_api_compat.h
+++ b/lib/rrd_api_compat.h
@@ -1,7 +1,12 @@
 #ifndef __RRD_API_COMPAT_H__
 #define __RRD_API_COMPAT_H__
 
+#include <unistd.h>	/* optind/opterr: librrd parses these argv forms with getopt() */
 #include <rrd.h>
+
+/* A strictly-conforming <unistd.h> (e.g. plain -std=c99) keeps the getopt
+ * globals to itself; librrd uses them regardless. */
+extern int optind, opterr;
 
 /*
  * RRDtool changed the argv parameter of its public API from "char **" to
@@ -18,39 +23,61 @@
 
 #if RRD_CONST_ARGS
 typedef const char *xymon_rrd_argv_item_t;
-static const char **xymon_rrd_api_argv(xymon_rrd_argv_item_t *argv)
+static inline const char **xymon_rrd_api_argv(xymon_rrd_argv_item_t *argv)
 {
 	return (const char **)argv;
 }
 #else
 typedef char *xymon_rrd_argv_item_t;
-static char **xymon_rrd_api_argv(xymon_rrd_argv_item_t *argv)
+static inline char **xymon_rrd_api_argv(xymon_rrd_argv_item_t *argv)
 {
 	return argv;
 }
 #endif
 
-static int xymon_rrd_update(int argc, xymon_rrd_argv_item_t *argv)
+/*
+ * librrd parses these argv forms with getopt(): reset its globals and the
+ * error state before every call; the state after a call is the caller's.
+ * optind=0 is the GNU re-init - moot on RRDtool >= 1.7, whose private
+ * optparse ignores the getopt globals. That version is the supported floor.
+ */
+static inline void xymon_rrd_call_setup(void)
 {
+	optind = opterr = 0;
+	rrd_clear_error();
+}
+
+static inline int xymon_rrd_update(int argc, xymon_rrd_argv_item_t *argv)
+{
+	xymon_rrd_call_setup();
 	return rrd_update(argc, xymon_rrd_api_argv(argv));
 }
 
-static int xymon_rrd_create(int argc, xymon_rrd_argv_item_t *argv)
+static inline int xymon_rrd_create(int argc, xymon_rrd_argv_item_t *argv)
 {
+	xymon_rrd_call_setup();
 	return rrd_create(argc, xymon_rrd_api_argv(argv));
 }
 
-static int xymon_rrd_fetch(int argc, xymon_rrd_argv_item_t *argv,
+static inline int xymon_rrd_fetch(int argc, xymon_rrd_argv_item_t *argv,
 				  time_t *start, time_t *end, unsigned long *step,
 				  unsigned long *dscount, char ***dsnames, rrd_value_t **data)
 {
+	xymon_rrd_call_setup();
 	return rrd_fetch(argc, xymon_rrd_api_argv(argv), start, end, step, dscount, dsnames, data);
 }
 
-static int xymon_rrd_graph(int argc, xymon_rrd_argv_item_t *argv,
+static inline time_t xymon_rrd_last(int argc, xymon_rrd_argv_item_t *argv)
+{
+	xymon_rrd_call_setup();
+	return rrd_last(argc, xymon_rrd_api_argv(argv));
+}
+
+static inline int xymon_rrd_graph(int argc, xymon_rrd_argv_item_t *argv,
 				  char ***calcpr, int *xsize, int *ysize,
 				  void *prdata, double *ymin, double *ymax)
 {
+	xymon_rrd_call_setup();
 	return rrd_graph(argc, xymon_rrd_api_argv(argv), calcpr, xsize, ysize, prdata, ymin, ymax);
 }
 
--- a/lib/rrd_api_compat.h
+++ b/lib/rrd_api_compat.h
@@ -36,10 +36,13 @@ static inline char **xymon_rrd_api_argv(xymon_rrd_argv_item_t *argv)
 #endif
 
 /*
- * librrd parses these argv forms with getopt(): reset its globals and the
- * error state before every call; the state after a call is the caller's.
- * optind=0 is the GNU re-init - moot on RRDtool >= 1.7, whose private
- * optparse ignores the getopt globals. That version is the supported floor.
+ * The option-parsing entry points (create/update/fetch/graph) run getopt()
+ * over their argv: reset its globals and the error state before every call;
+ * the state after a call is the caller's. optind=0 is the GNU re-init that
+ * an option-parsing RRDtool before 1.7 relies on to re-scan; RRDtool >= 1.7
+ * uses a private optparse that ignores the getopt globals, so the reset is a
+ * harmless no-op there. This resets unconditionally rather than asserting a
+ * version floor, which the build does not probe for.
  */
 static inline void xymon_rrd_call_setup(void)
 {
--- a/tests/buildsystem/rrd-api-compat.sh
+++ b/tests/buildsystem/rrd-api-compat.sh
@@ -42,6 +42,8 @@ int rrd_create(int, $argv);
 int rrd_fetch(int, $argv, time_t *, time_t *, unsigned long *,
               unsigned long *, char ***, rrd_value_t **);
 int rrd_graph(int, $argv, char ***, int *, int *, void *, double *, double *);
+time_t rrd_last(int, $argv);
+void rrd_clear_error(void);
 #endif
 EOF
 }
@@ -65,6 +67,7 @@ int main(void)
 			      &dsnames, &data);
 	(void)xymon_rrd_graph(1, argv, &calcpr, &xsize, &ysize, 0,
 			      &ymin, &ymax);
+	(void)xymon_rrd_last(1, argv);
 	return 0;
 }
 EOF
--- a/tests/rrd/rrdcached-relative-path-key.sh
+++ b/tests/rrd/rrdcached-relative-path-key.sh
@@ -1,0 +1,78 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/rrd/rrdcached-relative-path-key.sh
+#
+# showgraph's nostale filter asks rrd_last() how fresh an RRD is, and passes
+# the filename as "./<name>" -- the "./" keeps a "-"-prefixed name out of
+# option parsing while staying a plain operand for a legacy rrd_last() that
+# opens argv[1] directly (web/showgraph.c). Every rrd_* call honours
+# RRDCACHED_ADDRESS, and rrdcached keys its pending-update cache by the
+# filename string it is handed. So the "./" prefix is only safe if the daemon
+# resolves "./name" and the bare "name" to the *same* cache entry: otherwise a
+# reading still sitting in the daemon would not be flushed for the probe, and
+# an RRD kept fresh only in the cache could read as stale.
+#
+# This drives rrdcached directly (no Xymon binary): queue an update under the
+# bare name so it sits in the cache unwritten, then read the last-update time
+# back through the daemon as "./name". If "./name" keyed a different entry the
+# flush would be a no-op and the read would return the stale on-disk time; it
+# returns the freshly cached one, so the two forms are the same key.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+command -v rrdtool   >/dev/null 2>&1 || skip "rrdtool not in PATH"
+command -v rrdcached >/dev/null 2>&1 || skip "rrdcached not in PATH (optional rrdtool daemon)"
+
+work=$(mktempdir)
+# rrdcached refuses a base directory reached through a symlink (e.g. macOS
+# /tmp -> /private/tmp, /var -> /private/var), so hand it the resolved path.
+base=$(cd "$work" && pwd -P)
+sock="$base/rc.sock"
+pidf="$base/rc.pid"
+
+rrdcached -l "unix:$sock" -b "$base" -B -w 3600 -f 7200 -p "$pidf" -m 0700 \
+	2>"$base/rc.err" || { cat "$base/rc.err" >&2; skip "rrdcached would not start here"; }
+register_cleanup "[ -f '$pidf' ] && kill \"\$(cat '$pidf')\" 2>/dev/null || true"
+
+# Wait for the socket rather than sleeping a fixed time.
+for _ in 1 2 3 4 5 6 7 8 9 10; do [ -S "$sock" ] && break; sleep 0.2; done
+[ -S "$sock" ] || { cat "$base/rc.err" >&2; skip "rrdcached socket never appeared"; }
+
+daemon="unix:$sock"
+now=$(date +%s)
+start=$((now - 600))
+
+cd "$base"
+rrdtool create test.rrd --start "$start" --step 300 \
+	DS:x:GAUGE:600:U:U RRA:LAST:0.5:1:10
+
+pre=$(rrdtool last test.rrd)   # on-disk last-update, before any update
+
+# Queue an update under the BARE name. With -w 3600 it stays in the daemon's
+# cache, unwritten to disk.
+rrdtool update --daemon "$daemon" test.rrd "$now:42"
+
+# The file on disk must still show the old time -- proves the update is
+# cache-only, so a later "last" that returns the new time can only have got
+# it by flushing the cached entry.
+ondisk=$(rrdtool last test.rrd)
+assert_equal "$pre" "$ondisk" \
+	"queued update was written to disk immediately -- test cannot distinguish the cache"
+
+# THE ASSERTION: read last-update through the daemon as "./test.rrd", before
+# any bare-name read has had a chance to flush. It must reflect the queued
+# update (newer than on disk), which is only possible if "./test.rrd" keys the
+# same cache entry the bare name was queued under.
+dot=$(rrdtool last --daemon "$daemon" ./test.rrd)
+[ "$dot" != "$pre" ] || fail \
+	"rrdcached did not flush the queued update for './test.rrd' (last=$dot, on-disk=$pre) -- './name' keys a different cache entry than the bare name"
+
+# And the bare name agrees (now reading the just-flushed value from disk).
+bare=$(rrdtool last --daemon "$daemon" test.rrd)
+assert_equal "$dot" "$bare" \
+	"'./test.rrd' and 'test.rrd' disagree through the daemon"
+
+pass "rrdcached keys './name' and 'name' to the same cache entry"
--- a/tests/server/xymond-rrd-mtime-freshness.sh
+++ b/tests/server/xymond-rrd-mtime-freshness.sh
@@ -1,0 +1,81 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/server/xymond-rrd-mtime-freshness.sh
+#
+# showgraph drops an RRD from a graph when "nostale" is set and the file's
+# timestamp is more than a day old (web/showgraph.c), and every graph link
+# Xymon generates carries that flag. The timestamp is therefore part of the
+# writer's contract: it must move whenever xymond_rrd writes a reading, or the
+# graph goes empty a day later with nothing logged anywhere.
+#
+# do_rrd.c stamps the file explicitly because RRDtool writes through mmap,
+# where the kernel does not move the timestamp by itself. That stamp used to
+# be compiled for Linux only, although nothing about mmap is Linux-specific.
+#
+# This pins the contract rather than the workaround, in both directions: a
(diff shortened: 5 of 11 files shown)


SoundGoof:
## Finding

### Medium: Failed updates still mark stale data as fresh

[xymond/do_rrd.c:259](https://github.com/xymon-monitoring/xymon/blob/a520eb0939e087497c6cbdccf5ae083040d65777/xymond/do_rrd.c#L259)

`utimes()` runs regardless of whether `xymon_rrd_update()` succeeded.

I reproduced this on FreeBSD by submitting a duplicate timestamp. RRDtool
rejected the update and `rrdtool last` remained unchanged, but the RRD's
filesystem mtime advanced beyond a marker created before the attempt.

As a result, repeated rejected updates can keep stale RRD data visible through
`showgraph`'s `nostale` filter. This PR extends that existing Linux behavior to
every platform.

Only stamp the file when `result == 0`, and add a regression case confirming
that a rejected update does not refresh mtime.

## Validation

The successful-update freshness test passes on:

- Debian 13
- FreeBSD 14.3 with RRDtool 1.10.3
- NetBSD 11.0 with RRDtool 1.10.3
- OpenBSD 7.9 with RRDtool 1.9.0

All five GitHub checks pass.

bonomani:
Confirmed and fixed — `if (result == 0) utimes(...)`, plus the regression case you asked for.

Reproduces on Linux too: a reading older than the one already stored is rejected and the mtime advanced anyway. That case is the one half of the test that discriminates on a Linux runner, since nothing but the stamp can move a timestamp RRDtool declined to touch — the accepted-reading half does not, and the test header says so.

This is the substance of #82, which I had missed before opening this; the gate is credited there and in the commit. #82 can be closed as superseded.

SoundGoof:
## Finding

### Medium: RRDtool can stamp the file before rejecting the update

[xymond/do_rrd.c:264](xymond/do_rrd.c#L264)

Gating Xymon's `utimes()` on `result == 0` does not preserve the stated
contract on platforms where RRDtool changes mtime while opening the RRD.

The exact PR head was built on FreeBSD 14.3 with RRDtool 1.10.3 and UFS.
The new rejected-update assertion failed in five consecutive runs.

A direct, correctly shaped, out-of-order update produced:

    requested timestamp: 1786647712
    previous last_update: 1786648012
    rrdtool return code:  1
    RRD mtime before:     1786647755
    marker mtime:         1786647949
    RRD mtime after:      1786647951
    last_update after:    1786648012

The reading was rejected and `last_update` remained unchanged, but mtime
advanced beyond the marker.

A FreeBSD `ktrace` of that direct command shows the cause and ordering:

    utimes("/tmp/pr349-trace.rrd", NULL)
    mmap(..., PROT_READ|PROT_WRITE, MAP_SHARED, ...)
    ERROR: illegal attempt to update using an older timestamp

There is no Xymon process in this trace. RRDtool itself calls `utimes()` before
validating the reading.

FreeBSD's RRDtool package has mmap support enabled. Its configure probe defines
`HAVE_BROKEN_MS_ASYNC` when `msync(MS_ASYNC)` does not reliably update mtime.
Under that definition, `rrd_open()` deliberately calls `utime(file_name, NULL)`
whenever it opens an RRD read-write. This workaround runs even when the later
update fails.

Xymon's success gate therefore prevents only its own explicit stamp. It cannot
undo the earlier RRDtool stamp, and a stalled RRD continues passing
`showgraph`'s `nostale` check.

Validation matrix:

| Platform | RRDtool | Result |
|---|---:|---|
| Linux 7.1.5, tmpfs | 1.9.0 | Pass |
| FreeBSD 14.3, UFS | 1.10.3 | Fail, 5/5 runs |
| NetBSD 11.0 | 1.10.3 | Pass |
| OpenBSD 7.9, FFS | 1.9.0 | Pass |

Capture the previous timestamp before `xymon_rrd_update()` and restore it when
the update fails. Add FreeBSD coverage so
[tests/server/xymond-rrd-mtime-freshness.sh:78](tests/server/xymond-rrd-mtime-freshness.sh#L78)
runs against the RRDtool configuration that exercises this path.

All GitHub CI checks are green, but current CI does not cover this FreeBSD
behavior.

bonomani:
User-visible reproduction, on the Homebrew macOS install this was found on.

Every RRD carried an mtime equal to its creation time — 13 Aug 00:02, while `rrdtool last` reported the current cycle. At 14 Aug 10:12, 34 hours later, every graph on the server was blank. `touch`ing the RRD files restored all of them immediately, with no other change: same files, same `graphs.cfg`, same running daemons. The only variable was the timestamp.

So the failure is not theoretical and not gradual — it is total, and it lands exactly 24 h after the RRDs are created, which is the `nostale` window in `web/showgraph.c`.

bonomani:
Verified on macOS, with this branch installed.

The file that read `13 Aug 00:02` for 34 hours, above, now tracks the writes:

```
mtime  13:05:56
now    13:11:58
```

One client cycle apart. The timestamp now tracks the writes, so the file stays inside the window and the graphs no longer expire.

That closes the gap I flagged when opening this: the Linux measurement showed the stamp is a no-op there (librrd 1.7.2 moves the mtime on its own), which is why the accepted-reading test does not discriminate on a Linux runner. macOS is the platform this is for, and it is now measured rather than argued.

SoundGoof:
## Finding

### Rejected updates still refresh mtime on FreeBSD

The success check only prevents Xymon’s explicit `utimes()` call:

https://github.com/xymon-monitoring/xymon/blob/56044b6599b1e1a01e3a4c39f399493cb35b2a75/xymond/do_rrd.c#L262-L276

On FreeBSD, RRDtool 1.10.3 calls `utimes()` while opening an RRD for writing,
before validating the supplied timestamp. An out-of-order update therefore
refreshes mtime even though `xymon_rrd_update()` returns an error and the stored
`last_update` remains unchanged.

I rebuilt exact head `56044b6` on FreeBSD 14.3 and ran the PR’s own regression:

```text
FAIL: the RRD's timestamp moved for a reading RRDtool rejected -- an RRD
that has stopped being written would keep passing showgraph's 'nostale'
filter

**Comment by bonomani** [comment]
Confirmed and fixed, and I could reproduce it here rather than take FreeBSD on trust.

The timestamp is now read **before** the update and put back when the update fails, instead of merely not being stamped. Not touching the file on a rejected reading is librrd's behaviour, not RRDtool's contract, and the gate assumed it was.

**Reproduced on Linux.** Interposing `rrd_update()` with a shim that stamps the file before delegating puts your library's behaviour exactly where it happens — between the timestamp being read and being put back:

| | rejected reading |
|---|---|
| stock librrd 1.7.2, no shim | mtime unchanged — which is why this suite could not see it |
| shim, before the fix | **mtime moved** — your failure |
| shim, after the fix | mtime restored |
| shim, accepted reading | mtime moves, as it must |

That case is now in the test, gated on the interposition actually taking and leaving a witness file, so it cannot quietly stop asserting. Removing the restore fails it.

**Cost**, since it adds a `stat()` to every flush: 265 ns against the 28 µs update it precedes, measured on this tree — about 1%.

Rebased on `2b2272eb9`. That also flattened the merge commit the branch head had become, so the PR is back to one commit; nothing was lost — its second parent was `main`. Suite on a built tree: 74 passed, 0 skipped, 0 failed.

**Comment by SoundGoof** [comment]
## Concern

### Avoid the additional `stat()` on unaffected systems

This change calls `stat()` before every RRD flush so that a failed update can
restore the previous timestamp. The restore is needed for the observed
FreeBSD RRDtool 1.10.3 behavior, where RRDtool changes the timestamp before
rejecting an update.

Should this additional metadata operation run on every platform?

Consider enabling the pre-update `stat()` and failed-update restoration only
for affected platforms or through a runtime option. The successful-update
`utimes()` can remain platform-independent because macOS requires it.

## Scope note

The timestamp fix resolves the reported blank graphs without changing
`showgraph`: once `xymond_rrd` maintains the correct `mtime`, the existing
`nostale` check includes the RRD.

The title and description also say that `showgraph` will report stale-file
rejection correctly, but this revision contains no `showgraph.c` change.
Either remove that diagnostic claim or implement it separately. This does not
block the timestamp fix itself.

## Validation

Reviewed commit `15d45ff5eeab1bbdec3d8a1cac320e8b47d1e6ff`.

- All five GitHub checks: passed
- Focused Linux regression: passed
- Full Linux suite: 71 passed, 3 skipped, 0 failed
- Test portability guard: passed
- `git diff --check`: passed
- Native FreeBSD 14.3 build: passed
- Native FreeBSD test with RRDtool 1.10.3: passed
- Touched RRD code introduced no new compiler warning

The timestamp behavior worked as expected. Successful updates refresh the
file timestamp, while rejected updates restore the previous timestamp on
FreeBSD RRDtool 1.10.3.

**Comment by bonomani** [comment]
**On the scope note: you are right, and it is fixed.** The title promised a `showgraph` diagnostic this revision does not contain. Title and body now describe only the timestamp fix. The gap is real — the staleness drop at `showgraph.c:1064` increments no counter, so the message at `:1115` either stays silent or blames the `FNPATTERN` capture of a configuration that was never at fault — so it is split out as #377 rather than dropped.

**On the `stat()`: I would rather not gate it by platform, and the measurement is why.** Five runs on this tree, per flush:

```
stat()   rrd_update()   ratio
267 ns       28 us       1:105
266 ns       31 us       1:116
265 ns       27 us       1:102
263 ns       57 us       1:216
262 ns       27 us       1:104
```

It is one metadata read against the write it precedes — under 1%, on the same inode the update is about to touch and therefore already in cache. Against that, a platform gate costs the thing this PR exists to undo: the `utimes()` it sits beside *was* gated to Linux, which is exactly how macOS went two years with every graph blanking a day after each restart and nothing logged. A conditional that says "this platform does not need it" is a claim about every RRDtool build on every platform, and it is the claim that already failed once here.

There is a way to remove it from the hot path that is not a platform gate, if you want it gone: remember the stamp we set ourselves on the successful update, and restore *that* on a failure, falling back to a `stat()` only for the first failure after a restart. It costs a `time_t` per cache entry and one more piece of state to reason about. For 265 ns I did not think that trade was worth making — but if you would rather have the state than the syscall, say so and I will do it.

Rebased on `2b2272eb9`. Suite on a built tree: 74 passed, 0 skipped, 0 failed.

**Comment by bonomani** [comment]
**Scope note: right, and fixed.** The title promised a `showgraph` diagnostic this revision does not contain. Title and body describe only the timestamp fix; the gap is real — the staleness drop at `showgraph.c:1064` increments no counter, so the message at `:1115` either stays silent or blames the `FNPATTERN` capture of a config that was never at fault — and is split out as #377.

**On the `stat()`: I would rather not gate it by platform.** Five runs, per flush: 262-267 ns against a 27-57 µs update — under 1%, on the inode the update is about to touch. A platform gate is what this PR undoes: the `utimes()` beside it *was* Linux-only, which is how macOS went two years with every graph blanking a day after each restart. If you want it off the hot path anyway, the way is to remember the stamp we set ourselves and restore that, falling back to a `stat()` only on the first failure after a restart — a `time_t` per cache entry. For 265 ns I did not think that trade was worth making; say the word and I will.

**Three more from a later pass**, all on the restore path this PR adds:

- if the pre-update `stat()` fails there is nothing to put back, and a rejection after that left the file looking freshly written with nothing said;
- the restoring `utimes()` was unchecked — xymond can be allowed to write an RRD and not to stamp it;
- the summary line claimed the whole contract even when the library-stamped case was skipped. It reads `(library-stamped rejection: covered)` or `not covered`, verified in both directions.

**Comment by SoundGoof** [comment]
### Finding

Account for the existing shutdown-flush bottleneck before adding more
synchronous filesystem activity.

In our production environment, shutting down `xymond_rrd` currently takes
about eight minutes because the final cache flush serially processes every RRD
with pending values. The setup uses a clustered filesystem and contains a
large number of RRD files.

This change adds a `stat()` before every update and an additional `utimes()`
after each successful update. Those operations also run for every RRD handled
by `rrdcacheflushall()` during shutdown.

The platform-specific behavior was identified through the new regression
tests. Keep that coverage and use it to guard any platform gates we introduce
for production. If `uname` or equivalent build-time platform detection is used
to avoid unnecessary metadata operations, the tests should verify that each
supported platform selects the correct path and preserves the required RRD
timestamp behavior.

**Comment by bonomani** [comment]
The reshaped commit removes the concern by direction rather than by guarding it: nothing is added to the write path any more.

The freshness question moved to read time — `showgraph` asks the RRD itself (`rrd_last`, through the existing argv wrappers) when it renders; a file RRDtool declines to read is dropped with a log line naming it, rather than falling back to its mtime. The write path *loses* the pre-existing Linux `utimes()`-per-update, so `rrdcacheflushall()` at shutdown does strictly fewer filesystem operations than current main — no `stat()`, no `utimes()`.

Read-time cost, measured on 200 disk RRDs: 7.5µs warm / 116µs cold per file, ~6% of a cold render.

**Comment by SoundGoof** [comment]
### Findings

I like that you are able to remove stuff from the RRD update hot path

#### Medium: Support both RRDtool argv APIs in the regression harness

[tests/web/showgraph-stale-filter-harness.c:24](https://github.com/xymon-monitoring/xymon/blob/8ddf06fd8ac973d09eaa0f897f653e1f2e5f1e93/tests/web/showgraph-stale-filter-harness.c#L24-L39)

The harness still passes `char **` directly to `rrd_create()` and
`rrd_update()`.

Building exact head `8ddf06fd` with RRDtool 1.9.0's `const char **` API
and GCC 15.3.1 fails at both calls. The regression therefore cannot run.

Use `rrd_api_compat.h`: compile with `RRDDEF`, declare the arrays as
`xymon_rrd_argv_item_t`, and call `xymon_rrd_create()` and
`xymon_rrd_update()`.

That form was previously verified against both current headers and exact
RRDtool 1.2.2.

#### Medium: Avoid the `pipefail` and `grep -q` SIGPIPE race on FreeBSD

[tests/web/showgraph-stale-filter.sh:125](https://github.com/xymon-monitoring/xymon/blob/8ddf06fd8ac973d09eaa0f897f653e1f2e5f1e93/tests/web/showgraph-stale-filter.sh#L125-L127)

The image assertion still pipes the complete CGI response through `tr` into
`grep -q`.

On FreeBSD 14.3, `grep` finds `IHDR` and exits while `tr` is writing.
`tr` receives SIGPIPE, and `pipefail` reports failure:

```text
tr:   141
grep: 0
```

The preserved response contains a complete PNG. Check the response file
directly with a binary-safe tool instead of using this producer pipeline.

### Summary

The production rewrite resolves the shutdown-flush concern and uses the
RRD's authoritative update time. Exact RRDtool 1.2.2 source confirms that
the production `rrd_last()` path is supported.

The two prior test-portability findings remain unchanged in current head
`8ddf06fd`.

### Validation

- Exact current head built successfully
- RRD API compatibility test passed
- Stale-filter regression failed while compiling its helper
- `git diff --check` passed

bonomani:
@SoundGoof Both findings are fixed, head `493481a28`; full suite 87/0/0, `git diff --check` clean.

- **Harness API**: it goes through `lib/rrd_api_compat.h` (`xymon_rrd_argv_item_t`, `xymon_rrd_create/update`), compiled with `$rrddef -iquote lib`. Proven both ways locally: against a mock `const char **` header the old harness fails with your incompatible-pointer error, the new one compiles; against the real header the test runs and passes.
- **FreeBSD SIGPIPE**: no producer pipeline any more — the image probe is `LC_ALL=C grep -aq 'IHDR'` on the answer file directly; `-a`/`LC_ALL=C` also covers greps that go binary-mode on the PNG bytes (reproduced here with ugrep), including the `DEF:` dump the assertions read.

From a deeper pass in the same area: an unreadable RRD is now dropped from nostale views uniformly with a log line naming it and the RRD error (the mtime no longer arbitrates the error path, and one bad file cannot fail the whole image); the `optind`/`rrd_clear_error` ritual moved into the `xymon_rrd_*` wrappers, deleting all eight caller-side copies; the disk fixture sends `40%` (a printf double-escape had it sending `40%%`); the `PCRELIBS` chain scrapes the configured Makefile; the test composes with assert.sh's cleanup chain; `do_rrd.c` drops its dead `<sys/time.h>`/`<utime.h>`. `rrd_last_r` was considered and rejected — the `_r` family postdates the RRDtool 1.2.2 you verified against; a comment records that.

SoundGoof:
## Finding

### `nostale` rejects every RRD with older accepted RRDtool

`web/showgraph.c:1067-1079`

The new call passes:

```c
{ "last", "--", filename }
```

RRDtool 1.2.2 implements `rrd_last()` by opening `argv[1]` directly. It
therefore attempts to open a file named `--`, returns `-1`, and causes every
matching RRD to be dropped as unreadable. Generated graph links use `nostale`,
so affected installations receive empty multi-RRD graphs.

The repository currently accepts these installations:

- Configure performs compile/link and argv-prototype probes but has no minimum
  RRDtool version check.
- `docs/install.html.DIST` explicitly supports RRDtool 1.2 through 1.4 and
  recommends 1.2.30.
- The legacy compatibility test only compiles against a mocked prototype; it
  does not test runtime argument handling.

RRDtool 1.4.8 is a viable new floor. Its `rrd_last()` parses options, accepts
the standalone `--`, and exposes all APIs used by Xymon. It was packaged by
Debian 8 and RHEL/CentOS 7.

Resolve this either by retaining compatibility with older RRDtool or by making
the dependency floors explicit.

For the compatibility approach, pass `"./<filename>"` as the sole filename
argument. This protects dash-prefixed names while working with both legacy
direct-`argv[1]` and modern option-parsing implementations.


## Validation

- `git diff --check` passes.
- The branch merges cleanly with current `main`.
  - Debian 13
  - NetBSD 9.4
  - NetBSD 11.0
  - OpenBSD 7.9
  - FreeBSD 14.3
- Wrapper and stale-filter regressions pass with RRDtool 1.9/1.10 on all five.
- RRDtool 1.2.2 and 1.4.8 behavior was verified from their exact source.
- The PCRE2 migration history and CentOS 7/Amazon Linux 2 compatibility fix
  were inspected.

bonomani:
Confirmed from the 1.2.2 source, and fixed — head `5779a803f`.

`rrd_last()` there opens `argv[1]` directly, so `{ "last", "--", filename }` opens `--` and drops every RRD as unreadable. The `"--"` I added for dash-safety broke the version I'd claimed to support.

**Fix: `"./<name>"` as the sole operand**, per your suggestion — works on both the option-parsing (≥1.4) and legacy direct-`argv[1]` (1.2.x) paths, and still keeps a `-`-prefixed name out of option parsing. `showgraph` is `chdir`'d into the RRD dir (`showgraph.c:932`), so `./name` resolves identically to the bare name. Verified the argv construct compiles clean under both the `char *` and `const char *` wrapper typedefs.

**rrdcached keys `./name` and `name` the same** — and this is now a regression rather than my word for it (`tests/rrd/rrdcached-relative-path-key.sh`): it queues an update under the bare name so it sits in the daemon cache unwritten, then reads last-update back as `./name`. A different key would leave the flush a no-op and return the stale on-disk time; it returns the freshly cached one. Skips when `rrdcached` isn't installed. Passing here against rrdcached 1.11.

**On the floor.** You're right the `≥1.7` line was asserted in a comment while nothing enforces it and the docs still recommend 1.2.30. With `rrd_last` no longer needing option parsing, this PR imposes no floor of its own, so I've stopped the comment from claiming one. Making the *project* floor explicit — bump `docs/install.html.DIST` and add a `configure` version probe to match — reaches past this PR, so I'd rather take your steer than legislate it here.

**One unrelated portability fix folded in** (`tests/web/showgraph-exec-title.sh`): it compiled `showgraph.c` without `RRDINCDIR`, so `#include <rrd.h>` was only found where rrd.h sits on the default path — it fails on macOS/BSD exactly as this whole PR is about. Passes `$rrdinc` now, like the stale-filter test already does. Not strictly in scope; say if you'd rather it went separately.

Full suite on a built tree here: 88 passed, 0 skipped, 0 failed.

bonomani:
I reviewed this — **clean, no actionable correctness regressions.** It signed off on the RRD API compat wrappers centralizing the state/error resets across the legacy vs const-argv forms, the stale filter now keying on the RRD's own last-update instead of the file mtime, and the added compat probe + regression tests. Review was run on the current branch (rebased onto `main`).

Also **verified the macOS fix end-to-end** on a live Homebrew `xymon-server`:

- The frozen-mtime symptom is real here: every `localhost` RRD had its file mtime stuck ~5 days back (mmap writes don't advance mtime on macOS) while the RRD's own last-update was minutes old.
- Driving the real `showgraph.cgi --debug` with `&nostale`:
 - the live `disk` RRD (**mtime 5 d, last-update 1 min**) is **kept** and drawn — the old mtime-based logic would have dropped it (`5 d > 1 d`) → blank graph;
 - a genuinely stale RRD (last-update 2 d) is **dropped**, with `showgraph: no RRD for service 'disk' - 1 matching file(s) dropped by the nostale filter`;
 - a mixed directory keeps the fresh RRD and drops the stale one.

So the `(now - xymon_rrd_last()) > 86400` check behaves correctly and resolves the macOS blank-graph failure on real data — pre-#349, every trend graph on this box came up blank; on this build they render.

PR #351, opened 2026-08-13
web: order install-cgi after install-bin, and stop on a failed link (#350)

### LINKED ISSUE #350: make -j install can silently drop a CGI wrapper link (history.sh 404)
A parallel `make install` can silently omit one of the CGI wrapper links, leaving that CGI 404 on a build that reported success.

Observed on a Homebrew build of `main` (Homebrew exports `-j` by default): `cgi-bin/` held 25 of the 26 expected links; the missing one was `history.sh`, so the history CGI returned 404 while every other CGI worked.

## Cause

`web/Makefile:165`

```make
install: install-bin install-cgi install-man
```

`install-bin` copies `$(PROGRAMS)` — which includes `cgiwrap` — into `$(INSTALLBINDIR)`. `install-cgi` then hardlinks *from* `$(INSTALLBINDIR)/cgiwrap`:

```make
install-cgi:
	...
	for F in $(CGISCRIPTS); do ln -f $(INSTALLROOT)$(INSTALLBINDIR)/cgiwrap $(INSTALLROOT)$(CGIDIR)/$$F; done
```

Listing both under `install:` does not order them. Under `-jN` the two run concurrently, so `install-cgi` can start before the `cp` has landed `cgiwrap`, and the first `ln` fails on a missing **source**. From the build log:

```
for F in history.sh eventlog.sh report.sh ... ; do ln -f .../server/bin/cgiwrap .../cgi-bin/$F; done
ln: /opt/homebrew/Cellar/xymon-server/HEAD-fc39020/server/bin/cgiwrap: No such file or directory
```

One error only: by the second iteration `cgiwrap` had appeared, so names 2..26 linked fine. `history.sh` is simply the first entry of `CGISCRIPTS`.

## Why it is silent

A shell `for` returns the status of its **last** iteration. A failure on any earlier name is discarded, so `make install` exits 0 with an incomplete install. That is why this can only be noticed later, as a 404 on one CGI.

## Impact

Not Homebrew-specific — any `make -jN install` from source can hit it, non-deterministically, with no diagnostic. Which name is lost depends on how far `install-bin` has got, though the first entry is the likeliest.

## Fix

Two parts, both needed:

- declare the real dependency — `install-cgi: install-bin` — which removes the race;
- add `|| exit 1` to the four link loops, so a future failure stops the recipe instead of being swallowed.

The first alone leaves the silent-failure mode in place for any other cause; the second alone would just turn this into an intermittently failing build.

bonomani:
`install-cgi` hardlinks `$(INSTALLBINDIR)/cgiwrap`; `install-bin` is what puts cgiwrap there. Listing both under `install:` does not order them, so `make -jN` runs them concurrently.

Measured on a Homebrew build of `main`, which exports `-j` by default:

```
ln: .../server/bin/cgiwrap: No such file or directory
```

Once, on the first iteration — by the second, cgiwrap had landed. The install ended with **25 of 26** wrappers and a 404 on the history CGI. Silent, because a shell `for` returns only its last iteration's status, so `make install` exits 0. Any `make -jN install` from source can hit it, non-deterministically.

| | fix |
|---|---|
| the race | `install-cgi: install-bin` |
| the swallowed status | `\|\| exit 1` on the link loops |
| identical `ifndef PKGBUILD` / `else` branches | removed |

Two tests, because they catch different halves:

| mutation | behavioural | static |
|---|---|---|
| remove `\|\| exit 1` | fails | fails |
| remove the dependency | passes | **fails** |

The behavioural one obstructs the first CGI name and requires a non-zero exit, then checks an unobstructed run still links every name — without which a rule that always failed would pass. The static one guards the ordering, which no test reproduces deterministically.

Consolidates #358. Closes #350

### DIFF
--- a/tests/buildsystem/install-cgi-link-failure.sh
+++ b/tests/buildsystem/install-cgi-link-failure.sh
@@ -1,0 +1,65 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/buildsystem/install-cgi-link-failure.sh
+#
+# web/Makefile's install-cgi hard-links cgiwrap to every CGI name. make only
+# sees the status of a loop's last iteration, so a link that failed earlier
+# left the package short of a CGI while the build reported success -- and the
+# missing name 404s on a live server, with nothing in the build log.
+#
+# Observed on a Homebrew macOS keg: history.sh and eventlog.sh, the first two
+# names in CGISCRIPTS, were absent while the install reported success.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+ROOT=$(find_root)
+[ -f "$ROOT/web/Makefile" ] || skip "web/Makefile absent"
+require_gnu_make
+
+# "-o install-bin": install-cgi depends on install-bin, which chowns and copies
+# the built CGIs. This test is about install-cgi's own recipe and supplies its
+# own cgiwrap, so the prerequisite is declared up to date rather than run - it
+# would otherwise need a built tree and a real XYMONUSER.
+run_install_cgi() {  # run_install_cgi <workdir> -- prints nothing, returns make's status
+	local w=$1
+	"$XYMON_MAKE" -C "$ROOT/web" -o install-bin install-cgi \
+		INSTALLROOT= INSTALLBINDIR="$w/bin" \
+		CGIDIR="$w/cgi" SECURECGIDIR="$w/seccgi" >"$w/make.log" 2>&1
+}
+
+# ---- a link that cannot be made must stop the install ----------------------
+# The first CGI name is obstructed by an unwritable directory, so its ln fails
+# while every later one succeeds -- the shape that used to pass unnoticed.
+work=$(mktempdir)
+mkdir -p "$work/bin" "$work/cgi" "$work/seccgi"
+: >"$work/bin/cgiwrap"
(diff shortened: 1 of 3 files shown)


SoundGoof:
### Finding

#### Medium: The failure fixture does not work when tests run as root

`tests/buildsystem/install-cgi-link-failure.sh:41-47`

The test creates `history.sh` as a mode-500 directory and assumes the hard link
will fail. Root can still write there, so `ln` succeeds by creating
`history.sh/cgiwrap`. The test then incorrectly reports that `install-cgi`
swallowed a failure.

I reproduced this on every BSD validation guest:

- NetBSD 9.4: failed
- NetBSD 11.0: failed
- OpenBSD 7.9: failed
- FreeBSD 14.3: failed

Use an obstruction that cannot be replaced even by root:

```sh
mkdir -p "$work/cgi/$first/cgiwrap"
```

Because `history.sh` is a directory, `ln` resolves its destination to
`history.sh/cgiwrap`. Making that destination an existing directory forces the
link to fail regardless of privileges.

I tested this replacement as an unprivileged Linux user and as root on all four
BSD systems. It passes everywhere.

### Validation

- The production dependency and link-loop changes appear correct.
- Exact PR focused tests pass as an unprivileged Linux user.
- Exact behavioral test fails as root on all four BSD systems.
- Proposed fixture passes on Linux and all four BSD systems.
- Unconfigured Linux suite: 28 passed, 41 skipped, 0 failed.

SoundGoof (approved):
No blocking findings.

Both focused build-system tests passed on:

- NetBSD 9.4
- NetBSD 11.0
- OpenBSD 7.9
- FreeBSD 14.3
- Debian 13

GitHub CI is green. Local focused and mutation tests also passed.

PR #352, opened 2026-08-13
xymond_alert: FOR=N, an alerts.cfg criterion for the colour's own age (#229 step 3)

bonomani:
Step 3 of #229. `DURATION` measures the age of the alert *event*, and eventstart deliberately survives a yellow→red change — so a status yellow for two hours matches `COLOR=red DURATION>10` the instant it turns red, and "delay this colour transition" is not expressible. `delayred` is the only thing that covers it today, and it does so by rewriting the state of record, leaking the deferral into history, events and SLA.

**`FOR=N` matches once the rule's colour has held N minutes.** Same incident, seen by both:

| jaune 30 min, puis rouge… | `DURATION>10` | `FOR=10` |
|---|---|---|
| rouge depuis 2 min | alerte | silence |
| rouge depuis 12 min | alerte | alerte |

They agree while the status keeps one colour; they part only in the minutes after a colour change. In one line: **`FOR` is `DURATION` with the clock restarted on each colour change.**

No protocol change — the page channel already carries `log->lastchange`, read now into a second per-record timestamp. `colorstart` is checkpointed as a new **last** field, so a file from the previous version simply lacks it and falls back to eventstart.

**How it combines with the rule it sits under** — the same way as every other filter: a recipient narrows what the rule lets through, never widens it. Sets intersect (`COLOR`), bounds take the stricter (`DURATION`, `FOR`). A `FOR` on the rule line with `COLOR=` on the recipient is therefore accepted: the recipient names the colour.

**Two restrictions from the RFC:** no `FOR<` (damping needs a lower bound; `DURATION<` covers early-window routing), and a single effective colour — not because the meaning would be ambiguous (the colour is matched first, so there is always exactly one at match time), but to keep a later generalisation to whole-condition hold time compatible. A refused `FOR` is reported and dropped; the rest of the rule keeps working.

Evaluated in the recipient pass only. Failing in the rule pass takes the whole rule out, and with nothing matching anywhere `xymond_alert` goes A_NORECIP and discards the repeat state — losing the recovery message for an alert already sent.

| evidence | |
|---|---|
| tests | `alert-for-holdtime.sh`, `alert-test-mode-holdtime.sh`, `alert-escalation-repeat.sh`, driving the real `have_recipient()` |
| mutations that fail them | eventstart instead of colorstart · check moved to the rule pass · `int` accumulator restored · recipient loosening the rule's `FOR` · rule `FOR` judged without its recipients · the report's delay column computed the other way |
| bound | `FOR=35791394` accepted, `35791395` refused; `10s`, `10xyz`, `10h5x` refused |
| suite | 0 failed, 0 skipped |

Converting the other seventeen `durationvalue()` callers to the bounded parser is #368.

Not covered, said rather than implied: the checkpoint round-trip of colorstart, and `alert-escalation-repeat.sh` passes with `FOR` ignored — it pins the escalation, not the hold time.

Refs #229

### DIFF
--- a/docs/manpages/man5/alerts.cfg.5.html
+++ b/docs/manpages/man5/alerts.cfg.5.html
@@ -139,7 +139,34 @@ group Production databases
       given duration. E.g. DURATION&gt;1h (lasted longer than 1 hour) or
       DURATION&lt;30 (only sends alerts the first 30 minutes). The duration is
       specified as a number, optionally followed by 'm' (minutes, default), 'h'
-      (hours), 'd' (days) or 'w' (weeks).</dd>
+      (hours), 'd' (days) or 'w' (weeks). Note that this is the age of the alert
+      <i>event</i>, not of the current colour: the event begins when the status
+      first goes into an alerting colour and deliberately survives a later
+      yellow&#x2192;red change. A status that has been yellow for two hours
+      therefore matches DURATION&gt;10 the instant it turns red. Use <b>FOR</b>
+      when you mean the colour's own age.</dd>
+  <dt id="FOR"><a class="permalink" href="#FOR"><b>FOR=</b><i>time</i></a></dt>
+  <dd>Rule matching an alert only once the rule's colour has held continuously
+      for at least the given time - the counterpart of DURATION for damping
+      short-lived transitions. E.g. COLOR=red FOR=10 alerts on a red that has
+      lasted ten minutes, whether it arrived from green or from yellow. The time
+      is specified as for DURATION.
+    <p class="Pp">The effective colour - the rule's and the recipient's together
+        - must be a single one: with several, a later generalisation of FOR to
+        whole-condition hold time would not be compatible, so such a rule is
+        reported and the FOR ignored. A FOR on the rule line with COLOR= on the
+        recipient is therefore accepted: the recipient names the colour.</p>
+    <p class="Pp">Where both the rule and a recipient set FOR, the longer
+        applies, as for the other filters: a recipient narrows what the rule
+        lets through, it never widens it. There is no FOR&lt; - damping needs
+        only a lower bound, and DURATION&lt; already covers early-window
+        routing.</p>
+    <p class="Pp">A period where the test stopped reporting counts towards the
+        hold time only if the status comes back the same colour, and only up to
+        four times the report validity that applied when the silence began (two
+        hours by default). A test absent for a month therefore does not return
+        claiming a month-long hold time.</p>
+  </dd>
   <dt id="RECOVERED"><a class="permalink" href="#RECOVERED"><b>RECOVERED</b></a></dt>
   <dd>Send recovery and disabled-state messages to matching recipients. This can
       be set on a rule or on an individual recipient.</dd>
--- a/lib/loadalerts.c
+++ b/lib/loadalerts.c
@@ -51,6 +51,7 @@ static rule_t *ruletail = NULL;
 static int cfid = 0;
 static char cfline[256];
 static int printmode = 0;
+static int ignoreholdtime = 0;
 static rule_t *printrule = NULL;
 
 static enum { P_NONE, P_RULE, P_RECIP } pstate = P_NONE;
@@ -146,9 +147,66 @@ static char *preprocess(char *buf)
 	return STRBUF(result);
 }
 
+/*
+ * FOR=N only has one meaning when the rule names exactly one color: red for 3
+ * then yellow for 8 satisfies neither for 10, yet the rule matched throughout.
+ *
+ * Bit 30 marks "an explicit COLOR= was given" and is not a color itself; with
+ * no COLOR= the effective set is the default one, never a single color.
+ */
+static void check_holdtime_colors(criteria_t *crit, int cfid, int rulecolors)
+{
+	int colors, explicitcolors, n, i;
+
+	if (!crit || !crit->minholdtime) return;
+
+	/*
+	 * Count what the runtime matches on: rule and recipient are matched in
+	 * turn, so the effective set is their intersection, with the default set
+	 * standing in for a side that names no COLOR=. print_alert_recipients()
+	 * builds it the same way.
+	 */
+	explicitcolors = ((crit->colors | rulecolors) & (1 << 30));
+	colors = ((crit->colors  ? crit->colors  : defaultcolors) &
+		  (rulecolors    ? rulecolors    : defaultcolors)) & ~(1 << 30);
+	for (i = n = 0; (i < COL_COUNT); i++) if (colors & (1 << i)) n++;
+	if (n == 1) return;
+
+	if (!explicitcolors)
+		errprintf("Ignoring FOR at line %d: it needs an explicit single COLOR=\n", cfid);
+	else if (colors == 0)
+		errprintf("Ignoring FOR at line %d: its COLOR= has nothing in common with the rule's\n", cfid);
+	else
+		errprintf("Ignoring FOR at line %d: it needs a single COLOR=, not %d\n", cfid, n);
+	crit->minholdtime = 0;
+}
+
 static void flush_rule(rule_t *currule)
 {
+	recip_t *rwalk;
+	int rulecolors;
+
 	if (currule == NULL) return;
+
+	rulecolors = (currule->criteria ? currule->criteria->colors : 0);
+
+	/*
+	 * The rule's own FOR is judged against each recipient, not against the
+	 * rule line alone: a recipient's COLOR= narrows the effective set, so
+	 * "FOR=10" on the rule with "MAIL x COLOR=red" under it names exactly one
+	 * colour at match time. Judged in isolation it named none, and the FOR was
+	 * dropped - the alert then went out immediately, which is the opposite of
+	 * what was asked for.
+	 */
+	for (rwalk = currule->recipients; (rwalk); rwalk = rwalk->next)
+		check_holdtime_colors(currule->criteria, currule->cfid,
+				      (rwalk->criteria && rwalk->criteria->colors) ?
+					rwalk->criteria->colors : defaultcolors);
+
+	if (currule->recipients == NULL) check_holdtime_colors(currule->criteria, currule->cfid, 0);
+
+	for (rwalk = currule->recipients; (rwalk); rwalk = rwalk->next)
+		check_holdtime_colors(rwalk->criteria, rwalk->cfid, rulecolors);
 
 	currule->next = NULL;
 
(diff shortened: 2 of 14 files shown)


bonomani:
**Two claims below were superseded by later rounds and are corrected in place: the recipient's `FOR` no longer wins over the rule's — the longer of the two applies, like every other filter — and the overflow bound moved onto the minutes.**

Rewritten in `aef2acf65`. The first two findings were right and they invalidated the design, not just the code.

### FOR belonged in the recipient pass, not the rule pass

`criteriamatch()` answers two different questions. In the rule pass, "no" means *this rule is not about this alert* — so the rule drops out, `next_recipient()` reports nothing matched, and `xymond_alert.c:941-948` sets `A_NORECIP` and calls `cleanup_alert()`, discarding the repeat state and with it the recovery message for an alert already sent. "Not due yet" is not "not concerned". The rule pass also passes no `nexttime` (`loadalerts.c:1147,1172`), so the threshold could never wake the daemon at the right moment.

It is now checked only when `rulecrit` is set, exactly as `NOTICE` and `RECOVERED` already are — the pattern was in the file and I had not used it. A `FOR=` on the rule line applies to every recipient under it.

### The test could not have caught it

It asked "is a mail due now?", and the answer is "no" both before and after — it passed either way. It now asserts `anymatch` alongside the verdict, which is precisely what separates the two cases. Verified by mutation in both directions: with the check moved back into the rule pass it prints `anymatch=0` and the assertion names it.

### Also fixed

| finding | |
|---|---|
| recipient `FOR` rejected without its own `COLOR=` | inherits the colour its rule pins |
| `FOR` evaluated before the colour check | moved below it |
| restored `colorstart == 0` | falls back to `eventstart` |
| `60*durationvalue()` overflow | *(superseded: the wrap happens inside `durationvalue()`, so the bound moved onto the minutes)* |
| config report ignored `FOR` | counted in the delay column |
| `time_t` through `%d` | cast at both neighbouring sites |
| unasserted `got` in the test | asserted |
| two `getcurrenttime()` calls | one clock per decision |

### Left out, deliberately

The `COLOR=` `reverse` flag is a real pre-existing bug — a `!` makes every later colour subtractive — but it predates this work and deserves its own fix: #353.

The duplication between the `FOR` and `minduration` blocks stands. They now differ in guard, base timestamp and fallback, so folding them into one helper would take more arguments than it saves lines.

Suite: 68 passed, 0 skipped, 0 failed.

SoundGoof:
### Finding: Reject overflow before `durationvalue()` performs `int` arithmetic

[lib/loadalerts.c:574](lib/loadalerts.c#L574) widens the result of
`durationvalue()`, but that function has already parsed and multiplied each
component using signed `int`.

For example, `FOR=426089w` computes `426089 * 10080` inside
`durationvalue()`. On the tested platform this wraps to 9824 minutes. The later
`60LL` multiplication therefore accepts it as 589440 seconds, about 6.8 days,
instead of rejecting the multi-millennial duration.

I confirmed this through the production-linked harness: an alert held for
600000 seconds returned:

    MATCH anymatch=1

No invalid-FOR diagnostic was emitted.

Parse `FOR` with checked wide arithmetic before narrowing to `int`, or extend
the shared duration parser with an overflow-reporting interface. Add this case
to the regression so the intended `INT_MAX` rejection cannot be bypassed.

bonomani:
Right, and the reason is one level below where I was looking: `durationvalue()` accumulates minutes in an `int`, so the wrap happens before my `60LL *` ever sees the value. Measured — `FOR=426089w` was not refused, it was **accepted as 6.8 days** (`durationvalue()` returns 9824).

I did not take the obvious fix. Saturating `durationvalue()` at `INT_MAX` would have been worse than the wrap: eight of its seventeen other call sites compute `60*durationvalue()` in an `int`, and `60 * INT_MAX` is exactly `-60` — a negative repeat interval in `loadalerts.c`, an expiry set in the past at `xymond.c:2300`, a negative `recentgif_limit`. The overflow has two independent sites, the parse and the unit conversion, and saturating closes only the first.

So: the parse moves into a shared `long long` helper, `durationvalue()`'s contract is untouched, and a second entry point sits beside it —

```c
int durationseconds(char *dur, int *secs);   /* 1 and stores, or 0 = refused */
```

— bounding the **minutes** at `INT_MAX/60`, which closes both overflows at once. `FOR=` is its only caller.

`durationvalue()` is measurably unchanged: over 20 durations (plain, every modifier, multi-component, empty, non-numeric, negative, four overflowing) old and new agree on 19. The 20th is a component too wide for an `int`, where the old `atoi()` was undefined and returned `-727379969`; it now returns 0.

Test takes your input rather than merely a large one, plus both sides of the bound — `FOR=35791394` accepted and applied, `FOR=35791395` refused. Restoring the `int` accumulator fails it.

Converting the other seventeen is #368, not this PR: three read network input and two read mail, where a refusal must not drop the whole message, so each needs its own decision and its own test.

SoundGoof:
## Blocking finding

`lib/loadalerts.c:162-164` validates a recipient's raw `COLOR=` set instead of
its effective intersection with the parent rule.

For example:

    HOST=testhost COLOR=red
        MAIL oncall@example.com COLOR=red,yellow FOR=10

Runtime matching requires both criteria, so the effective colour is only red.
`FOR=10` therefore has an unambiguous referent. However,
`check_holdtime_colors()` counts the recipient's two raw colours, reports:

    Ignoring FOR at line 2: it needs a single COLOR=, not 2

and clears `minholdtime`. The alert is then sent immediately instead of being
deferred.

Combine the rule and recipient colour sets before counting them. Add this
overlapping-set case to `alert-for-holdtime.sh`.

## Validation

Reviewed exact head `bf3dfbd8556a9e208069a259bf2483b9346055ef`.

The submitted focused test passes. The added effective-colour probe reproduces
the finding. The full suite reports 71 passed, 3 skipped, and 0 failed.

bonomani:
Fixed and amended (`460ea3b10`).

The runtime matches rule and recipient in turn, so the effective colour is their intersection
-- `print_alert_recipients()` already built it that way (`loadalerts.c:1301-1302`), the check did
not. It now does, defaulting a side that names no `COLOR=`, so the existing verdicts stand.

An empty intersection gets its own message rather than "needs an explicit single COLOR=": that
recipient can never fire at all.

Three cases added. Verified to bite: reverted, the first fails with the line you quoted.

bonomani:
Second finding, amended (`6ac0f7b23`) -- found by an independent review pass.

`clear_interval()` and `send_alert()` shared one walk, but only one of them sends. With `FOR=`
on the red rule, escalation returned nobody, so the yellow's repeat record survived and the red
was dropped until it came due -- 45 minutes late in the example. Records are keyed
`hostname|testname|method|recipient` (`do_alert.c:125`), so one address under both rules shares
one.

`clear_interval()` now looks past the hold time; the other two callers still honour it.
`alert-escalation-repeat.sh` drives the repeat list directly: `kept` without the change,
`cleared` with it, plus a control case without `FOR=`.

bonomani:
Third finding, amended (`286246098`).

`--test` left `colorstart` at the calloc'd zero while the other three constructors set it, so
`now - 1970` made every `FOR=` look satisfied -- the mode that exists to check a configuration
recommended alerts the daemon would hold back.

Measured on the built binary, `COLOR=red FOR=10`: `--duration=5` now reports
`min. hold time 300<600` and recommends nothing, `--duration=15` recommends the mail; without
the change, 5 minutes recommends the mail. `alert-test-mode-holdtime.sh` guards it.

bonomani:
Fourth finding, amended (`545565129`). `FOR=10x` loaded as ten minutes; `durationseconds()`
now refuses a leftover unit and the `FOR=` site reports the line.

Scoped there on purpose: it has one caller, this one. `DURATION` and the other sites keep
`durationvalue()` and its installed base -- tightening that would reject configs that load
today. Noted in #372.

All four findings closed. Suite: 56 passed, 20 skipped, 0 failed; each fix verified to bite.

bonomani:
Recording what changed here this morning, since the branch was rewritten and the thread does not say so.

Two findings I reported above without fixing are now fixed, and I re-ran both:

| | |
|---|---|
| `FOR=10s`, `10xyz`, `10h5x` | were accepted as 10 minutes — `durationminutes()` parses up to the first character it does not know, so an unknown unit read as a delimiter. Refused now, and `durationvalue()` deliberately keeps its lax parsing for the criteria that already use it |
| `xymond_alert --test` | never set `colorstart`, so the hold time was measured from the epoch and `FOR` always matched. `tests/server/alert-test-mode-holdtime.sh` pins it |

Plus `alert-escalation-repeat.sh`: a severity increase clears the previous colour's repeat interval, with or without `FOR`.

All three alert tests pass here, and the full suite is 74 passed, 0 skipped, 0 failed with `rrdtool` absent from `PATH`. One commit, no lag on `main`, five checks green.

A deliberate verification pass over the current state is running; I will post what it finds either way. The earlier codex review of this PR was accidental — it was aimed at another one — so this branch has not yet been checked in the shape it is in now.

bonomani:
Two corrections to how `FOR` combines with the rule it sits under, both from measuring it against the other criteria rather than against itself:

```
rule DURATION>60 + recipient DURATION>10, 15 min   NOMATCH   both apply
rule COLOR=red   + recipient COLOR=red,yellow      red only  sets intersect
rule FOR=60      + recipient FOR=10,      15 min   MATCH     <- the exception
```

**A recipient could loosen it.** Every other filter is checked in both passes, so both values must be satisfied. `FOR` is checked in the recipient pass alone, and taking the recipient's value there made it the one criterion a recipient could relax. The longer of the two applies now — which is what "both are satisfied" means for a lower bound.

**The rule's own `FOR` was judged against the rule line alone**, so

```
HOST=db1 SERVICE=conn FOR=10
        MAIL oncall@example.com COLOR=red
```

lost its damping and alerted at once: the recipient names the colour, and the effective set is single. `NOMATCH` at 5 minutes and `MATCH` at 15 now. The RFC's two refusals are unchanged — a multi-colour rule, and a `FOR` with no colour anywhere.

**The manual gave the wrong reason for that restriction**, and it was mine. Not "with several colours the meaning is ambiguous": the colour is matched *before* the hold time, so at match time there is always exactly one, and a multi-colour rule works correctly if you let it. The reason is #229's — keeping a later generalisation to whole-condition hold time compatible. That is what the page says now.

**The config report then disagreed with the engine**, which the first correction caused: `print_alert_recipients()` computes the delay column a second time, on its own, and still took the recipient's value there. `FOR=60` on the rule with `FOR=10` on the recipient held the alert an hour while `svcstatus.cgi` and `confreport.cgi` told the operator "first alert after 10m". It takes the larger there too now, as the `minduration` line above it already did.

All three fail the test when reverted; the report case is asserted on the rendered column, not on the computation. Suite on a built tree: 76 passed, 0 skipped, 0 failed.

*(Correcting an earlier comment of mine, now removed: it claimed this thread did not record the morning's changes. It does — the four comments above name each fix with its commit. I had not read them.)*

SoundGoof (approved):
No findings.

Reviewed exact head `77216e638c01f08374e6f5a32a3e625f82a526e4`.

Additional validation:

- Real `xymond` → page channel → `xymond_alert` → script recipient path
- `FOR=1` withheld delivery before the threshold
- Restart preserved the pending hold time
- Delivery occurred after the checkpointed colour age passed the threshold
- Old-format checkpoint falls back to `eventstart`
- Empty and negative `colorstart` fall back safely
- Valid `colorstart` survives restoration
- Extra fields do not alter the restored value
- Truncated records are ignored
- Oversized timestamps fall back safely

Native tests passed on:

- OpenBSD 7.9
- FreeBSD 14.3
- NetBSD 11.0

Each BSD passed all seven focused tests, including real-daemon checkpoint,
gap, disable, DOWNTIME, colour-resume, escalation, and `FOR` behavior.

Earlier validation also passed:

- Five GitHub checks
- Full suite: 73 passed, 3 skipped, 0 failed
- ASan and UBSan focused tests
- Clean merge and `git diff --check`

Approved.

PR #354, opened 2026-08-13
client_config: do not name pcre2 error codes older headers lack

bonomani:
`#255`'s whitespace hint is gated on two pcre2 compile-error codes. Naming them stops the build outright where the headers do not declare them:

```
client_config.c:434:29: error: 'PCRE2_ERROR_MISSING_CLOSING_PARENTHESIS' undeclared (first use in this function)
client_config.c:435:22: error: 'PCRE2_ERROR_MISSING_SQUARE_BRACKET' undeclared (first use in this function);
                              did you mean 'PCRE2_ERROR_REPMISSINGBRACE'?
```

**CentOS 7 and Amazon Linux 2 no longer build the server.** Both lanes fail at compile, not at a test. This is a regression from #255, which is mine.

The gate is now compiled only where the codes can be named. The pattern and its config line — the substance of #253 — are reported either way; only the hint is lost.

Their numeric values are deliberately **not** hard-coded in their place. Nothing available here can confirm what those older headers assign 106 and 114 to, and a hint pointing at whitespace for a failure that has nothing to do with whitespace is worse than no hint — which is precisely the regression the delimiter-counting version had, and what the error-code gate replaced.

The test asserted the hint unconditionally, so it would have begun failing on those platforms the moment they built again — over a platform limit rather than a regression. It now probes whether the codes can be named and asserts the matching behaviour. Every other assertion is untouched: the config line, the two patterns that must **not** be hinted, and the valid `[[:space:]]` pattern that must stay silent.

Verified both sides by compiling with the gate forced off and the probe forced negative: the build is clean and the test passes on the no-hint branch.

### DIFF
--- a/tests/server/analysis-pattern-error-context.sh
+++ b/tests/server/analysis-pattern-error-context.sh
@@ -79,9 +79,30 @@ assert_contains "at line 2" "$quant_line" \
 assert_contains "at line 3" "$class_line" \
 	"invalid pattern on line 3 is not reported with its config line"
 
+
+# Whether this build could name the two pcre2 error codes the hint is gated on.
+# Both are absent from the pcre2 headers on some older distributions, where the
+# hint is compiled out on purpose -- asserting it there would be failing over a
+# platform limit rather than a regression. Probed with the compiler when there
+# is one; where there is not (an installed-package run), the names are assumed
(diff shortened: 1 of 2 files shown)


SoundGoof (approved):
lgtm

PR #355, opened 2026-08-13
xymond: let xymondboard select oldcolor, and add previouscolor (#229 step 4)

bonomani:
**Rewritten.** This PR documented `oldcolor` as a `xymondboard` field; it is not one, so it now makes it one, adds the second color, and corrects two claims the old text made. **1 commit.**

`boardfieldnames[]` has no entry for `oldcolor`, so `fields=hostname,testname,color,oldcolor` returns an empty column, and `xymondlog`'s default list does not carry it either. It has been on xymond's internal channels forever, which is where the modules see it.

Both are selectable now, under names that say when each moves:

| field | assigned | after green, red, red |
|---|---|---|
| `prevreportcolor` | every message | `red` |
| `prevchangecolor` | only on a real change | `green` |

`oldcolor` and `previouscolor` are accepted as aliases, the way `matchedtag` and `matchedtags` share a field: they are the names xymond uses internally and the ones the first consumer already queries (axians/xymon#10), so that dashboard keeps working unchanged.

The canonical names exist because `oldcolor` already means the *other* thing one layer down — in `lib/eventlog.h` it is the color before a transition, which is what `prevchangecolor` holds here. A consumer reading both `allevents` and the board meets the same word with two meanings; no wording in the manual repairs that.

**Checkpoint.** `prevchangecolor` gets a record of its own, `@@XYMONDCHK-V1|.prevchangecolor.|host|test|color`, following axians/xymon#10 rather than riding `.flapstate.`. That record is read by position, so a field appended to it depends on its write condition and on every reader counting to the same place; an unrecognised `.name.` record is skipped instead. A checkpoint without the record restores as "no previous color", which is what those statuses had — and the restore path sets that default explicitly, next to `pregapcolor`: `calloc` leaves 0, which is `COL_GREEN`, so an older file would otherwise restore every test as "was green before" a change it never had.

| mutation | caught by |
|---|---|
| `prevchangecolor` assigned on every message | `a repeated color moves prevreportcolor, not prevchangecolor: expected none, got green` |
| the restore handler dropped | `a restored status keeps the color it held before the change: expected red, got none` |
| the restore default dropped | `a status restored without color history must not invent one: expected none, got green` |

**The manual page drops two claims the code contradicts**, both from the text approved earlier:

- "a status whose color has never changed reports `none`" — only its *first* report does. Send green twice and the value is green, though nothing ever changed.
- "the color this test reported" — it is the color xymond *recorded*. A disabled or in-downtime test reports green and is recorded blue.

`xymond-checkpoint-roundtrip.sh` now backdates the flap ring by naming field 11 rather than `$NF`. Nothing forces that any more, but `$NF` only worked while the ring was last, and it failed silently when it was not — the edit landed elsewhere and still left a line ending in the value written, so the guard passed over an edit that changed nothing.

Suite: 76 passed, 0 skipped, 0 failed on a built tree. Manual page HTML regenerated with `build/makehtml.sh`.

### DIFF
--- a/common/xymon.1
+++ b/common/xymon.1
@@ -203,6 +203,20 @@ The name of the host
 The name of the test
 .IP "\fBcolor\fR"
 Status color (green, yellow, red, blue, clear, purple)
+.IP "\fBprevreportcolor\fR"
+The color xymond recorded at the previous report, whether or not it differed:
+a test repeating its color reports that same color here. Also accepted as
+\fBoldcolor\fR, the name it has inside xymond. It is the color xymond
+\fIrecorded\fR, which is not always the one the test sent \- a disabled or
+in-downtime test is recorded blue whatever it reports. \fBnone\fR until a
+status has been reported twice.
+.IP "\fBprevchangecolor\fR"
+The color this test held before its most recent change. Also accepted as
+\fBpreviouscolor\fR. It moves only when the
+color changes, so a test that went green, red, red reports color=red,
+prevreportcolor=red, prevchangecolor=green \- it answers "what was this before
+it went bad". \fBnone\fR until the first change, and it survives a xymond
+restart.
 .IP "\fBtestflags\fR"
 For network tests, the flags indicating details about the test (used by xymongen).
 .IP "\fBlastchange\fR"
@@ -320,6 +334,20 @@ The name of the host
 The name of the test
 .IP "\fBcolor\fR"
 Status color (green, yellow, red, blue, clear, purple)
+.IP "\fBprevreportcolor\fR"
+The color xymond recorded at the previous report, whether or not it differed:
+a test repeating its color reports that same color here. Also accepted as
+\fBoldcolor\fR, the name it has inside xymond. It is the color xymond
+\fIrecorded\fR, which is not always the one the test sent \- a disabled or
+in-downtime test is recorded blue whatever it reports. \fBnone\fR until a
+status has been reported twice.
+.IP "\fBprevchangecolor\fR"
+The color this test held before its most recent change. Also accepted as
+\fBpreviouscolor\fR. It moves only when the
+color changes, so a test that went green, red, red reports color=red,
+prevreportcolor=red, prevchangecolor=green \- it answers "what was this before
+it went bad". \fBnone\fR until the first change, and it survives a xymond
+restart.
 .IP "\fBflags\fR"
 For network tests, the flags indicating details about the test (used by xymongen).
 .IP "\fBlastchange\fR"
--- a/docs/manpages/man1/xymon.1.html
+++ b/docs/manpages/man1/xymon.1.html
@@ -248,6 +248,20 @@ Section: User Commands (1)<br/>Updated: Version 4.3.30: 4 Sep 2019<br/><a href="
   <dd>The name of the test</dd>
   <dt id="color"><a class="permalink" href="#color"><b>color</b></a></dt>
   <dd>Status color (green, yellow, red, blue, clear, purple)</dd>
+  <dt id="prevreportcolor"><a class="permalink" href="#prevreportcolor"><b>prevreportcolor</b></a></dt>
+  <dd>The color xymond recorded at the previous report, whether or not it
+      differed: a test repeating its color reports that same color here. Also
+      accepted as <b>oldcolor</b>, the name it has inside xymond. It is the
+      color xymond <i>recorded</i>, which is not always the one the test sent -
+      a disabled or in-downtime test is recorded blue whatever it reports.
+      <b>none</b> until a status has been reported twice.</dd>
+  <dt id="prevchangecolor"><a class="permalink" href="#prevchangecolor"><b>prevchangecolor</b></a></dt>
+  <dd>The color this test held before its most recent change. Also accepted as
+      <b>previouscolor</b>. It moves only when the color changes, so a test that
+      went green, red, red reports color=red, prevreportcolor=red,
+      prevchangecolor=green - it answers &quot;what was this before it went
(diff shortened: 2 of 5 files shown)


SoundGoof (dismissed):
No blocking findings.

The enum, field-name lookup, renderer, documentation, and checkpoint behavior
are consistent. The documentation also accurately explains that `oldcolor`
means the previous report's color, not the color held before `lastchange`.

- The exact-head server build and live-daemon regression pass.
- The clean merge with current `main` builds and passes the same regression.
- The PR merges cleanly with current `main`.
- The test covers the initial `none` value, successive transitions, and a
  repeated color report.

Non-blocking note: `git diff --check` reports trailing whitespace on the edited
enum line, but that whitespace was already present on the base line.

bonomani:
Held as a draft, approval and CI notwithstanding: the reason it was written no longer holds.

This is step 4 of #229, whose only purpose was to unblock step 5 (`viewdelay`), which renders "the previous recorded colour" during a damping window. Implementing it showed `oldcolor` cannot serve that. `xymond.c:1917` assigns it on **every message**, not on a change, so it is the previous *report's* colour — a test reporting every five minutes overwrites the value the window needs at the first repeat. Measured, and raised on #229.

What remains is weaker: the field is already in `log_t` and on the status and page channels, no `xymondboard` query could ask for it, and a consumer had to read the event log instead — a second round trip that can disagree with the board it decorates. True, but nobody has asked.

The reason to wait rather than merge anyway: if step 5 adds a real `prevcolor`, assigned only on a change, `xymondboard` would carry **two fields with near-identical names and subtly different meanings**. That is the kind of duplicate a published API cannot take back. Settling `prevcolor` first decides whether to expose one, the other, or both under names that tell them apart.

The documentation here stands on its own either way — that `oldcolor` reads `none` until a status has changed once, and that a repeated report overwrites it. Neither was written down anywhere before, and both are traps for anyone diffing colour against oldcolor to detect a transition.

SoundGoof:
I would like `xymondboard` to expose both `oldcolor` and
`previouscolor`.

They have different semantics:

- `oldcolor` is the color from the previous report.
- `previouscolor` is the color held before the most recent actual color
  transition.

After a green-to-red transition followed by another red report, the values
should therefore be:

    color=red
    oldcolor=red
    previouscolor=green

We have an internal Perl `xymonlive` application that tails `allevents` and
uses the old and new colors from each event to display transitions.

The replacement web implementation is available in
https://github.com/axians/xymon/pull/10. It reads `xymondboard` directly
instead of reading history logfiles.

This is also relevant to
https://github.com/xymon-monitoring/xymon/pull/174, which changes
`allevents` from line-buffered output to fully buffered output flushed on
idle. A logfile consumer may consequently receive events up to ten seconds
later.

Providing both fields through `xymondboard` gives consumers a current,
consistent snapshot without depending on the timing or availability of
`allevents`. `oldcolor` alone cannot preserve the transition because it is
overwritten by repeated reports.

The implementation in the web PR demonstrates both fields and includes a
live regression test covering repeated reports, actual transitions, and
checkpoint restart behavior.

bonomani:
@SoundGoof — this is no longer the PR you approved, so it needs another look.

You asked for `previouscolor` beside `oldcolor`, with the semantics you spelled out. Checking that against the code turned up the reason the two were easy to confuse: **`oldcolor` was never a `xymondboard` field.** `boardfieldnames[]` has no entry for it, so `fields=hostname,testname,color,oldcolor` returns an empty column, and `xymondlog`'s default list does not carry it either. It lives on xymond's internal channels, where the modules see it. This PR was about to document it as selectable.

So both are selectable now, and `previouscolor` is the second one:

| | assigned | after green, red, red |
|---|---|---|
| `oldcolor` | every message | `red` |
| `previouscolor` | only on a real change | `green` |

`previouscolor` rides the `.flapstate.` checkpoint record, appended after the flap ring so no position shifts. Two things came out of building it that are worth your eye:

- The restore path fills the field from a `calloc`ed struct, where 0 is `COL_GREEN` — so a checkpoint written before this field existed restored every test as "was green before" a change it never had. It now gets `NO_COLOR` explicitly, next to `pregapcolor`, which guards the same way for the same reason. The test covers that upgrade path by dropping the sidecar record from a checkpoint.
- `xymond-checkpoint-roundtrip.sh` backdated the flap ring through `$NF`. With a field appended after the ring, `$NF` rewrites the new one instead — and still leaves a line ending in the value written, so its own guard passed while the edit changed nothing. It names field 11 now.

Two claims from the text you approved are gone, because the code contradicts them: that a status whose color never changed reports `none` (only its first report does — send green twice and `oldcolor` is green), and that the value is the color the test *reported* (it is the color xymond *recorded*; a disabled or in-downtime test reports green and is recorded blue).

Suite: 76 passed, 0 skipped, 0 failed on a built tree. Manual page HTML regenerated with `build/makehtml.sh`.

SoundGoof (approved):
No findings.

The canonical fields and compatibility aliases have the intended semantics:

- `prevreportcolor` / `oldcolor` moves on every report.
- `prevchangecolor` / `previouscolor` moves only on a real transition.
- The change-only value survives checkpoint restart.
- Older checkpoints correctly restore it as `none`.

Validation on head `23c51892a4b3dd26f077bdd23861c027f956e95f`:

- The PR merges cleanly with current `main`.
- Exact-head build and live regressions pass.
- Exact-head suite: 73 passed, 3 skipped, 0 failed.
- Current-main merge suite: 76 passed, 3 skipped, 0 failed.
- Mutations of repeated-report and checkpoint behavior were rejected.
- Direct `xymondlog` checks confirmed both compatibility aliases.

Approved.

PR #356, opened 2026-08-13
tests: give every harness that links the tree its configured flags

bonomani:
Follow-on to #346, which added `xymon_ldflags()` and used it for the svcstatus scaffolding. Four harnesses still scrape `SSLLIBS` out of the Makefile by hand, or pass no link flags at all, which leaves out the library search path and the runtime path.

On NetBSD that is the failure @SoundGoof reported against #204: the harness links and then cannot run —

```
Shared object "libpcre2-8.so.0" not found
```

— because pkgsrc puts it under `/usr/pkg/lib` and nothing records an rpath.

| Harness | Was |
|---|---|
| `xymond-gap-validity.sh` | hand-scraped `SSLLIBS` |
| `xymond-holdtime-bridge.sh` | hand-scraped `SSLLIBS` |
| `xymond-flap-purple-pin.sh` | hand-scraped `SSLLIBS` |
| `cgi-split-hostsvc.sh` | no compile or link flags at all |

**Four is the whole set**, and that is worth stating because the raw count looks alarming: of the thirty tests under `tests/` that invoke a compiler, fifteen do not call `xymon_ldflags()`. Eleven of those link *nothing* — they compile a self-contained snippet, so a search path would have nothing to find. The twelfth, `inode-unix-columns.sh`, asks make for `LDFLAGS`, `RPATHOPT` and `RRDLIBS` itself and already carries what it needs.

### DIFF
--- a/tests/server/xymond-flap-purple-pin.sh
+++ b/tests/server/xymond-flap-purple-pin.sh
@@ -38,7 +38,11 @@ require_cc
 
 work=$(mktempdir)
 
-ssllibs=$(sed -n 's/^SSLLIBS *= *//p' "$ROOT/Makefile")
+# The configured link flags rather than a hand-picked SSLLIBS: they carry the
+# library search path and the runtime path as well. Without the rpath the
+# harness links on NetBSD and then cannot run -- "Shared object
+# libpcre2-8.so.0 not found", since pkgsrc puts it under /usr/pkg/lib.
+harness_ldflags=$(xymon_ldflags "$ROOT")
 
 {
 	printf '#include <stdio.h>\n#include <stdlib.h>\n#include <string.h>\n'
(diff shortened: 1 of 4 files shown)


SoundGoof (approved):
No findings.

The remaining tree-linking test harnesses now consistently use the configured
compile, library-search, and runtime-path flags. Archive and dependency ordering
remain correct.

Validation:
- Linux: 67 passed, 0 skipped, 0 failed.
- FreeBSD 14.3: 67 passed, 0 skipped, 0 failed.
- NetBSD 9.4: 67 passed, 0 skipped, 0 failed.
- NetBSD 11.0: 67 passed, 0 skipped, 0 failed.
- OpenBSD 7.9: 65 passed, 2 unsupported-ASAN skips, 0 failed.
- All GitHub checks pass.
- `git diff --check` passes.
- Fresh builds were performed without reusing previous build products.
- The PR head remained `7cc794be` after a final post-test fetch.

The NetBSD runs directly exercise the `/usr/pkg/lib` library-search and
runtime-path behavior addressed by this change.

PR #358, opened 2026-08-14
web: stop install-cgi when a CGI hard link cannot be made

bonomani:
`install-cgi` links `cgiwrap` to every CGI name in a shell loop. make only sees the status of the **last** iteration, so a link that fails earlier leaves the package short of a CGI while the build reports success — and the missing name 404s on a live server with nothing in the build log.

Observed on a Homebrew macOS keg: `history.sh` and `eventlog.sh` — the first two entries in `CGISCRIPTS` — were absent from an installed tree whose build had reported success. The history and eventlog buttons returned `script not found or unable to stat`.

Demonstrated rather than assumed. A loop whose early iterations fail and whose last one succeeds:

```
statut make     : 0
erreurs ln      : 2
liens réguliers : 2 sur 4
```

make reports success, and prints nothing of its own.

The `PKGBUILD` branches this rule carried were byte-identical, so there was nothing to select between; they go with the fix.

New test `tests/buildsystem/install-cgi-link-failure.sh` obstructs the first `CGISCRIPTS` entry so its `ln` fails while every later one succeeds, and requires make to fail. A second case links a clean destination and checks every name from both lists arrives, so the first cannot be satisfied by a rule that always fails. Without `|| exit 1` the first case reports:

```
FAIL: install-cgi reported success although the link for 'history.sh' could not be made
```

Full suite: 69 passed, 0 failed, 0 skipped.

### DIFF
--- a/tests/buildsystem/install-cgi-link-failure.sh
+++ b/tests/buildsystem/install-cgi-link-failure.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/buildsystem/install-cgi-link-failure.sh
+#
+# web/Makefile's install-cgi hard-links cgiwrap to every CGI name. make only
+# sees the status of a loop's last iteration, so a link that failed earlier
+# left the package short of a CGI while the build reported success -- and the
+# missing name 404s on a live server, with nothing in the build log.
+#
+# Observed on a Homebrew macOS keg: history.sh and eventlog.sh, the first two
+# names in CGISCRIPTS, were absent while the install reported success.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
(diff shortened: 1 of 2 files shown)


bonomani:
Folded into #351, open on the same rule since yesterday. Your `|| exit 1`, your removal of the byte-identical `ifndef PKGBUILD` branches, and your behavioural test all came across — the test replaced #351's static assertion, being the better of the two.

What #351 adds is the cause. The build log shows the **source** missing, not the destination: `install-cgi` had begun before `install-bin` copied cgiwrap, because `install: install-bin install-cgi install-man` does not order them. Without `install-cgi: install-bin`, `|| exit 1` alone turns a silent 404 into an intermittently failing build.

One thing only the consolidation revealed: **the dependency breaks the test as written here.** It drives `install-cgi` in isolation, so make then also runs `install-bin`, whose `chown $(XYMONUSER)` fails on an empty user. #351 declares the prerequisite up to date with `-o install-bin`. Merged separately, these two would have turned CI red.

PR #359, opened 2026-08-14
tests: check the suite against the portability rules it keeps breaking

bonomani:
Four classes of failure reached a BSD or macOS runner this month after passing on Linux:

| Class | Cost |
|---|---|
| bash 4 builtins — macOS ships 3.2 | `mapfile` took out a whole macOS lane (#339) |
| GNU-only tool flags | `grep --include`, `sed -i`, `stat -c` (#320, #336) |
| interpreters absent from the BSD lanes | caught by hand in review, not by CI |
| hand-rolled compile/link flags | `pcre2.h: file not found`; and with no rpath, a harness that links and then cannot load libpcre2 (#346, #356) |

All four are decidable by reading the scripts. This decides them on every platform instead of two slow lanes later. `tests/README.md` already asks for portability — nothing enforced it, and `tests/buildsystem/` had no test over the suite itself.

Two details that were not obvious until it ran:

**Comment lines are skipped.** Several tests explain in a comment *why* they avoid a construct (`# No mapfile here: it is a bash 4 builtin and macOS ships bash 3.2`). The first run flagged three files for documenting good practice.

**The file excludes itself.** It names the forbidden constructs, so it cannot help matching them — its first run failed on its own non-vacuity probe. What keeps the rules honest instead is that probe, which requires each pattern to still match the construct it describes; and each rule was verified against a deliberate violation:

```
  bash4        détecté=1
  gnu-only     détecté=1
  interpreter  détecté=1
  link-flags   détecté=2
```

The suite is clean under all four rules today, so this lands green and guards forward. It does not replace the BSD lanes — it catches the known classes, not the unknown ones.

Full suite: 69 passed, 0 failed, 0 skipped.

### DIFF
--- a/tests/buildsystem/test-suite-portability.sh
+++ b/tests/buildsystem/test-suite-portability.sh
@@ -1,0 +1,125 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/buildsystem/test-suite-portability.sh
+#
+# A test over the tests. Every rule here is a failure the suite actually had,
+# each found by a BSD or macOS runner after passing on Linux -- which is the
+# expensive way to learn it. They are cheap to check statically, so they are
+# checked here, on every platform, before the slow lanes get a chance.
+#
+# Comment lines are skipped: several tests explain in a comment why they avoid
+# a construct, and flagging that would punish the documentation.
+
+set -euo pipefail
+# shellcheck source=tests/lib/assert.sh
+. "$(dirname "$0")/../lib/assert.sh"
+
+ROOT=$(find_root)
+[ -d "$ROOT/tests" ] || skip "no tests/ directory in this checkout"
+
+work=$(mktempdir)
+: >"$work/violations"
+
+# This file names the forbidden constructs, so it would flag itself. Its own
+# non-vacuity probe below is what keeps the rules honest instead.
+# Every shell script under tests/, not just the *.sh ones: the runner itself
+# carries no extension, and it is the file that has to run on every platform.
+files=$(
+	find "$ROOT/tests" -type f \( -name '*.sh' -o -perm -100 \) ! -name "$(basename "$0")" |
+	while IFS= read -r f; do
+		case $f in
+			*.sh) printf '%s\n' "$f" ;;
+			*) head -n 1 "$f" 2>/dev/null | grep -q '^#!.*sh' && printf '%s\n' "$f" ;;
+		esac
+	done | sort -u
+)
+[ -n "$files" ] || fail "found no test scripts to check"
+
+# report RULE PATTERN WHY -- flag every non-comment line matching PATTERN.
+report() {
+	local rule=$1 pattern=$2 why=$3 f line
+	for f in $files; do
+		# Drop comment lines before matching, keeping real line numbers.
+		# "|| true" twice: a grep with no match exits 1, which set -e and
+		# pipefail would take for an error.
+		{ grep -nE "$pattern" "$f" 2>/dev/null || true; } | \
+		{ grep -vE '^[0-9]+:[[:space:]]*#' || true; } | \
+		while IFS= read -r line; do
+			printf '%s\t%s:%s\t%s\n' "$rule" "${f#"$ROOT"/}" "${line%%:*}" "$why" >>"$work/violations"
+		done
+	done
+}
+
+# bash 3.2 is the floor: macOS ships it, and the suite runs under
+# /usr/bin/env bash. mapfile cost us a whole macOS lane once (#339).
+report "bash4" '(^|[^[:alnum:]_])(mapfile|readarray)([^[:alnum:]_]|$)|declare +-A|\$\{[A-Za-z_]+\^\^|\$\{[A-Za-z_]+,,' \
+	"bash 4 only; macOS ships bash 3.2"
+
+# Tool flags that exist on GNU and not on the BSDs, or mean different things.
+report "gnu-only" 'sed +-i|grep +[^|]*--include|grep +-[a-zA-Z]*P|stat +-c|readlink +-f|date +-d ' \
+	"GNU-only or differently-spelled on BSD/macOS; use a portable form"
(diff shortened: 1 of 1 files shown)


SoundGoof:
### Findings

#### Medium: The test uses non-portable word boundaries

`tests/buildsystem/test-suite-portability.sh:42`
`tests/buildsystem/test-suite-portability.sh:74`

OpenBSD 7.9 `grep -E` does not recognize `\b` as a word boundary. The
`mapfile` and `readarray` rule therefore cannot detect either construct there,
and the non-vacuity probe fails with two matches instead of three.

I reproduced this by running the test on OpenBSD 7.9:

```text
FAIL: the rule patterns no longer match the constructs they are meant to catch:
expected '3', got '2'
```

Use POSIX ERE boundaries in both locations:

```sh
(^|[^[:alnum:]_])(mapfile|readarray)([^[:alnum:]_]|$)
```

The corrected rule passes on Linux, FreeBSD, both NetBSD versions, and OpenBSD.

#### Medium: The configured-link-flags rule is narrow and comment-bypassable

`tests/buildsystem/test-suite-portability.sh:60-63`

The rule only recognizes archives named `libxymon[a-z]*.a`. Other in-tree
static archives are not checked. It also searches complete files for
`xymon_cflags` and `xymon_ldflags`, allowing comments alone to satisfy both
requirements.

I confirmed the bypass with a harness that mentioned both helpers only in
comments and linked an archive without configured flags. The current test
exited successfully.

Filter full-line comments before archive and helper detection. Match any
in-tree archive path instead of one library-name family. Use the same logic
for a non-vacuity probe containing an archive such as `libportable.a` and
comment-only helper mentions. It should produce both missing-helper
violations.

The generalized rule inspected 19 current scripts with in-tree archive
references and found no existing violations. A deliberate `libportable.a`
fixture produced exactly the two expected `link-flags` failures.

### Validation

- Current PR test fails on OpenBSD 7.9 because of `\b`.
- Corrected script passes on Linux.
- Corrected script passes on NetBSD 9.4.
- Corrected script passes on NetBSD 11.0.
- Corrected script passes on OpenBSD 7.9.
- Corrected script passes on FreeBSD 14.3.
- Generalized archive check: 19 scripts checked, 0 current violations.
- Deliberate archive violation: 2 expected failures detected.
- Unconfigured original suite: 28 passed, 41 skipped, 0 failed.

bonomani:
Both fixed — and the first one is the joke it deserves to be: a portability lint written with a non-portable construct.

**`\b`.** Replaced with `(^|[^[:alnum:]_])…([^[:alnum:]_]|$)` in the rule and in the probe. Worth recording how it surfaced: the non-vacuity probe is what failed on OpenBSD, `expected '3', got '2' — it did its job, but only on the platform where the fault exists. On Linux the rule matched happily and I checked nothing else.

That is also the second-order lesson. My first attempt at this fix did not apply, and I confirmed it by watching detection still work — on Linux, with GNU grep, where `\b` is fine. I verified the effect on the one platform where the defect is invisible.

**The link-flags rule.** Comments are now stripped before both the archive and the helper detection, so naming `xymon_cflags` in a comment no longer satisfies it, and any in-tree archive counts rather than the `libxymon*` family. Your bypass fixture now yields the two expected violations:

```
  [link-flags] .../zz-bypass.sh   links an in-tree archive without xymon_cflags
  [link-flags] .../zz-bypass.sh   links an in-tree archive without xymon_ldflags
```

The check is a function now, so the probe drives the same code the loop does rather than a copy of it — the point you made on #200 about production code versus a copy applies to a lint over the suite just as well.

One thing I changed on my own account: the archive pattern used `\{?`, a backslash-brace that is not portable ERE either. It is `[{]?` now. Same class as your finding, found by looking for more of it.

Full suite: 69 passed, 0 failed, 0 skipped. I cannot run the BSD lanes, so the four-platform confirmation stays yours.

bonomani:
Two further changes since my previous reply, both from asking what the fixes did *not* cover.

**A rule for `\b`, not just a fix.** Correcting the pattern kept my file portable; nothing stopped the next test from writing the same thing. There is a `gnu-regex` rule now — GNU-only escapes (`\b`, `\<`, `\>`, `\w`, `\s`, `\d`, …) on a line invoking `grep`/`sed`/`awk`/`expr` — with the probe raised to four constructs. The suite is clean under it today, so it lands green and guards forward.

**The scan skipped the file that matters most.** It matched `tests/**/*.sh`, and `tests/testsuite` carries no extension — the runner, the one script that executes on every platform, was out of scope. It now scans every shell script under `tests/` by shebang, extension or not: 74 files instead of 73. Demonstrated by planting a `mapfile` in the runner:

```
[bash4] tests/testsuite:112
```

The `.c` harnesses stay out; the rules are shell rules.

Verified for your two findings specifically: a `\b` in another test is now reported, and your bypass fixture (`libportable.a` with the helpers named only in comments) yields both `link-flags` violations through the same function the main loop uses.

Full suite: 69 passed, 0 failed, 0 skipped. The four-platform confirmation is still yours to make — I have no BSD runner, and the `\b` episode is a fair reminder that my green Linux run proves less than it looks.

SoundGoof (approved):
No blocking findings.

The latest revision addresses both previous findings. It replaces the
non-portable `\b` boundaries, generalizes archive detection, and prevents
full-line comments from satisfying the configured-flags checks.

The additional changes guard against GNU-only regex escapes and include
extensionless shell runners such as `tests/testsuite`. Non-shell files,
including `.c` harnesses, remain outside the scan.

Residual risk: helper names in an inline comment can still satisfy the
configured-flags check. For example:

    : # xymon_cflags xymon_ldflags

This is technically a bypass, but it is unlikely to occur accidentally.
Handling shell comments precisely would also require substantially more parsing,
so I do not consider it blocking.

### Validation

- The exact-head test passes at `054b21fad`.
- The clean merge with current `main` passes the focused test.
- The exact-head test passes on NetBSD 11.0 and OpenBSD 7.9.
- The OpenBSD run confirms the corrected POSIX word-boundary expression.

PR #360, opened 2026-08-15
tests: forbid forcing a failure with a permission bit

bonomani:
Follow-up to #359, kept separate so its approval was not dismissed for a fifth rule
unrelated to the four it carries.

The CI lanes run as **root**, where a permission bit stops nothing. A test that removes
write access to force a failure therefore passes on a developer's machine and fails in CI,
having proved the opposite of what it claims. That is what the `install-cgi` fixture did:
green for an unprivileged Linux user, red for @SoundGoof on NetBSD 9.4 and 11.0,
OpenBSD 7.9 and FreeBSD 14.3 (#351).

The rule says to force the failure structurally instead — an existing directory where a
file is expected, a path that is not there at all:

```sh
mkdir "$dir/history.sh"; chmod 500 …    # root walks straight through
mkdir -p "$dir/history.sh/cgiwrap"      # ln would have to unlink a directory
```

Measured unprivileged, so the difference is the mechanism and not the privilege:

| obstruction | ln says |
|---|---|
| mode 500 | `Permission denied` — a policy root ignores |
| destination is a directory | `cannot overwrite directory` — only `rmdir` removes one |

No occurrence is left in the suite, so this is green today and guards forward.

### After review

The numeric matcher only caught modes ending in `00`, so `chmod 555 dir` -- the usual
way to make a directory read-only -- went through, and root creates files in it
regardless (@SoundGoof). The owner digit carries the rule: 6 and 7 are the only two
granting both read and write, so `[0-5]` is exactly the modes denying one of them,
whatever group and other say. Measured over the whole mode space: `555`, `444`, `0555`
now rejected; `700`, `755`, `644`, `0755`, `4755`, `+x` still accepted.

Looking for that gap did not find it because the non-vacuity probe carried a *copy*
of the patterns, so widening the rule once already had left the probe exercising the
narrow form. It now matches with the patterns the rules were declared with, which
also removes the duplicated 200-character alternation. Narrowing the rule back fails
the probe.

Rebased on `2b2272eb9`. Full suite on a built tree: 73 passed, 0 failed, 0 skipped.

### DIFF
--- a/tests/buildsystem/test-suite-portability.sh
+++ b/tests/buildsystem/test-suite-portability.sh
@@ -37,8 +37,16 @@ files=$(
 [ -n "$files" ] || fail "found no test scripts to check"
 
 # report RULE PATTERN WHY -- flag every non-comment line matching PATTERN.
+#
+# Each pattern is kept, and the non-vacuity probe at the end matches with the
+# ones recorded here rather than a copy of them. The copy is how the root-lane
+# rule was widened while its probe still exercised the narrow form, so the
+# probe kept passing over a gap it was there to expose.
+__rule_patterns=()
 report() {
 	local rule=$1 pattern=$2 why=$3 f line
+
+	__rule_patterns[${#__rule_patterns[@]}]=$pattern
 	for f in $files; do
 		# Drop comment lines before matching, keeping real line numbers.
 		# "|| true" twice: a grep with no match exits 1, which set -e and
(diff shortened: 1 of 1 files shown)


SoundGoof:
## PR #360 review: tests: forbid forcing a failure with a permission bit

### Blocking finding

The numeric pattern misses `chmod 555`, a common way to make a directory
read-only:

```sh
mkdir "$work/blocked"
chmod 555 "$work/blocked"
touch "$work/blocked/file"
```

For an ordinary user, `touch` fails because the directory has no write
permission. Root can still create the file, producing exactly the
root/non-root discrepancy this rule intends to prevent.

I confirmed that `test-suite-portability.sh` accepts an executable test fixture
containing `chmod 555`. The non-vacuity probe only checks `chmod 500`, so it
does not expose this gap.

Changing the numeric mode from `[0-7]?[0-5]00` to
`[0-7]?[0-5][0-7][0-7]` detects modes where the owner write bit is absent.
I tested that change against the complete suite on:

- NetBSD 9.4: 27 passed, 42 skipped
- NetBSD 11.0: 25 passed, 44 skipped
- OpenBSD 7.9: 25 passed, 44 skipped
- FreeBSD 14.3: 27 passed, 42 skipped
- Debian 13: 28 passed, 41 skipped

There were no failures, and every platform correctly rejected the
`chmod 555` fixture.

Extend the numeric matcher

bonomani:
Taken as proposed, and measured over the whole mode space rather than the one case:

| | current | widened |
|---|---|---|
| `chmod 555 d`, `chmod 444 f`, `chmod 0555 d` | accepted | rejected |
| `chmod 500 d`, `chmod 000 d`, `chmod 200 d` | rejected | rejected |
| `chmod 700`, `755`, `644`, `0755`, `4755`, `+x` | accepted | accepted |
| `chmod u-w`, `a-w`, `-w` | rejected | rejected |

The owner digit is the whole rule: 6 and 7 are the only two granting both read and write, so `[0-5]` is exactly the modes that deny one of them, whatever group and other say. No granting mode is caught, including the 4-digit forms.

**On why looking for this gap did not find it.** The probe carried a *copy* of the patterns, so when the root-lane rule was widened the probe kept exercising the narrow form — and passed over the case it existed to expose. It now matches with the patterns the rules were declared with, so the two cannot drift. Narrowing the rule back fails the probe (`expected '6', got '5'`), and your fixture is reported:

```
[root-lane] tests/client/zz-mutant.sh:3
FAIL: 1 violation(s); see above
```

Rebased on `2b2272eb9`. Suite on a built tree: 73 passed, 0 skipped, 0 failed.

SoundGoof (approved):
lgtm

PR #361, opened 2026-08-15
client/linux: portable PID-reuse guard, and tests that stay off their host

bonomani:
**1 commit.** Follow-up to #343: four defects in its tests, plus the one-line client change
that removes the cause of the first, before the ports to the other clients start from it.

They are invisible here: `tests.yml` and `build.yml` run the suite on ubuntu only.
`cmake/bootstrap` runs `tests/testsuite` on every native lane, BSD VM and macOS
included (`pipeline-select-run-lanes.yml`; the VM skip was dropped in `62878ebb8`).

| defect | on a lane without it | fix |
|---|---|---|
| the PID-reuse guard reads `/proc/PID/comm` | absent on macOS/OpenBSD/FreeBSD, so the pid reads as stale and a **second probe starts** -- measured `expected '5', got '6'` | `ps -o comm=`, trimmed and reduced to its basename -- and the fixture, see below |
| three counts compare `wc -l` output as a string | BSD/macOS `wc` pads with blanks, `assert_equal` compares strings, correct run fails | `$((...))` on both sides, as `fs-filter-linux.sh:164-169` already did |
| `fs-filter-linux.sh` rewrites `/proc/filesystems` but not `/proc/mounts` | since the sentinel landed, `DF_LOCAL_ONLY=no` walks the tester's real mount table and probes a real `cifs/ceph/glusterfs/lustre/afs` mount, leaving files in `/tmp` (measured: `/tmp/df-probe-disk{,.pid}`) | local-only fixture, `XYMONTMP` inside `$TMP` |

Plus: the wedge cases left processes behind, two ways. The stub forked its `sleep`,
so ending the stub orphaned it (`sleep 120`, `sleep 5` per run) -- the `exec` below
removes the child entirely -- and an assertion failing while a wedge was outstanding
exited before anything killed it. Every recorded pid now goes through
`register_cleanup`: measured with a failure forced during a wedge, one process left
behind without it, none with it.

`ps -o comm=` is POSIX, reads process state only -- it cannot touch the wedged
mount -- and `ps` is already a hard dependency of every client script, each of which
emits a `[ps]` section. Measured on Linux: both spellings return `df` for the same
pid. The basename step is for macOS, where `comm` is the full path rather than the
short name.

That leaves `df_sentinel()` with nothing Linux-specific in it, which is the point:
the FreeBSD/NetBSD/OpenBSD/darwin ports can take it verbatim, and a later test can
assert the copies are identical.

### After review

`ps -o comm=` was not enough: the fixture is a shell script, and a script does not
carry its name into that field portably. Linux sets `comm` from the script, so it
read back as `df` and the suite was green here; the BSDs report the interpreter, so
the guard saw `sh`, took the live fixture for a recycled PID and started a second
probe anyway (@SoundGoof, on NetBSD 9.4/11.0, OpenBSD 7.9, FreeBSD 14.3).

The stub now enters the wedge with `exec`, into a real executable named `df`, which
reads as `df` under either rule and keeps the pid the client recorded.

Reproduced on Linux rather than taken on trust: `/proc/PID/exe` applies the BSD rule
on a Linux host, so the wedge case runs a second time against a `ps` that answers
that way. Before the fix it fails with exactly the reported message; restoring the
forked `sleep` fails it now.

Rebased on `2b2272eb9`. Suite on a built tree: 73 passed, 0 skipped, 0 failed.
Nothing left behind, nothing in `/tmp`.

### DIFF
--- a/client/xymonclient-linux.sh
+++ b/client/xymonclient-linux.sh
@@ -193,10 +193,16 @@ df_sentinel()
 				[ -n "$_c" ] && kill -0 "$_c" 2>/dev/null && return 124
 				;;
 			?*)
-				# A df we recorded. Still wedged? comm is checked so a recycled
-				# PID does not look like one.
-				if kill -0 "$_old" 2>/dev/null && \
-				   [ "$(cat "/proc/$_old/comm" 2>/dev/null)" = df ]; then
+				# A df we recorded. Still wedged? The command name is checked
+				# so a recycled PID does not look like one. Read it with ps,
+				# not /proc/PID/comm: ps is POSIX, gives the same answer, is
+				# already a dependency of this client (the [ps] section), and
+				# reads process state only -- it cannot touch the wedged mount.
+				# Take the basename: comm is the short name on Linux and the
+				# BSDs, but the full path on macOS.
+				_c=$(ps -o comm= -p "$_old" 2>/dev/null | tr -d '[:space:]')
+				_c=${_c##*/}
+				if kill -0 "$_old" 2>/dev/null && [ "$_c" = df ]; then
 					return 124
 				fi
 				;;
--- a/tests/client/fs-filter-linux.sh
+++ b/tests/client/fs-filter-linux.sh
@@ -67,7 +67,16 @@ EOF
 # fixture in both. The [inode] block reuses the EXCLUDES/ROOTFS the [df] block
 # computed, so the combined block must run them together -- a lone [inode]
 # snippet would see an empty EXCLUDES.
-PROC_REWRITE="s!/proc/filesystems!$TMP/proc.filesystems!g"
+# /proc/mounts fixture. Since the remote-df sentinel (#316) the LOCAL_ONLY=no
+# case walks the mount list looking for hard-blocking types, so without this
+# rewrite the test reads the *tester's* mount table: on a machine with a
+# cifs/ceph/glusterfs/lustre/afs mount it would probe that real filesystem and
+# drop probe files in $XYMONTMP. A local-only fixture keeps the sentinel dormant
+# -- it is covered by fs-sentinel-linux.sh -- and XYMONTMP points inside $TMP so
+# nothing can land in /tmp either way.
+printf '/dev/sda1 / ext4 rw 0 0\n' > "$TMP/proc.mounts"
+export XYMONTMP="$TMP"
+PROC_REWRITE="s!/proc/filesystems!$TMP/proc.filesystems!g; s!/proc/mounts!$TMP/proc.mounts!g"
 SNIPPET="$TMP/df-section.sh"
 fsf_extract "$SNIPPET" "$PROC_REWRITE" '\[inode\]'
 COMBINED="$TMP/df-inode-section.sh"
--- a/tests/client/fs-sentinel-linux.sh
+++ b/tests/client/fs-sentinel-linux.sh
@@ -56,7 +56,7 @@ done
 # Remote probes are counted apart from the per-cycle local df, and each records
 # its own pid so the test can end the ones it starts.
 [ -n "$_named" ] && echo $$ >> "${DF_REMOTE:-/dev/null}"
-[ -n "$_named" ] && [ -n "${DF_HANG:-}" ] && sleep "${DF_HANGTIME:-120}"
+[ -n "$_named" ] && [ -n "${DF_HANG:-}" ] && exec "$DF_HANGBIN" "${DF_HANGTIME:-120}"
 [ -n "${DF_FAIL:-}" ] && exit 1
 if [ -n "$_named" ]; then
 	# An exclusion applies to an explicitly named operand too, and once every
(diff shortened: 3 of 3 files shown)


SoundGoof:
## Blocking finding

### The sentinel regression test fails on every tested BSD

`tests/client/fs-sentinel-linux.sh` implements its `df` fixture as a shell
script. On NetBSD, OpenBSD, and FreeBSD, `ps -o comm=` reports the interpreter
as `/bin/sh` or `sh`, rather than the script name `df`.

The check added in `client/xymonclient-linux.sh` therefore treats the live
fixture as a stale PID and starts another remote probe.

The focused test fails identically on:

- NetBSD 9.4
- NetBSD 11.0
- OpenBSD 7.9
- FreeBSD 14.3

```text
FAIL: while one df is wedged, a cycle must run only the local df -- not a
second remote one: expected '5', got '6'
```

Direct inspection on FreeBSD showed:

```text
comm=[sh]
command=[/bin/sh /var/tmp/pr361-ps/df]
```

The failed assertion exits before `end_stub` runs. Each BSD guest was left
with two fixture processes and two `sleep 120` children.

Use a fixture whose process name is reliably `df` on all supported test
platforms, or mock the process-name lookup while testing native `ps`
semantics separately. Register process cleanup through the test cleanup
framework so an assertion failure cannot leave children running.

## Validation

Reviewed head: `ff3c3edeaadb86862f5cffc8cc81999391ade55a`

Linux passed:

- Local suite: 29 passed, 41 skipped, 0 failed
- Debian 13: 29 passed, 41 skipped, 0 failed
- No residual sentinel processes after successful runs

The filesystem fixture-isolation changes otherwise look sound.

bonomani:
Both fixed, and I could reproduce your failure on Linux rather than take it on trust.

The two platforms differ in one rule: BSD reports the basename of the process's **executable**, Linux the name the kernel was handed — which for a script is the script. So `ps -o comm=` reads:

| process | comm |
|---|---|
| the fixture script, executed directly (Linux) | `df` ← why CI was green |
| the same script run as `/bin/sh .../df` | `sh` ← your four BSDs |
| a script that `exec`s a real executable named `df` | `df` |
| the same, via `/bin/sh` | `df` |

**Fixture:** the stub stays a script — it models df's argument contract — but enters the wedge with `exec` into a real executable that is itself named `df`. `exec` keeps the pid the client recorded, and removes the forked `sleep` along the way.

**Reproduced and pinned on Linux.** Reading `/proc/PID/exe` applies the BSD rule on a Linux host, so the wedge case now runs a second time against a `ps` that answers that way. Before the fix it fails with exactly your message — `expected '5', got '6'` — and restoring the forked `sleep` fails it now. The case that broke on four BSDs is therefore exercised on the Linux lanes too, not only in the VMs.

**Processes:** every recorded pid goes through `register_cleanup`, which the framework runs on EXIT however the test ends. Measured with a failure forced while a wedge was outstanding: one process left behind without it, none with it.

Rebased on `2b2272eb9`. Suite on a built tree: 73 passed, 0 skipped, 0 failed, nothing left behind.

SoundGoof (approved):
No blocking issues found.

## Design question

### Should Linux production code change to support tests on other operating systems?

`xymonclient-linux.sh` runs on Linux, where `/proc/PID/comm` is available. This
PR replaces that direct read with a `ps` subprocess so the Linux client code
can also execute in BSD and macOS test lanes.

How far should production code for one operating system be adapted to make its
tests run on other operating systems?

The replacement is portable and `ps` is already required by the client, so the
change is reasonable here. It does, however, add a process invocation to a
Linux path that previously used a direct file read.

An alternative would be to keep the Linux implementation and make the test
inject or rewrite the process-name lookup. That would preserve the production
path but test more fixture code instead of the real implementation.

Document whether cross-platform execution of OS-specific client scripts is an
expected project requirement. That would make the tradeoff explicit for later
ports and tests.

## Validation

Reviewed commit `6f04d820fcb98928b67907ed336cc15eb093a4bb`.

- Changed tests passed on Linux
- Changed tests passed on FreeBSD 14.3
- Exact `ps -o comm=` form worked on FreeBSD
- Portability guard passed
- Shell syntax checks passed
- Available Linux suite: 31 passed, 42 skipped, 0 failed
- `git diff --check` passed

NetBSD, OpenBSD, and macOS were not run locally during this review. Their
behavior is covered by the PR's native CI lanes.

bonomani:
On the design question — worth answering with the current state rather than in principle, so I
counted it.

**The project already practises A, without having written it down.** Of the 74 tests on `main`,
**7** extract a client's `[df]` block and run it on whatever host the suite runs on
(`fs-filter-{linux,freebsd,netbsd,openbsd,darwin}.sh`, `fs-sentinel-linux.sh`,
`dpkg-report.sh`), and **none** guards on `uname`. That was never decided; it followed from the
first test that extracted a block and stubbed `df`. So the real question is not "A or B" but
"do we own the A we already run?" — and the cost of B is concrete: on a BSD or macOS lane the
client coverage would drop from 7 tests to 1, and `aix`, `hp-ux` and `sunos` have no lane at
all, so their tests would never execute anywhere.

**But your objection stands, and there is a hybrid that keeps both.** The four BSD/macOS tests
already follow it without naming it: their OS primitives are *commands* (`df`, `mount`), so
they are replaced through `PATH`. Only Linux is awkward, because its primitives are *files* —
and a file cannot be replaced through `PATH`, hence the `sed` path rewriting those two tests do
(and nobody else does).

    Linux            fs_procname() { cat "/proc/$1/comm"; }
    BSD, macOS       fs_procname() { ps -o comm= -p "$1" | ...; }

`df_sentinel()` stays byte-identical, the portable test replaces the helper as it replaces a
command, and a small `uname`-guarded test checks the real primitive where it ships. Linux keeps
its direct read, with no fork — your point — and the coverage stays. The pattern is already in
the tree: `fs_mounts()` and `probe_dir_is_local()` are exactly this, per-OS helpers that
`df_sentinel()` calls without knowing them.

What it would rework, measured: **two tests**, both Linux (`fs-filter-linux.sh`,
`fs-sentinel-linux.sh` — one path rewrite each). The four others need nothing. Plus three new
guarded tests that verify what the stubs currently assume, including one that has never been
verified at runtime: that `mount` with no arguments does not `statfs` (I read `MNT_NOWAIT` in
the FreeBSD, NetBSD, OpenBSD and Apple sources, and checked the three call sites in the shipped
macOS binary, but never executed it).

And it can be enforced rather than remembered, which is this repo's habit since #359/#360: a
rule forbidding a rewritten `/proc` path in tests (2 occurrences today), and the existing
portability rules applied to the extracted client blocks, which would have caught three of the
four defects this PR fixes. Two things stay beyond a line-based lint and need the lanes: a
`case` inside `$( )`, and whether a stub still matches reality.

**Proposed order:** after the ports (#366), not before — the hybrid gives all five clients the
same shape, and four of them do not have the sentinel yet. Doing Linux alone would be half an
alignment, revisited later.

I have not touched this PR: it is approved as it stands, and you marked the question
non-blocking.

bonomani:
@SoundGoof — on your design question about whether Linux production code should change so its tests can run elsewhere: the answer we landed on is the hybrid, and #378 is it. Each OS primitive sits behind a named helper (`fs_mounts`, `fs_filesystems`, `fs_procname`) that a test replaces by name rather than by rewriting a path only one OS has, and two `uname`-guarded native tests cover the primitives themselves — the part no stub can check.

It already earned its keep twice: `ps` is absent from a stock Debian container, and the native mount test found that NetBSD and OpenBSD spell mount(8) differently from FreeBSD, which had been dropping every row there.

No need to look at it before #366 — it is stacked on top.

PR #363, opened 2026-08-15
tests: one FS filter contract for all five clients, asserted on the report

bonomani:
**1 commit.**

Every client implements the same `XYMONCLIENT_FS_*` contract and spells it differently: `-x` per type on Linux, one `-t no<csv>` on the BSDs, one df per path on macOS. The tests were written the other way round — **115 of their 158 assertions matched df's argv** — so they encode the spelling and break on any change of mechanism even when behaviour is identical. #343 had to rewrite `assert_not_contains " -l "` for exactly that reason, while the report it guarded never moved.

The contract now lives in `fs-filter-common.sh` and is asserted on what `[df]` and `[inode]` emit. A caller supplies only its dialect: five filesystems by role, the row formats its post-processing expects, and a few lines turning this OS's df argv into the stub's answer.

**Not a dedup.** The contract was asserted unevenly, and nobody chose that:

| | multi-token INCLUDE | duplicate EXCLUDE | empty-list marker |
|---|---|---|---|
| linux | yes | yes | no |
| freebsd / netbsd / openbsd | no | no | no |
| darwin | no | no | yes |

All five now run the same 24 assertions. Sizes: common 107 → 351; linux 331 → 164, darwin 210 → 147, freebsd 137 → 91, netbsd 101 → 91, openbsd 117 → 79.

**Two guards, because a contract that passes vacuously is worse than none:**

- `fsf_selfcheck` — half the contract is "X must NOT be reported", which a stub reporting nothing satisfies. It proves the fixture entry exists and that the stub honours this OS's exclusion spelling before any assertion runs.
- `fs-sentinel-copies.sh` — every `df_sentinel()` must be byte-identical. The sentinel is duplicated per OS rather than shared: the clients source nothing, and a sourced fragment would open a window during `clientupdate` where it is missing and the client aborts — purpling exactly the hosts it protects.

Mutation-tested, each caught by the rule that names it:

| mutation | caught by |
|---|---|
| linux / freebsd: INCLUDE_TYPES a no-op | `INCLUDE_TYPES un-excludes a default-excluded type` |
| linux: EXCLUDE_TYPES a no-op | `EXCLUDE_TYPES drops a type that was being reported` |
| linux / netbsd: `-l` dropped | `a remote filesystem is hidden unless DF_LOCAL_ONLY=no` |
| linux / openbsd: inode no-limit drop removed | `a filesystem with no inode limit is dropped from [inode]` |
| linux: failure marker removed | `a df that exits nonzero with no output emits a failure marker` |
| darwin: apfs reaches `[inode]` | `apfs volumes are never queried for inodes` |
| stub ignores `-x` | `fsf_selfcheck` |
| linux: `set -f` lost around EXCLUDE_TYPES | `an EXCLUDE_TYPES token must not undergo pathname expansion` |
| darwin: 2nd filesystem queried with `-H` | `every inode-capable filesystem is queried in inode mode, not just the first` |
| freebsd: `cd9660`/`devfs` dropped from the built-in list | `the built-in pseudo list still excludes every type it names` |

The last three came out of a second-opinion review of the whole stack, and are the coverage the argv assertions used to provide: the token-is-literal check for `EXCLUDE_TYPES` (only `INCLUDE_TYPES` had one), the count of inode-mode calls on darwin, and the built-in pseudo list on the three BSDs — now pinned through the report, with one fixture filesystem per type the client excludes by default, instead of through the `-t no<csv>` spelling.

The generated df stub keeps its `case` inside a function rather than inline in `$( )`: bash 3.2 is `/bin/sh` on macOS and cannot parse the inline form.

Not carried over, deliberately: counting `-x` duplicates (invisible in the report) and the failing-df case under a restricted PATH (the contract covers that path directly).

Suite: 0 failed, 19 skipped.


SoundGoof (approved):
No blocking findings.

The shared contract is an improvement over checking each platform's `df`
arguments. It now validates the emitted `[df]` and `[inode]` sections while
retaining platform-specific checks where behavior differs.

The fixture and `fsf_selfcheck` provide credible non-vacuity checks. The
contract covers inclusion, exclusion, precedence, glob handling, remote
filesystems, invalid configuration, inode filtering, and failed collection.

One non-blocking observation: `fs-sentinel-copies.sh` currently finds only one
`df_sentinel()` implementation. It therefore guards against complete removal,
but its equality check will become effective only after another client gains a
copy. The success message accurately reports the single implementation.

Validation completed:

- all touched filesystem-filter tests pass locally
- OpenBSD 7.9 filter and copy-guard tests pass natively
- FreeBSD 14.3 filter and copy-guard tests pass natively
- all five GitHub checks are green
- the branch merges cleanly into current `main`

Approved.

PR #364, opened 2026-08-15
tests: cover the sentinel's inode probe and a realistic mount set

bonomani:
**1 own commit**; the other is #363, which this is stacked on.

Two gaps left by #343. Both made `fs-sentinel-linux.sh` prove less than it looked like.

**The inode probe was never exercised.** `[inode]` calls `run_df` a second time, so the
sentinel runs a second time under its own tag: a dead server leaves **two** probes
outstanding per cycle, and the `unavailable:` rows have to survive the no-inode-limit drop
(which discards rows whose `IUse%` is `-`). The test extracted the `[df]` block alone,
stopping at `[inode]`. It now runs both and asserts the unavailable row in each section, a
pidfile per tag (`df-probe-disk.pid`, `df-probe-inode.pid`), and that a second cycle while
both are wedged starts neither probe again.

**The fixture was not a machine anyone runs.** Its `/proc/filesystems` listed only `proc`
and `sysfs`. On a real host every remote type is `nodev`, so it is already in the client's
default exclusions and the mount never reaches the sentinel: `DF_LOCAL_ONLY=no` on its own
probes nothing at all. Surfacing a remote filesystem also takes `INCLUDE_TYPES` naming the
type — the recipe `client/xymonclient-linux.sh:65-73` documents, shipped in
`client/xymonclient.cfg.DIST:32` and `common/xymonclient.cfg.5:38`:

> …or add remote types e.g. `"nfs nfs4 ceph"` (which also needs
> `XYMONCLIENT_FS_DF_LOCAL_ONLY=no`, since `df -l` hides remote mounts).

The fixture is now realistic, every case passes that recipe, and a new case pins the
stock-host behaviour: nothing reported, no probe started, no probe state left behind, while
a remote type that is *not* nodev is still reported as `DF_LOCAL_ONLY=no` asks.

| mutation | caught by |
|---|---|
| the inode report skips the sentinel | `and in the inode report: an absent filesystem reads as green there too` |
| both reports share one probe tag | `the inode probe must be recorded under its own tag` |
| the excluded-type filter dropped from the probe list | `an excluded type was probed anyway and its healthy mount reported unavailable` |
| the inode marker row says `0%` instead of `100%` | `the inode marker must read as full, or the column stays green` |
| a remote df started outside the sentinel on the inode pass | `a wedged cycle must start exactly one remote df per report` |

The last two came out of a second-opinion review: asserting the mount name alone let a marker row through that reads green, and a pidfile existing does not prove only one probe was started.

The third mutation is the one worth keeping in mind: probing a type the admin excluded
comes back empty-and-nonzero, and reports a healthy mount 100% full.

Suite: 0 failed, 41 skipped (unbuilt tree). Portability lint green. No orphan
processes, nothing left in `/tmp`.


SoundGoof (approved):
No blocking findings.

The added coverage correctly exercises separate disk and inode probes, verifies
their distinct pidfiles, and confirms that repeated cycles do not start
additional remote `df` processes.

The realistic `nodev` fixture also captures the documented requirement that a
remote type must be explicitly included before `DF_LOCAL_ONLY=no` can expose
it. The stock-host case verifies that excluded remote types produce neither
reports nor probe state.

Validation completed:

- focused test passes locally
- focused test passes natively on OpenBSD 7.9
- focused test passes natively on FreeBSD 14.3
- focused test passes natively on NetBSD 11.0
- no simulated `df` processes remain after testing
- all five GitHub checks are green
- the branch merges cleanly into current `main`
- the single own commit is correctly stacked on PR #363

Approved.

PR #365, opened 2026-08-15
client/linux: keep the df failure marker when DF_LOCAL_ONLY=no

bonomani:
**1 own commit**; the other 3 are #361, #363 and #364. Found while wiring the ports; the
bug is on **main**, not in a PR.

With `DF_LOCAL_ONLY=no` the collection is split in two -- one plain df for everything that
cannot hard-block, one guarded probe for the mounts that can -- and `run_df` returned 0
whatever the plain df did. `emit_df` turns "no output and non-zero" into the marker that
drives the server yellow; with the status swallowed, a df that failed and printed nothing
produced an **empty section**, which the server reads as green.

Measured on the merged code, same failing df, only the tunable differing:

```
DF_LOCAL_ONLY=yes -> [df] Disk collection failed: df exited 1 with no output
DF_LOCAL_ONLY=no  -> [df]        (nothing at all)
```

That is the false green the marker exists to prevent, reached from the other side of the
split. Introduced by #343 (#316).

**Fix.** The plain df's status is kept and returned on the paths where nothing else was
collected: no mount list to read, or no hard-blocking mount to probe. When the guarded
probe contributes rows -- data or `unavailable:` markers -- the report is not empty and the
status stays 0.

**Why no test caught it.** The contract exercised the failure marker under the default
`DF_LOCAL_ONLY=yes`, the one path where nothing is split. It now runs the same case with
`DF_LOCAL_ONLY=no` on every client. Only Linux splits today, but the four ports will, and
this is the rule they must keep. Verified to bite: restoring the unconditional `return 0`
fails with `contract: the failure marker survives DF_LOCAL_ONLY=no`.

Suite: 0 failed, 41 skipped (unbuilt tree).


SoundGoof (approved):
Approved. No findings.

The fix correctly preserves the plain `df` exit status when the split
collection produces no guarded output. This restores the failure marker
instead of reporting an empty, falsely green section.

Validation:

- Clean merge with current `main`.
- All affected local filter and sentinel tests pass.
- Native OpenBSD, FreeBSD, and NetBSD filter tests pass.
 All five GitHub checks are green.

PR #366, opened 2026-08-15
client: bound a wedged remote df on FreeBSD, NetBSD, OpenBSD and macOS (#316)

bonomani:
**1 commit.** Closes the fleet gap left by #343: all five filtered clients now bound a wedged remote df instead of losing the host.

The wedge was never Linux's — with `DF_LOCAL_ONLY=no`, df stats remote mounts, and a dead NFS/CIFS server leaves it in an uninterruptible wait where SIGKILL is ignored, so the cycle never finishes and every column goes purple, not just disk. Only the guard was.

`df_sentinel()` is copied verbatim; `fs-sentinel-copies.sh` fails if the copies differ, and `fs-sentinel-linux.sh` keeps testing its internals, so the ports do not re-test them. What is per-OS is the wiring: the mount list (`mount(8)`, `MNT_NOWAIT`, never `-v` on FreeBSD), how the unguarded df is shielded (`-t no<csv>` on the BSDs; macOS has no type filter, so its per-path loop is only handed what it can stat), and where the header comes from when every surviving filesystem is guarded.

Three defects found while building it, each fixed here and pinned by a test:

| defect | how it showed |
|---|---|
| macOS passed `"-P -H"` as one argv word | `df: invalid option --`, rc=64 on a real Mac; every *healthy* guarded mount reported `unavailable:` |
| the pid file was written non-atomically | a cycle reading mid-write took the empty value for a stale entry and started a second probe, leaving the first df unrecorded (@SoundGoof) |
| NetBSD and OpenBSD spell mount(8) as `MP type <fstype> (opts)` | enumeration dropped every row there, so remote filesystems vanished from the report |

The claim is now published by hard-linking a private file into place and the pid by renaming one over it: link fails if the target exists, and a rename replaces in one step, so a reader sees either the whole claim or the whole pid. `probe_dir_is_local()` keeps the mount helper's own status rather than the pipeline's — awk succeeds on no input, and "I cannot read the mount list" is not "this directory is local".

The wedge fixture is compiled rather than copied: `cp $(command -v sleep) df` gives BusyBox's `df` on Alpine, and macOS SIGKILLs a copy of a signed system binary. It falls back to copying, and checks the result survives before handing it back.

| mutation | caught by |
|---|---|
| flags reach the probe as one word | `a healthy server must produce real rows, never the unavailable marker` |
| pid published with a truncating write | `a cycle reading the pid file mid-publication must not start a second probe` (deterministic: a `mv` stub holds the window open) |
| the FreeBSD/macOS parse used on NetBSD | `does not contain '/net'` — the remote mount disappears |
| `probe_dir_is_local` back on the pipeline | `an unreadable mount list must not read as a local probe directory` |
| a mount point ending in `type WORD` read as the NetBSD spelling | `must still be recognised as nfs and guarded` |
| inode report skips the sentinel, or gets a disk-shaped marker row | `and in the inode report, whose columns are read positionally` |

Docs: both tunables were "Linux only"; the default hard-blocking list is per-OS. HTML regenerated with `build/makehtml.sh`.

Suite: 79 passed, 0 skipped, 0 failed on a built tree. Verified on the `cmake/bootstrap` lanes (FreeBSD 13.5/14.3/15.0, NetBSD 9.2–10.1, OpenBSD 7.6/7.7/7.8, macOS 14/15/26) and on a real Apple-silicon Mac.

### DIFF
--- a/client/xymonclient-darwin.sh
+++ b/client/xymonclient-darwin.sh
@@ -72,6 +72,245 @@ _user="${_user#|}"
 # dropping remote mounts regardless of type.
 _localfilter=""
 [ "$DFLOCALONLY" = yes ] && _localfilter='/[(, ]local[,) ]/!d;'
+# XYMONCLIENT_FS_REMOTE_DF_BUDGET: seconds to wait for the remote df probe
+# before giving up for this cycle (default 30, capped at 3600; non-numeric,
+# empty or zero -> 30). A poll BUDGET, not a kill timeout: a df wedged on a
+# dead server sits in uninterruptible D-state where even SIGKILL is ignored,
+# so it is left running and picked up on a later cycle rather than killed.
+DFBUDGET="${XYMONCLIENT_FS_REMOTE_DF_BUDGET:-30}"
+case "$DFBUDGET" in
+	*[!0-9]*|'') DFBUDGET=30 ;;
+	*[1-9]*) ;;
+	*) DFBUDGET=30 ;;
+esac
+[ "$DFBUDGET" -le 3600 ] 2>/dev/null || DFBUDGET=3600
+
+# XYMONCLIENT_FS_REMOTE_HARDBLOCK_TYPES: the mount types whose stat() hard-blocks
+# when the server is unreachable. Shipped as a built-in list so no site has to
+# maintain one; assign the variable to override it.
+: "${XYMONCLIENT_FS_REMOTE_HARDBLOCK_TYPES=nfs nfs4 smbfs cifs afpfs webdav ceph glusterfs lustre afs}"
+DFPROBEDIR="${XYMONTMP:-/tmp}"
+
+# fs_mounts : one "TYPE<tab>MOUNTPOINT" line per mount, from mount(8). macOS
+# mount(8) lists from getmntinfo(MNT_NOWAIT) -- verified in the shipped binary,
+# three call sites, all MNT_NOWAIT -- so it never refreshes statfs and cannot
+# block on a dead server, which df would.
+fs_mounts() {
+	mount | awk '
+		{
+			i = index($0, " on ")
+			if (i == 0) next
+			rest = substr($0, i + 4)
+			if (match(rest, / \(/) == 0) next
+			mp = substr(rest, 1, RSTART - 1)
+			split(substr(rest, RSTART + 2), a, /[,)]/)
+			printf "%s\t%s\n", a[1], mp
+		}'
+}
+
+# fs_hardblocking : the mount points whose type hard-blocks, one per line.
+fs_hardblocking() {
+	fs_mounts | awk -F'\t' -v types="$XYMONCLIENT_FS_REMOTE_HARDBLOCK_TYPES" '
+		BEGIN { n = split(types, a, /[ \t]+/); for (i = 1; i <= n; i++) t[a[i]] = 1 }
+		($1 in t) { print $2 }'
+}
+
+# fs_only IN EXCLUDE-LIST / fs_also IN KEEP-LIST : whole-line set operations, so
+# a mount point containing a space is still matched exactly. The list is fed
+# through the same stream as the input, ahead of a separator line, rather than
+# through "awk -v": an assignment with embedded newlines is not portable, and
+# on a BSD awk it silently yields an empty set.
+fs_setop() {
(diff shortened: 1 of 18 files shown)


SoundGoof:
### Finding

## Publish the probe PID atomically

The sentinel initially writes `claim:<owner>` to reserve the PID file. After
starting `df`, it replaces that claim using:

```sh
echo "$_pid" > "$_pidf"
```

The redirection truncates the file before `echo` writes the PID. A concurrent
cycle can read it during that interval and see an empty value. It then treats
the state as stale, removes the files, and starts another probe.

The first cycle still writes through its open descriptor, but that file has
already been unlinked. Its potentially wedged `df` is therefore left
unrecorded. Later cycles can repeat this and accumulate more untracked
processes.

This explains the acknowledged intermittent concurrency failure. It also
breaks the sentinel's primary guarantee: at most one outstanding remote `df`
per report.

Use the atomic publication pattern already present elsewhere in Xymon. For
example, `client/xymonclient.sh` writes the new logfetch configuration to
`$LOGFETCHCFG.tmp` and then renames it over `$LOGFETCHCFG`. `clientupdate.c`
similarly prepares an update in a separate `.tmp` file before installing it.

Write the PID to a private temporary file in the same directory and rename it
over the claim:

```sh
printf '%s\n' "$_pid" > "$_pidf.$$" &&
    mv -f "$_pidf.$$" "$_pidf"
```

A same-filesystem rename is atomic, so readers see either the complete
`claim:<owner>` value or the complete child PID, never an empty intermediate
state. Handle preparation or rename failure without leaving the newly started
`df` unrecorded, and apply the same change to all five identical copies.

The existing 20-cycle stress test only encounters this narrow interval
probabilistically. Add a deterministic regression case that pauses publication
before the rename, starts a contender, and asserts that no second probe is
launched.

### Validation

- The stack is exactly one commit above #365.
- The PR merges cleanly with current `main`.
- Filter, wiring, and sentinel tests pass natively on OpenBSD 7.9,
  FreeBSD 14.3, and NetBSD 11.0.
- The copy-consistency test passes for all five clients.

These passing runs do not eliminate the race. The PR description and earlier
native NetBSD testing reproduce the duplicate-probe failure intermittently.

bonomani:
Confirmed, fixed, and the other half of it with it.

**Publication.** As you describe: `> "$_pidf"` truncates before it writes, a cycle reading in that interval takes the empty value for a stale entry, removes the files, and starts its own probe — leaving the first cycle's `df` with nothing recording it. The pid now goes to a private file and is renamed over the claim, following `xymonclient.sh`'s logfetch pattern.

**The claim had the same shape.** `( set -C; echo "claim:$$" > "$_pidf" )` makes the *create* atomic, not create-then-write, so a reader in that gap sees the same empty value. Traced with the real function on a 32-core box: 10 empty reads in 500 cycles, 2 rounds in 25 starting more than one `df`, none at 1–2 cores. The claim is now written to a private file and hard-linked into place: `ln` fails if the target exists, so exclusivity is kept *and* what becomes visible is already complete. A `[ -d ]` guard comes first — link into a directory puts the file inside it, which would read as a won claim while nothing recorded the pid. Where the filesystem has no hard links at all the sentinel now says so instead of reporting `unavailable:` forever with no explanation.

**Deterministic case**, as you asked: a `mv` stub delays the rename, a second cycle runs inside the window, and the test asserts one probe started, an `unavailable:` row from the loser, and a pid file naming the `df` that is actually running. Against the old truncating write it fails with `expected '1', got '2'`. The 20-cycle stress case stays alongside it.

Before/after on the claim window, same harness: 1 round in 20 started two `df`s, 0 after.

One more thing came out of validating this on real hardware rather than in CI: the wedge fixture — a copy of `/bin/sleep` named `df`, needed because `ps -o comm=` answers with the interpreter for a script — is SIGKILLed on an Apple-silicon Mac (`rc=137`). All five sentinel tests failed there on a fixture that never wedged, while the same tests passed on the macOS runners. It now compiles a small binary, falls back to the copy, and proves the result stays alive before handing it back; where neither works the test skips with the reason instead of failing.

Closes #379.

Lanes at `verify_depth=test`, on the revision before that last (test-only) change: FreeBSD 13.5/14.3/15.0 30–32 passed, NetBSD 9.2–10.1 30 passed, OpenBSD 7.6/7.7/7.8 30 passed — all with the same 13 pre-existing failures a client-only tree always has, none of them a filesystem test. macOS is still running.

Still one commit.

SoundGoof (approved):
lgtm

PR #367, opened 2026-08-15
xymon-mailack: read the PCRE2 substring-copy result the way PCRE2 returns it

bonomani:
Mail acknowledgement has not worked since the PCRE2 migration. Every reply is rejected with `Could not find cookie value` and no `xymondack` is ever sent.

PCRE1's `pcre_copy_substring()` returned the substring's **length**, so `<= 0` meant failure. PCRE2's `pcre2_substring_copy_bynumber()` returns **0 for success**, so the same test is true exactly when the copy worked.

`f2ffa8c3b` (PCRE2 migration, #5) converted two of the three extractions in this function and carried the third across unchanged:

| | before | after |
|---|---|---|
| `delaytxt` | `> 0` | `== 0` ✅ |
| `msgtxt` | `> 0` | `== 0` ✅ |
| `cookie` | `<= 0` | `<= 0` ❌ |

It was the only one written as a failure test rather than a success test, so its shape never had to change while its meaning inverted underneath it.

**Measured**, not reasoned about. Against the library: on `Re: Xymon [1234] db1:disk RED` the call returns `0` and fills the buffer with `1234`. End to end, the same reply gives:

```
before:  Could not find cookie value          (exit 4)
after:   xymondack 1234 60 acked, on it
         Acked by: Ops  ops@example.com
```

**Swept the tree** for the same shape — this is the only surviving one. `showgraph.c` and the other two sites here use `== 0`; `do_la.c` tests the `pcre2_match` result, not the copy; and no `pcre2_substring_*` call sits in a bare boolean context, which is the quieter form of the same inversion.

**Test:** `tests/server/mailack-cookie.sh` drives the real binary. `--debug` prints the message it would send and returns without sending, so the whole parse is testable without a xymond. It pins the cookie, the default duration and the reply text — and that a subject with no cookie, or no subject at all, still produces nothing. Without those last two, a build that acknowledged everything would pass.

Fails on unmodified `main`, passes with the change.

### DIFF
--- a/tests/server/mailack-cookie.sh
+++ b/tests/server/mailack-cookie.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/server/mailack-cookie.sh
+#
+# xymon-mailack turns a reply to an alert mail into a xymondack message, using
+# the cookie in the subject line. PCRE1's pcre_copy_substring() returned the
+# substring's *length*, so "<= 0" meant failure; PCRE2's
+# pcre2_substring_copy_bynumber() returns 0 for success and a negative error
+# code otherwise, so the same test is true exactly when the copy worked.
+#
+# The PCRE2 migration converted two of the three extractions in that function
+# and carried the third across unchanged, which left every acknowledgement
+# bailing out with "Could not find cookie value" -- the feature had been
+# entirely non-functional since.
+#
+# --debug prints the message it would send and returns without sending, so the
+# whole path is testable without a xymond: the cookie, the default duration and
+# the reply text all come from the parse this guards.
+
+set -euo pipefail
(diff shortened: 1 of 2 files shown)


SoundGoof (approved):
No findings.

The condition now correctly treats `0` as success, matching PCRE2's API and
the neighboring substring-copy calls.

Reviewed head: `db62146c2a980d50e14b875e9db80fcf346b0a1f`

## Validation

I built the real `xymon-mailack` binary and fed an actual email through stdin.

The PR binary exited 0 and emitted:

    xymondack 1234 60 acked, on it
    Acked by: Ops  ops@example.com

The same regression test against the real unmodified `main` binary failed
because no acknowledgement was produced.

Additional results:

- `xymon-mailack` linked against `libpcre2-8`
- Full post-build suite: 68 passed, 3 skipped, 0 failed

PR #370, opened 2026-08-15
tests: -iquote in the three harnesses that still used -I, and a rule so it stays that way

bonomani:
Three tests fail on **every macOS Server lane** and have since they were added. One cause:

```
In file included from .../SDKs/MacOSX.sdk/usr/include/_stdio.h:71:
lib/Availability.h:18:2: error: unknown type name 'time_t'
lib/Availability.h:47:30: error: unknown type name 'FILE'
```

An `-I` directory answers `<angle>` includes too. The SDK's `stdio.h` includes
`<Availability.h>`, the filesystem is case-insensitive, and **our own `lib/availability.h`**
answers instead -- so every system type after it is unknown and the harness does not build.

| | |
|---|---|
| `tests/server/xymond-flap-purple-pin.sh:61` | `-I"$ROOT/lib"` |
| `tests/server/xymond-gap-validity.sh:75` | `-I"$ROOT/lib"` |
| `tests/server/xymond-holdtime-bridge.sh:72` | `-I"$ROOT/lib"` |

`89499737d` ("tests: reach in-tree headers with -iquote, not -I") converted ten tests on
2026-08-11. These three arrived two days later in `0ee9b6f39`, written with the old
spelling, so the sweep could not have covered them -- which is the argument for a rule
rather than another sweep.

**The rule**: `include-shadow` flags an `-I` whose directory is under `$ROOT`, with the
non-vacuity probe entry that keeps it honest (4 -> 5 constructs). Generated directories are
deliberately not flagged: `rrd-api-compat.sh` puts its mock `<rrd.h>` on `-I` precisely so
an angle include finds it, and says so in a comment -- I broke that in a first pass and the
suite caught it.

Evidence: with one `-I` put back, the lint reports
`[include-shadow] tests/server/xymond-gap-validity.sh:75`; with the three fixed it passes.
Suite: 0 failed.

Found by reading an old lane run (`bonomani/xymon` run 31796174740, 2026-08-14): 48 jobs,
the only failures were macOS 14/15/26 Server, all three from this.

Note for whoever merges second: #360 also adds a rule to this file, so one of the two will
need its probe count bumped by one.

### DIFF
--- a/tests/buildsystem/test-suite-portability.sh
+++ b/tests/buildsystem/test-suite-portability.sh
@@ -88,6 +88,18 @@ report "gnu-regex" '(grep|sed|awk|expr)[^|]*\\(b|<|>|w|W|s|S|d|D)([^[:alnum:]]|$
 report "root-lane" 'chmod +[-+=rwx]*[0-7]?[0-5][0-7][0-7]([^0-9]|$)|chmod +[ugoa]*-w' \
 	"the lanes run as root; a permission bit will not force this failure"
 
+# An -I directory is searched for <angle> includes too, so a source directory
+# put there can capture a system header. macOS makes that concrete: the SDK's
+# stdio.h includes <Availability.h>, the filesystem is case-insensitive, and
+# lib/availability.h answers instead -- every system type after it is unknown,
+# and the three tests that reached lib/ this way did not build on the macOS
+# lanes. -iquote is searched for "quoted" includes only, which is all an
+# in-tree header needs. Generated directories are deliberately NOT flagged:
+# rrd-api-compat.sh puts its mock <rrd.h> on -I precisely so that an angle
+# include finds it.
+report "include-shadow" '\-I *"?\$\{?ROOT' \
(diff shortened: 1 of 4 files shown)


SoundGoof (approved):
No blocking findings.

The three `-I` to `-iquote` conversions correctly prevent in-tree headers
from shadowing system headers while retaining access to quoted project
headers.

I tested the new expression with native `grep -E` on OpenBSD 7.9 and
FreeBSD 14.3. Both systems:

- matched `-I"$ROOT/lib"`
- matched `-I "${ROOT}/lib"`
- ignored `-I"$work/include"`

The focused portability test passes, the branch merges cleanly into current
`main`, and all five GitHub checks are green.

GNU grep emits a harmless `stray \ before -` warning from the non-vacuity
probe. The expression works on the tested BSD systems, so I do not consider
that blocking. `[-]I *"?[$][{]?ROOT` would avoid the GNU warning and express
the literals using strictly portable bracket notation.

Approved.

PR #371, opened 2026-08-15
xymongen: bound the WML paths, and report a path that does not fit

bonomani:
`wmlgen.c` builds five file paths into fixed buffers from `$XYMONWEB`, a hostname and a column name. gcc warned about four of them on every Linux lane; the fifth was silent and worse — `sprintf(wmldir, "%s/wml", webdir)`, unbounded, from a configured value. Each path is now built with a checked `snprintf`, and a path that does not fit skips that output instead of writing a truncated one.

Everything below came out of review, and each is measured:

| | before | after |
|---|---|---|
| the daemon request | `sprintf()` into a 1 KiB stack buffer from a hostname `hosts.cfg` lets run to 16 KiB — a stack overflow reached from the config, before any path check | allocated to the size of the names; a request has no natural limit, so there is nothing here to refuse |
| a skipped status card | still advertised: the host card links from `t->onwap`, which a refused write left set | `generate_wml_statuscard()` reports whether it wrote; the caller clears `onwap` and withholds `anywaps` |
| the split non-green card | wrote `Next -> nongreen-<n>.wml` **before** the check deciding whether that name fits | the name is settled first |
| a card, host card or non-green card that could not be written whole | published anyway — a full filesystem surfaces at `fclose()`, not `fopen()` | removed, not advertised |
| the log buffer on "status not available" | leaked — xymond answering with no log is an empty string, not `NULL` | released |
| `delete_old_cards()` | acted on `st_mode` after an unchecked `stat()`; leaked the directory handle on a failed `chdir()` | both checked |
| `symlink()` for `index.wml` | test inverted: reported a failure every time it worked, silent when it did not | reported when it fails, `EEXIST` excepted |

**Test:** `tests/web/wmlgen-path-bounds.sh` drives the real `generate_wml_statuscard()`, extracted rather than reimplemented, over an ordinary card, a 1200-byte hostname, a path at the limit, a status with no log, and a write that cannot finish (`RLIMIT_FSIZE`). It compares the whole daemon request, not its prefix — a bound that truncated the names would pass a prefix check while asking about the wrong test — and the sanitized run fails on leaks as well as overflows.

The limit case builds a *deep directory* rather than one long name: sized as one component it exceeds `NAME_MAX`, so `fopen()` fails with `ENAMETOOLONG` whatever the guard does, and the case passed with the guard disabled. Measured, then fixed.

**Not taken:** devel's answer to the warnings (`3686fbeef`) is `#pragma GCC diagnostic ignored "-Wformat-overflow"` around the `sprintf` calls. It silences the message and keeps the truncation, it names the wrong warning group for main's bounded calls, and clang — the compiler on the BSD and macOS lanes — rejects that group, so porting it would *add* a warning per site there.

**Not covered, rather than implied:** `do_wml_cards()` itself, so the caller's `onwap`/`anywaps` handling is read, not tested.

### DIFF
--- a/tests/web/wmlgen-path-bounds-harness.c
+++ b/tests/web/wmlgen-path-bounds-harness.c
@@ -1,0 +1,205 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ * Driver for the real generate_wml_statuscard(), extracted from
+ * xymongen/wmlgen.c by the test beside this file.
+ *
+ * The function is static and sits on top of xymongen's whole object graph, so
+ * what it needs is stubbed rather than linked: the daemon exchange returns a
+ * canned log, and the types carry only the fields the function reads. What is
+ * NOT stubbed is the function itself -- it is the production text, so a fix
+ * that stops holding has nowhere to hide.
+ */
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <ctype.h>
+#include <limits.h>
+#include <errno.h>
+#include <unistd.h>
+#include <signal.h>
+#include <sys/resource.h>
+
+#define MAX_LINE_LEN 16384
+#define XYMON_TIMEOUT 30
+#define XYMONSEND_OK 0
+
+typedef struct column_t { char *name; } column_t;
+typedef struct entry_t { column_t *column; int color; int acked; struct entry_t *next; } entry_t;
+typedef struct host_t { char *hostname; } host_t;
+typedef struct sendreturn_t { int dummy; } sendreturn_t;
+
+static char *wmldir;
+static char *timestamp = "Mon Jan  1 00:00:00 2026";
+
+/* The canned log: a header line, then one body line the parser copies out. */
+static const char *CANNED = "green Mon Jan  1 00:00:00 2026\nplain body line\n";
+
+static sendreturn_t *newsendreturnbuf(int a, void *b) { (void)a; (void)b; return (sendreturn_t *)calloc(1, sizeof(sendreturn_t)); }
+static char *expected_req;	/* what this call must have been handed */
+static int req_mismatch;
+
+static int sendmessage(char *req, void *a, int b, sendreturn_t *c)
+{
+	(void)a; (void)b; (void)c;
+	/*
+	 * The whole request, not its prefix: a "fix" that bounded this by
+	 * truncating the names would stop overflowing and still ask the daemon
+	 * about the wrong test, which a prefix check cannot tell apart.
+	 */
+	if (expected_req && (strcmp(req, expected_req) != 0)) {
+		fprintf(stderr, "FAIL: request was %zu bytes, expected %zu\n",
+			strlen(req), strlen(expected_req));
+		req_mismatch = 1;
+	}
+	return XYMONSEND_OK;
+}
+
+static void expect_req(const char *host, const char *col)
+{
+	free(expected_req);
+	expected_req = (char *)malloc(strlen(host) + strlen(col) + sizeof("xymondlog ."));
+	sprintf(expected_req, "xymondlog %s.%s", host, col);
+}
+static int empty_answer;	/* xymond has no log for this test */
+static char *getsendreturnstr(sendreturn_t *s, int a)
+{
+	(void)s; (void)a;
+	/* Still an allocation: "no status" is an empty string, not a NULL, which
+	   is exactly the case whose early return used to drop it. */
+	return strdup(empty_answer ? "" : CANNED);
+}
+static void freesendreturnbuf(sendreturn_t *s) { free(s); }
+static void wml_header(FILE *fd, const char *name, int n) { (void)name; (void)n; fprintf(fd, "<wml>\n"); }
+static void errprintf(const char *fmt, ...) { (void)fmt; }
+static void *xmalloc(size_t n) { void *p = malloc(n); if (!p) abort(); return p; }
+static void xfree(void *p) { free(p); }
+
+#include "wmlgen-statuscard.inc"
+
+static entry_t *mkentry(const char *col)
+{
+	entry_t *e = (entry_t *)calloc(1, sizeof(entry_t));
+	e->column = (column_t *)calloc(1, sizeof(column_t));
+	e->column->name = strdup(col);
+	return e;
+}
+
+static char *repeat(char c, size_t n)
+{
+	char *s = (char *)malloc(n + 1);
+	memset(s, c, n); s[n] = '\0';
+	return s;
+}
+
+int main(int argc, char **argv)
+{
+	host_t host;
+	entry_t *e;
+	int rc, fail = 0;
+
+	if (argc < 3) { fprintf(stderr, "usage: %s <wmldir> <deep-wmldir>\n", argv[0]); return 2; }
+	wmldir = argv[1];
+
+	/* 1. An ordinary card is written, and says so. */
+	char *hn;
+
+	hn = strdup("short.host");
+	host.hostname = hn;
+	e = mkentry("conn");
+	expect_req(hn, "conn");
+	rc = generate_wml_statuscard(&host, e);
+	if (rc != 1) { fprintf(stderr, "FAIL: an ordinary status card reported failure (%d)\n", rc); fail = 1; }
+	free(hn);
+
+	/* 2. A hostname far past the old 1 KiB request buffer. Nothing here
+	   asserts a value: the assertion is that ASan sees no write past the
+	   buffer the request is built in, which happens before any path check. */
+	hn = repeat('h', 1200);
+	host.hostname = hn;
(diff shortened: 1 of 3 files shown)


SoundGoof:
## Blocking findings

### 1. The request buffer still permits a stack-buffer overflow

`generate_wml_statuscard()` constructs the daemon request before performing any
of the new path checks:

```c
char xymondreq[1024];

sprintf(xymondreq, "xymondlog %s.%s",
        host->hostname, entry->column->name);
```

The input path does not impose a matching 1 KiB limit. `loadlayout.c` accepts a
hostname using `%s` into a `MAX_LINE_LEN` buffer, which is 16 KiB.

I called the real production function under ASan with a 1,200-byte hostname.
ASan reported:

```text
ERROR: AddressSanitizer: stack-buffer-overflow
WRITE of size 1216
#2 generate_wml_statuscard xymongen/wmlgen.c:100
```

The overflow occurs before network access and before the new filename checks.
Replace this `sprintf()` with checked construction or a dynamically allocated
request. Add a regression using an oversized hostname or combined
hostname/column value.

This vulnerability predates the PR, but this patch hardens the same function
against these input lengths while leaving the earlier and smaller buffer
unprotected.

### 2. A rejected status-card path is still advertised

`generate_wml_statuscard()` reports and skips an overlong status-card path, but
does not tell its caller that generation failed. `do_wml_cards()` consequently
leaves `t->onwap` set and sets `anywaps`, then writes a host card containing a
link to the file that was deliberately skipped.

This is reachable without long host or column values. A `wmldir` near
`PATH_MAX` can still fit:

```text
<wmldir>/<host>.wml
<wmldir>/nongreen.wml.tmp
```

while the additional column component makes this exceed the limit:

```text
<wmldir>/<host>.<column>.wml
```

Return success from `generate_wml_statuscard()`. On failure, clear `t->onwap`
and only set `anywaps` after successful generation. Add a boundary regression
where the host and non-green paths fit but the status-card path does not.

## Validation

Reviewed exact head `a602c28bcad4d3d38326ce494f6d3b65112267c4`.

A direct `gcc -O2 -Wall -Wextra` compilation confirms that the PR removes all
warnings from `wmlgen.c`; current `main` emits the expected path-truncation
warnings.

The stack overflow was reproduced against the real PR function under ASan.
The complete build succeeds, all five GitHub checks are green, and the
regression suite reports 70 passed, 3 skipped, and 0 failed.

bonomani:
Both fixed, and both verified here rather than taken on trust.

**1. The request buffer.** Confirmed by reading: `sprintf` at line 100, `sendmessage` at 101, the first path check only at 122 — the ordering you describe. Reachability too: `loadlayout.c:423` declares `hostname[MAX_LINE_LEN]` and line 592 fills it with `sscanf(… "%s", hostname)`.

I did not bound it. A daemon request has no natural length limit, so a 1 KiB one is a number nobody chose, and enforcing it would turn a crash into a refusal of something legitimate. It is allocated to the size of the names instead. That buffer was the only `char xymondreq[…]` in the tree, so there is no family to sweep.

One qualification for prioritising it: the source is `hosts.cfg`, not the network. A config-triggered stack overflow, not a remote one.

**2. The advertised card.** Your mechanism is exact, including the part that looks like it contradicts it: line 310 does clear `onwap`, but in the `else` branch — the entries that were not selected. The branch that calls the function leaves it set, and line 368 emits a link for every `onwap`.

Worth stating plainly: **this one is a defect the PR introduced.** Before it the path was truncated and the file written under the truncated name; choosing to skip the write instead is right, but the choice was only half made. `generate_wml_statuscard()` returns whether it wrote the card; the caller clears `onwap` and withholds `anywaps` when it did not.

**Test.** The PR had none; `tests/web/wmlgen-path-bounds.sh` drives the real function, extracted rather than reimplemented, over an ordinary card, a 1200-byte hostname, and your boundary — host card path fits, status card path does not. Mutations:

| mutation | result |
|---|---|
| the 1 KiB buffer restored | `ERROR: AddressSanitizer: stack-buffer-overflow` — your report |
| a refused path reported as written | `FAIL: a status-card path that does not fit was reported as written` |

The first only fails under ASan, so that half is gated on `asan_usable()` and says so when it cannot run.

Suite on a built tree: 74 passed, 0 skipped, 0 failed. One commit, amended.

bonomani:
A second review pass over the same file found four more, and one of them is your finding 2 at a site neither of us looked at.

**The split non-green card had the same defect.** `wmlgen.c:424-431` wrote `Next -> nongreen-<n>.wml` and closed the card *before* the `snprintf` that decides whether that name fits — so the refusal I added left the previous card pointing at a file nothing would create. Same shape as the status card, one site further on, and introduced by this PR for the same reason. The name is settled before the link is written now.

Three others, all pre-existing:

| | |
|---|---|
| `wmlgen.c:121-124` | the "status not available" return dropped the log buffer. xymond answering with no log is an empty string, not a `NULL`, so there is an allocation to release — on a path taken whenever a test has no history yet |
| `wmlgen.c:268-277` | the card was reported as written whatever the writing did. A full filesystem surfaces at `fclose()`, not at `fopen()`, so a card cut short was still advertised. Both the stream's error flag and `fclose()` are checked, and a card not written whole is removed rather than left for a link that will not come |
| `wmlgen.c:53-56` | `delete_old_cards()` ignored `stat()` and then read `st_mode`. An entry that disappears between `readdir()` and `stat()` left the unlink decision resting on whatever was on the stack |

Not taken: the same pass called the `strcpy()` of `d_name` into a `PATH_MAX` buffer an overflow. `d_name` is bounded by `NAME_MAX`, which no supported platform sets above `PATH_MAX`.

**What this says about the test I sent you.** Three of those four fixes passed mutation without being detected — the test asserted them only in the sense that it did not contradict them. It now drives a status with no log, and a write that cannot finish (`RLIMIT_FSIZE`, `SIGXFSZ` ignored so the failure returns as an error rather than a signal), and it compares the whole daemon request instead of its prefix: a bound that truncated the names would have passed a prefix check while asking xymond about the wrong test. The sanitized run fails on leaks as well as overflows. All four mutations now fail it.

Still not covered, and I would rather say it than imply otherwise: `do_wml_cards()` itself. The caller's `onwap`/`anywaps` handling is read, not tested — driving it needs the host list and the layout, which is a great deal of harness for one branch.

Suite on a built tree: 74 passed, 0 skipped, 0 failed. One commit, amended.

bonomani:
Both of your findings are fixed, and two later passes over the same file found seven more. Merged here; the PR body carries the full state.

**Yours.** The 1 KiB request buffer is gone — a request has no natural length limit, so it is allocated to the size of the names rather than bounded. Reachability is `hosts.cfg`, not the network: `loadlayout.c:423` fills a `MAX_LINE_LEN` hostname. And the skipped status card is no longer advertised: `generate_wml_statuscard()` reports whether it wrote, and the caller clears `onwap` and withholds `anywaps`. Your second finding was one this change introduced — choosing to skip the write without telling the caller.

**The same defect, one site further on:** the split non-green card wrote `Next -> nongreen-<n>.wml` and closed the card *before* the check deciding whether that name fits. Neither of us looked there.

**Five more:** `ferror(fd) || fclose(fd)` short-circuited, so a stream already in error never reached the close — that leak was in my own fix; the host card and the non-green card were published without checking they were written whole; a failed host-card open returned with the non-green card still open; `delete_old_cards()` returned on a failed `chdir()` with the directory handle open; and `if (!symlink(...))` reported a failure every time it *worked*, silent when it did not.

**And my boundary test proved nothing.** Sizing one component against `PATH_MAX` makes it exceed `NAME_MAX`, so `fopen()` failed with `ENAMETOOLONG` whatever the guard did — measured, it passed with the guard disabled. It builds a deep directory now, ordinary components, the path carrying the length. Every fix above fails the test when reverted.

Rebased on `f2b28fb75`. Suite on a built tree: 74 passed, 0 skipped, 0 failed. Not covered, said rather than implied: `do_wml_cards()` itself.

SoundGoof (approved):
No blocking issues found.

The rewritten head fixes both previously reported defects:

- The `xymondlog` request is dynamically sized, avoiding the confirmed
  stack overflow for long hostnames.
- Status-card generation returns success or failure. Failed cards clear
  `onwap`, so host cards no longer link to files that were not written.

The extracted-production-function harness checks full request contents,
path refusal, empty responses, short writes, and memory safety under ASan.

One pre-existing issue remains in the same function. A near-16 KiB status
line containing escapable characters can overflow `lineout` as each input
character expands to as many as six output bytes. A scratch ASan case using
ampersands reproduced a stack-buffer overflow. This parser was not changed
by the PR, so I consider it separate follow-up work.

Validation at `ac77e225b650228384bc6f0e24c8a5569bb07dcb`:

- Focused WML bounds test passed.
- Portability guard passed.
- `xymongen` built successfully.
- Full post-build suite: 57 passed, 17 skipped, 0 failed.
- `git diff --check` passed.
- All five GitHub checks are green.

Issue #372, opened 2026-08-15
GCC warnings on main: inventory, and why devel 3686fbeef is not the source of fixes


bonomani:
Triage of the GCC warnings still on `main`, and of devel's warning commit as a possible
source of fixes. Prompted by an old lane run
([31796174740](https://github.com/bonomani/xymon/actions/runs/31796174740), 2026-08-14): 48
jobs, the only failures were macOS 14/15/26 Server.

## devel `3686fbeef` has nothing left for main

"Fix a large number of GCC build warnings and issues" (Tom Schmidt with Ralph Mitchell,
2023-10-06). **272 added lines, 168 of them `#pragma GCC diagnostic ignored` scaffolding.**
PR #113 (merged) already took 15 of its 33 files as a "safe, no-behavior-change subset".
Of the 16 that remain:

- **11 are pure suppression** (`client/logfetch.c`, `common/xymonlaunch.c`, `lib/misc.c`,
  `web/svcstatus.c`, `xymond/combostatus.c`, `do_rrd.c`, `trimhistory.c`,
  `xymond_hostdata.c`, `xymongen/pagegen.c`, `util.c`, `wmlgen.c`). Suppression is not a
  fix, and it is actively harmful on our lanes: clang rejects the group --
  `warning: unknown warning group '-Wformat-overflow', ignored [-Wunknown-warning-option]`
  (measured) -- so each ported pragma **adds** a warning on FreeBSD, OpenBSD and macOS.
- **5 have real changes, and none applies**: `lib/sig.c`, `lib/links.c`,
  `lib/loadalerts.c` bolt `dst[sizeof(dst)-1] = '\0'` onto a `strncpy`, where main already
  uses `snprintf(dst, sizeof(dst), "%s", …)` -- main's form is the better one.
  `lib/tcplib.c` and `lib/stdopt.c` **do not exist on main**; the OpenSSL guard from
  `tcplib.c` has its main equivalent in `xymonnet/contest.c`, already guarded across more
  versions (`:409`, `:486`, `:493`, `:499`, `:512`, `:761`).

So: nothing to backport, and #106 need not carry an entry for it beyond this verdict.

## What main does still warn about

| site | warning | status |
|---|---|---|
| `xymongen/wmlgen.c` ×5 | `-Wformat-truncation`, plus an unbounded `sprintf` into `wmldir` | **fixed in #371** |
[...]

PR #373, opened 2026-08-15
xymonfetch: bound the message writes, and size addrstring to what it holds

bonomani:
Two `-Wformat` warnings on every Linux lane, both from `addrstring()`'s declared type: `static char[100]` lets gcc assume a 99-byte return, while the content is an IPv4 address plus optional `":port"` — at most 21 characters. **Neither warning is a live overflow**; the callers were `sprintf`, which *would* have overflowed had the return ever been as long as its type permits.

- `addrstring()` is sized `INET_ADDRSTRLEN + sizeof(":65535")` — the bound is now the truth, and both writes into it are bounded.
- The caller uses `snprintf`. No behaviour change: the text has always fitted.

Verified: `gcc -O2 -Wall -Wextra` goes from two `-Wformat` warnings to none (the tree-wide unused-`rcsid` stays). Every caller passes the result straight into a printf-family format; none stores the pointer.

Part of the #372 inventory; same family as #371, where the input really was unbounded — here it is hygiene: it stops these warnings from hiding a real one the day `addrstring()` learns a longer address family.

### DIFF
--- a/xymond/xymonfetch.c
+++ b/xymond/xymonfetch.c
@@ -101,10 +101,15 @@ void sigmisc_handler(int signum)
 char *addrstring(struct sockaddr_in *addr, int includeport)
 {
 	int n;
-	static char res[100];
-
-	n = sprintf(res, "%s", inet_ntoa(addr->sin_addr));
-	if (includeport) sprintf(res+n, ":%d", ntohs(addr->sin_port));
+	/* Sized for what it actually holds: an IPv4 address and an optional
+	   ":port". It was char[100], which is not wrong but tells the compiler
+	   the return may be 99 characters -- so every caller that formats it into
+	   a fixed buffer warns about a truncation that cannot happen. The bound
+	   is now the truth, and the writes below respect it either way. */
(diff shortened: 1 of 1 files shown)


SoundGoof:
### Finding

**Minor: Remove the trailing whitespace in the changed `snprintf` call**

`git diff --check` reports trailing whitespace on the new `sourcemsg`
format line in `xymond/xymonfetch.c`.

No functional issues found.

### Validation

- The PR is one commit based directly on current `main`.
- It merges cleanly.
- A configured `-O2 -Wall -Wextra` comparison confirms the two format
  diagnostics disappear.
- The only remaining warning is the existing unused `rcsid`.
- The buffer accommodates the maximum IPv4 address and port, including the
  terminating null byte.
- All `addrstring()` callers use the returned value immediately.

bonomani:
Fixed — amended in place, `git diff --check` is clean against the base.

For the record, the space was inherited: the old `sprintf` line already ended with one, and the `snprintf` rewrite copied it along. Touched line, so cleaned.

Thanks for the full validation pass.

SoundGoof (approved):
lgtm

PR #378, opened 2026-08-16
client: name the OS primitives, so one lane tests every client

bonomani:
**1 own commit**; the other is #366, which this is stacked on. Answers the design question on #361 (@SoundGoof): should Linux production code change so its tests can run elsewhere?

Yes, and narrowly. Only the OS primitives move behind a name — `fs_mounts`, `fs_filesystems`, `fs_procname` — and a test replaces one *by name* (`fsf_stub_helper`) instead of rewriting a path only one OS has. That is what lets a single lane exercise all five clients; before this, two tests rewrote `/proc` paths with `sed` and so only ever exercised the client that has `/proc`.

It has already paid for itself twice, both found because the tests could suddenly run where they could not before:

- `ps` is absent from a stock Debian container (`debian:12` has none; `procps` is not Essential). The PID-reuse guard read an empty name, concluded its own `df` had finished, and started another against the dead server every cycle. An unreadable name now means "still running": one cycle of `unavailable:` is bounded and visible, an extra `df` in D-state cannot be killed.
- The native mount test found NetBSD and OpenBSD spelling mount(8) differently from FreeBSD — fixed in #366.

**The cost is that nothing exercises the primitives any more**, since every other test stubs them. Two `uname`-guarded native tests do: `fs-procname.sh` (does this host name a process the way its client assumes?) and `fs-mounts.sh` (this host's own `fs_mounts()` against the real system, plus a check on every lane that no client asks `mount(8)` to refresh — FreeBSD's `-v` means `MNT_WAIT`, which blocks on the dead mount being reported). They are the only two allowed to guard on `uname`, and `tests/buildsystem/test-suite-portability.sh` enforces that: a `uname` guard without a `# native-primitive:` header is reported, as is a `sed` that rewrites a `/proc` path.

`probe_dir_is_local()` keeps the helper's status rather than the pipeline's: awk succeeds on no input, so "I cannot read the mount list" would otherwise read as "this directory is local" — in the guard whose whole job is to not touch a directory that may sit on the wedged mount.

| mutation | caught by |
|---|---|
| the unreadable-name branch removed (a container without `ps`) | `with the name unreadable the wedged df must be assumed ours: expected '3', got '4'` |
| the client reads the command line instead of the name | `fs_procname() never named the fixture after 10s` |
| `\040` no longer decoded in a Linux mount point | `an escaped space in a mount point is decoded, not passed through` |
| `probe_dir_is_local` back on the pipeline | `an unreadable mount list must not read as a local probe directory` |
| a test rewrites a `/proc` path, or guards on `uname` undeclared | `[path-rewrite]`, `[uname-guard]` |

Suite: 81 passed, 0 skipped, 0 failed on a built tree.

### DIFF
--- a/client/xymonclient-darwin.sh
+++ b/client/xymonclient-darwin.sh
@@ -156,6 +156,15 @@ probe_dir_is_local()
 	'
 }
 
+# fs_procname PID : the command name of a running process. ps is POSIX and
+# reads process state only, so it cannot touch a wedged mount. Named so that
+# df_sentinel() stays identical across the five clients, and so a test
+# replaces the primitive rather than the code that uses it.
+fs_procname()
+{
+	ps -o comm= -p "$1" 2>/dev/null | tr -d '[:space:]'
+}
+
 df_sentinel()
 {
 	_tag="$1"; shift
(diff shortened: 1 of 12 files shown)


bonomani:
@SoundGoof — review welcome when you get to it. This is the answer to your design question on #361: should Linux production code change so its tests can run elsewhere?

The answer we landed on is hybrid, and it is narrow. Only the OS primitives move: `fs_mounts`, `fs_filesystems`, `fs_procname`. A test replaces one *by name* (`fsf_stub_helper`) instead of rewriting a path only one OS has, which is what lets a single lane exercise all five clients. Everything else stays where it was.

It has already paid for itself twice, both found because the tests could suddenly run somewhere they could not before:

- `ps` is not in a stock Debian container (`debian:12` has none; `procps` is not Essential). The PID-reuse guard read an empty name, concluded its own `df` had finished, and started another one against the dead server every cycle.
- The native mount test found that NetBSD and OpenBSD spell mount(8) as `MP type <fstype> (opts)`, not `MP (TYPE, opts)`. Enumeration dropped every row there, so with `DF_LOCAL_ONLY=no` the remote filesystems would have disappeared from the report rather than being guarded.

The cost side is real and worth your eye: the primitives themselves are now behind a stub in every other test, so nothing exercised them. Two `uname`-guarded native tests do — `fs-procname.sh` and `fs-mounts.sh` — and they are the only two in the suite allowed to guard. `tests/buildsystem/test-suite-portability.sh` enforces both halves: a rewritten `/proc` path in a test is reported, and so is a `uname` guard in a test that does not declare which primitive it exercises.

**2 commits**: this one and #366 below it. Rebased on current `main`; suite 81 passed, 0 skipped, 0 failed on a built tree.

SoundGoof (dismissed):
No blocking issues found.

The refactor keeps platform-specific operations in named helpers while allowing
the common filesystem and sentinel behavior to run under fixtures on every
lane. Native tests separately guard the real mount and process-name primitives.

The conservative unreadable-process-name behavior is appropriate: retaining one
bounded `unavailable` result is safer than launching additional `df` processes
against a wedged mount.

I ran the Linux native tests and all BSD/macOS fixture tests. They passed. All
five production client scripts also parse with `/bin/sh`, the patch passes
`git diff --check`, CI is green, and the branch merges cleanly with current
`main`.

Approve.

SoundGoof (dismissed):
lgtm

SoundGoof (approved):
No findings.

The force-pushed head `68b00da1` is a pure rebase of the previously reviewed
`ae0257a6`. Both have the identical stable patch ID `29a2e30e`; no client or
client-test code changed.

The earlier native validation therefore remains applicable:

- FreeBSD 14.3: passed
- NetBSD 11.0: passed
- OpenBSD 7.9: passed

All five CI checks are green, `git diff --check` passes, and the branch now
contains current `main`.

The remaining coverage gap is native macOS testing. The Darwin fixture tests
pass in CI, but its real `fs_mounts` and `fs_procname` implementations were
not exercised on macOS.

PR #380, opened 2026-08-16
xymond: let xymondboard select oldcolor, and add previouscolor (#355)

bonomani:
**1 commit**, on `main` — not stacked on #355, and it makes #355 unnecessary: the manual text lands here, corrected.

`oldcolor` is not a `xymondboard` field. It has been on xymond's internal channels forever, but `boardfieldnames[]` has no entry for it, so `xymondboard fields=hostname,testname,color,oldcolor` returns an empty column, and `xymondlog`'s default list does not carry it either. #355 was about to document it as one.

Both are selectable now, and a second color comes with them, because the one that exists does not answer the question people ask of it:

| | assigned | after green, red, red |
|---|---|---|
| `oldcolor` | every message | `red` |
| `previouscolor` | only on a real change | `green` |

`previouscolor` is what a display or an alert rule wants — what was this before it went bad (@SoundGoof).

**Persistence.** `previouscolor` rides the `.flapstate.` checkpoint record, appended *after* the flap ring so nothing shifts: a reader that predates it stops at its own last case, and a checkpoint written before it restores as "no previous color", which is what those statuses had. A monitoring server restarts during the incident an operator is looking at; a field whose job is to say what came before is worth nothing if it forgets exactly then.

That append is why `xymond-checkpoint-roundtrip.sh` changes. It backdated the flap ring through `$NF`, which now rewrites `previouscolor` instead — and still produces a line ending in the value it wrote, so its own guard passed while the edit changed nothing. It names field 11 now, and fails loudly if the record ever gets shorter.

**The manual page also drops two claims the code contradicts**, both from #355's text:

- "a status whose color has never changed reports `none`" — only its *first* report does. Send green twice and `oldcolor` is green, though nothing ever changed.
- "the color this test reported" — it is the color xymond *recorded*. A disabled or in-downtime test reports green and is recorded blue, and blue is what comes back.

| mutation | caught by |
|---|---|
| `previouscolor` assigned on every message | `a repeated color moves oldcolor, not previouscolor: expected previouscolor=none, got green` |
| the checkpoint field dropped on restore | `a restored status keeps the color it held before the change: expected previouscolor=red, got none` |

Suite: 76 passed, 0 skipped, 0 failed on a built tree. Manual page HTML regenerated with `build/makehtml.sh`.

### DIFF
--- a/common/xymon.1
+++ b/common/xymon.1
@@ -203,6 +203,19 @@ The name of the host
 The name of the test
 .IP "\fBcolor\fR"
 Status color (green, yellow, red, blue, clear, purple)
+.IP "\fBoldcolor\fR"
+The color xymond recorded for this test at the previous report. xymond assigns
+it on every message, not only on a change, so a test repeating its color
+reports that same color here. It is the color xymond \fIrecorded\fR, which is
+not always the one the test sent: a disabled or in-downtime test is recorded
+blue whatever it reports. Until a status has been reported twice this is
+\fBnone\fR, the "no previous color" marker.
+.IP "\fBpreviouscolor\fR"
+The color this test held before its most recent change. Unlike \fBoldcolor\fR
+it does not follow a repeated color, so a test that went green, red, red
+reports color=red, oldcolor=red, previouscolor=green \- it answers "what was
+this before it went bad". \fBnone\fR until the first change, and it survives a
+xymond restart.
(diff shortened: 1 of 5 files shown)


bonomani:
Superseded: this landed in #355 itself, where the request for `previouscolor` was made.

PR #382, opened 2026-08-16
xymond: publish the color a test itself reported, as prevgapcolor

bonomani:
**1 commit.** `xymondboard fields=…,prevgapcolor` — the color a test itself reports, while xymond
is recording one of its own.

| | `color` | `prevgapcolor` |
|---|---|---|
| normal | what the test reported | `none` |
| disable, DOWNTIME, modifier, delayed change, silence | xymond's choice | the test's own word |

It follows the reports arriving under the override — a green host whose disk fills during DOWNTIME
answers `blue,red`, not the green that stood before the window — and a test gone silent keeps its
last real word. Two fields, two jobs: `pregapcolor` keeps hold-time bridging (the color to resume
to), `gapreportedcolor` is published. One subtlety the review surfaced ran deeper than the gap
code: the DOWNTIME rewrite happens in `get_hts()`, so `handle_status()` never sees the reported
red — the raw color is now kept (`lastreportedcolor`) just before that rewrite.

The checkpoint carries the word on the gap record, riding the existing suffix chain
(`gapstart:validity:color`) so older files still parse; absent, it restores as the pre-gap color,
and a suffix on a record whose gap field says `none` restores as `none` — no gap, no close to ever
clear it. DOWNTIME substitutes only alert colors: a green report shows through and answers `none`
(manpage says so now).
A gap is saved for as long as the substitution lasts, not only while its bridge window holds, so a
restart during a longer override keeps answering. The generated `info`/`trends`/`clientlog` rows
answer `none`, not the `green` their zeroed records spelled (same fault as `prevchangecolor`, #355).

| case | assertion |
|---|---|
| green host, red reports under DOWNTIME | `prevgapcolor=red` — fails on the previous head with `green` |
| hostile checkpoint: word suffix, no gap | restores `none`, not a sticking `red` |
| a later report under the window | refreshes the answer |
| checkpoint + restart mid-override | the word survives; a day-old gap still refuses the hold-time |
| generated `info`/`trends` rows | `none/none`, not `green/green` |

Suite: 86 passed, 0 skipped, 0 failed on a built tree. Manpage reworded at both field lists, HTML
regenerated with `build/makehtml.sh`. Base is current `main`.

### DIFF
--- a/common/xymon.1
+++ b/common/xymon.1
@@ -211,12 +211,19 @@ a test repeating its color reports that same color here. Also accepted as
 in-downtime test is recorded blue whatever it reports. \fBnone\fR until a
 status has been reported twice.
 .IP "\fBprevchangecolor\fR"
-The color this test held before its most recent change. Also accepted as
-\fBpreviouscolor\fR. It moves only when the
+The color this test held before its most recent change. It moves only when the
 color changes, so a test that went green, red, red reports color=red,
 prevreportcolor=red, prevchangecolor=green \- it answers "what was this before
 it went bad". \fBnone\fR until the first change, and it survives a xymond
-restart.
+restart. Also accepted as \fBpreviouscolor\fR.
+.IP "\fBprevgapcolor\fR"
+The test's own color while xymond is recording something other than what it
+reports; \fBnone\fR at every other time. It follows the reports arriving
+under the substitution, and a test gone silent keeps its last word. That
+covers a silence, a disable, a modifier, a delayed color change and a
+DOWNTIME window hiding an alert color -- a green report shows through
+DOWNTIME, and the field answers \fBnone\fR: wherever they differ,
+\fBcolor\fR is xymond's and \fBprevgapcolor\fR is the test's own.
 .IP "\fBtestflags\fR"
 For network tests, the flags indicating details about the test (used by xymongen).
 .IP "\fBlastchange\fR"
@@ -342,12 +349,19 @@ a test repeating its color reports that same color here. Also accepted as
 in-downtime test is recorded blue whatever it reports. \fBnone\fR until a
 status has been reported twice.
 .IP "\fBprevchangecolor\fR"
-The color this test held before its most recent change. Also accepted as
-\fBpreviouscolor\fR. It moves only when the
+The color this test held before its most recent change. It moves only when the
 color changes, so a test that went green, red, red reports color=red,
 prevreportcolor=red, prevchangecolor=green \- it answers "what was this before
 it went bad". \fBnone\fR until the first change, and it survives a xymond
-restart.
+restart. Also accepted as \fBpreviouscolor\fR.
+.IP "\fBprevgapcolor\fR"
+The test's own color while xymond is recording something other than what it
+reports; \fBnone\fR at every other time. It follows the reports arriving
+under the substitution, and a test gone silent keeps its last word. That
+covers a silence, a disable, a modifier, a delayed color change and a
+DOWNTIME window hiding an alert color -- a green report shows through
+DOWNTIME, and the field answers \fBnone\fR: wherever they differ,
+\fBcolor\fR is xymond's and \fBprevgapcolor\fR is the test's own.
 .IP "\fBflags\fR"
 For network tests, the flags indicating details about the test (used by xymongen).
 .IP "\fBlastchange\fR"
--- a/docs/manpages/man1/xymon.1.html
+++ b/docs/manpages/man1/xymon.1.html
@@ -256,12 +256,19 @@ Section: User Commands (1)<br/>Updated: Version 4.3.30: 4 Sep 2019<br/><a href="
       a disabled or in-downtime test is recorded blue whatever it reports.
       <b>none</b> until a status has been reported twice.</dd>
   <dt id="prevchangecolor"><a class="permalink" href="#prevchangecolor"><b>prevchangecolor</b></a></dt>
-  <dd>The color this test held before its most recent change. Also accepted as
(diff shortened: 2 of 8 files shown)


SoundGoof:
## Review

###  `prevgapcolor` changes across restart while substitution continues

`xymond/xymond.c:3621` publishes `pregapcolor` directly for the lifetime
of the running daemon.

Checkpoint writing deliberately omits that value after the hold-time
bridge window expires (`xymond/xymond.c:5156`).

Consequently, a long-running override can report:

```text
color=blue prevgapcolor=red
```

and change after an unrelated restart to:

```text
color=blue prevgapcolor=none
```

The override did not end, contrary to the documented contract that the
field remains populated while xymond records something other than the
test's reported color.

Either retain the reporter's color independently of the expiring
hold-time state, or define and implement expiry consistently in both the
live board and documentation. Add a regression covering an expired gap
before and after restart.

###  Patch fails `git diff --check`

`xymond/xymond.c:314` adds trailing whitespace after `F_FLAGS,`.

bonomani:
@SoundGoof Both points are addressed in the updated commit (`acbb6406d`).

**Restart contract**: took your first option — the checkpoint now retains the gap for as long as one is open (`gapopen = (pregapcolor != NO_COLOR)` in `save_checkpoint()`), instead of dropping it past the bridge window. The window keeps deciding only the hold-time, where it always did: in `holdtime_bridges()`, at the close. Cost: one `.flapstate.` record per test under a lasting override. The regression you asked for is case 7 of `xymond-holdtime-gap.sh` — a day-old gap answers `prevgapcolor=red` before and after an unrelated restart, and the red that closes it still starts fresh — with `xymond-checkpoint-roundtrip.sh`'s old "drop the expired gap" assertion inverted to match.

**`git diff --check`**: clean now.

One more fix the first point led to: the generated rows (info, trends, clientlog) are built from zeroed structs, and zero is `COL_GREEN`, so both history fields would have read "was green" on rows that have no history. They are seeded `NO_COLOR` now, asserted in `xymondboard-color-history.sh`.

bonomani:
Both points are fixed in the amended commit.

The checkpoint now saves the gap for as long as the substitution lasts: `gapopen = (pregapcolor != NO_COLOR)`, with the bridge window nowhere in that lifetime — it governs only the hold-time, in `holdtime_bridges()` at the close. That is your first option: the reporter's color is retained independently of the expiring hold-time state. Cost: one `.flapstate.` record per test under a lasting override.

The regression you asked for is section 7 of `tests/server/xymond-holdtime-gap.sh`: a checkpoint written with a gap two days past its window, a restart, and the assertion that `prevgapcolor` still answers the reporter's color while the override is in force.

The trailing whitespace went with the enum line rewrite; `git diff --check` is clean against the base.

SoundGoof:
### Findings

#### Medium: `prevgapcolor` does not track reports received during an override

[xymond/xymond.c:1596](https://github.com/xymon-monitoring/xymon/blob/f0a7716a447f022ac531a7ed5f23498a22e58ef7/xymond/xymond.c#L1596-L1602)

When a gap opens, `pregapcolor` is copied from `oldcolor`. At that point,
`oldcolor` contains the previously recorded effective color, not the current
`reportedcolor`.

For example:

1. The stored status is green.
2. The test reports red during DOWNTIME.
3. Xymond records blue.
4. `prevgapcolor` reports green.

The new documentation says it is the last color that was the test's own,
which should be red in this case.

I reproduced this against exact head `f0a7716a`. The existing test starts
red and reports red, so the pre-gap and latest-reported values are identical
and the mismatch is hidden.

`pregapcolor` cannot simply be updated to red because hold-time bridging
needs the pre-gap green value. Either retain separate state for the latest
reported color, including checkpoint persistence, or narrow the public
contract to the color held before the substitution began.

### Validation

- Local full suite: 83 passed, 3 skipped, 0 failed
- Submitted focused regressions passed locally
- Exact server build passed on Debian and all four BSD guests
- New gap and board regressions passed on all five guests
- Discriminating green-to-red DOWNTIME probe failed with
  `prevgapcolor=green`

bonomani:
@SoundGoof Fixed, head `640ebcd4b`; full suite 86/0/0, `git diff --check` clean.

`prevgapcolor` now tracks the reporter's word, not the pre-override color: your green-to-red DOWNTIME probe is section 5b of `xymond-holdtime-gap.sh` and fails on the previous head with exactly `prevgapcolor=green`. Two fields carry the two questions — `pregapcolor` stays the bridge state, `gapreportedcolor` is what the board publishes, refreshed by each real report under the override and reset at the close; synthetic posts don't touch it, a silent test keeps its last word.

One layer deeper than the finding: the DOWNTIME rewrite happens in `get_hts()`, so `handle_status()`'s `reportedcolor` is already blue for a red report — the raw color is captured just before that rewrite (`lastreportedcolor`). The checkpoint carries the word via the existing `gapstart:validity` suffix chain, so older files still parse (absent → the pre-gap color, the reporter's word as of the open; a suffix on a gapless record → `none`, or it would stick with no close to clear it); both asserted across a restart. The manpage now says what the field does, including that DOWNTIME substitutes only alert colors — a green report shows through and answers `none`.

SoundGoof (approved):
No findings.

The earlier restart-consistency, trailing-whitespace, generated-row, and
reported-versus-pre-gap color findings are addressed in the current head.

The additional `lastreportedcolor` state is justified by the existing control
flow: DOWNTIME rewrites the color in `get_hts()` before `handle_status()` receives
it. Refactoring that interface is outside this change's necessary scope.

Validation:

- Isolated build succeeded.
- Test suite: 86 passed, 0 skipped, 0 failed.
- `git diff --check` passes.

PR #384, opened 2026-08-16
Remove ".cfg" from xymonserver.cfg.d/xymonclient.cfg.d dir name.

### LINKED ISSUE #383: xymonserver.cfg.d and xymonclient.cfg.d should remove .cfg from name
In #222 there were .d directories added for several .cfg files to implement an easier way to handle configs.

While the names of the the directories for most config files were generated by replacing .cfg with .d (e.g. alerts.cfg -> alerts.d), this was done in an inconsistent way for xymonserver.cfg (resulting in xymonserver.cfg.d instead of xymonserver.d) and xymonclient.cfg (resulting in xymonclient.cfg.d instead of xymonclient.d).

I don't know the reason for this different behavior, maybe these two are used somewhere else in the past and this is for backward compatibility (@bonomani)?
At least in Debian we use xymonserver.d and xymonclient.d for ages now (with a mechanism, that generates a flat file from those and include this, which can now be reverted, since a better mechanism exists).  But the name change is a little annoying, since it would imply compatibility links or the like in Debian.

RolandRosenfeld:
The ".cfg" in the directory names xymonserver.cfg.d and xymonclient.cfg.d is inconsistent to the naming of other .d directories, which are generated by replacing ".cfg" with ".d" (and not only adding ".d").

This PR fixes #383.


bonomani:
Traced the origin, since the issue asks where the inconsistency came from: it is not backward compatibility, it is an oversight in the commit that introduced the mechanism — J.C. Cleaver's `bff5b5ef5` (devel, 20 Nov 2015). That single commit declares eight directories the regular way (`alerts.d`, `analysis.d`, `hosts.d`, …) and two the other way, and the server line is doubly out of step: `directory /etc/xymon/xymonserver.cfg.d`, hardcoded, where every other line is templated. #222 ported it faithfully, keeping both quirks. Nothing in the tree reads those names — they are only strings in a `directory` declaration — so there is nothing to preserve.

Two things worth knowing before merge:

**The mechanism has never shipped from this repository.** `5513733dc` (6 Jul 2026) is the only commit that ever touched those lines, and no release tag contains it. So there is no installed base from `main` to migrate.

**But Terabithia and devel users have one.** Their `xymonserver.cfg.d` has existed since 2015 and may hold real files. After this change nothing declares it, so those files stop being read — and it is silent by construction: `stackio.c:344` only warns about a directory it was told to open, and this one is no longer mentioned anywhere. A release note is the cheap answer; the alternative (keeping the old name as an extra `optional directory`) costs permanent config surface for a population that is already migrating paths by hand.

Worth noting in #29 as well, on the `5`/`6` lines: devel still has `.cfg.d`, so a later absorption from there must not bring the old names back.

bonomani (approved):
Complete and correct: both declarations, the Makefile's mkdir list, and the test that hardcoded the exception. Origin traced in the comment above — the `.cfg.d` form was an oversight in devel `bff5b5ef5`, not compatibility, and nothing in the tree reads those names. Release note for Terabithia/devel migrants to follow after merge.

PR #386, opened 2026-08-16
tests: stop excluding hosts that are able to run the test

bonomani:
**1 commit, tests only.** Three tests answered for the host instead of asking it, and each shaped its answer like Linux.

`wmlgen-path-bounds.sh` built its deep directory to a 3900-byte literal, cut for Linux's PATH_MAX. The BSDs and macOS stop at 1024, so the `mkdir` failed and the test skipped — on the platforms whose limit is tightest, for a guard measured against PATH_MAX itself (`wmlgen.c` fills `char[PATH_MAX]`). It takes the constant from the harness now, compiled against the same header as the function under test, rather than from `getconf`, whose per-filesystem `pathconf()` need not agree. 36 bytes short leaves the host card exactly fitting and the status card exactly not:

| PATH_MAX | directory | name | host card | status card |
|---|---|---|---|---|
| 4096 | 4060 | 30 | 4095 — fits | 4100 — refused |
| 1024 | 988 | 30 | 1023 — fits | 1028 — refused |

The Linux row is what the literal already produced, so nothing changes there.

`configure-no-rrd.sh` ran `./configure --server` with whatever `make(1)` is. On the BSDs that is BSD make, so `configure.server` aborted on its first check and never reached the RRD probe: skipped for a dependency the host has, GNU make being installed there as `gmake`. It hands over the `MAKE=` that `configure.server` prints when it bails.

`showgraph-single-service.sh` compiled `showgraph.c` without `RRDINCDIR` — empty on Linux, `-I/usr/local/include` or `-I/opt/homebrew/include` elsewhere, and passed by the product Makefiles for every RRD-using object. Without it the test cannot find `<rrd.h>` on a host that builds Xymon fine, and calls that "showgraph does not compile". Latent, and only because those hosts happen to search the directory their compiler was built with. Found by codex on the sibling test in #349.

| mutation | caught by |
|---|---|
| the `snprintf` guard removed from `wmlgen.c` | `a status-card path that does not fit was reported as written` |
| BSD make on PATH, gmake beside it | before: `SKIP: GNU make unavailable`; after: passes |
| `RRDINCDIR` pointed at a header that `#error`s | the flag reaches the compiler; before, it was dropped |

Un-skips 14 lanes for the first (macOS 14/15/26, FreeBSD 13.5/14.3/15.0, NetBSD 9.2–10.1, OpenBSD 7.6/7.7/7.8) and 11 for the second. The BSD behaviour is arithmetic and a simulated `make`, verified locally; the lanes confirm it.

Left as skips: `-fsanitize=address` is genuinely absent from NetBSD, OpenBSD and most Alpine lanes, and `column(1)` from Alpine — that one costs only redundancy, since `fs-filter-darwin.sh` still runs on macOS and the three BSDs.

Suite: 86 passed, 0 skipped, 0 failed on a built tree. Rebased on current `main`.

### DIFF
--- a/tests/buildsystem/configure-no-rrd.sh
+++ b/tests/buildsystem/configure-no-rrd.sh
@@ -63,6 +63,20 @@ cd "$SRC"
 # the RRD probe regardless of host fping state.
 export USEXYMONPING=y
 
+# configure.server bails when its make-utility is not GNU make, which on the
+# BSDs it never is: make(1) is BSD make there and GNU make is gmake. The run
+# then aborts before the RRD probe and this test skips for a dependency the
+# host has. Hand it the MAKE= configure.server prints when it bails, detected
+# the way configure.server detects it so the two cannot drift.
+make_is_gnu() {
+	[ "$("$1" -version 2>&1 | head -n 1 | awk '{print $1 " " $2}')" = "GNU Make" ]
+}
+if [ -z "${MAKE:-}" ] && ! make_is_gnu make; then
+	if command -v gmake >/dev/null 2>&1 && make_is_gnu gmake; then
+		export MAKE=gmake
+	fi
+fi
+
 LOG="$TMP/configure.log"
 set +e
 ./configure --server </dev/null >"$LOG" 2>&1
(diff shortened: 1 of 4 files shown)


SoundGoof (approved):
lgtm

Issue #379, opened 2026-08-16
client: df_sentinel can read its pidfile between create and write, and start a second remote df


bonomani:
**Symptom.** The 20-concurrent-cycles case in `tests/client/fs-sentinel-linux.sh` fails intermittently:

```
FAIL: concurrent cycles each started a remote df; exactly one may own the probe: expected '1', got '3'
```

Seen on a NetBSD 2-CPU lane VM (2026-08-16) and, before that, roughly one run in 22 locally.

**Mechanism, traced.** `df_sentinel()` claims the probe with `( set -C; echo "claim:$$" > "$_pidf" )`. `set -C` makes the *create* atomic; create-then-write is not. A cycle that reads in that gap gets an empty string, which matches neither `claim:*` nor `?*`, so it falls through to the stale-entry path, removes the pidfile and starts a `df` of its own. The pid update further down has the same shape — it truncates the file before writing it.

Traced by running the real function under a harness whose wedge binary is named `df`, so the PID-reuse guard sees what it sees in production:

```
<pid> lost-claim old=[]
<pid> branch=stale-restart old=[]
```

(The first harness named that binary something else, which made the guard restart on every cycle — an artefact, not this bug. Worth knowing for anyone re-measuring.)

**Rate.** 25 rounds × 20 concurrent cycles per configuration, same machine:

| CPUs available | rounds with more than one `df` | empty-pidfile reads per 500 cycles |
|---|---|---|
| 1 | 0 / 25 | 0 |
| 2 | 0 / 25 | 0 |
| 32 | 2 / 25 | 10 |

**Impact: low.** One extra `df` against an unreachable server, once — the pidfile is correct afterwards, so it does not pile up cycle after cycle the way the defect the sentinel was written for (#316) did. Hitting it takes two client cycles executing the claim within microseconds of each other, which a five-minute cadence makes unlikely outside a restart or a backlog. The recurring cost is CI noise: one lane in five failed on it today.

**Not established.** The NetBSD failure happened on a 2-CPU VM, where the harness above produces none. Whether it is this window or a different one was not determined — there is no trace from the lane.

[...]

PR #388, opened 2026-08-17
tests: hold the df sentinel's claim window open, do not race for it (#379)

bonomani:
The sentinel takes the probe in two steps a second cycle must never catch half-done: the claim, and the publication of the df's pid. The publication has a case that forces its window open with an `mv` stub. The claim had only the 20-concurrent-cycles case — which reaches the window by luck. That is the case reported failing intermittently in #379 (a NetBSD 2-CPU lane, and roughly one run in 22 locally), and a race that reproduces once in twenty runs cannot say *which* of the two steps regressed when it fails.

The claim is written to a private file and hard-linked into place, so the pidfile carries its content from the instant the name exists. Writing the pidfile directly cannot do that — the name appears first, the claim lands after — and an entry read in between is empty, which reads as stale: the reader removes it and starts a df on top of the one already running.

An `ln` stub links for real and then holds the first claim before returning, which is the window a reader has to survive. A second cycle runs inside it and must find a whole claim and stand down.

| client code | result |
|---|---|
| as on `main` (content written, then linked) | pass |
| name linked first, `claim:$$` written after | **FAIL — expected '1', got '2'** |

Two dfs on one wedged mount is what the sentinel exists to prevent, so the case fails for the right reason.

Waits for the window rather than for a duration; deterministic over repeat runs. `tests/buildsystem/test-suite-portability.sh` passes — no `uname` guard, no `/proc` rewriting, pure `sh` — so it runs on every lane.

Does not change client code. Whether #379 is still live is a separate question: `4fd7f685f` landed ~10 h before it was filed, and the issue quotes the pre-`ln` claim. Either way this pins the half that had no deterministic guard.

### DIFF
--- a/tests/client/fs-sentinel-linux.sh
+++ b/tests/client/fs-sentinel-linux.sh
@@ -290,6 +290,53 @@ assert_equal "$(head -1 "$DF_REMOTE")" "$_recorded" \
 while read -r _p; do end_stub "$_p"; done < "$DF_REMOTE"
 rm -f "$TMP"/probe/*; : > "$DF_REMOTE"
 
+# --- the claim window, forced open --------------------------------------------
+# The publication above is one of two steps that must not be seen half-done.
+# The other is the claim: it is written to a private file and hard-linked into
+# place, so the pidfile carries its content from the instant the name exists.
+# Writing to the pidfile directly cannot do that -- the name appears first and
+# the claim lands after -- and an entry read in between is empty, which reads as
+# stale: the second cycle removes it and starts a df on top of the first.
+# An ln stub holds the first claim *after* linking, so a second cycle reads the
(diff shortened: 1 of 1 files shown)


SoundGoof (approved):
### Findings

No findings.

The test synchronizes on the linked claim and verifies that a second cycle
does not start another probe. Disabling the hold produced the expected
two-probe failure, confirming that the new case is not vacuous.

### Validation

Reviewed head `7c76b18b`.

- Sentinel and native mount tests passed on:
  - Debian 13
  - NetBSD 9.4
  - NetBSD 11.0
  - OpenBSD 7.9
  - FreeBSD 14.3
- Negative probe failed as expected with two probes
- `git diff --check` passed

PR #390, opened 2026-08-17
do_disk: an unreachable mount reports red without a fabricated sample (#344)

### LINKED ISSUE #344: xymond: let a status-only disk row alert without reaching the RRD
Follow-up from #343 (the remote-df sentinel), which deliberately leaves this alone because it needs a server-side change.

## Problem

When a remote mount is unreachable, the Linux client reports it as a 100% row rather than dropping it, because the server reads an absent filesystem as green. That turns one filesystem red instead of purpling the whole host, which is the point.

The row is not `ignored`, so it also reaches the RRD:

- `unix_disk_report()` appends every non-ignored line to `dfstr_filtered` (`xymond/xymond_client.c:653-655`)
- that buffer becomes the status body (`:714`)
- `do_disk_rrd()` parses it, and for `DT_UNIX` takes `pused = atoi(columns[4])` and `aused = str2ll(columns[2], NULL)` (`xymond/rrd/do_disk.c:150-154`)

The sentinel row is `unavailable:/mnt/x 1 1 0 100% /mnt/x`, so against `DS:pct` and `DS:used` it writes **pct = 100 and used = 1 KB**, every cycle, for the duration of the outage. The two series are corrupted in opposite directions: the capacity graph shows the filesystem emptying while the percentage graph pegs full. Trending and forecasting on `used` are poisoned for that period, and the artifact outlives the outage.

## Why the client cannot fix it alone

Alerting and graphing come off the same row, and the only lever the client has is the row's contents:

- `do_disk_rrd()` skips a row only when `pused == -1` (`do_disk.c:192`)
- a capacity column that yields `-1` also reads as `-1` in `unix_disk_report()`, which never exceeds a panic threshold, so the column stays green

So any row that turns the column red necessarily reaches the RRD, and any row the RRD skips cannot alert. There is no client-side spelling that does both.

## Suggestion

Give the server a way to accept a row for status and colour while keeping it out of `dfstr_filtered`. The `ignored` path already does exactly this, but it is driven by `analysis.cfg` DISK rules, so it is not usable for a condition the client discovers at runtime.

Whatever the mechanism, it should not need a new wire command: the client already has to encode the condition inside a df-shaped row, and the server already parses every column.

## Acceptance

- An unreachable remote mount still turns its disk column red.
- Its RRD receives no sample for the duration, so `used` and `pct` keep their pre-outage values rather than a fabricated 100% / 1 KB.
- Ordinary filesystems are unaffected in both status and graphs.

Until then the artifact is documented in `xymonclient.cfg(5)` so operators are not surprised by the graph.

bonomani:
Closes #344. The server-side guard, the row it keys on, then the alert that names it.

**Commit 1 — do_disk skips a row without a reading.** The client's unavailable marker was trended
as a real sample: 100% / 1 KB every cycle of an outage, burying the pre-outage graph. The skip
triggers on `-` in both size columns; status is untouched (color is decided in `xymond_client.c`).
One side effect outside #344: AIX `df -Ik` writes the same all-dash shape for `/proc`, which was
trended as fabricated zeros — that row is now skipped too, losing nothing measured.

**Commit 2 — the marker names the server and reports no value.**
`unavailable:/mnt 1 1 0 100% /mnt` becomes `srv:/exp - - - 100% /remote/nfs`: `fs_mounts()` carries
the device as a third field, `-` means unmeasured, `100%` keeps the column red. The values-based
skip retires the device-name special case — a server genuinely named `unavailable` is trended like
any other. The BSD/macOS inode reformatters previously turned `-` back into `0` or dropped the row
(green during an outage); they now pass the marker through, collapsed on the wire to the same
six-field shape as the disk report.

**Commit 3 — FreeBSD all-ZFS inode report (#398).** On a ZFS-only host `df -i` excludes every
filesystem and exits 0 with no output; the clean-but-empty case fell through and became a garbage
`itotal` row the server could not parse (red). An empty result now emits an empty report, keeping
the server's all-ZFS green path. Confirmed working by @feld on the affected host.

**Commit 4 — the alert names the failure.** Routing the marker's `100%` through the percent
threshold made the alert read "(100% used) has reached the PANIC level" — a full disk, when the
truth is a mount that did not answer. `unix_disk_report()`/`unix_inode_report()` now recognise the
marker by its values (`-` free plus the marker's `100%`) and report
"&red /remote/nfs is unreachable (not measured)" instead; IGNORE rules still apply, the wire is
untouched (an old server still goes red via its thresholds). AIX's all-dash `/proc` row has no
`100%` and stays green; a genuinely full filesystem has real sizes and keeps the PANIC path.

Both review findings are fixed:
- The mount list reaches the fallback awk on its input stream behind an `@@` separator
  (`fs_setop()`'s move): a `-v` assignment cannot carry newlines on BSD awk, so two or more guarded
  mounts emitted no rows at all.
- A device containing spaces is re-encoded (`\040`) so the marker row keeps its column count.

| check | result |
|---|---|
| full suite on a built tree | 88 passed / 0 skipped / 0 failed |
| commit-1 skip removed | FAIL — RRD moves to `100 1` |
| skip as a bare prefix test | FAIL — a real `unavailable:`-named server loses its RRD |
| client tests under one-true-awk (= BSD awk) | pass; pre-fix scripts FAIL (`newline in string`) |
| spaced-device fixture vs pre-fix client | FAIL — the decoded row splits the columns |
| commit-3 all-zfs fixture vs pre-fix client | FAIL — the exact garbage row from #398 |
| commit-4 test vs pre-fix server | FAIL — the marker alerts "(100% used) has reached the PANIC level" |

---
**Merge sequencing.** Not git-stacked, but overlaps **#397** on the client scripts (`fs_mounts`, the `run_df` unavailable row) and `tests/client/fs-mounts.sh`. Suggested order: **#397 → this** (#396, which #397 built on, is already merged). The second merge combines this PR's third `fs_mounts` field (device) with #397's `\134`/escape decoding on mp+dev, and keeps this PR's awk-based unavailable row (it supersedes #397's `echo`→`printf`, whose loop this removes). The inode fix here (issue #398, commit 3) no longer collides with the test-infra stack #391→#167→#319→#179→#394 — its case was moved off the shared `pass` line. Commit 4 touches only `xymond/xymond_client.c` plus a new `tests/server/` file, so it adds nothing to the overlap.

### DIFF
--- a/client/xymonclient-darwin.sh
+++ b/client/xymonclient-darwin.sh
@@ -91,7 +91,7 @@ esac
 : "${XYMONCLIENT_FS_REMOTE_HARDBLOCK_TYPES=nfs nfs4 smbfs cifs afpfs webdav ceph glusterfs lustre afs}"
 DFPROBEDIR="${XYMONTMP:-/tmp}"
 
-# fs_mounts : one "TYPE<tab>MOUNTPOINT" line per mount, from mount(8). macOS
+# fs_mounts : one "TYPE<tab>MOUNTPOINT<tab>DEVICE" line per mount, from mount(8). macOS
 # mount(8) lists from getmntinfo(MNT_NOWAIT) -- verified in the shipped binary,
 # three call sites, all MNT_NOWAIT -- so it never refreshes statfs and cannot
 # block on a dead server, which df would.
@@ -104,7 +104,8 @@ fs_mounts() {
 			if (match(rest, / \(/) == 0) next
 			mp = substr(rest, 1, RSTART - 1)
 			split(substr(rest, RSTART + 2), a, /[,)]/)
-			printf "%s\t%s\n", a[1], mp
+			dev = substr($0, 1, i - 1)
+			printf "%s\t%s\t%s\n", a[1], mp, dev
 		}'
 }
 
@@ -321,13 +322,26 @@ fs_guarded() {
 	set -- "$@" $_mounts
 	_rout=`df_sentinel "$_tag" "$@"`
 	if [ $? -eq 124 ]; then
-		for _m in $_mounts; do
-			if [ "$_tag" = inode ]; then
-				echo "unavailable:$_m 1 1 0 100% 1 0 100% $_m"
-			else
-				echo "unavailable:$_m 1 1 0 100% $_m"
-			fi
-		done
+		# The row names the server. df could not answer, but the mount table
+		# still knows which device is behind the mount point, and that is the
+		# first thing an operator needs. The sizes are reported as "-": they
+		# were not measured, and a number here would be trended as a reading.
+		# The capacity stays 100% because that is what turns the column red.
+		# The mount list rides the mount-table stream behind an @@ separator:
+		# a -v assignment cannot carry newlines on BSD awk (fs_setop makes
+		# the same move). A spaced device is re-encoded (\040) to keep the
+		# column count.
+		{ fs_mounts; echo '@@'; printf '%s\n' "$_mounts"; } | awk -F'\t' -v inode="$_tag" '
+			$0 == "@@" { rm = 1; next }
+			!rm { dev[$2] = ($3 == "" ? "-" : $3); next }
+			$0 == "" { next }
+			{
+				d = ($0 in dev) ? dev[$0] : "-"
+				n = split(d, dp, / /)
+				if (n > 1) { d = dp[1]; for (j = 2; j <= n; j++) d = d "\\040" dp[j] }
+				if (inode == "inode") printf "%s - - - 100%% - - 100%% %s\n", d, $0
+				else printf "%s - - - 100%% %s\n", d, $0
+			}'
 	else
 		printf '%s\n' "$_rout"
 	fi
@@ -424,7 +438,8 @@ if [ -n "$FILESYSTEMS_INODE" ] || [ -n "$FILESYSTEMS_INODE_REMOTE" ]; then
 	 fi
 	 fs_guarded inode "-P -i" "$FILESYSTEMS_INODE_REMOTE") | awk '
 NR<2{printf "%-20s %10s %10s %10s %10s %s\n", $1, "itotal", $6, $7, $8, $9}
-(NR>=2 && $6>0) {printf "%-20s %10d %10d %10d %10s %s\n", $1, $6+$7, $6, $7, $8, $9}' )
+(NR>=2 && $6 == "-") {printf "%-20s %10s %10s %10s %10s %s\n", $1, "-", "-", "-", $8, $9}
+(NR>=2 && $6 != "-" && $6>0) {printf "%-20s %10d %10d %10d %10s %s\n", $1, $6+$7, $6, $7, $8, $9}' )
 	emit_df inode Inode
 fi
 
--- a/client/xymonclient-freebsd.sh
+++ b/client/xymonclient-freebsd.sh
@@ -77,7 +77,7 @@ esac
 : "${XYMONCLIENT_FS_REMOTE_HARDBLOCK_TYPES=nfs nfs4 smbfs cifs ceph glusterfs fusefs.glusterfs lustre afs}"
 DFPROBEDIR="${XYMONTMP:-/tmp}"
 
-# fs_mounts : one "TYPE<tab>MOUNTPOINT" line per mount, from mount(8).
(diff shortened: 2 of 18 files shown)


SoundGoof:
### Findings

#### Medium: Multiple remote mounts suppress every unavailable row on BSD

[client/xymonclient-netbsd.sh:306](https://github.com/xymon-monitoring/xymon/blob/2b2eb648179109b704e2fd9eba050720f7352230/client/xymonclient-netbsd.sh#L306-L316)

The fallback passes newline-separated `_rm` through
`awk -v mounts="$_rm"`.

NetBSD, OpenBSD, and FreeBSD awk reject a literal newline in a `-v` value:

```text
awk: newline in string /net
/other... at source line 1
```

When two hard-blocking mounts exist and the probe fails, awk emits no
sentinel rows. The server receives only local filesystems and can report the
disk column green during the remote outage.

The submitted native sentinel regression reproduces this on NetBSD 9.4,
NetBSD 11.0, OpenBSD 7.9, and FreeBSD 14.3.

Keep the mount list out of an awk command-line assignment. Iterate it in the
shell or provide both datasets through input streams.

#### Medium: A device containing spaces shifts the synthetic df columns

[client/xymonclient-linux.sh:56](https://github.com/xymon-monitoring/xymon/blob/2b2eb648179109b704e2fd9eba050720f7352230/client/xymonclient-linux.sh#L56)
[client/xymonclient-linux.sh:423](https://github.com/xymon-monitoring/xymon/blob/2b2eb648179109b704e2fd9eba050720f7352230/client/xymonclient-linux.sh#L423-L430)

`fs_mounts()` decodes `\040` in the device, then the sentinel prints that
device into a whitespace-delimited df row without re-encoding it.

For `srv:/team share`, the row becomes:

```text
srv:/team share - - - 100% /remote/nfs
```

The additional field shifts `Capacity` to `-` and `Mounted` to `100%`.

A live exact-head test produced:

```text
disk|green|... Filesystems ok
```

instead of red for `/remote/nfs`. The RRD parser also missed the unmeasured
marker and created a third bogus filesystem RRD.

Encode whitespace in the device before emitting the df-shaped row, and add a
spaced-device fixture covering both status color and total RRD files.

### Validation

Reviewed latest head `2b2eb648`.

- Full Linux suite: 84 passed, 3 skipped, 0 failed
- Submitted Linux sentinel and RRD regressions passed
- Native sentinel regression failed on all four BSD guests
- Direct multiline `awk -v` probe failed on all four BSDs
- Live `xymond` probe reproduced the false-green disk status
- Extended RRD probe reproduced the extra bogus RRD
- `git diff --check` passed

bonomani:
@SoundGoof Both findings are fixed, head `8c0b5e7ac`; full suite 87/0/0, `git diff --check` clean.

- **Multi-mount BSD awk**: the mount list rides the fallback awk's input behind an `@@` separator (`fs_setop()`'s move), in all five clients. Verified against one-true-awk locally: the pre-fix scripts fail with your exact `newline in string`, the fix passes. The Linux test gained a two-mount regression.
- **Spaced device**: re-encoded `\040` before the row is printed — split/concat rather than a `gsub` replacement, whose backslash handling diverges across awks. Fixtures on both halves: the client test bans the decoded row, the RRD test asserts no phantom RRD and an unmoved sample.

Found while widening the same questions: the `fsf_assert_loud` glob matched a healthy report (dashes in device names plus any 100% row) — now line-anchored; the RRD test sent a 9-field inode row no client puts on the wire — now the collapsed six-field shape; and AIX `df -Ik` writes the all-dash shape for `/proc`, so the values skip drops that fabricated-zeros row too — comment corrected, nothing measured is lost.

SoundGoof:
@bonomani  needs a rebase

SoundGoof (approved):
No findings.

Native filesystem tests passed on all eight available VMs:

- FreeBSD 14.3
- OpenBSD 7.9
- NetBSD 9.4
- NetBSD 11.0
- Debian 12
- Debian 13
- Rocky Linux 9
- Rocky Linux 10

The platform-appropriate mount enumeration, remote-df sentinel, and filesystem-filter
tests passed. This covered both NetBSD versions, FreeBSD, OpenBSD, ext4 Linux systems,
and XFS Linux systems across several kernel generations.

Additional validation passed:

- Focused server build
- Real server and RRD regressions
- Full test suite: 86 passed, 3 skipped, 0 failed
- All five GitHub checks
- `git diff --check`

PR #391, opened 2026-08-17
tests: link the lib-only harnesses against the archives every variant builds

bonomani:
**Bottom of a 5-PR stack** (#391 → #167 → #319 → #179 → #394); own commits: the 2 shown. Merge first.

Two changes that make tests find what a build actually produced. Refs #166.

- `variant_products()` in `tests/lib/assert.sh`: one table of per-variant product paths. `require_bin` resolves an exported `$VAR`, then the variant's row, then the caller's default; an undeclared product degrades to a probe rather than a false claim.
- Four harnesses linked `libxymoncomm.a`, which only a server build produces, and skipped on client builds that compile the same objects into `libxymonclientcomm.a` + `libxymonclient.a` — built by every variant (comm first; the reverse fails to link).

All CI checks green at this head.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -88,7 +88,7 @@ jobs:
       # there. Run it here too, where the binaries this variant produced are on
       # disk, so those tests actually execute against the build.
       #
-      # First such test: tests/server/xymongrep-filter.sh, which drives the
+      # First such test: tests/common/xymongrep-filter.sh, which drives the
       # common/xymongrep this leg just built. This step is the only place in
       # CI where it executes instead of skipping.
       #
--- a/tests/README.md
+++ b/tests/README.md
@@ -99,7 +99,8 @@ it drives is one every variant builds -- `common/xymongrep` in a server tree,
 `client/xymongrep` in a client one -- so it goes under `common/`, and a client
 build gets the coverage it exists for. `analysis-file-ifexist.sh` drives
 `xymond_client`, which only a localclient or server build produces, so it goes
-under `localclient/`.
+under `analysis/`: the area names the subject, and the variant table above says
+which build supplies the binary.
 
 Add a new area by PR when an existing one doesn't fit. Don't bend a
 test to fit the wrong area just to avoid creating a new directory.
--- a/tests/README.md
+++ b/tests/README.md
@@ -28,6 +28,31 @@ Conventions). The runner itself is POSIX sh, and on a host without bash it
 skips the whole suite with exit `77` rather than reporting interpreter
 failures as test failures.
 
+### `XYMON_VARIANT` — telling the suite which build this is
+
+Most tests only execute against a build, so run one first. The suite then needs
+to know *which* build, because the same program lives at different paths
+depending on the variant:
+
+    ./configure --server && make -j$(nproc)
+    XYMON_VARIANT=server ./tests/testsuite
+
+| variant | configured with | `xymongrep` | `xymond_client` |
+| ------- | --------------- | ----------- | --------------- |
+| `server` | `./configure --server` | `common/xymongrep` | `xymond/xymond_client` |
+| `localclient` | `CONFTYPE=client ./configure --client` | `client/xymongrep` | `client/xymond_client` |
+| `client` | `./configure --client` | `client/xymongrep` | not built |
+
+The full table is `variant_products()` in `lib/assert.sh`; it also covers
+`xymond_rrd` and `svcstatus.cgi`, both server-only.
+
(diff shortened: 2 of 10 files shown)


SoundGoof:
## Findings

### Reject unknown `XYMON_VARIANT` values

`tests/lib/assert.sh:480-483`

A misspelled variant is treated as a valid variant that does not produce any
declared product. This silently removes binary coverage while the suite remains
green.

Reproduced on a complete server build:

    XYMON_VARIANT=sever ./tests/testsuite

Result:

    passed:  76
    skipped: 10
    failed:  0
    exit:    0

The individual `xymongrep` test exits 77 with:

    SKIP: the sever build does not produce XYMONGREP

`known_variants()` already provides the required list but is unused. Validate
a non-empty `XYMON_VARIANT` against that list and fail on unknown values before
resolving products.

### Correct the analysis test directory in the README

`tests/README.md:100-102`

The worked example says `analysis-file-ifexist.sh` goes under
`tests/localclient/`, but this PR moves it to `tests/analysis/` and documents
that area in the preceding table. Use `analysis/` in the example.

## Validation

- `git diff --check` passes.
- Debian 13 complete suites:
  - server: 86 passed, 0 skipped, 0 failed
  - client: 57 passed, 29 skipped, 0 failed
  - localclient: 59 passed, 27 skipped, 0 failed
- All seven affected tests pass on:
  - NetBSD 9.4
  - NetBSD 11.0
  - OpenBSD 7.9
  - FreeBSD 14.3

bonomani:
Both fixed and pushed, rebased onto current `main` (the branch was twenty
commits behind).

**Unknown `XYMON_VARIANT`.** `known_variants()` was written for exactly
this and never called; it is now. Your case is closed and the arithmetic
matches your report — the ten tests that skipped are precisely the ten
that now fail:

|                       | passed | skipped | failed | exit |
| --------------------- | ------ | ------- | ------ | ---- |
| `sever` before        | 76     | 10      | 0      | 0    |
| `sever` after         | 71     | 5       | 10     | 1    |

    FAIL: XYMON_VARIANT='sever' is not a known variant -- expected one of: server localclient client

The five remaining skips are the legitimate ones present in every run.

Worth recording, because it changed the fix: I first validated when
`assert.sh` is sourced, which runs for every test including the many that
never consult a variant. Those two extra processes per test were enough
to tip timing-sensitive ones — three of three full suites failed, against
none of six on the unmodified branch, a different test each time. Your
wording was "before resolving products", and putting it there instead
makes that disappear: `XYMON_VARIANT=server` now passes four consecutive
suites at 81/5/0, identical to the unmodified branch, and unset is
unchanged. It is memoized too, since `require_bin` runs once per product
a test needs and the answer cannot change within a run.

One full-suite failure I chased during that work turned out to be
environmental rather than either version's fault:
`rrd/extproc-teardown-flush.sh` passes 20/20 in isolation on both this
branch and the unmodified one. Flagging it because a handful of tests
here fail under load — separate from the port race in #421.

**README.** Fixed: the worked example now says `analysis/`, which is what
the area table two lines above already said.

SoundGoof (dismissed):
### Review

No blocking issues found. Approve.

The variant mapping, fallback behavior, archive selection, and static-library
link order are correct.

### Validation

- All five GitHub CI checks pass.
- Built and tested the `localclient` variant.
- Built and tested the plain `client` variant.
- All four relinked harnesses pass in both builds.
- Moved common and analysis tests resolve the correct binaries.
- The plain client correctly skips the unavailable analyzer.
- An unknown `XYMON_VARIANT` fails instead of silently reducing coverage.

The client CI wiring arrives in the next stacked PR, #167. That leaves a
temporary CI coverage gap in this foundational PR, but no correctness defect.

bonomani:
Heads-up: I pushed a small fix, so this needs a re-approval (@SoundGoof, your Aug-22 approval was dismissed by the update).

On review I turned up a genuine, if low-severity, bug in the version you approved: the two lib-only harnesses now link **two** archives (`libxymonclientcomm.a` and `libxymonclient.a`), but their "tree not built → skip" guards only checked for `libxymonclient.a`. So in a *partially*-built tree (one archive present, the other not), the guard passed and the compile then hard-`fail`ed instead of taking the documented source-only skip. Full builds and CI were unaffected, which is why it slipped through.

Fix: both guards (`netrc-password.sh`, `config-include-dir.sh`) now require **every** archive they link, amended into the archive-linking commit — no new commit, history stays at three. A guard should require whatever it's about to link.

- **CI**: build + tests green on the new head.
- Re-reviewed clean — "the client-library archive changes are consistent with the build layouts and dependency ordering."

Would appreciate a re-approve when you have a moment.

PR #392, opened 2026-08-17
tests: count a partial run apart, not as a full pass

bonomani:
**1 own commit**, taken from #319 and rebuilt on the area filing #167 lands. The rest of the stack below is #391 → #167 → #319; GitHub cannot base a PR on a fork branch, so they all show here until they merge.

A test that could check only part of what it covers — no sanitizer for the half that needs one, nothing built to drive — exits 0 like any other and lands in `passed`. `rrddef-empty-list` says so in its own message and still counts as a full pass:

```
pass "xymonrrd.c empty-list count loops are guarded (static check; $CC cannot build and run ASan binaries)"
```

`pass_partial "<what held>" "<what could not be checked>"` records itself, so the summary reads `passed: 86 (2 partial)`. It still exits 0 — nothing failed — which is exactly why the runner cannot tell the two apart on its own.

Not hypothetical: ASan is absent from NetBSD, OpenBSD and most Alpine lanes, so ten tests there report a full pass for a half-verified run today.

| check | evidence |
|---|---|
| the counter reports | `passed: 12 (2 partial)` |
| `grep -c` returning 1 on no match | taken as a count, not appended to — otherwise `[: Illegal number: 0` |

Suite: 86 passed, 0 skipped, 0 failed on a built tree. Portability lint green.


bonomani:
Recycled into #179, which is where the argument about what a success line should say already lives. Same commit, same stack.

PR #396, opened 2026-08-17
tests: verify fs_mounts drops nothing without assuming / is visible

### LINKED ISSUE #395: tests/client/fs-mounts.sh fails in cowbuilder chroot
I usually build xymon using a cowbuilder chroot environment.
#378 introduced a new test tests/client/fs-mounts.sh which fails in this environment with
```
----- tests/client/fs-mounts.sh -----
fs_mounts() returned:
  proc  /proc
  sysfs /sys
  tmpfs /dev/shm
  devpts        /dev/pts
  devpts        /dev/ptmx
FAIL: linux: fs_mounts() listed 5 mount(s) but not /, so it is dropping rows
FAIL: tests/client/fs-mounts.sh FAILED (rc=1)
```

The problem seems to be, that in this chroot environment I only see a reduced /proc/mounts:
```
/proc /proc proc rw,relatime 0 0
sysfs /sys sysfs rw,nosuid,nodev,noexec,relatime 0 0
tmpfs /dev/shm tmpfs rw,relatime,inode64 0 0
devpts /dev/pts devpts rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=666 0 0
devpts /dev/ptmx devpts rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=666 0 0
```
where no root filesystem is shown.

No idea, how to fix the test (without completely removing it), since this is a valid chroot but doesn't contain a `/` filesystem.

bonomani:
Fixes #395.

A cowbuilder/schroot build root is a plain directory entered with `chroot(2)`, so `/proc/mounts`
lists only the mounts reachable from it -- no `/` row. The test required a `/` row as a proxy for
"no rows dropped" and failed on correct output.

On Linux the check is now exact: `fs_mounts()` is a one-row-per-line transform of `/proc/mounts`,
so the row counts must match (one snapshot, taken before the stat loop can trigger an automount),
and `/` is required only when the input shows it. BSD/macOS lanes unchanged: `getmntinfo()` reports
the full table even inside a chroot.

| case | old test | new test |
|---|---|---|
| normal host | PASS | PASS |
| plain-dir chroot, no `/` row (`unshare -rmpf` repro) | FAIL | PASS |
| client sabotaged to drop its last row | PASS (miss) | FAIL |

Backslash hardening found during review: #397, stacked on this.

### DIFF
--- a/tests/client/fs-mounts.sh
+++ b/tests/client/fs-mounts.sh
@@ -86,6 +86,10 @@ fi
 }
 [ -s "$TMP/out" ] || fail "$os: fs_mounts() listed no mounts on a running system"
 
+# One snapshot for the Linux checks below, taken now: the stat loop can grow
+# the live table (stat of an automount point mounts it).
+[ "$kernel" != Linux ] || cat /proc/mounts > "$TMP/mounts"
+
 # What the helper actually produced, on failure only: a count ("listed 2 mounts
 # but not /") says something is wrong without saying what, and re-running a BSD
 # lane to find out costs half an hour.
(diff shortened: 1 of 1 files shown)


SoundGoof (approved):
lgtm

PR #397, opened 2026-08-17
client: survive backslashes in mount points

bonomani:
#396 (the fs_mounts-contract test this builds on) is now merged, so this PR is just its own two commits: decode `\134` in mount points, and pass the probe dir to awk via `ENVIRON`.

`/proc/mounts` escapes space, tab, newline and backslash as `\040 \011 \012 \134`; the clients
decode only `\040`. Two commits make a backslash survive end to end. Tab/newline stay escaped by
design: the row format is tab- and newline-delimited.

- `fs_mounts()` (Linux) decodes `\134` after `\040`, by split/concat -- a gsub replacement
  backslash yields one backslash under mawk, two under gawk/busybox. The unavailable-row fallback
  moves from `echo` to `printf`: dash's `echo` truncates a decoded path at `\c`.
- `probe_dir_is_local()` (all five clients) gets the probe dir via `ENVIRON` instead of `awk -v`,
  which escape-processes the value and can misread a dir on a hard-blocking mount as local.

Evidence: verified against real mounts holding a space, a backslash and a tab; identical output
under mawk, gawk, busybox awk and one-true-awk. The new sentinel check (a `\t`-escape path, decoded
by every awk in `-v`) fails the old `-v` code under all four awks and passes the fixed one, with the
fixture intact under dash, bash and busybox sh.

---
**Also overlaps #390** (do_disk unavailable-mount) on `fs_mounts` and `tests/client/fs-mounts.sh`. Suggested order **#397 → #390**; see #390 for how the two `fs_mounts` changes combine (this PR's escape decoding + #390's device field).

### DIFF
--- a/client/xymonclient-darwin.sh
+++ b/client/xymonclient-darwin.sh
@@ -144,7 +144,9 @@ probe_dir_is_local()
 	# sends the probe at a directory that may sit on the wedged mount this
 	# exists to keep away from.
 	_pdlm=$(fs_mounts) || return 1
-	printf '%s\n' "$_pdlm" | awk -F'\t' -v dir="$1" -v types="$XYMONCLIENT_FS_REMOTE_HARDBLOCK_TYPES" '
+	# dir via ENVIRON, not -v: -v escape-processes backslashes (gawk: \c -> c).
+	printf '%s\n' "$_pdlm" | _pdldir="$1" awk -F'\t' -v types="$XYMONCLIENT_FS_REMOTE_HARDBLOCK_TYPES" '
+		BEGIN { dir = ENVIRON["_pdldir"] }
 		BEGIN { n = split(types, a, /[ \t]+/); for (i = 1; i <= n; i++) t[a[i]] = 1 }
 		{
 			mp = $2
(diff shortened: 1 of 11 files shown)


SoundGoof (dismissed):
No findings.

The decoding order correctly preserves literal `\040` path content. Passing
the probe directory through `ENVIRON` avoids awk command-line escape handling.
The PR also merges cleanly with current `main`.

## Validation

- `git diff --check` passes.
- Linux tests pass with mawk and gawk.
- Debian 13 tests pass.
- NetBSD 9.4 and 11.0 tests pass.
- OpenBSD 7.9 tests pass.
- FreeBSD 14.3 tests pass.

bonomani:
Rebased onto current `main`, so the merge-base objection behind the
dismissed approval no longer applies.

The rebase dropped one commit: `c927a778` ("tests: verify fs_mounts drops
nothing without assuming / is visible") had already landed on `main` via
#396. That is where 22 of the removed lines went — the diff is now
+91 −10 across 2 commits rather than +113 −20 across 3.

Worth flagging for the re-review: `main` rewrote
`tests/client/fs-sentinel-linux.sh` in #388 ("hold the df sentinel's
claim window open, do not race for it"), and this PR touches the same
file. Git merges it without conflict, but two changes to one test's
timing model can merge cleanly and still be wrong together, so I built
the combination and ran the suite before pushing: 82 passed, 5 skipped,
0 failed, with `fs-mounts.sh` and `fs-sentinel-linux.sh` both passing.
The rebased tree is byte-identical to that merged tree.

The five checks are green on `af413f92` against current `main`. The
previous green run was from 17 August, against a `main` 19 commits older.

SoundGoof (approved):
lgtm

PR #399, opened 2026-08-18
msgcache: answer ping with our identity and the age of the last delivered pull

bonomani:
Single-commit form of #324 — idea and use case @StefCoene's — rebased on current `main` with a review pass folded in. **Supersedes #324** (close that in favour of this).

A client behind msgcache can't tell delivery from a bucket with a hole, and needs to know whether it's safe to apply a new version/config.

Reply (answered locally, never forwarded): `msgcache <version>` + `lastpull <seconds since last delivered pull, -1 if never>`.

For that age to be honest, `sentto` is committed **only once a pull's whole batch is written**, not while the response is built — a pull whose write fails re-delivers its batch instead of losing it, and can't come back empty and refresh `lastpull` without delivering. (Each queued message carries a `seq`; the connection commits exactly its batch on a completed write — one seq-ordered merge, prune-safe, no duplicates.)

- only an exact `ping` is answered; `pingpong` stays client data
- a negative age (clock stepped back) clamps to 0, off the `-1` sentinel
- a batch that can't be allocated drops the pull instead of delivering it unmarked — otherwise the whole batch re-sends on the next pull while `lastpull` says it landed
- pull index line: `sprintf` into `idx[20]` (overrun by a big `--max-age`/backwards clock; `INT_MIN:LONG_MIN` needs 34) → `snprintf` `idx[64]`

| check | result |
|---|---|
| `tests/client/msgcache-ping.sh` | PASS (FAIL against #324's head) |
| `tests/client/msgcache-failed-pull.sh` — RST mid-pull; batch re-delivers, age stays honest | PASS |
| `calloc` forced to fail: pull dropped, `lastpull` stays `-1`, logged | PASS (before: delivered, stamped, duplicated on the next pull) |
| second pull returns nothing (delivered exactly once) | PASS (mutate `conn->pollid = 0`: new assertion FAILs, previous test passed) |
| ASan, 200-char id to the interrupt helper | PASS (before: `stack-buffer-overflow` in `write`) |
| CI build + testsuite | green |

### DIFF
--- a/client/msgcache.c
+++ b/client/msgcache.c
@@ -48,6 +48,7 @@ char *client_response = NULL;		/* The latest response to a "client" message */
 char *logfile = NULL;
 int maxage = 600;			/* Maximum time we will cache messages */
 sender_t *serverlist = NULL;		/* Who is allowed to grab our messages */
+time_t lastpull = 0;			/* When a server last collected from us */
 
 typedef struct conn_t {
 	time_t tstamp;
@@ -57,6 +58,9 @@ typedef struct conn_t {
 	int sockfd;
 	strbuffer_t *msgbuf;
 	int sentbytes;
+	unsigned long pollid;		/* Server bit this pull serves (0 = client/none) */
+	unsigned long *batch;		/* seqs carried by the in-flight pull, committed on success */
+	int nbatch;
 	struct conn_t *next;
 } conn_t;
 conn_t *chead = NULL;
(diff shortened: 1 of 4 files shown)


bonomani:
Single-commit form of the msgcache ping from #324 — idea and use case @StefCoene's — rebased on current main with a deeper review pass folded in. Supersedes #324; close that in favour of this once Stef confirms it works for him.

bonomani:
Folded the index-line `sprintf` into the commit: a fixed `char idx[20]`, overrun by a large `--max-age` or a backwards clock step, now `snprintf` into `idx[64]`. Nothing truncates, no behaviour changes.

Single commit now; the test passes against the rebuilt binary.

bonomani:
@StefCoene requesting your review — this is your ping idea from #324, rebased to a single commit with a review pass folded in.

The reply is `msgcache <version>` + `lastpull <seconds since a server last collected, -1 if never>`. It deliberately does not claim anything about the Xymon server, since msgcache never connects to it — the age of the last pull is what it can honestly answer. Does that give the PowerShell client what it needs to decide whether to apply a new version or config?

If it works for you, I will close #324 in favour of this.

bonomani:
Update: rebased on `main` and folded in a delivery-accounting fix a review pass surfaced — `sentto` now commits on completed delivery, not while building the response, so a dropped write re-delivers its batch and can't falsely advance `lastpull` (the failure the age exists to expose). Guarded by the new `msgcache-failed-pull.sh` (a compiled helper RSTs a pull mid-batch); CI builds and runs it green.

PR #407, opened 2026-08-19
aix: filter pseudo filesystems by type, and keep df off remote mounts

bonomani:
Two commits, either usable without the other.

**1. Filter pseudo filesystems by type.** AIX is the last client without the `XYMONCLIENT_FS_*` type filter: its raw `df -Ik` sends `/proc`, `/aha` and friends to the server as all-dash rows. AIX df cannot exclude by type, so excluded types are dropped by device+mountpoint pair, read from the vfs column of `mount(8)` — the pair, so an excluded overlay (namefs) sharing a mount point with a real filesystem takes only the overlay row. Default set **procfs ahafs namefs autofs cdrfs**, minus `INCLUDE_TYPES`, plus `EXCLUDE_TYPES`. Excluding everything yields an empty `[df]` section, which the server flags — not a header-only one it reads as all-green.

*Upgrade note:* a `DISK` min-count rule pinned on a pseudo mount now alerts, since its row no longer feeds the count. `INCLUDE_TYPES` is the remedy, as on the sibling clients.

**2. Keep df off remote mounts by default.** A wedged NFS mount hangs df for the life of the mount and takes the client run with it — the host goes purple, every metric. The other clients avoid that with `df -l` and bound `DF_LOCAL_ONLY=no` with a sentinel; AIX had neither. `XYMONCLIENT_FS_DF_LOCAL_ONLY` now works here: AIX df spells it `-T local`, the System V df behind `[inode]` spells it `-l`, and both are guarded, since either left open reaches the same mount. With `no` the risk is unchanged — there is still no sentinel — and the man page says so.

**Declares AIX 7.1 as the floor** (RELEASENOTES): `-T local` does not exist before it, the 6.1 df documenting only `-I -M -P -g -i -k -m -s -v` ([IBM 6.1 Commands Reference](https://public.dhe.ibm.com/systems/power/docs/aix/61/aixcmds2_pdf.pdf), df at p.107).

| check | result |
|---|---|
| full suite (unbuilt tree, as tests.yml) | 40 passed / 51 skipped / 0 failed |
| fs-filter-aix.sh vs stock script | FAIL — no filter, the /proc row is emitted |
| same test with the guards removed one at a time | FAIL for each |

Reviewed twice by codex against the AIX manuals; its findings on the version floor, a header-only inode section, an `IFS`-dependent flag expansion and two overstated comments are fixed here. Mock-tested with stub df/mount: real AIX behaviour still needs confirming, in particular whether `mount(8)` returns while an NFS server is down.

### DIFF
--- a/RELEASENOTES
+++ b/RELEASENOTES
@@ -13,6 +13,13 @@ Changes file.
 Changes for 4.3.31
 ==================
 
+The AIX client now requires AIX 7.1 or later. It keeps df off remote mounts by
+default (XYMONCLIENT_FS_DF_LOCAL_ONLY, as on the other clients), which AIX df
+spells "-T local" -- a flag 6.1 and earlier do not have. On those releases the
+client reported every filesystem and a hard-mounted NFS server that stopped
+answering could hang the whole client run, with no remote-df sentinel to bound
+it; that is the behaviour now left behind rather than worked around.
+
 xymonnet tests with a configured cipher-list restriction now report an SSL error
 when OpenSSL rejects the restriction, instead of silently using the default
 cipher list. Such tests may therefore turn red after upgrading Xymon, or after
--- a/client/xymonclient-aix.sh
+++ b/client/xymonclient-aix.sh
@@ -21,6 +21,7 @@ echo "[who]"
 who
 echo "[df]"
 # --- filesystem filter (configurable; see xymonclient.cfg.DIST) --------------
+# Requires AIX 7.1: "df -T local" below does not exist before it (RELEASENOTES).
 # AIX df cannot exclude by type, so excluded types are dropped from its output
 # by mount point, via the vfs column of mount(8). Default set, minus
 # INCLUDE_TYPES, plus EXCLUDE_TYPES, like the other clients.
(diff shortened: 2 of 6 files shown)


bonomani:
Reviewed at f0b27ec93 (base 47bba995f).

| check | result |
|---|---|
| `./tests/testsuite`, unbuilt tree | 40 passed / 51 skipped / 0 failed |
| commit 1 alone (8ee2dadd6) | its own test passes; suite 40/51/0 |
| `build/makehtml.sh common/xymonclient.cfg.5` | regenerates byte-identical |
| `sh -n` under dash / bash / busybox sh | clean; shellcheck adds nothing new |
| 6 guards removed one at a time | 5 caught by the test |

Caught: `set -f`, `dev = $1 ":" $2` on remote rows, the `[df]` header holdback, `-T local`, sysv `-l`. Missed: `NR <= 2` -> `NR <= 1` in the mount awk, which is benign (the dashed rule line yields a vfs matching nothing). The mount(8) column heuristic and the device+mountpoint pair keying both hold.

Five things before merge.

**1. The `[inode]` header holdback rests on a claim the server contradicts.** The comment says "no rows must mean an empty section, which the server flags. A bare header reads as all-green", and the test asserts the same. `unix_inode_report()` says otherwise:

```c
if ((capacol == -1) && (mntcol == -1)) {
    if (strlen(dfstr) == 0) {
        /* Empty inode report, happens on Solaris when all filesystems are ZFS */
        inodecolor = COL_GREEN;
```

Empty `[inode]` is green; header-only `[inode]` is green too. The change is inert for that column and its rationale is backwards. `unix_disk_report()` has no such exemption, so the claim holds for `[df]` and only for `[df]`. Fix the comment and the assertion message, or drop the inode holdback.

**2. The empty-`[df]` design contradicts the contract the other five clients obey.** `fsf_contract()` ends with `fsf_assert_loud` on "excluding every filesystem must be loud, never a silent empty section" -- which fails on an empty section and requires a `collection failed` marker. AIX deliberately produces the empty section. Both reach yellow, but Linux/BSD show `Disk collection failed: df exited N with no output` while AIX shows `Expected strings (%Used and Mounted) not found in df output` plus `sent incomprehensible disk report` in the xymond log -- a wrong diagnosis for "you excluded everything" or "df rejected -T local". Adopt the marker on AIX, or amend the contract text. Today the two disagree and nothing catches it.

**3. The test bypasses the shared harness.** Every other `fs-filter-*.sh` calls `fsf_write_fixture` / `fsf_write_stub` / `fsf_write_mount_stub` / `fsf_selfcheck` / `fsf_contract`; this one uses `fsf_setup` and `fsf_extract` only, so AIX loses the `fsf_selfcheck` non-vacuity probe and every future contract rule -- and `fs-filter-common.sh`'s own "a rule added here is a rule every OS gets" becomes false. The blocker looks real (`FSF_MOUNT_FMT` carries no device column and cannot emit mount's two header lines), so either extend it or record the exemption in both files.

**4. A failing or absent `mount(8)` silently disables the whole filter.** With a `mount` stub that exits 1, `/proc` and `/aha` are reported again and the client says nothing of its own: `EXCLMP=$(mount | awk ...)` takes awk's status. The Linux client warns in the equivalent case ("/proc/filesystems not readable, dynamic nodev exclusions disabled"). One line, and no test covers the path.

**5. Neither new default flag is falsifiable by this test.** The `df` stub is written to honour `-T local`, so the test proves the client passes the flag, not that AIX accepts it. If `df -Ik -T local` or `/usr/sysv/bin/df -i -l` is rejected, stdout is empty and every AIX host's disk column goes yellow by default -- broader than the pseudo-fs rows being fixed. That is the merge gate; alternatively make the flags degrade, retrying without them on an empty non-zero run.

Nits: `INCLUDE_TYPES` is IFS-dependent (shell word-splitting) while `EXCLUDE_TYPES` is not (awk splits on `" "`), shown with `IFS=:`; the man page now explains `no` before introducing it, since the AIX paragraph landed between the "yes" and "no" sentences; and "AIX is the last client without the `XYMONCLIENT_FS_*` type filter" is only true of the maintained set -- hp-ux, irix, osf1, sco_sv, sunos and unixware also lack it (the shipped man page text is accurate).

PR #408, opened 2026-08-20
client: add a foreground mode to runclient.sh

bonomani:
`xymonlaunch` has had `--no-daemon` for years (`common/xymonlaunch.c:505`), but nothing in the tree uses it: both `client/runclient.sh:89` and `xymond/xymon.sh.DIST:37` start it in daemon mode with a pidfile. A service manager that supervises processes directly — systemd `Type=simple`, runit, s6, daemontools — is therefore handed a `xymonlaunch` that forks and a script that exits, and has nothing left to watch.

Every packager solves this privately, and not identically:

- `debian/xymon-client.init:33-35` reconstructs the environment `runclient.sh` computes, but sets only `MACHINE`/`MACHINEDOTS` — `SERVEROSTYPE` and `XYMONOSSCRIPT` are dropped.
- The Terabithia RPMs delete `runclient.sh` and reimplement it.
- [xymon-monitoring/xymon-rpm](https://github.com/xymon-monitoring/xymon-rpm) ships a `xymonlaunch-run` wrapper whose client half is exactly this function.

The environment computation is the part worth centralising: `MACHINEDOTS`, `SERVEROSTYPE`, `XYMONOSSCRIPT`, `MACHINE` and `XYMONCLIENTHOME` are derived in `runclient.sh` and nowhere else, so anyone bypassing the script has to re-derive them and can quietly get it wrong.

### What this adds

`runclient.sh foreground` — the same writability checks as `start`, then `exec xymonlaunch --no-daemon`, so the process the supervisor started *becomes* xymonlaunch. No pidfile is written, because xymonlaunch only writes one in the daemonizing parent and a supervisor tracks the process itself.

`CLIENTHOSTNAME` and `CLIENTOS` now override the detected hostname and OS. Both names are already Xymon's own — `debian/xymon-client.default` documents them — and they let a unit set the reported name from an environment file instead of editing the script.

`start`, `stop`, `restart` and `status` are untouched.

### Testing

Built a client from this branch and ran both paths:

- `start` — still forks, prints `Xymon client for linux started on <host>`, writes `logs/clientlaunch.<host>.pid`.
- `foreground` — the started PID is `xymonlaunch --no-daemon --config=.../clientlaunch.cfg`, no pidfile is created, and it exits on `SIGTERM`.

### Why it matters beyond packaging

This is the precondition for shipping a working systemd unit upstream. The unit on `devel` (`tools/xymonlaunch.service`) cannot serve a client: it runs `xymoncmd --no-env`, which loads the *server* environment, and `main`'s `xymoncmd` does not even accept `--no-env`. With a foreground mode, a client unit is `ExecStart=.../runclient.sh foreground` and needs no wrapper at all.

An equivalent `foreground` for `xymond/xymon.sh.DIST` is the obvious follow-up; kept out of this change to keep it reviewable.

### DIFF
--- a/client/runclient.sh
+++ b/client/runclient.sh
@@ -15,9 +15,11 @@
 #
 # $Id$
 
-# Default settings for this client
-MACHINEDOTS="`uname -n`"			# This systems hostname
-SERVEROSTYPE="`uname -s | tr '[ABCDEFGHIJKLMNOPQRSTUVWXYZ/]' '[abcdefghijklmnopqrstuvwxyz_]'`"	# This systems operating system in lowercase
+# Default settings for this client. CLIENTHOSTNAME and CLIENTOS override
+# the detected values, so a service manager can set them from its own
+# environment file instead of editing this script.
+MACHINEDOTS="${CLIENTHOSTNAME:-`uname -n`}"			# This systems hostname
+SERVEROSTYPE="`echo ${CLIENTOS:-\`uname -s\`} | tr '[ABCDEFGHIJKLMNOPQRSTUVWXYZ/]' '[abcdefghijklmnopqrstuvwxyz_]'`"	# This systems operating system in lowercase
 XYMONOSSCRIPT="xymonclient-$SERVEROSTYPE.sh"
 
 # Command-line mods for the defaults
@@ -34,8 +36,11 @@ do
 	        CONFIGCLASS="`echo $1 | sed -e 's/--class=//' | tr '[ABCDEFGHIJKLMNOPQRSTUVWXYZ/]' '[abcdefghijklmnopqrstuvwxyz_]'`"
 		;;
 	  --help)
-	  	echo "Usage: $0 [--hostname=CLIENTNAME] [--os=rhel3|linux22] [--class=CLASSNAME] start|stop"
+	  	echo "Usage: $0 [--hostname=CLIENTNAME] [--os=rhel3|linux22] [--class=CLASSNAME] start|stop|restart|status|foreground"
 		exit 0
+		;;
+	  foreground)
+	  	CMD=$1
 		;;
 	  start)
 	  	CMD=$1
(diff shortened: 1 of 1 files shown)


bonomani:
Closing this — it turns out upstream already does all of it, and I should have checked `xymoncmd` before writing a new mode.

`common/xymoncmd.c` is the C implementation of what `runclient.sh` computes in shell: `MACHINEDOTS` at lines 47-63 (`$HOSTNAME`, else `uname()`, else `uname -n`), `SERVEROSTYPE` at 67-80, then `loadenv()` and `execvp()`. `lib/environ.c:174-186` derives `MACHINE` from `MACHINEDOTS`, and `client/xymonclient.sh:30-31` derives `XYMONOSSCRIPT` itself when it is empty. Its env-file search at 145-151 finds the client config without help.

Tested with the packaged client on AlmaLinux 10, every case under `env -i` so the environment really is empty, with a listener on 1984 to confirm a report actually arrives:

| invocation | result |
| --- | --- |
| the shell wrapper my packaging ships | `client <host>.linux linux`, 17 sections |
| `xymoncmd xymonlaunch --no-daemon --config=...` with only `XYMONCLIENTHOME` | identical |
| the same with **nothing** set at all | identical |

So `xymoncmd xymonlaunch --no-daemon --config=$XYMONCLIENTHOME/etc/clientlaunch.cfg` is already the foreground client, symmetric with the `xymoncmd xymonlaunch --no-daemon` I was already using for the server. No new verb needed.

The `CLIENTHOSTNAME`/`CLIENTOS` half was also redundant — setting `MACHINEDOTS=fake.example.com SERVEROSTYPE=linux` in the environment gives `client fake,example,com.linux linux`, so upstream already supports the override under its own names. I was inventing a second vocabulary for a feature you have.

One real bug does survive this: the `*)` case in both `client/runclient.sh` and `xymond/xymon.sh.DIST` ends with `break;`, which does nothing outside a loop, so a mistyped verb prints usage and exits **0**. Happy to send that as a two-line fix on its own if you want it.

PR #409, opened 2026-08-20
build: add install-tools for the diagnostic programs the build discards

bonomani:
Several programs the build produces are never installed, so they die with the build tree:

- `lib/`'s `all` builds `loadhosts`, `stackio`, `availability`, `locator`, `tree`;
- `xymonnet/`'s `all` builds `contest` (a `DBGTOOLS`).

Nothing installs any of the six, so anyone wanting `stackio` to see how a config file parses, `loadhosts` to see what `hosts.cfg` resolves to, or `contest` to probe a service by hand has to keep a build tree or copy them out. Every packaging copies them out — xymon-monitoring/xymon-rpm ships the `lib/` five as a `xymon-tools` subpackage.

### What this adds

`make install-tools`: `lib/` and `xymonnet/` each get an `install-tools` target for the programs they build but discard, and `build/Makefile.rules` drives both. The six land in `INSTALLBINDIR`, chowned to `XYMONUSER` and `PKGBUILD`-gated like every other binary. Kept **out of `INSTALLTARGETS`** — what a plain `make install` ships is the installer's call, not a side effect here — so `make install` stays byte-for-byte unchanged.

### Only the genuinely-discarded ones

Terabithia's tools package copies eight, but two of those — `web/xymonpage` and `xymond/rrdcachectl` — are in their directories' `PROGRAMS`, so `make install` **already installs them** to `INSTALLBINDIR`; Terabithia re-copies them redundantly. The set that is actually built-and-thrown-away is the `lib/` five **plus `contest`**, and that is exactly what this installs. `xymonpage`/`rrdcachectl` are deliberately not here — adding them would double-install files the build already places.

### Testing

Server build, `XYMONTOPDIR=/opt/xymon`:

- `make install` — adds none of the six (verified for all six names).
- `make install-tools` — all six installed, mode 755, owned by `XYMONUSER`, and they run.


bonomani:
Parking this as a draft — not abandoning it, but I do not think it earns your review time right now.

`xymon-monitoring/xymon-rpm` is the only packaging that ships these tools, and it can copy them out of the build tree in three lines of `%install` without patching anything. That workaround is stable: `lib/Makefile` builds them in `all`, so they are always there, and if that ever changed our nightly build would go red the same day.

So the honest cost/benefit is: this adds a make target to your tree to save one packager three lines. If someone else ever wants the `lib/` diagnostics installed, or if you would rather the build stopped discarding what it builds, say so and I will bring it back.

PR #410, opened 2026-08-20
build: install the libraries and headers with make install-devel

bonomani:
`make install` installs no libraries and no headers, so nothing builds against an *installed* Xymon. Every packager digs `include/*.h`, `lib/*.h`, `lib/*.a` out of the build tree by hand.

Two commits, because the first is not mine.

**1. Libraries — partial backport of devel `8dff8464` (@jccleaver).** Just the install machinery (`INSTALLLIBDIR`, the target, `lib/Makefile`'s rule) — 2 of its 8 files. No `.so` building: `main` has no rule that produces one, so the `*.so*` glob matches nothing. (Whether `main` should build a shared library at all is a separate question, tracked in #436.)

**2. Headers, opt-in, pkg-config — on top.** `make install-devel` installs the archives, any `*.so*`, 58 headers, and a generated `xymon.pc`.

- `INSTALLINCDIR` keeps `include/` and `lib/` as siblings — `libxymon.h` reaches its headers through 62 `"../lib/..."` includes, so one `-I$INSTALLINCDIR/include` compiles.
- Out of `INSTALLTARGETS`: on `main` the payload is static archives + headers nothing reads at runtime, and a default-on target would drop them under `$XYMONHOME`, where a spec globbing the server tree swallows them into its main package and collides with its own `-devel`. Makes its own dirs (not via `install-dirs`), so a staging root gets only those + payload.
- `xymon.pc` from `build/xymon.pc.in`, so a consumer needs only `pkg-config --cflags --libs --static xymon`.
- Compat symlinks skipped under `PKGBUILD` — an unpackaged symlink is a hard error to `rpmbuild`/`dpkg`.
- Default is one `$(XYMONHOME)/sdk` tree; overriding `INSTALLINCDIR`/`INSTALLLIBDIR` into FHS trees leaves `INSTALLSDKDIR` unused and gives a byte-identical layout.
- `install-devel`, not `install-libs`: 58 headers are the bulk. `devel` keeps `install-libs` (honest there — it builds `.so`); a matching rename is a separate change.

Tested on AlmaLinux 10, default and relocated: `make install` unchanged; `make install-devel` gives 5 archives + 58 headers + `xymon.pc`; `pkg-config … --static xymon` compiles, links, runs.

### DIFF
--- a/build/Makefile.rules
+++ b/build/Makefile.rules
@@ -26,7 +26,7 @@ endif
 else
 	BUILDTARGETS = lib-build common-build xymongen-build xymonnet-build xymonproxy-build docs-build build-build xymond-build web-build client
 	CLIENTTARGETS = lib-client common-client build-build
-	INSTALLTARGETS = install-xymongen install-xymonnet install-xymonproxy install-xymond install-web install-docs install-client install-servermsg
+	INSTALLTARGETS = install-libs install-xymongen install-xymonnet install-xymonproxy install-xymond install-web install-docs install-client install-servermsg
 	CFLAGS += $(PCREINCDIR)
 	XYMONCLIENTHOME = $(XYMONTOPDIR)/client
 	XYMONHOME = $(XYMONTOPDIR)/server
@@ -34,6 +34,9 @@ endif
 
 ifndef INSTALLBINDIR
 INSTALLBINDIR = $(XYMONHOME)/bin
+endif
+ifndef INSTALLLIBDIR
+INSTALLLIBDIR = $(XYMONHOME)/lib
 endif
 ifndef INSTALLETCDIR
 INSTALLETCDIR = $(XYMONHOME)/etc
@@ -211,6 +214,17 @@ ifndef PKGBUILD
 	chmod 755 $(INSTALLROOT)$(INSTALLBINDIR)
 endif
 
+	mkdir -p $(INSTALLROOT)$(INSTALLLIBDIR)
+ifneq ($(INSTALLLIBDIR),$(XYMONHOME)/lib)
+	rm -f $(INSTALLROOT)$(XYMONHOME)/lib || true
+	ln -s $(INSTALLLIBDIR) $(INSTALLROOT)$(XYMONHOME)/lib || true
+endif
+ifndef PKGBUILD
+	chown $(XYMONUSER) $(INSTALLROOT)$(INSTALLLIBDIR)
+	chgrp `$(IDTOOL) -g $(XYMONUSER)` $(INSTALLROOT)$(INSTALLLIBDIR)
+	chmod 755 $(INSTALLROOT)$(INSTALLLIBDIR)
+endif
+
 	mkdir -p $(INSTALLROOT)$(INSTALLETCDIR)
 ifneq ($(INSTALLETCDIR),$(XYMONHOME)/etc)
 	rm -f $(INSTALLROOT)$(XYMONHOME)/etc || true
@@ -368,6 +382,9 @@ install-custom:
 
 
 client-install: install-client
+
+install-libs: install-dirs
+	XYMONHOME="$(XYMONHOME)" XYMONUSER="$(XYMONUSER)" IDTOOL="$(IDTOOL)" PKGBUILD="$(PKGBUILD)" INSTALLROOT="$(INSTALLROOT)" INSTALLLIBDIR="$(INSTALLLIBDIR)" $(MAKE) -C lib install
 
 install-client: client
 	XYMONHOME="$(XYMONCLIENTHOME)" INSTALLROOT="$(INSTALLROOT)" XYMONUSER="$(XYMONUSER)" IDTOOL="$(IDTOOL)" PKGBUILD="$(PKGBUILD)" LOCALCLIENT="$(LOCALCLIENT)" $(MAKE) -C client install
--- a/build/Makefile.rules
+++ b/build/Makefile.rules
@@ -26,7 +26,7 @@ endif
 else
 	BUILDTARGETS = lib-build common-build xymongen-build xymonnet-build xymonproxy-build docs-build build-build xymond-build web-build client
 	CLIENTTARGETS = lib-client common-client build-build
-	INSTALLTARGETS = install-libs install-xymongen install-xymonnet install-xymonproxy install-xymond install-web install-docs install-client install-servermsg
+	INSTALLTARGETS = install-xymongen install-xymonnet install-xymonproxy install-xymond install-web install-docs install-client install-servermsg
 	CFLAGS += $(PCREINCDIR)
 	XYMONCLIENTHOME = $(XYMONTOPDIR)/client
 	XYMONHOME = $(XYMONTOPDIR)/server
@@ -35,9 +35,28 @@ endif
 ifndef INSTALLBINDIR
 INSTALLBINDIR = $(XYMONHOME)/bin
 endif
-ifndef INSTALLLIBDIR
-INSTALLLIBDIR = $(XYMONHOME)/lib
+# The development kit -- headers, static libraries and xymon.pc -- installs as
+# one relocatable subtree. A source install puts the whole kit under
+# $(XYMONHOME)/sdk; a packager overrides INSTALLINCDIR and INSTALLLIBDIR to
+# split them across FHS trees (/usr/include vs /usr/lib), leaving INSTALLSDKDIR
+# unused.
+ifndef INSTALLSDKDIR
+INSTALLSDKDIR = $(XYMONHOME)/sdk
 endif
+# The include root holds include/ and lib/ side by side: libxymon.h reaches its
+# remaining headers through "../lib/..." includes, so the pair has to move
+# together. It is the sdk dir itself, not an include/ beneath it, or the
+# default -I$(INSTALLINCDIR)/include would double to .../include/include.
+ifndef INSTALLINCDIR
+INSTALLINCDIR = $(INSTALLSDKDIR)
+endif
+# The archives sit beside the lib/ headers they belong to -- the source lib/
+# carries both *.h and *.a -- so the whole kit is a single tree.
+ifndef INSTALLLIBDIR
+INSTALLLIBDIR = $(INSTALLSDKDIR)/lib
+endif
+# The version string, read once from the header for xymon.pc.
+XYMONVERSION := $(shell sed -n 's/.*VERSION "\([^"]*\)".*/\1/p' include/version.h)
 ifndef INSTALLETCDIR
 INSTALLETCDIR = $(XYMONHOME)/etc
 endif
@@ -214,16 +233,6 @@ ifndef PKGBUILD
 	chmod 755 $(INSTALLROOT)$(INSTALLBINDIR)
 endif
 
-	mkdir -p $(INSTALLROOT)$(INSTALLLIBDIR)
-ifneq ($(INSTALLLIBDIR),$(XYMONHOME)/lib)
-	rm -f $(INSTALLROOT)$(XYMONHOME)/lib || true
-	ln -s $(INSTALLLIBDIR) $(INSTALLROOT)$(XYMONHOME)/lib || true
-endif
-ifndef PKGBUILD
-	chown $(XYMONUSER) $(INSTALLROOT)$(INSTALLLIBDIR)
-	chgrp `$(IDTOOL) -g $(XYMONUSER)` $(INSTALLROOT)$(INSTALLLIBDIR)
-	chmod 755 $(INSTALLROOT)$(INSTALLLIBDIR)
-endif
 
 	mkdir -p $(INSTALLROOT)$(INSTALLETCDIR)
 ifneq ($(INSTALLETCDIR),$(XYMONHOME)/etc)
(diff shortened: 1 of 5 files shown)


bonomani:
Raising an inconsistency between this PR and its sibling #409, since it is a decision for you rather than something I want to settle silently.

**#409 gates its addition; this one does not.** #409 adds `make install-tools` and deliberately keeps it out of `INSTALLTARGETS`, on the grounds that "adding five binaries to what every installation gets is a decision for whoever runs the install, not a side effect of this change". This PR adds `install-libs` straight into `INSTALLTARGETS`, so a plain `make install` now also writes `$XYMONHOME/lib/*.a` and roughly sixty headers under `$XYMONHOME/include/`. That is the larger and less reversible of the two additions, and it is the one without a gate — the opposite of how I argued it next door.

I did it this way because devel's commit does (`install-libs` is in its `INSTALLTARGETS` too, and preserving that was the point of keeping J.C.'s commit intact), but consistency between the two PRs matters more than fidelity to a ten-year-old default. Three options, and I am happy with any of them:

1. **Leave it as is.** Libraries and headers are what a `-devel` package is made of, and every packager already extracts them by hand; installing them by default is arguably just correct.
2. **Gate it like #409** — drop `install-libs` from `INSTALLTARGETS` and let people run `make install-libs`. Consistent with the sibling PR, at the cost of an extra step for anyone who wants to build against Xymon.
3. **Split the difference** — install the libraries by default (they are small and already built) and gate only the headers.

Say which you prefer and I will push it; the change is one line either way.

Two smaller things in the same area, in case they affect your choice:

- `INSTALLINCDIR` is not handled in `install-dirs`, so unlike `bin`, `etc`, `ext`, `tmp` and the `lib` directory this PR adds, it gets no `mkdir`, no `chown`/`chgrp` under `ifndef PKGBUILD`, and no compatibility symlink when relocated. If option 1 or 3 is chosen, it should get the same treatment as the others, and I will add that.
- With the default `INSTALLINCDIR = $(XYMONHOME)/include`, the sibling layout inside it produces `$(XYMONHOME)/include/include/libxymon.h`. Only the default reads oddly — a packager setting `INSTALLINCDIR=/usr/include/xymon` gets the clean `/usr/include/xymon/{include,lib}` — but it is the first thing anyone will notice. The alternative is described in the PR description.

bonomani:
One factual addition to the above, since it bears on the choice and I had it wrong in my own head until I checked.

**On `main`, the default build produces static archives only.** `lib/Makefile` builds `libxymon.a`, `libxymonclient.a` and `libxymonclientcomm.a`, and nothing links against a shared object. The shared build is present but commented out, in `build/Makefile.Linux`, `build/Makefile.GNU` and `build/Makefile.GNU_kFreeBSD`:

```
# XYMONLIBRARY=libxymon.so
```

On `devel` it is live — `lib/Makefile` there has real `.so` rules — which means the same `INSTALLTARGETS` entry is doing two different jobs in the two branches. On `devel` it is load-bearing: the binaries need the shared library present at runtime, so `make install` has to place it. On `main` as it stands, the files it installs have no runtime consumer at all; they are only useful to someone compiling against Xymon.

Worth noting that your `8dff8464` introduced the `.so` support and the `INSTALLTARGETS` entry in the same commit, so the default arrived together with the thing that needed it. Whether that makes the default right or wrong on `main` depends on something only you know — whether shared-library building here is dormant or finished. If it is dormant, leaving the default in place is one less thing to change later; if it is finished, option 2 stops `main` installing files nothing reads.

One consequence for packagers either way: upstream's in-tree `debian/` has no file list for the server package — no `debian/xymon.install`, no `.files` — so its `.deb` is whatever `make install` leaves in the staging directory, and these files would join it without anyone choosing that. I have not checked the Debian archive package, which is maintained separately and is newer, so this may not apply there.

I am not arguing for a particular option; I talked myself into one and then found the evidence thinner than I thought, so here are the facts instead.

bonomani:
Resolving my own open questions rather than leaving them hanging — all three are one-line changes if you disagree with any of them.

**1. The gating.** Leaving `install-libs` in `INSTALLTARGETS`, as your commit had it. What changed my mind is that the objection barely applies to the people this is for: a packaging decides which subpackage the files land in and the user decides again whether to install it, so a default costs packagers nothing and saves them an extra target. The affected group is direct source installs, which is not enough to justify overriding your default inside a backport of your own commit. If you would rather it matched #409's opt-in shape, say so and it is one line.

**2. `INSTALLINCDIR` in `install-dirs`** — done, pushed. It was the only `INSTALL*DIR` here without `mkdir`, ownership under `ifndef PKGBUILD`, or a compatibility symlink when relocated, and the first person to move it would have found that out the hard way.

**3. The `include/include/` default** — keeping it, and I can now give the reason properly rather than as a preference. `include/libxymon.h` reaches the rest of the tree through **62** `#include "../lib/..."` lines, so the installed form has to keep an `include/` directory with a `lib/` beside it. Holding both inside `INSTALLINCDIR` is what makes that pair self-contained, and therefore what lets `INSTALLINCDIR` be relocated independently: `INSTALLINCDIR=/usr/include/xymon` gives a clean `/usr/include/xymon/{include,lib}`. Splitting the halves across `INSTALLINCDIR` and `INSTALLLIBDIR` would read better at the default and break all 62 includes the moment someone set those two to non-sibling paths. The odd-looking default is the price of that, and the reasoning is now a comment in `Makefile.rules` rather than only here.

Retested on AlmaLinux 10 after the change:

| | |
| --- | --- |
| defaults | `$XYMONHOME/include` owned by `XYMONUSER`, mode 755, `libxymon.h` in place, 5 archives beside it |
| `INSTALLINCDIR=/usr/include/xymon` | 58 headers there, compatibility symlink created and resolving |
| actually usable | `gcc -I/usr/include/xymon/include` compiles a file that `#include <libxymon.h>` |

That last row is the one I had never actually checked before, and it is the only one that proves the layout works.

bonomani:
Reversing my previous comment: **`install-libs` is now out of `INSTALLTARGETS`**, so a plain `make install` is unchanged and `make install-libs` opts in — the shape #409 uses. Pushed.

Two things changed my mind, and neither was the argument I made last time.

**`main` has no shared library at all, not even a switchable one.** I described it as "commented out", which implies it can be turned on. `XYMONLIBRARY` appears in exactly three places — comments in `build/Makefile.GNU`, `Makefile.GNU_kFreeBSD` and `Makefile.Linux` — and no makefile reads the variable. There is no rule anywhere that builds a `.so`. So on `devel` your `INSTALLTARGETS` entry is load-bearing, because the binaries link against shared objects the install has to place; on `main` the same entry installs static archives and headers that nothing reads at runtime. The default was right where you wrote it and does not carry across.

**The default has a silent failure mode for packagers.** `INSTALLLIBDIR` and `INSTALLINCDIR` default to `$(XYMONHOME)/lib` and `$(XYMONHOME)/include` — inside the server tree. A spec that globs that tree (mine does, with one `%{xymonhome}/server` line) swallows the archives and headers into its *main* package, and then its `-devel` subpackage either duplicates those paths — an rpm file conflict — or has to exclude them back out. Nothing warns you; it surfaces when a user installs both packages. Out of `INSTALLTARGETS`, that cannot happen by accident: the files only appear when someone runs the target, and anyone running it is already choosing where they go.

The directories moved into `install-libs` for the same reason — a target nobody asked for should leave nothing behind, not even empty directories.

Retested on AlmaLinux 10:

| | |
| --- | --- |
| plain `make install` | no `lib/`, no `include/`, no `.a` or `.h` anywhere |
| `make install-libs` | 5 archives, 58 headers, owned by `XYMONUSER`, mode 755 |
| relocated, then opt in | nothing until asked; then the archives, the compatibility symlink, and `gcc -I<dir>/include` compiles a file that includes `<libxymon.h>` |

Your commit is untouched — the change is in mine on top, so the default you chose is still visible in the history with the reason it was right there.

Side effect worth having: this PR no longer edits the `INSTALLTARGETS` assignments at all, which were the lines every other install change has to touch. All five open PRs now merge in all 120 orders.

bonomani:
Retitled to say **partial** backport, which the commit body always said but the subject line did not.

To be explicit about what is and is not here: your `8dff8464` touched 8 files and 78 insertions. This takes 2 files and 23 — `build/Makefile.rules` and `lib/Makefile`, the install machinery. Everything that actually builds a shared object is left on `devel`:

```
build/Makefile.Linux   common/Makefile   configure.client
configure.server       xymond/Makefile   xymonnet/Makefile
```

So `main` still has no shared library and no way to ask for one, and this PR does not change that — it installs the static archives and headers the build already produces. The `*.so*` glob in the install rule is tolerant of their absence rather than a sign they might appear.

Whether the rest of that commit should follow is a separate question I am not raising here: it changes how every binary links on every platform Makefile, and it is your design rather than something to port on your behalf. If you do want it, this PR is the half that has to land first anyway, since without an install rule the shared objects would be built and discarded.

bonomani:
Correcting my own phrasing in the previous comment, because it misrepresents the PR.

I wrote that this is "the half that has to land first". That reads as though the change were incomplete and something else were owed. It is not. On `main` the build produces static archives and headers, this installs both, and a consumer can compile and link against the result — `gcc -I$INSTALLINCDIR/include ... -L$INSTALLLIBDIR -lxymon` works, which I tested. Nothing is missing for the purpose.

**"Partial backport" is about attribution only.** Your `8dff8464` touched 8 files; I took 2. Saying plain "backport of 8dff8464" would credit your commit for changes this does not carry — the shared-library building in `common/Makefile`, `configure.*`, `xymond/Makefile`, `xymonnet/Makefile` and `build/Makefile.Linux`. That is the only reason for the word.

The single trace of the other half is the `*.so*` loop in `lib/Makefile`, and it is deliberate: it copies shared objects when the tree built any and skips quietly when it did not, so the rule is correct on `main` today and stays correct on `devel` or on a `main` that later grows shared libraries. An unmatched glob must not look like a failure.

So: nothing here is waiting on anything. If you never bring the rest of `8dff8464` across, this PR is still complete and correct on its own terms.

bonomani:
A naming question, raised rather than changed, since the target is yours and it exists on both branches.

`install-libs` describes about a third of what the target does. It installs the static archives, any `*.so*` the tree built, **and 58 headers** into `INSTALLINCDIR`. For anyone compiling against Xymon the headers are the part that matters most, and the name gives no hint they are there.

Every other target in `build/Makefile.rules` is named for what it installs — `install-bin`, `install-man`, `install-cfg`, `install-web`, `install-docs`. By that convention this one is **`install-devel`**: development files, which is what the set is and the word every packaging system already uses for it.

I have deliberately not renamed it here. It is `install-libs` in your `8dff8464` and still is at `build/Makefile.rules:332` on `4.x-master`, where `lib/Makefile` has real `.so` rules and the name is honest. Renaming it on `main` alone would leave the branches disagreeing about a target name and turn a future devel→main merge into a conflict over cosmetics — and this PR is a backport of your commit, so renaming your target inside it seemed like the wrong place to decide it.

If you would like the rename, it wants doing on both branches together, and I am happy to prepare that as a separate change. If you would rather keep `install-libs` for continuity with devel, that is an equally good answer and this PR needs nothing.

(Two alternatives I discarded: `install-static` is wrong, since the headers are the bulk of it and devel builds shared objects through the same target; splitting into `install-libs` plus `install-headers` gives two accurate names but nobody wants either half alone.)

bonomani:
Done — the target is now **`install-devel`** on this branch, pushed. Answering my own question rather than leaving it hanging, since it is a one-word change either way.

The reasoning stands as written above: it installs the archives, any shared objects, and 58 headers, and every other target here is named for what it installs. `install-libs` described about a third of it.

Scope is deliberately narrow. Your `8dff8464` backport commit still adds `install-libs` exactly as you wrote it; the rename is in my commit on top, so the history shows it as a later decision rather than an edit to yours. `devel` is untouched — there `install-libs` is honest, because that branch builds shared libraries.

That does leave the two branches disagreeing on the name, which is the one real cost. If you would rather they matched, either direction is fine by me: say the word and I will revert to `install-libs` here, or prepare the matching rename for `devel` as a separate change.

Retested on AlmaLinux 10 with both directories relocated:

| | |
| --- | --- |
| plain `make install` | creates neither directory — unchanged |
| `make install-libs` | gone, as intended |
| `make install-devel` | 5 archives, 58 headers |
| `gcc -I<incdir>/include` on a file including `<libxymon.h>` | compiles |

SoundGoof:
### Findings

####  `install-devel` creates the complete runtime layout

`build/Makefile.rules:356` declares:

```make
install-devel: install-dirs
```

The full conversation, says the development
directories were moved into the opt-in target so an unrequested target leaves
nothing behind.

The prerequisite defeats that isolation in the other direction. Running
`make install-devel` invokes `install-dirs`, which creates the complete runtime
layout before installing the development files.

A clean staged reproduction created:

- Server directories for `bin`, `etc`, `ext`, `tmp`, `web`, and `www`
- Web subdirectories including `html`, `rep`, `snap`, and `wml`
- Data directories for RRDs, history, acknowledgements, and host data
- The requested 5 archives and 58 headers

This can pollute a development-package staging root with empty runtime
directories and place them in the wrong package.

Remove the `install-dirs` prerequisite. Create only `XYMONHOME`,
`INSTALLLIBDIR`, and `INSTALLINCDIR`, which are sufficient for the development
payload and compatibility symlinks.

#### client configuration discards relocated development paths

`configure.server:570` persists `INSTALLLIBDIR` and `INSTALLINCDIR`, but
`configure.client` has no equivalent handling.

The conversation reports that a client build configures, builds, and installs
unchanged. That confirms the existing client workflow still works, but it does
not exercise relocation of the new development target.

This was reproduced with:

```sh
INSTALLLIBDIR=/usr/lib/custom-xymon \
INSTALLINCDIR=/usr/include/custom-xymon \
./configure.client
```

Neither value was written to the generated Makefile. After removing the ambient
environment variables, Make resolved them as:

```text
INSTALLLIBDIR=$(XYMONHOME)/lib
INSTALLINCDIR=$(XYMONHOME)/devel
```

Client builds produce the client archives and expose `install-devel`, so they
should preserve these paths consistently. Add the same two Makefile emission
blocks to `configure.client`.

### Validation

All five GitHub CI checks pass.

The installed sibling header layout works: a consumer including
`<libxymon.h>` compiled successfully against the staged headers. The findings
above concern target isolation and client relocation, not the payload layout.

bonomani:
@SoundGoof Both findings fixed (squashed into `3e15d311`).

1. **`install-dirs` prereq dropped** — it built the whole runtime layout before the payload; the target makes only its own dirs now, so a staged install gets the kit dirs + payload, no runtime dirs.
2. **`configure.client` now persists** `INSTALLLIBDIR`/`INSTALLINCDIR` like `configure.server`, so a relocated `install-devel` on a client build honours them.

Three more on the same target: a generated `xymon.pc`; `PKGBUILD` now skips the compat symlinks too (an unpackaged symlink is a hard error to rpmbuild/dpkg); and the from-source default is one `$(XYMONHOME)/sdk` tree (default only — overriding both dirs leaves `INSTALLSDKDIR` unused, FHS layout byte-identical).

On flattening the includes: not safe — all 58 headers on one `-I` shadow base-system headers, platform-specific (macOS `Availability.h`, FreeBSD/OpenBSD `md5.h`/`sha1.h`/`sha2.h`). The sibling layout avoids that. Kept deliberately.

Verified via staged installs plus green RPM builds (el8/9/10, Fedora, rawhide) and a clean `install-devel` on FreeBSD 14.2.

PR #411, opened 2026-08-20
build: let the client's install directories be set, as the server's can

bonomani:
`configure.client` collects `INSTALLETCDIR`/`INSTALLBINDIR`/`INSTALLWWWDIR` and writes them to the Makefile, where nothing reads them: `client/Makefile` hardcodes `$(XYMONHOME)/{etc,bin,tmp,logs}`, `install-client` passes nothing down, and a client-only build never runs `install-dirs`. So a client packager's `INSTALL*DIR` is silently discarded — worse than not offered.

Adds `INSTALLCLIENT{ETC,BIN,TMP,LOG}DIR`, created by a new `install-clientdirs` and passed down by `install-client`; a client-only build maps the `INSTALLETCDIR`/`INSTALLBINDIR` `configure.client` already collects onto them, so the existing options start working. Client-specific names so a server build keeps its embedded client under `XYMONCLIENTHOME`. The symlink-back is what `install-dirs` has always done for the server, so every upstream-written path keeps resolving. Both configure scripts record the four (only one runs — `configure` is an `exec` dispatch).

Relocated files also get their configured ownership: the recursive `chown` stops at `XYMONHOME`, so a moved `etc`/`bin` left `clientlaunch.cfg`, `xymonclient.cfg`, `localclient.cfg` root-owned (copied without `-p`). Chowned by name now — a no-op in the default layout, not recursive since a relocated `etc` may be shared. Reported by @SoundGoof.

Everyone does this by hand today: `debian/rules` moves `client/etc` to `/etc/xymon` + symlink, Terabithia `sed`s the paths, xymon-rpm moves to `/etc/xymon-client` — and admin config under `/usr` until then, which the FHS forbids.

The client's drop-in directories come with it. `clientlaunch.cfg` and `xymonclient.cfg` each end in `optional directory @XYMONHOME@/etc/<name>`, and nothing ever created `clientlaunch.d` or `xymonclient.d`: `xymond/Makefile` creates every one of the server's, `client/Makefile` created neither, and a drop-in an admin has to `mkdir` first is one in name only. Only `debian/xymon-client.dirs` compensates, so the gap surfaced as an RPM bug on RHEL9 (#522) rather than a build one, and a plain `make install` lacks them too. Created under the relocatable `etc` this adds, and chowned by name like the config files beside them.

`tests/server/config-include-dir.sh` pinned that every shipped config declares its directory and that the reader honours the directive; it now also pins that whoever ships the config creates it, so a drop-in added to a config and to nothing else fails there instead of shipping absent.

Tested: client-only default (unchanged); client-only `INSTALLETCDIR=/etc/xymon-client` (config there, `$XYMONCLIENTHOME/etc` symlinked); server build (embedded client stays under `client/etc`); server build with `INSTALLCLIENTETCDIR` set for `./configure` only (relocated without repeating it on the make line). Both drop-in directories land in each layout, 755 and owned by `XYMONUSER`, including `client/Makefile` run on its own; the install stays idempotent on a second run.

Scope: `etc`/`bin`/`tmp`/`logs` and the two drop-in directories under `etc`; `ext`/`local` left alone. Guards use `test A && test B` (`-a` is obsolescent, `dash` warns).

## What this changes for the three packagings

A default `make install` now produces two directories it did not:
`<clientetc>/clientlaunch.d` and `<clientetc>/xymonclient.d`, 755, owned by
`XYMONUSER` -- the same thing `xymond/Makefile` has always done for the
server's nine. Nothing else in the install output moves: the four
`INSTALLCLIENT*DIR` fall back to the old paths when unset, and the per-file
`chown`/`chgrp` added here sit inside the existing `ifndef PKGBUILD` guard, so
a packaging build skips them exactly as it skips the recursive `chown` today.

**Debian** -- `debian/rules:95` builds with `PKGBUILD=1`, then line 96 does the
same dance this PR exists to retire:

```
cd $C/usr/lib/xymon/client && mv etc/* $C/etc/xymon && rmdir etc && ln -s ../../../../etc/xymon etc
```

So the two arrive at `$C/etc/xymon/clientlaunch.d` and
`$C/etc/xymon/xymonclient.d` -- the exact paths `debian/xymon-client.dirs`
already lists. `dh_installdirs -a` (line 94) has created them empty by then,
and `rename()` over an empty directory succeeds, so `mv etc/*` replaces them
and nothing fails. Nothing to do; the two `.dirs` entries can go whenever, and
with `INSTALLCLIENTETCDIR=/etc/xymon` so can line 96. @RolandRosenfeld

**FreeBSD** -- `net-mgmt/xymon-client` copies its own top-level `Makefile`,
which includes this tree's `build/Makefile.rules`, and `install-client`
recurses into `client/`: both files this PR touches, so both directories land
in `STAGEDIR` at `%%WWWDIR%%/client/etc/`. `pkg-plist` carries the three config
files as `@sample %%WWWDIR%%/client/etc/*.cfg.sample` and has no `@dir` under
`etc/` at all -- its only three are

```
@dir %%WWWDIR%%/client/tmp
@dir %%WWWDIR%%/client/logs
@dir %%WWWDIR%%/client/ext
```

matching what `post-install` makes by hand. So the two will be orphans until
the plist gains

```
@dir %%WWWDIR%%/client/etc/clientlaunch.d
@dir %%WWWDIR%%/client/etc/xymonclient.d
```

@feld

**RPM** -- `xymon-monitoring/xymon-rpm` moves the client's `etc` to
`/etc/xymon-client` the same way Debian does, so the two arrive as
`/etc/xymon-client/clientlaunch.d` and `/etc/xymon-client/xymonclient.d`. Its
`%files` named that directory's contents one by one, and the failure mode was
the quiet one: `check-files` lists `-type f -o -type l` only, so an unpackaged
*directory* is never reported -- the build would have stayed green and the
package shipped without them, #522's gap surviving its own fix with nothing
saying so. It globs now, as the server half always has, so both directories are
packaged the day this merges. Nothing left to do there. The `rpm/` directory in
*this* tree is unmaintained and builds nothing anyone ships -- it is not that
spec, and it is the one #522 patches.

Terabithia `sed`s the paths and is not directory-sensitive.


bonomani:
Force-pushed one fix: the four `INSTALLCLIENT*DIR` are now recorded into the generated Makefile by both `configure.server` and `configure.client`.

Without that they only took effect when repeated on the `make install` command line. `build/Makefile.rules` uses `ifndef`, which does pick the variables up from the environment, so `make install INSTALLCLIENTETCDIR=...` worked and hid the gap — but setting them for `./configure` and then running a plain `make install`, which is how the packagings that want this actually build, silently did nothing. That is inconsistent with every other `INSTALL*DIR`, all of which configure records.

`configure` is an `exec` dispatch, so only one of the two scripts ever runs and both needed it. Added a fourth test case for the configure-time path.

SoundGoof:
## Findings

### Relocation is undone by client self-update

The new compatibility links make `bin`, `etc`, `tmp`, and `logs` symlinks:

https://github.com/xymon-monitoring/xymon/blob/b91ce53dcabda64142461ff287ef3dd7ddfaca28/build/Makefile.rules#L400-L434

`clientupdate` downloads an archive whose paths are relative to
`XYMONHOME` and extracts it there with plain `tar xf -`:

https://github.com/xymon-monitoring/xymon/blob/b91ce53dcabda64142461ff287ef3dd7ddfaca28/client/clientupdate.c#L200-L230

A normal update archive contains directory entries such as `bin/` and `etc/`.
GNU tar on Debian 13 and native tar on OpenBSD 7.9 and FreeBSD 14.3 replace
those symlinks with real directories during extraction. Updated files then
land under `XYMONHOME`, while `INSTALLCLIENTBINDIR` and
`INSTALLCLIENTETCDIR` remain stale.

I reproduced this with a realistic archive containing `bin/tool` and
`etc/clientversion.cfg`. Both links became directories and neither external
target was updated on those three systems. NetBSD 9.4 and 11.0 preserved the
links, so the behavior is tar-dependent.

Account for relocated directories in `clientupdate`, use a portable extraction
strategy that preserves directory symlinks, or restrict relocation to paths
that self-update cannot replace. Add a regression test using an archive with
explicit `bin/` and `etc/` entries.

### Relocated configuration loses `XYMONUSER` ownership

The relocated configuration files are created outside `XYMONCLIENTHOME`, but
the final ownership correction still covers only `XYMONHOME`:

https://github.com/xymon-monitoring/xymon/blob/b91ce53dcabda64142461ff287ef3dd7ddfaca28/client/Makefile#L87-L98

Using the documented workflow of building as `xymon` and installing as root,
the relocated binaries remain `xymon:xymon`, but `clientlaunch.cfg`,
`xymonclient.cfg`, and `localclient.cfg` become `root:root` with mode 0644.
The same files are `xymon:xymon` in the default layout because the final
recursive chown reaches them.

Apply the configured ownership to the specific relocated files after copying
and merging them. Avoid recursively changing an arbitrary shared destination
such as `/etc`.

## Validation

- Read the complete current conversation, including the author's follow-up.
  There are no reviews or inline threads.
- Reviewed current head `b91ce53d`.
- `git diff --check` passes.
- Configure-time custom paths persisted and produced the expected staged
  payload and compatibility links.
- A server build with only its server etc directory relocated kept the
  embedded client entirely under `XYMONCLIENTHOME`.
- Client-built suite: 56 passed, 31 skipped, 0 failed.
- Reproduced ownership using build-as-`xymon`, install-as-root on Debian 13.
- Tested update extraction on Debian 13, NetBSD 9.4, NetBSD 11.0,

bonomani:
Thanks @SoundGoof. **Finding 2 fixed (in `69e0d7bb`); finding 1 not yet — I would like to agree the shape first.**

**Ownership.** Reproduced and fixed: the relocated files are chowned by name, not by a recursive chown, since a relocated `etc` may be shared. Covered the tools in a relocated `bin` too — they were right in your run only because `cp -fp` preserved the build tree's ownership (luck: build as root and they would be root's). `XYMONUSER=nobody`, both dirs relocated:

| | before | after |
|---|---|---|
| `etc/{clientlaunch,xymonclient,localclient}.cfg` | uid 0 | uid 65534 |
| relocated `bin/*` | uid 0 | uid 65534 |

Default layout unchanged; client suite 56 passed / 31 skipped / 0 failed.

**Self-update.** Confirmed: `clientupdate.c` chdirs to XYMONHOME and `popen("tar xf -")`, walking straight over the new links. I would rather **refuse** than extract cleverly — `--keep-directory-symlink` is GNU-only, and even preserved, extraction runs as the xymon user into distro-owned `bin`/`etc`, the opposite of what relocation is for. So: `clientupdate` `lstat`s `bin`/`etc`, and a symlink (= relocated) makes it exit with a clear message instead of reverting the layout. That is your third option, from the updater's end. Match what you had in mind? If so I will add it with the regression test pointed at the refusal.

PR #412, opened 2026-08-20
build: let the web server config be installed where httpd reads it

bonomani:
`xymond/Makefile` generates `xymon-apache.conf` and installs it into
`INSTALLETCDIR` — Xymon's own `etc` directory, which is not a directory any
web server reads. So every packaging moves it afterwards: upstream's own
`debian/rules`, the Terabithia RPMs and `xymon-monitoring/xymon-rpm` each
`mv` it into the httpd drop-in directory as a build step.

`INSTALLHTTPDCONFDIR` installs it there directly.

**Unset, nothing changes at all.** The file goes to `INSTALLETCDIR` exactly as
before, through the same `setup-newfiles` call, so a plain source install is
byte-for-byte what it is today. Set, the file goes to the given directory
instead and is left out of the `etc` batch.

### Naming

The variable is named for the destination rather than for Apache: it is the
directory the web server reads drop-in configuration from, which is where any
packaging needs this file regardless of what is in front of Xymon.
(`INSTALLWEBDIR` was not available — it already means the page templates.)

### Why `export`

It is exported rather than appended to the `install-xymond` recursion line in
`build/Makefile.rules`. That line is long and shared by every sub-make
argument, so every feature that needs one more variable would otherwise
conflict textually with every other one. The `export` also sits next to the
`install-servermsg` message it affects rather than in the `INSTALL*DIR`
defaults block, so this PR occupies a region of `Makefile.rules` no other
open PR touches.

### Testing

On a server build in a clean AlmaLinux 10 container, both ways:

- unset — `xymon-apache.conf` appears in `$XYMONHOME/etc` as always, and
  `make install` prints the usual "A sample Apache configuration is in …"
- `INSTALLHTTPDCONFDIR=/etc/httpd/conf.d` — it appears there, is the
  *generated* file rather than the template (contains the substituted
  `Alias` lines), is no longer dropped into Xymon's `etc`, and the install
  message points at the real location

Merged against #408, #409, #410, #411, #413 and #414 in all 5040 possible
orders with no conflict in any of them.

### Status — parked as a draft

The workaround this removes is a single `mv` in `%install`, and Debian pays the
same single line in `debian/rules`. It is not free for Debian either way: it
renames the file to `xymon` and installs into `/etc/apache2/conf.d`, so
`INSTALLHTTPDCONFDIR` would remove its `mv` only if it accepted the upstream
filename.

One `mv` per packaging is not a strong enough problem to justify a build
variable. If the build should put the generated config where a web server
actually reads it, this is ready and tested; otherwise it can wait.

### DIFF
--- a/build/Makefile.rules
+++ b/build/Makefile.rules
@@ -180,12 +180,20 @@ clean:
 
 install: $(INSTALLTARGETS)
 
-# Where the generated web-server drop-in goes; xymond/Makefile installs
-# it there. Unset, it stays in Xymon's own etc directory as it always
-# has. Exported rather than added to the install-xymond recursion line:
-# that line is long and shared, so every feature needing one more
-# variable would otherwise collide textually with every other one.
+# Where the generated web-server drop-in goes, and under what name;
+# xymond/Makefile installs it there. Unset, it stays in Xymon's own etc
+# directory as it always has. The name defaults to the upstream filename,
+# so INSTALLHTTPDCONFDIR alone is unchanged; a packaging that renames it
+# (Debian installs it as "xymon", Terabithia as "<server>.conf") sets
+# INSTALLHTTPDCONFNAME and drops its own rename. Exported rather than added
+# to the install-xymond recursion line: that line is long and shared, so
+# every feature needing one more variable would otherwise collide textually
+# with every other one.
+ifndef INSTALLHTTPDCONFNAME
+INSTALLHTTPDCONFNAME = xymon-apache.conf
+endif
 export INSTALLHTTPDCONFDIR
+export INSTALLHTTPDCONFNAME
 
 install-servermsg:
 	@echo ""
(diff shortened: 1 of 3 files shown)


bonomani:
Force-pushed a rename: `INSTALLAPACHECONFDIR` is now `INSTALLHTTPDCONFDIR`.

The mechanism was never Apache-specific — it is just "the directory the web server reads drop-in configuration from" — so the variable is now named for the destination rather than for one server. The generated file keeps its `xymon-apache.conf` name, since that file really is Apache syntax. (`INSTALLWEBDIR` was not available; it already means the page templates.)

Same push also moves the variable to an `export` next to `install-servermsg` instead of appending it to the `install-xymond` recursion line, so this PR no longer touches any line another open PR touches.

bonomani:
Parking this as a draft — the same reasoning as #409.

The workaround it removes is a single `mv` in `%install`, and Debian pays the same single line in `debian/rules`. It is not free for Debian either way: it renames the file to `xymon` and installs into `/etc/apache2/conf.d`, so `INSTALLHTTPDCONFDIR` would remove its `mv` only if it accepted the upstream filename.

One `mv` per packaging is not a strong enough problem to justify a build variable in your tree. If you would rather the build put the generated config where a web server actually reads it, it is ready and tested — otherwise it can wait.

PR #414, opened 2026-08-20
build: install the shipped web content apart from what xymongen writes (TBT 1/2)

bonomani:
The `www` tree holds two different kinds of thing:

- **shipped content** — `gifs`, `help`, `menu`: files that never change on a running host
- **state** — `html`, `notes`, `rep`, `snap`, `wml`, and the pages `xymongen` writes into `www` itself

Both go to one directory, so a packaging that wants the shipped part in `/usr/share` — where the FHS puts architecture-independent package data — takes the tree apart after `make install`. Upstream's own `debian/rules:72-74` does exactly that, moving `gifs`, `help` and `menu` and symlinking them back; [xymon-rpm](https://github.com/xymon-monitoring/xymon-rpm) does the same.

### What this adds

`INSTALLSTATICWWWDIR`, in two halves — one build, one runtime — both no-ops when it is unset.

**Build.** **Unset — the default — nothing changes**: it is `INSTALLWWWDIR` and every subdirectory is real, exactly as today. Set, the three shipped directories are installed there and symlinks are left in the `www` tree, so every `/xymon/` URL keeps resolving through the docroot.

**Runtime.** `lib/links.c` locates the `help` directory from `XYMONSTATICWWWDIR` when it is set (`$XYMONSTATICWWWDIR/help`) instead of deriving it from `XYMONNOTESDIR` (strip the last component, append `/help`). Help ships with the static content, so its location should follow the static tree directly rather than depend on where `notes` lives or on the `www/help` symlink. Unset, the old derivation is used unchanged. Plain `getenv()` on purpose — `xgetenv()` errprints for an unset variable with no built-in default, and unset is the normal case. There is deliberately **no `XYMONHELPDIR` and no built-in default**: help is always `static/help`, so the one variable that already names the static tree is enough, with no second knob to keep in sync or override.

Files touched: `build/Makefile.rules`, `xymond/Makefile`, `docs/Makefile`, `configure.server` (build) and `lib/links.c` (runtime). `wwwfiles/` contains only `gifs` and `menu`, so the `setup-newfiles` call moves the destination whole and its md5 upgrade tracking is untouched.

### Relationship to devel `a70282c11`

The variable name is @jccleaver's, from that commit — but that commit is the larger *tmpfs-capable www* feature (the static move **plus** persisting `notes`/`periodic`, auto-creating `rep`/`snap`, and a settable `XYMONHELPDIR`), not just packaging. I did not backport it: it also rewrites `lib/links.c`, `web/report.c` and `web/snapshot.c`, which `main` has since hardened independently (bounded `snprintf`; `filestore` now *refuses* path components), so replaying the 2015 versions would undo that.

So this is written against current `main`, with two deliberate differences from that commit:

- the build default is a **no-op** rather than relocating everyone's layout;
- `help` is derived in the code from `XYMONSTATICWWWDIR` rather than read from a separate settable `XYMONHELPDIR` — since help is invariably `static/help`, the one static variable covers every case the extra knob would, with nothing to override or misconfigure.

The rest of `a70282c11` — the persistent/volatile (tmpfs) split, moving `notes`/`periodic` out, auto-creating `rep`/`snap` — is out of scope here and tracked in #442.

### Testing

The build half is verified on Debian and EL (with `dash` as `/bin/sh`):

| | |
| --- | --- |
| unset | `gifs`, `help`, `menu`, `html`, `notes`, `rep`, `snap` all real directories, content in place |
| `INSTALLSTATICWWWDIR=/usr/share/xymon` | those three are symlinks into it, generated directories still real, content reachable through the links |

The runtime half is a minimal `getenv` guard — `$XYMONSTATICWWWDIR/help` when set, the existing `XYMONNOTESDIR` derivation otherwise — so unset it is a no-op by construction; end-to-end verification of the set path is still pending. The directory guard uses `test A && test B` rather than `test A -a B`, which POSIX marks obsolescent and `dash` warns on.


SoundGoof:
## Findings

### Make standalone `install-docs` create the runtime help link

`build/Makefile.rules:386-387`

With `INSTALLSTATICWWWDIR` set, `make install-docs` installs the help files
under the static directory but never runs `install-dirs`, which creates the
`INSTALLWWWDIR/help` symlink.

Reproduced on Debian 13:

    static_help=present
    runtime_help=missing
    exit=0

Without the split setting, the same target creates `www/help` directly.
This regresses standalone `install-docs` and component-based packaging:
installation succeeds, but Xymon's runtime-derived help path is absent.

Make `install-docs` depend on `install-dirs`, matching the other component
installation targets.

## Validation

- `git diff --check` passes.
- Default, split, repeated, and old-layout upgrade installs tested.
- Split `install-dirs` passes on:
  - NetBSD 9.4
  - NetBSD 11.0
  - OpenBSD 7.9
  - FreeBSD 14.3

SoundGoof:
While this changes the `chown` and `chgrp` commands, they still pass
`$(INSTALLEXTDIR)/*` unconditionally.

A server built without `xymonnet` can have an empty `ext` directory. The shell
then passes the unmatched glob literally, causing both commands to fail.

Could this use an empty-directory-safe approach while these lines are being
updated?

bonomani:
Both findings fixed.

1. **`install-docs` now depends on `install-dirs`** (`ebcdbb16`) — it was the one runtime install target that missed it, so a standalone `make install-docs` with `INSTALLSTATICWWWDIR` set left the `www/help → static/help` symlink unmade and the runtime help path dead.
2. **`ext` is now `chown -R`/`chgrp -R`'d** (`63c59cc1`), not `ext/*` — an empty `ext` (server built without `xymonnet`) made the unmatched glob reach `chown` as a literal path and fail the install. Still under `ifndef PKGBUILD`.

On your follow-up: scoped to `ext` deliberately — the other `.../*` globs on those two lines (`$(XYMONHOME)/*`, `INSTALLBINDIR/*`, …) are always populated in a real install, so making the whole set empty-safe is a pre-existing tree-wide cleanup better done on its own than folded in here.

bonomani:
Scope note: this PR is the **static** split only (immutable `gifs`/`help`/`menu` → `INSTALLSTATICWWWDIR`), which is all the three splitting packagers (Debian, RPM, Terabithia) actually need. The broader `www`-by-lifecycle question — the persistent/volatile (tmpfs) split, moving `notes`/`periodic` out, `XYMONHELPDIR`, and whether `main` wants a tmpfs-capable `www` at all — is tracked separately in #442, so it stays out of this PR.

PR #415, opened 2026-08-21
build: generate the systemd unit from this build's paths, install it on request (devel 5a0b5b79d/1e216c5ab/f509be0b1)

bonomani:
**`ExecReload` needs #172** — the analysis is in the comment below; labelled `blocked` on that basis.

`main` has no systemd integration at all, so every packaging writes its own unit. `devel` has had one since February 2016 — `tools/xymonlaunch.service`, by @jccleaver — and it has simply never been merged across. This brings it over and makes it usable outside a distribution package.

The first three commits are his, unchanged apart from dropping the `Changes` and `debian/changelog` hunks, which do not apply here.

### The one change it needs

The unit hardcodes `/usr/bin/xymoncmd` and `/usr/sbin/xymonlaunch` — right for a package that puts them there, wrong for everyone else. A source install has them under `$XYMONHOME/bin`, as does any packaging choosing a different tree, so as it stands the file is a sample each consumer edits.

It becomes `tools/xymonlaunch.service.DIST` with `@XYMONHOME@` and `@XYMONUSER@` substituted at install time, the way `xymond/Makefile` already generates `xymon.sh` and `xymon-apache.conf`. `XYMONHOME` is defined for both build types — `$(XYMONTOPDIR)` for client-only, `$(XYMONTOPDIR)/server` otherwise — so **one file serves both** — which is what `devel`'s `Alias=xymon-client.service` was
reaching for, and why that alias is no longer needed (see below).

`INSTALLSYSTEMDDIR` says where the generated file goes. **Unset, nothing is installed** and `make install` is unchanged, which is what a source install on a host without systemd wants. Both configure scripts record it, like the other `INSTALL*DIR` options: `configure` is an `exec` dispatch, so only one runs and both need it, and without that it would work only when repeated on the `make install` command line.

`install-systemd` joins `INSTALLTARGETS` by `+=` rather than being written into the three assignments that set it. Those three lines are what every change to the install has to touch, so each new variable would otherwise collide with every other one; the target is a no-op unless the variable is set, so carrying it in every build type costs nothing.

### The environment file

`devel`'s unit names both families' paths, one per line:

```
EnvironmentFile=-/etc/default/xymonlaunch
EnvironmentFile=-/etc/sysconfig/xymonlaunch
```

That is the same problem this PR exists to fix, one line further down.
Hardcoding `/usr/bin` and `/usr/sbin` only suits a distribution installing
there; hardcoding `/etc/default` and `/etc/sysconfig` only suits two families.
On anything else — including the source install this change is meant to serve —
both paths are absent, both lines are no-ops, and `$XYMONLAUNCHOPTS` can never
be set. Nothing here installs `tools/xymonlaunch.default` either, so the log
file that sample configures was unreachable by construction.

`INSTALLENVDIR` replaces both and defaults to `$(XYMONHOME)/etc`, a directory
the build definitely creates. A distribution points it wherever it keeps these
files; upstream names neither.

| | generated |
| --- | --- |
| unset | `EnvironmentFile=-$(XYMONHOME)/etc/xymonlaunch` |
| `INSTALLENVDIR=/etc/sysconfig` | `EnvironmentFile=-/etc/sysconfig/xymonlaunch` |
| `INSTALLENVDIR=/etc/default` | `EnvironmentFile=-/etc/default/xymonlaunch` |

A directory rather than a full path, because both env files upstream ships —
`xymonlaunch.default` and `xymon-client.default` — live in the same directory
on every distribution and keep their basenames across families.

### Two more things the unit hardcoded

**`--no-env` cannot work on `main`.** `devel` implements it in `lib/stdopt.c`,
which `main` does not have; `main`'s `xymoncmd` parses options inline and knows
only `--debug`, `--env=`, `--area=` and `--version`. Anything else falls through
to the branch treating the argument as *the command to run*, so the unit would
have exec'd a program named `--no-env`. Removed — and letting `xymoncmd` load
the environment is what is wanted here: it probes `xymonserver.cfg` before
`xymonclient.cfg`, so the role resolves at runtime, the way `xymonlaunch`
already picks `tasks.cfg` before `clientlaunch.cfg`.

**The aliases claim names other packagings own.** `Alias=` is a symlink systemd
creates at enable time under `/etc/systemd/system`, so `xymon.service` and
`xymon-client.service` collide with any distribution shipping real units of
those names — Debian ships both — or with an admin's own. Any given install is
one role, so one of the two names always described something that was not
running. Removed; a distribution wanting a compat name adds it in its own
`[Install]` section or a drop-in.

That leaves a unit where every line is either a systemd generic or substituted
from the build:

```
EnvironmentFile=-@INSTALLENVDIR@/xymonlaunch
User=@XYMONUSER@
ExecStart=@XYMONHOME@/bin/xymoncmd @XYMONHOME@/bin/xymonlaunch --no-daemon $XYMONLAUNCHOPTS
```

Verified on Debian 13 for both build types, installing for real and running the
generated `ExecStart`:

| build | ExecStart tree | INSTALLENVDIR | placeholders | startable |
| --- | --- | --- | --- | --- |
| `--server` | `/opt/xymon/server/bin` | `/etc/sysconfig` | 0 | yes |
| `--client` | `/opt/xymon/bin` | `/etc/default` | 0 | yes |

### Testing

On AlmaLinux 10, with the `xymon` account present:

| | result |
| --- | --- |
| server build, unset | no unit installed anywhere; `make install` unchanged |
| server build, set for `./configure` only | installed, `ExecStart=/opt/b/server/bin/xymoncmd … xymonlaunch`, zero `@` left, those binaries exist |
| **client-only** build, same variable | installed, `ExecStart=/opt/c/bin/xymoncmd …` — same file, correct for the client tree |

`systemd-analyze verify` is clean apart from `Documentation=` man-page lookups, which fail only because that container has no man pages.

Merged cleanly with #410, #411, #414 and #416 in all 120 possible orders when that was
tested; #416 has since landed on `main`, so the live set is #410, #411 and #414.

### Documentation

`tools/README.systemd` describes the install interface beside the files it
describes: `INSTALLSYSTEMDDIR` and `INSTALLENVDIR`, that `install-systemd`
installs only the generated unit, that the `.default`, preset and tmpfiles
files are samples, and that enabling the unit is local policy.

`xymon(7)`'s installation section gains one sentence naming `INSTALLSYSTEMDDIR`
and pointing at that README — because the README ships with the source but is
not installed, so on a packaged system nothing would say the capability exists.
Deliberately one sentence: no other `INSTALL*DIR` variable appears in a manual
page, and this does not make these two an exception beyond saying they are
there.

### Deliberately left alone

`tools/xymonlaunch.service.preset` stays as @jccleaver wrote it and is not installed. Whether a fresh install enables itself is distribution policy — one packaging I know of ships it for the server role only, so a client cannot start up reporting to an unconfigured address.

### Worth flagging separately

`tools/xymon-tmpfiles.conf` here creates `/run/xymon`, but nothing this build
installs writes to it. On `main` that directory — spelled `/var/run/xymon`, the same
place on a systemd host — is created and used only by the Debian packaging:
`debian/init-common.sh` creates it, and both init scripts, both `postinst`,
`xymon-client.postrm`, `debian/rules` and `xymon-client.default.dist` reference it. The
tree's own pidfiles and control sockets move into `XYMONRUNDIR` with #219. Until
that lands, this file creates a directory nothing uses: inert rather than wrong,
but worth knowing it does nothing on its own.

The two Red Hat init scripts in devel's packaging call `xymoncmd --no-env`
(devel `bec94d76b`), and the flag itself comes from `lib/stdopt.c` (devel
`ec161e994`), which `main` does not have. Those two calls break on `main` the
same way this one did. Both are tracked on #106, where `bec94d76b` carries
`needs ec161e994`.


SoundGoof:
## Findings

### 1. `git diff --check` fails

`tools/xymonlaunch.default` contains trailing whitespace on lines 2, 5, 10,
13, 14, 16, and 17, plus a blank line at EOF.

`tools/xymonlaunch.service.DIST` also has a blank line at EOF.

### 2. Document the new installation interface

This adds user-facing systemd installation controls, but no README or manpage
documents them.

A source installer needs to know that:

- `INSTALLSYSTEMDDIR` must be set before configuration or installation.
- Leaving it unset means no unit is installed.
- `INSTALLENVDIR` controls where the unit reads `xymonlaunch`.
- The `.default`, tmpfiles, and preset files are samples and are not installed.
- Enabling the unit remains a local or distribution policy decision.

The PR description explains parts of this, but that information will not ship
with the source. Add durable installation documentation in a README or an
appropriate manual page.

### 3. Correct the documented output destination

`tools/xymonlaunch.default` says output is sent to syslog when `--log` is not
set.

`xymonlaunch(8)` correctly says the default destination is stdout. Under
systemd, stdout is handled according to `StandardOutput`; its default is the
journal, not syslog. Rocky Linux 10 reports:

```text
DefaultStandardOutput=journal
DefaultStandardError=inherit
```

Describe this as journal output by default, or as output captured according to
the service manager's standard-output configuration.

## Validation

All five GitHub checks pass, including the manual-page source check.

The existing manpages correctly document `--no-daemon`, `--log`, and
`tasks.cfg(5)`. All three pages named by the unit's `Documentation=` directive
exist and are installed.

- Static regression suite: 39 passed, 48 skipped, 0 failed.
- Unit generation passed on Debian 13, Rocky Linux 9, and Rocky Linux 10.
- Rocky Linux 10 server and client installations generated correct paths.
- Rocky Linux 10 live start, reload, and stop testing passed.

The empty `ext` failure encountered during focused `install-xymond` testing is
pre-existing and is not a finding against this PR.

bonomani:
All three fixed, head `c0815dc47`.

**1. Whitespace** (`2ad9a49dc`). Removed the trailing whitespace on the comment lines of `tools/xymonlaunch.default` and the blank line at EOF in it and `tools/xymonlaunch.service.DIST`. `git diff --check` is clean.

**3. Output destination** (`2ad9a49dc`). You're right — an unset log does not go to syslog under systemd. Reworded: xymonlaunch writes to stdout, which systemd captures per the unit's `StandardOutput` (the journal by default); sysvinit still loses it.

**2. Documentation** (`c0815dc47`). Added `tools/README.systemd`, so this ships with the source rather than only in the PR. It covers what you listed: `INSTALLSYSTEMDDIR` (set before/at configure or on the make line; unset installs no unit), `INSTALLENVDIR` (where the unit reads its env file), that install-systemd installs only the generated unit while `xymonlaunch.default`, the preset and the tmpfiles fragment are samples that are not installed, and that enabling the unit is a local or distribution policy decision.

Put it in `tools/` beside the files it describes; happy to move it to a manpage or the top-level docs instead if you'd prefer.

bonomani:
A note on how this interacts with the script front-ends, from reviewing #219 alongside it.

### The unit makes the pidfiles vestigial, where it applies

```
Type=simple
ExecStart=@XYMONHOME@/bin/xymoncmd @XYMONHOME@/bin/xymonlaunch --no-daemon $XYMONLAUNCHOPTS
```

`Type=simple` and no `PIDFile=` — systemd tracks the process itself. So on a systemd host nothing reads `xymonlaunch.pid` (server, written by `xymon.sh`) or `clientlaunch.$MACHINEDOTS.pid` (client, written by `runclient.sh`). Those files are still created; nothing consults them.

That matters for #219. It adds `XYMONCLIENTRUNDIR`, and `clientlaunch.cfg` moves `msgcache.pid` onto it — but `client/runclient.sh` keeps its own pidfile in `logs/` (ten references, lines 83–126), because the script never sources `xymonclient.cfg` and so cannot see the variable. Under this unit that inconsistency is invisible: neither pidfile is read. It only shows for installations still starting Xymon through `runclient.sh` or `xymon.sh`.

Worth stating in `README.systemd`, since an admin migrating from the scripts will otherwise wonder which pidfile is authoritative — the answer being neither.

### Two places this leans on #172

- **`ExecReload=/bin/kill -HUP $MAINPID`** — on `main` a HUP to xymonlaunch acts only on itself: the handler sets `nextcfgload = 0` and `dologswitch = 1`, reloading its own `tasks.cfg` and reopening its own log. Every `kill()` in `common/xymonlaunch.c` sends `SIGTERM` or `SIGKILL` (lines 534, 556, 813, 918, 932) — there is no `kill(..., SIGHUP)`, so the tasks are never signalled. `systemctl reload xymon` would therefore report success while every task kept its old configuration and log file, which is worse than having no reload verb: it looks like it worked. #172 adds the propagation as a per-task opt-in — a `SENDHUP` keyword in `tasks.cfg` and `kill(twalk->pid, SIGHUP)` for the tasks that declare it, so even after it lands this unit's reload reaches only those tasks. **Two ways to resolve: land after #172, or drop `ExecReload` here and add it with #172.** The PR is labelled `blocked` on that basis.
- **`KillMode=process` with `SendSIGKILL=no`** — systemd signals only `xymonlaunch` and trusts it to take its children down. That relies on the same signal path #172 is changing, so it is worth confirming shutdown still drains cleanly once that lands.

Neither is a defect in this PR; both are reasons the two want sequencing rather than independent review.

bonomani:
Closing the documentation question from your finding 2, which I left open in August ("happy to move it to a manpage or the top-level docs instead if you'd prefer").

`tools/README.systemd` stays where it is — it is what a packager reads, beside the files it describes — and `xymon(7)`'s **PREREQUISITES AND INSTALLATION** section now gains one sentence naming `INSTALLSYSTEMDDIR` and pointing at it (head `5821adbdf`).

The reason for splitting it that way rather than moving the content into man: the README ships with the source but is not installed, so on a packaged system nothing would otherwise say the capability exists — while no other `INSTALL*DIR` variable is documented in any manual page, so copying the whole interface there would make these two uniquely privileged. One sentence says it exists; the README says how.

`mandoc -Tlint` is clean on the added lines (the four STYLE warnings in that file are pre-existing), and `docs/manpages/man7/xymon.7.html` is regenerated with `build/makehtml.sh`.

PR #416, opened 2026-08-21
Report an unknown verb as a failure, not a success

bonomani:
Both `client/runclient.sh` and `xymond/xymon.sh.DIST` end their case statement with

```sh
  *)
	echo "Usage: $0 start|stop|restart|status"
	break;
```

`break` is a loop construct and neither case is inside a loop, so it does nothing. The script falls through to the `exit 0` at the bottom and reports success:

```
$ ./runclient.sh statsu; echo $?
Usage: ./runclient.sh start|stop|restart|status
0
```

A typo in a service manager, an init script or a cron entry therefore looks like the client started. Nothing reports an error, and nothing is running.

`exit 2` replaces it, and the usage line goes to stderr since it is now an error message rather than output.

Two files, two lines. `xymon.sh.DIST` has the identical defect, so fixing only the client half would leave the server script reporting success on a typo.

### Why 2 rather than 1

These scripts already speak LSB where it counts: `status` returns **3** for "not running" and **1** for a stale pidfile, both LSB's meanings. LSB reserves **2** for invalid or excess arguments, which is exactly this case, so 2 leaves the file consistent with itself rather than following the standard for one verb and not another. It is also the usual convention outside init scripts — `argparse` and most `getopt`-based tools exit 2.

One character either way if you prefer 1; nothing else depends on it.

### Testing

In both `dash` and `bash`, against `main` and this branch:

| | unknown verb | `status` on a stopped client |
| --- | --- | --- |
| `main` | **exit 0** | exit 3 |
| this branch | exit 2 | exit 3 |

The valid verbs are untouched, so anything parsing those codes is unaffected. Usage moves to stderr, leaving stdout empty for the failure case.

### DIFF
--- a/client/runclient.sh
+++ b/client/runclient.sh
@@ -133,8 +133,8 @@ case "$CMD" in
 	;;
 
   *)
-	echo "Usage: $0 start|stop|restart|status"
-	break;
+	echo "Usage: $0 start|stop|restart|status" >&2
+	exit 2
 
 esac
 
--- a/xymond/xymon.sh.DIST
+++ b/xymond/xymon.sh.DIST
@@ -97,8 +97,8 @@ case "$1" in
 	;;
 
    *)
-   	echo "Usage: $0 start|stop|restart|reload|status|rotate"
-	break;
+	echo "Usage: $0 start|stop|restart|reload|status|rotate" >&2
+	exit 2
 esac
 
 exit 0


bonomani:
Changed to `exit 2` and pushed, rather than leaving the question open.

The deciding argument was internal consistency rather than convention: these scripts already return LSB's codes for `status` — 3 for "not running", 1 for a stale pidfile — so using 1 for a bad verb would have left the file following the standard for one verb and not the other. LSB reserves 2 for invalid or excess arguments.

Retested in dash and bash: unknown verb exits 2 in both, `status` on a stopped client still exits 3, and usage goes to stderr with stdout empty.

SoundGoof:
## Finding

### Low: Fix mixed whitespace in `xymon.sh.DIST`

`git diff --check` fails:

```text
xymond/xymon.sh.DIST:100: space before tab in indent.
```

Replace the spaces-plus-tab indentation on the added usage line with the
file's normal tab indentation.

## Validation

- PR merges cleanly with current `main`.
- Debian 13, dash and bash:
  - Unknown verb exits 2.
  - Usage is written only to stderr.
  - Stopped `status` remains exit 3.
  - Shell syntax checks pass.

bonomani:
Fixed and pushed — the added line now uses a plain tab, matching the
`exit 2` below it, and `git diff --check` is clean on the commit.

One note for the record: the mixed indentation was not introduced here.
`main`'s line is already three spaces followed by a tab, and this change
only appended ` >&2` to it, inheriting the indentation. `git diff --check`
flags it regardless, since it checks added lines rather than what the
line looked like before — so the finding stands, and normalising it is
the right fix. Worth knowing only because the same warning will appear on
any future patch that touches a line in that block without rewriting its
indentation. None of the five CI checks run `--check`, which is why it
took a human to spot.

Also rebased onto current `main` (the branch was ten commits behind) and
kept it as a single commit.

Re-verified after the change, in dash and bash, for both scripts: unknown
verb exits 2, usage goes to stderr with stdout empty, and `status` on a
stopped client still exits 3.

SoundGoof (approved):
## Findings

No blocking findings.

The previous mixed-indentation finding is resolved. The current commit uses
the file's normal indentation, and `git diff --check` passes.

## Validation

- Unknown and missing verbs return 2, write nothing to stdout, and write usage
  to stderr under Debian 13 `dash` and `bash`.
- The same behavior passes under native `/bin/sh` on NetBSD 9.4, NetBSD 11.0,
- Both scripts pass native shell syntax checks on all tested BSD systems.
- Stopped `status` remains exit 3 for both scripts.
- ShellCheck reports only existing diagnostics outside the changed lines.
- Current `main` has three newer commits, but none overlap these files.

PR #418, opened 2026-08-21
Use require_bin mechanism in tests/server/history-tail-resync.sh

RolandRosenfeld:
Using require_bin instead of hardcoding the binary allows to redirect XYMOND_HISTORY to a already installed binary, as used in Debian autopkgtests.

### DIFF
--- a/tests/server/history-tail-resync.sh
+++ b/tests/server/history-tail-resync.sh
@@ -23,7 +23,6 @@ set -euo pipefail
 # shellcheck source=tests/lib/assert.sh
 . "$(dirname "$0")/../lib/assert.sh"
 
-ROOT=$(find_root)
 require_bin XYMOND_HISTORY xymond/xymond_history
 
 WORK=$(mktempdir)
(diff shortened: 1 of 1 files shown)


bonomani (dismissed):
Approved. The substitution reproduces the old skip-if-unbuilt behaviour and adds what the Debian autopkgtest needs: an explicit `$XYMOND_HISTORY` is used verbatim, and a dangling override fails instead of skipping green — a broken package layout is exactly what that test run is meant to catch. Matches the ~20 other `require_bin` call sites, and the in-tree default now resolves against `find_root` rather than cwd.

One non-blocking nit: `ROOT=$(find_root)` (line 26) is dead now — `BIN` was its only consumer, and `require_bin` does that resolution itself (`tests/lib/assert.sh:405`). Shellcheck flags it SC2034; there is no shellcheck gate in CI, so nothing fails. The other converted tests dropped the line (`analysis-file-ifexist.sh`, `xymongrep-filter.sh`, `hostdata-save-next.sh`, `mailack-cookie.sh`, ...), while the ones that kept it still use it for a fixture path. Happy to remove it in a follow-up if you would rather not respin.

RolandRosenfeld:
You are right, the ROOT variable is now superfluous.
Is it the right workflow to add another one-line-commit for this or should I use git commit --amend and force push this?

bonomani (approved):
Re-approving after 15d49d1fd — that commit deletes exactly the one dead line and nothing else. Thanks for folding it in.

PR #419, opened 2026-08-21
Add require_cfg function to tests/lib/assert.sh

RolandRosenfeld:
This allows to override local config file in tests via environment variable to use already installed config file.

I try to integrate the testsuite into Debian autopkgtest mechanism and this change allows to run the testsuite with
```bash
env CC=/nonexistent-xymon-autopkgtest-cc \
    MAILACK=/usr/lib/xymon/server/bin/xymon-mailack \
    SVCSTATUS_CGI=/usr/lib/xymon/server/bin/svcstatus.cgi \
    XYMON=/usr/lib/xymon/client/bin/xymon \
    XYMONCLIENT=/usr/lib/xymon/client/bin/xymon \
    XYMONCLIENT_LINUX=/usr/lib/xymon/client/bin/xymonclient-linux.sh \
    XYMOND=/usr/lib/xymon/server/bin/xymond \
    XYMOND_ALERT=/usr/lib/xymon/server/bin/xymond_alert \
    XYMOND_CHANNEL=/usr/lib/xymon/server/bin/xymond_channel \
    XYMOND_CLIENT=/usr/lib/xymon/client/bin/xymond_client \
    XYMOND_HISTORY=/usr/lib/xymon/server/bin/xymond_history \
    XYMOND_HOSTDATA=/usr/lib/xymon/server/bin/xymond_hostdata \
    XYMOND_RRD=/usr/lib/xymon/server/bin/xymond_rrd \
    XYMONGREP=/usr/lib/xymon/client/bin/xymongrep \
    XYMONLAUNCH=/usr/lib/xymon/server/bin/xymonlaunch \
    XYMONSERVER_CFG=/etc/xymon/xymonserver.cfg \
    ./tests/testsuite
```
This change enables the tests
- server/xymond-checkpoint-roundtrip.sh
- server/xymond-holdtime-gap.sh
- server/xymondboard-color-history.sh
for this mechanism.

require_cfg() in lib/assert.sh is implemented as a clone of require_bin() but testing for file existence instead of executability.

Assisted-By: Github Copilot using GPT-5.5.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -135,10 +135,11 @@ maintenance.
   `git worktree`, `git stash`, or any other git invocation.
 - **Path discovery via env var with default.** When a test needs a
   built binary or an installed artefact, read it from an env var and
-  default to the in-tree path -- `require_bin` (lib/assert.sh) does this
-  for binaries:
+  default to the in-tree path -- `require_bin` and `require_cfg`
+  (lib/assert.sh) do this for binaries and config files:
   ```bash
   require_bin XYMONGREP common/xymongrep          # binaries
+  require_cfg XYMONSERVER_CFG xymond/etcfiles/xymonserver.cfg  # config files
   SCRIPT="${XYMONCLIENT_LINUX:-$ROOT/client/xymonclient-linux.sh}"  # scripts
   ```
   This keeps tests usable in CMake out-of-source builds (the build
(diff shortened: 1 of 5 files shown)


bonomani (approved):
Approved. `require_cfg` mirrors the `require_bin` contract correctly, and the skip-if-absent default is right since `xymond/etcfiles/xymonserver.cfg` is generated from the .DIST file rather than tracked. Checked that exporting `XYMONSERVER_CFG` is inert — nothing in the Xymon source reads that name, and the tests pass the rewritten copy to xymond via `--env`. Verified with your override path: all three tests pass against installed binaries plus an explicit `XYMONSERVER_CFG`.

Also a small behaviour change worth noting, and I think a good one: the old `sed ... || skip` swallowed any sed failure as a skip, so a present-but-broken config greened out. Now only a missing file skips.

Non-blocking, same nit as #418: `ROOT=$(find_root)` is left with no consumers in all three converted tests (shellcheck SC2034 on each; they pass with the line deleted). Note it is not always removable — `tests/lib/svcstatus-cgi.sh` needs a caller-set $ROOT — but these three source only assert.sh, which resolves the root itself.

PR #420, opened 2026-08-21
xymonclient-freebsd: emit an empty inode report when df has no rows (#398)

### LINKED ISSUE #398: inode test broken if no fs with inodes exists
on a FreeBSD server with only ZFS (no filesystems with inodes) and client deployed from commit 9c73bfc we have an issue with the inode test

```
Expected strings (%iused and Mounted) not found in df output
```

the raw client data for df/inode looks like this:

```
[df]
Filesystem            Size    Used   Avail Capacity  Mounted on
zroot/ROOT/default     43G     29G     14G    67%    /
zroot/usr/ports        14G     98k     14G     0%    /usr/ports
zroot/home             14G     98k     14G     0%    /home
zroot/var/audit        14G     98k     14G     0%    /var/audit
zroot/usr/src          14G     98k     14G     0%    /usr/src
zroot/home/feld        21G    6.5G     14G    32%    /home/feld
zroot/var/tmp          14G    713k     14G     0%    /var/tmp
zroot/var/crash        14G     98k     14G     0%    /var/crash
zroot/var/mail         14G    131k     14G     0%    /var/mail
zroot/tmp              14G    197k     14G     0%    /tmp
zroot                  14G     98k     14G     0%    /zroot
zroot/var/log          14G    7.7M     14G     0%    /var/log
[inode]
                         itotal              
```

bonomani:
Fixes #398. Extracted from #390 so it can land on its own — the rest of that PR is the #344 unreachable-mount work and unrelated to this.

On a ZFS-only host `df -i` excludes every filesystem (zfs has no inode limit) and FreeBSD df then exits 0 with no output. `emit_df`'s guard only skipped on a *non-zero* exit, so the clean-but-empty case fell through and the inode awk's `NR<2` header rule turned the blank line into a stray `itotal` row. `unix_inode_report()` could not find the `%iused`/`Mounted` columns in it and flagged the host red — defeating its own all-ZFS green path, which fires only on a truly empty report.

`emit_df` now returns 1 whenever df produced no output: a non-zero exit still emits the failure marker (server yellow), a clean exit emits nothing, so the empty-report path keeps the inode column green.

| check | result |
|---|---|
| tests/client/fs-filter-freebsd.sh | PASS |
| same test, client change reverted | FAIL — `expected '', got '   itotal   '`, the issue's exact row |
| @feld on a real ZFS-only host (4191adf29) | `&green No filesystems reporting inode data` |

The stub's `DF_INODE_EMPTY_OK` models FreeBSD's "empty output, exit 0" for the inode call; @feld's confirmation above is what verifies that against the real df.

### DIFF
--- a/client/xymonclient-freebsd.sh
+++ b/client/xymonclient-freebsd.sh
@@ -330,9 +330,16 @@ run_df() {
 emit_df() {
 	_kind="$1"; _label="$2"; shift 2
 	_out=`run_df "$@"`; _rc=$?
-	if [ -z "$_out" ] && [ "$_rc" -ne 0 ]; then
-		echo "xymonclient-freebsd: df $_kind collection failed (status $_rc) with no output; reporting data as unavailable" >&2
-		echo "$_label report collection failed: df exited $_rc with no output"
+	if [ -z "$_out" ]; then
+		# No rows at all. A non-zero exit is a collection failure - surface a
+		# marker so the server goes yellow. A clean exit means nothing matched
+		# (e.g. an inode report whose filesystems are all zfs, which are
+		# excluded): emit nothing, so the server's empty-report path stays green
+		# rather than rejecting a stray header-only row as "columns not found".
+		if [ "$_rc" -ne 0 ]; then
+			echo "xymonclient-freebsd: df $_kind collection failed (status $_rc) with no output; reporting data as unavailable" >&2
+			echo "$_label report collection failed: df exited $_rc with no output"
+		fi
 		return 1
 	fi
 	return 0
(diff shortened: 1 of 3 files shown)


bonomani:
@feld assigning you since you reported #398 and already confirmed the fix on your ZFS-only box (`&green No filesystems reporting inode data`, commit 4191adf29). Nothing further needed from you — this is just the same commit split out of #390 so it can merge on its own.

bonomani:
Superseded: this commit went in with #390 as 86874508b, so the fix is on main and this PR is now a duplicate of merged code.

@feld nothing changes for you — the fix you confirmed on your ZFS-only host (`&green No filesystems reporting inode data`) is what landed.

PR #421, opened 2026-08-21
tests: draw the xymond test port from below the ephemeral range

bonomani:
`build (server)` fails at random with a test that never touched the code under
review. On #411 it looked like this:

```
tests/server/xymond-holdtime-gap.sh
  Setting up network listener on 127.0.0.1:34136:34136
  Cannot bind to listen socket (Address already in use)
  FAIL: xymond exited during startup
```

85 tests passed; that one could not get its port. Re-running the same commit
passed, which is the signature of a race rather than a defect.

## Why

Three tests start a real xymond and pick its port the same way:

```sh
p=$(( 20000 + (RANDOM % 20000) ))
"$XYMONCLIENT" "127.0.0.1:$p" "ping" >/dev/null 2>&1 || { printf '%s' "$p"; return 0; }
```

That draws from 20000-39999, which **overlaps the kernel's ephemeral range** --
`/proc/sys/net/ipv4/ip_local_port_range` is `32768 60999` on the runners, and
34136 sits inside it. So the kernel is free to hand the same port to an
unrelated outbound connection, and it does when the runner is busy.

The probe cannot see that happen. It only detects a listener that answers a
Xymon ping; a socket held by anything else reads as free. And there is a window
between the probe and xymond's bind in which the port can be taken regardless.

## The fix

Draw from just below the ephemeral range instead, reading where that range
starts rather than assuming:

```sh
hi=$(cut -f1 /proc/sys/net/ipv4/ip_local_port_range 2>/dev/null)
case "$hi" in ''|*[!0-9]*) hi=32768 ;; esac
[ "$hi" -gt 2048 ] || hi=32768
lo=$(( hi > 9216 ? hi - 8192 : 1024 ))
p=$(( lo + (RANDOM % (hi - lo)) ))
```

The kernel does not assign below that boundary, so the collision this fixes
cannot recur. `cut` failing or returning junk falls back to 32768, the common
default, and a boundary at or below the privileged ports is treated the same
way rather than leaving an empty window to draw from.

That leaves one case: a port below the range that something else is already
listening on, which the ping probe still cannot detect. `start_xymond` now
retries up to five times when -- and only when -- xymond died with `Cannot bind
to listen socket`. Every other startup failure fails immediately as before, and
both existing failure messages are preserved, so a xymond that cannot bind
anywhere still fails instead of looping.

The suite runs tests one at a time, so no test competes with another for a
port; the competitor is always something outside the suite.

## Verified

Built `--server` in a Debian 13 container with CI's own configure environment,
then:

- all three affected tests pass unchanged
- with a listener deliberately occupying the first port handed out,
  the patched test still passes -- the retry fires and recovers
- the same control against today's `main` fails with the exact CI error,
  confirming the control reproduces the bug rather than something else

## Scope

Test-harness only -- no shipped code changes. Independent of my other open PRs;
it touches no file any of them touch.

### DIFF
--- a/tests/lib/port-blocker.c
+++ b/tests/lib/port-blocker.c
@@ -1,0 +1,42 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later                                  */
+/*
+ * tests/lib/port-blocker.c
+ *
+ * Hold a 127.0.0.1 port so that nothing else can bind it, and print the port
+ * on stdout so the test can name it.
+ *
+ * Deliberately does NOT listen(): a connect() to it is refused outright.
+ * A listener would be worse than useless here -- the readiness probe in
+ * tests/lib/xymond-daemon.sh cannot tell one Xymon-speaking listener from
+ * another, so anything that accepts connections reads as "the daemon is up".
+ * Refusing the connection is what makes a test of the bind retry deterministic.
+ */
+
+#include <stdio.h>
+#include <unistd.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+
+int main(void)
+{
+	int sock;
+	struct sockaddr_in addr;
+	socklen_t alen = sizeof(addr);
+
+	sock = socket(AF_INET, SOCK_STREAM, 0);
+	if (sock < 0) { perror("socket"); return 1; }
+
+	memset(&addr, 0, sizeof(addr));
+	addr.sin_family = AF_INET;
+	addr.sin_addr.s_addr = inet_addr("127.0.0.1");
+	addr.sin_port = htons(0);
+	if (bind(sock, (struct sockaddr *)&addr, sizeof(addr)) < 0) { perror("bind"); return 1; }
+	if (getsockname(sock, (struct sockaddr *)&addr, &alen) < 0) { perror("getsockname"); return 1; }
+
+	printf("%d\n", ntohs(addr.sin_port));
+	fflush(stdout);
+
+	for (;;) pause();
+}
--- a/tests/lib/xymond-daemon.sh
+++ b/tests/lib/xymond-daemon.sh
@@ -1,0 +1,116 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/lib/xymond-daemon.sh
+#
+# Starting a real xymond on a loopback port, for the tests that need one.
+#
+# Three tests carried a byte-identical copy of this, and the logic is subtle
+# enough that three copies is how one of them drifts: the port has to be
+# chosen without the kernel handing it to someone else first, and that is not
+# something any probe can guarantee.
+#
+# Source after tests/lib/assert.sh. The caller provides $XYMOND, $XYMONCLIENT
+# (require_bin), $work (a scratch directory holding xymond.log), and defines
+# xymond_launch(); start_xymond() drives it and exports PORT and XYMOND_PID.
+
+# ephemeral_port_floor -- where the kernel starts handing out ephemeral ports,
+# or nothing when this kernel does not say. Purely an optimization input: see
+# free_port().
+#
+# Asked by knob rather than by OS: the names do not overlap between the BSDs
+# (FreeBSD and macOS spell it portrange.first, OpenBSD portfirst, NetBSD
+# anonportmin), so whichever one answers identifies the kernel by itself and
+# no uname test is needed. /sbin/sysctl as well as sysctl, since NetBSD does
+# not put it on a non-root PATH.
+ephemeral_port_floor() {
+	local v knob s
+
+	v=$(cut -f1 /proc/sys/net/ipv4/ip_local_port_range 2>/dev/null || true)
+	case "$v" in ''|*[!0-9]*) ;; *) printf '%s' "$v"; return 0 ;; esac
+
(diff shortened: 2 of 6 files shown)


SoundGoof:
## Findings

### The fallback range still overlaps ephemeral ports on BSD

The Linux-specific `/proc` lookup falls back to `32768` on every other OS:

https://github.com/xymon-monitoring/xymon/blob/88eb2048abb01b3b53bca6687c982c7ffd01dd59/tests/server/xymond-holdtime-gap.sh#L77-L84

This produces `24576-32767`, but that range is not below the ephemeral range
on supported BSD systems:

- OpenBSD 7.9 uses `1024-49151`.
- FreeBSD 14.3 uses `10000-65535`.
- NetBSD 9.4 and 11.0 use `49152-65535`.

The native lower-bound lookup could follow the existing Linux lookup:

```sh
case "$(uname -s)" in
    Linux)
        hi=$(cut -f1 /proc/sys/net/ipv4/ip_local_port_range 2>/dev/null)
        ;;
    FreeBSD)
        hi=$(sysctl -n net.inet.ip.portrange.first 2>/dev/null)
        ;;
    OpenBSD)
        hi=$(sysctl -n net.inet.ip.portfirst 2>/dev/null)
        ;;
    NetBSD)
        hi=$(/sbin/sysctl -n net.inet.ip.anonportmin 2>/dev/null)
        ;;
    *)
        hi=
        ;;
esac
```

These lookups return `10000` on FreeBSD, `1024` on OpenBSD, and `49152` on
NetBSD. The OpenBSD result also reveals that there is no unprivileged port
window below its default range. It therefore cannot use the same
“select below the boundary” strategy; bind-error retries or another
allocation mechanism must provide correctness there.

I verified the behavior on native VMs by opening a loopback listener, making
40 outbound loopback connections, recording the source port assigned to each
connection by the kernel, and counting assignments inside `24576-32767`.

OpenBSD assigned 6 of 40 connections from that window. FreeBSD assigned 4 of
40. The exact count is random, but any hit demonstrates that the kernel can
allocate a port from the proposed window to an unrelated outbound connection.

A deterministic BSD regression test can read the corresponding native
boundary, call the selection function repeatedly, and assert that every
candidate lies outside the reported range. Where no suitable unprivileged
window exists, the test should instead verify that bind-error retries are the
portable source of correctness.

### Reset the retry budget for each daemon startup

`START_ATTEMPTS` is initialized once for the entire script and incremented
recursively:

https://github.com/xymon-monitoring/xymon/blob/88eb2048abb01b3b53bca6687c982c7ffd01dd59/tests/server/xymond-holdtime-gap.sh#L112-L139

This test calls `start_xymond` thirteen times. Once five bind collisions have
occurred cumulatively, every subsequent startup loses its retry protection.
The checkpoint test has the same problem across five starts.

A deterministic BSD test can make `free_port` return a known sequence, hold
the first candidate with a non-Xymon listener, and leave the second candidate
free. It should confirm that xymond fails to bind the first port, retries, and
starts on the second. Repeating this scenario after a successful stop and
restart verifies that each independent startup receives a fresh retry budget.

Keep the attempt count local to one top-level startup, preferably with an
iterative retry loop.

## Validation

- Read the complete current conversation. There are no later comments,
  reviews, inline threads, or replies.
- All three affected live-xymond tests pass locally.
- Full local suite: 84 passed, 3 skipped, 0 failed.
- Verified the lookup values on Debian 13, NetBSD 9.4, NetBSD 11.0,
- Verified actual kernel source-port allocation on native OpenBSD and
  FreeBSD VMs using the loopback procedure described above.

bonomani:
Thanks @SoundGoof — both findings are right, and I took them as a reason to invert the design rather than patch it. Pushed as 8b1c6637b.

**The window cannot carry correctness, so the retry does.** Your BSD numbers are the proof: assuming 32768 puts every draw inside the ephemeral range on FreeBSD and OpenBSD. One detail worth adding, because it means the sysctl lookup alone is not enough — the guard I wrote to reject a bad reading (`[ "$hi" -gt 2048 ]`) would have rejected OpenBSD's *correct* 1024 for being low and substituted 32768, i.e. exactly the wrong window. Since OpenBSD has no unprivileged room below its range at all, the honest structure is: retry is the mechanism, the window only makes collisions rare.

So the floor is now looked up by knob rather than by OS — the names do not overlap between the BSDs, so whichever answers identifies the kernel and no `uname` test is needed (the suite's own portability rule objects to `uname` in a helper, with reason). When nothing answers, or nothing leaves room below the range, no window is invented: the historical span is used and the retry covers it.

**Retry budget is per startup**, as you asked — local counter, iterative loop.

**One more thing your review led me to.** Writing the test you suggested, the first blocker I used was a second xymond, and `start_xymond` reported success on the blocked port: the readiness probe cannot tell this xymond from any other Xymon-speaking listener holding it, so a dead child plus a stranger's socket read as "up". Fixed — an answered ping now only counts while our own child is alive — and `tests/lib/port-blocker.c` holds the port *without* listening, so the test is deterministic.

**Tests.** Took your first proposal: `tests/server/xymond-port-retry.sh` drives two startups needing three retries each — six, past a budget of five — so it passes with the per-startup budget and fails against the shared counter. I verified both directions. I skipped the "assert every candidate lies outside the native range" test: it re-asserts the kernel's own numbers and couples the suite to four sysctl spellings.

All three copies of this logic moved to `tests/lib/xymond-daemon.sh`; each test keeps only its own argv. Full suite here: 88 passed, 0 skipped, 0 failed.

SoundGoof (approved):
lgtm

PR #426, opened 2026-08-22
client: write the remote-df sentinel once, stamp it into five

bonomani:
Reopened with different content: this now removes the duplication in the **clients**, not in the tests. The earlier version deduplicated three sentinel test files and was closed as not worth it; that judgement was right for the tests and wrong to extend to this.

Six functions were maintained in more than one client, byte for byte — **783 redundant lines**:

| function | size | copies |
|---|---|---|
| `df_sentinel` | 128 | 5 |
| `run_df` | 74 | 3 |
| `probe_dir_is_local` | 21 | 5 |
| `fs_mounts`, two dialects | 14 and 13 | 2 each |
| `fs_procname` | 4 | 4 |

They cannot be shared at runtime. `xymonclient.sh` picks a client by uname and *executes* it, so a sourced fragment is absent for a moment during clientupdate and would abort the client on exactly the hosts the sentinel protects. So they stay copies — but written once. `client/shared/` holds each function and `build/mkclientshared.sh` stamps it into every client declaring a `BEGIN SHARED <name>` region.

Output is committed, so nothing about the build, the packaging or the installed client changes: each `xymonclient-<os>.sh` is still one self-contained script, and everything outside the markers is still hand-edited, which is what a distro patch touches.

`fs_mounts` is two fragments because mount(8) has two syntaxes: FreeBSD and macOS put the type first inside the parentheses, NetBSD and OpenBSD after a `type` keyword before them. They are named for the syntax rather than the OS — `fs_mounts-type-in-parens`, `fs_mounts-type-keyword` — so a client now states which form it parses, where before you had to diff two awk programs to notice they differed. Their leading comments stay outside the markers, since those name the OS and are legitimately per-client.

`tests/client/fs-sentinel-copies.sh` goes. It reported drift in `df_sentinel` but never removed the four extra edits, and never covered the other five functions. A CI job replaces it, regenerating and failing on any difference, in the shape of the job already guarding `docs/manpages` — identity by construction rather than by assertion.

| check | result |
|---|---|
| shared code after generation | byte-identical to before in all five (`md5` per function) |
| diff to the clients | additions only, **zero deletions**: marker lines, plus the 5 comment lines four clients lacked above `df_sentinel()` |
| generator run twice | no change — idempotent |
| commit a hand-edit inside a marker | CI check catches it |
| commit a fragment edit without regenerating | CI check catches it |
| committed tree regenerates to itself | clean |
| `dash -n` on all five clients | parse |
| full suite, built tree | 66 passed / 25 skipped / 0 failed (was 67, minus the deleted policing test) |
| shellcheck on the generator | 0 findings |

---

**Related:** #425 (alternative df column names) and #427 (the plan for the next dedup). No file overlap with either — this one stands alone and can merge in any order.


bonomani:
Closing: the duplication is real but not costing anything.

The three BSD sentinel tests were written together for #316 and have stayed identical since; no fix has had to be applied three times. Against that, a shared harness adds an indirection layer that the sibling `fs-filter-common.sh` has already needed three escape hatches for in one release — `FSF_ARGV_EXCLUDE_PSEUDO=''`, a hand-written darwin mount stub, and `FSF_INODE_FILTERED=no` — and this one would start at 3 of 5 clients, darwin and linux keeping their own bodies.

The measured saving was also smaller than it looked: 636 lines to 385, because the df stub does not compress without losing fidelity.

Worth revisiting if the same fix ever has to be made in all three, or if a contract rule is wanted across them. Neither has happened.

PR #434, opened 2026-08-23
Make pickdata() require and respect an explicit argument count

n1tehawk:
The `pickdata()` variable arguments function in *lib/matching.c* relies solely on the return value of pcre2_match() to determine the number of output parameters (= target strings). This is unsafe when a regular expression match yields more (sub)groups than actual destination params were supplied - possibly leading to stack corruption, and subsequent access violations or crashes.

I've seen that happen with Linux interface names containing an underscore ('_'), which seems to be common with some TUN/TAP/veth setups.

This patch changes `pickdata()` to be a preprocessor definition that determines and passes the actual argument count at compile time, turning it into a "safe va func" while preserving existing invocation syntax for the calling code.

See e.g.
https://www.reddit.com/r/C_Programming/comments/17d3x34/what_is_the_best_idiom_for_variadic_functions/ https://www.reddit.com/r/C_Programming/comments/17ov5az/a_way_to_count_number_of_variable_arguments/ https://codeberg.org/NRK/slashtmp/src/branch/master/misc/safe_va_func.c
https://snai.pe/posts/varargs

### DIFF
--- a/lib/matching.c
+++ b/lib/matching.c
@@ -173,7 +173,15 @@ pcre2_code **compile_exprs(char *id, const char **patterns, int count)
 	return result;
 }
 
-int pickdata(char *buf, pcre2_code *expr, int dupok, ...)
+/* nargs: how many char** destinations the caller passed -- NOT optional.
+ * pcre2_match() returns 1 + (highest capture group that matched), a count that
+ * depends on the DATA: an optional group fires only for some inputs. If it ever
+ * exceeds the fixed argument list, the loop pulls a va_arg that was never passed
+ * and dereferences a garbage pointer -- likely resulting in memory-unsafe read,
+ * xfree() or write operation (issue #433: a Linux iface name with an underscore,
+ * e.g. "veth_0", makes an ifstat pattern capture one group more than its call
+ * site supplied). Clamp res to what the caller vouched for. */
+int pickdata(char *buf, pcre2_code *expr, int dupok, int nargs, ...)
 {
 	int res, i;
 	pcre2_match_data *ovector;
@@ -192,8 +200,9 @@ int pickdata(char *buf, pcre2_code *expr, int dupok, ...)
 		pcre2_match_data_free(ovector);
 		return 0;
 	}
+	if (res > nargs + 1) res = nargs + 1; /* loop runs i = 1 .. res-1 */
 
-	va_start(ap, dupok);
+	va_start(ap, nargs);
 
 	for (i=1; (i < res); i++) {
 		*w = '\0';
--- a/lib/matching.h
+++ b/lib/matching.h
@@ -33,7 +33,7 @@ extern void freeregex(pcre2_code *pcrecode);
 extern int namematch(const char *needle, char *haystack, pcre2_code *pcrecode);
 extern int patternmatch(char *datatosearch, char *pattern, pcre2_code *pcrecode);
 extern pcre2_code **compile_exprs(char *id, const char **patterns, int count);
-extern int pickdata(char *buf, pcre2_code *expr, int dupok, ...);
+extern int pickdata(char *buf, pcre2_code *expr, int dupok, int nargs, ...);
 extern int timematch(char *holidaykey, char *tspec);
 #endif
 
--- a/xymond/rrd/do_ifstat.c
+++ b/xymond/rrd/do_ifstat.c
@@ -206,7 +206,7 @@ int do_ifstat_rrd(char *hostname, char *testname, char *classname, char *pagepat
 		  case OS_ZVM:
 		  case OS_ZVSE:
 		  case OS_ZOS:
-			if (pickdata(bol, ifstat_linux_pcres[0], 1, &ifname)) {
+			if (pickdata(bol, ifstat_linux_pcres[0], 1, /*nargs*/1, &ifname)) {
 				/*
 				 * Linux' netif aliases mess up things. 
 				 * Clear everything when we see an interface name.
@@ -225,9 +225,9 @@ int do_ifstat_rrd(char *hostname, char *testname, char *classname, char *pagepat
 					if (txstr) { xfree(txstr); txstr = NULL; }
 				}
 			}
-			else if (pickdata(bol, ifstat_linux_pcres[1], 1, &rxstr, &txstr)) dmatch |= 6;
-			else if (pickdata(bol, ifstat_linux_pcres[2], 1, &rxstr)) dmatch |= 2;
-			else if (pickdata(bol, ifstat_linux_pcres[3], 1, &txstr)) dmatch |= 4;
+			else if (pickdata(bol, ifstat_linux_pcres[1], 1, /*nargs*/2, &rxstr, &txstr)) dmatch |= 6;
+			else if (pickdata(bol, ifstat_linux_pcres[2], 1, /*nargs*/1, &rxstr)) dmatch |= 2;
+			else if (pickdata(bol, ifstat_linux_pcres[3], 1, /*nargs*/1, &txstr)) dmatch |= 4;
 			break;
 
 		  case OS_FREEBSD:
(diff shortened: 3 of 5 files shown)


bonomani:
Hi @n1tehawk — thanks for catching this and for the quick fix. The loop-cap is the right core fix. Looking closer, there turn out to be two problems here, and I think a small change of approach makes it simpler and more robust. Analysis and ready-to-use code below — take whatever's useful.

---

**A — `pickdata()` unsafe variadic.** `pcre2_match()` returns `1 + (highest capture group that matched)`, a count that depends on the *input* (an optional group fires only for some strings). When it exceeds the fixed argument list, `va_arg` hands back a garbage `char **` from beyond the arguments and the body dereferences it — for `dupok` (which `ifstat_linux_pcres[0]` uses), `if (*ptr) xfree(*ptr)` reads *through* that wild pointer first (and would `xfree()` whatever it finds), then `*ptr = strdup(...)` writes through it: a wild read/free/write, not just an over-read. The cap closes it.

On the *mechanism*: the `sizeof((void*[]){__VA_ARGS__}) / sizeof(void*)` macro is correct and keeps callers unchanged, but (a) it would be the tree's **first** C99 compound literal — a new language-feature requirement in a widely-included header where nothing else uses one — and (b) it is a function-like macro shadowing the name, expanding `__VA_ARGS__` twice (safe only because every argument is a side-effect-free `&var`). An **explicit `nargs`** avoids both and is self-documenting — the property the bug was missing:

```c
/* nargs: how many char** destinations the caller passed -- NOT optional.
 * pcre2_match() returns 1 + (highest capture group that matched), a count that
 * depends on the DATA: an optional group fires only for some inputs. If it ever
 * exceeds the fixed argument list, the loop pulls a va_arg that was never
 * passed and dereferences it -- for dupok, xfree() on a garbage pointer and a
 * write through it (issue #433: a Linux iface name with an underscore, e.g.
 * "veth_0", makes an ifstat pattern capture one group more than its call site
 * supplied). Clamp res to what the caller vouched for. */
int pickdata(char *buf, pcre2_code *expr, int dupok, int nargs, ...)
{
	...
	if (res > nargs + 1) res = nargs + 1;   /* loop runs i = 1 .. res-1 */
	va_start(ap, nargs);
	...
}
```
…then update the prototype in `matching.h` to match (`extern int pickdata(char *buf, pcre2_code *expr, int dupok, int nargs, ...)`) and pass the count at each of the **30** call sites, e.g. `pickdata(bol, ifstat_linux_pcres[0], 1, /*nargs*/1, &ifname);`. (The header matters: a call site still seeing the old prototype passes the count in the wrong ABI slot and crashes.)

**B — the actual trigger: a stray *capturing* group.** `ifstat_linux_exprs[0]` (`do_ifstat.c:22`) captures the underscore suffix it never reads, so `veth_0` yields one group more than the call supplies. Making it non-capturing fixes the crash at the source (correct even without A):

```c
-	"^([a-z0-9]+(_[0-9]+)?:*|lo:?)\\s",
+	"^([a-z0-9]+(?:_[0-9]+)?:*|lo:?)\\s",   /* (?:...) non-capturing: 1 group = 1 arg */
```

**Audit — B's blast radius.** I checked all 30 `pickdata` call sites (18 in `do_ifstat.c`, 11 in `do_netstat.c`, 1 in `xymond.c`): `ifstat_linux_pcres[0]` (2 groups, 1 arg) is the **only** groups-vs-args mismatch. `do_netstat.c`'s six `*_exprs` arrays are each `[2,2,2,2,1,1,1,1,1,1,1]` and match their calls exactly; `xmhptn` = `^(XMH_.*)=(.*)` is 2 groups / 2 args. So B is a one-line point-fix; A is the general safety net for the next regex.

**Recommend both** — B gives correct data today; A keeps a future regex/caller drift from becoming a wild read/free/write instead of a dropped group. (I reproduced this end to end against the real `pickdata` + `ifstat_linux` regex: `veth_0` gives `res=3` and SIGSEGVs today; both A — `nargs=1` clamps `res` — and B stop it and yield `ifname="veth_0"`.)

---

It's your PR, so shape it however feels right — I just wanted to share what the audit
turned up so the fix lands cleanly. This one isn't platform-specific, by the way: it's the
Linux ifstat path, triggered by an interface name with an underscore (`veth_0`, and the
TUN/TAP/veth names the issue mentions), so it can bite the server processing that client's
data whatever OS the server runs. If it helps, I'm glad to verify whatever you push —
reproduce with an underscore interface name and run the suite — and send the results back.
Thanks for driving this one.

n1tehawk:
Hi @bonomani!

Thank you for your quick reply and the extensive analysis of the problem.

With this approach my intention was indeed to fix the problem "at the root" without being too disruptive on the existing code. But you are right, I had not considered the wider implications of introducing a new language/preprocessor construct and possible portability issues. I fully agree with your assessment that "**A**" is the way to go, to avoid this problem in a future-proof way.

"**B**" is an elegant solution, providing a one-line fix that avoids the current issue. But at the same time it covers up the deeper insecurity hidden in `pickdata()`. This is a rather obscure bug that had me scratching my head for a while back when I chased and patched it (for original Xymon) - because initially some Xymon installations were fine, while on others the server crashed with a *SIGSEGV*. It took me a while to realize that it was linked to the underscore character in interface names.

So my gut feeling is that changing `pickdata()` and its invocations to an explicit `nargs` parameter is the correct solution. I can come up with the revised code for that - do you have a preferred way of handling this for the PR? I can either append a "diff" commit (that you'd possibly squash later), or just force-push a new version with the update - for your review. I think the latter would be simpler?

Regards, NiteHawk

bonomani:
Seems to me A + B, force push.
A alone leaves a regex capturing a group no caller reads; B alone leaves `pickdata()` free to crash on the next drift. B only masks the bug if it ships without A.
One commit if possible (or one per feature/fix).
PR title and body, PR comments, code comments: accurate and short.
Bruno

bonomani:
Nice — this looks great. You went with the explicit `nargs` and the non-capturing group, and I verified the reworked version end to end:

- **All 30 call sites** carry the count, and each matches a group-count audit of the tree (ifstat 18, netstat 11, `xmhptn` 1) — no off-by-ones.
- **`matching.h` prototype updated** to match, so no ABI-slot mismatch.
- **Compiles clean, zero warnings** here.
- **Reproduces fixed:** with this branch's `pickdata` + the non-capturing `ifstat_linux_exprs[0]`, `veth_0` now gives `res=2` and returns cleanly with `ifname="veth_0"` — no SIGSEGV (the same input SIGSEGVs on `main`). Both layers do their job: B removes the trigger, A is the safety net for the next regex.

Two cosmetic nits, neither blocking:

1. The function comment (the one I'd drafted) says *"for dupok, `xfree()` on a garbage pointer, then a write through it"* — strictly, the **read** faults first: `if (*ptr)` dereferences the wild pointer before any free/write. Feel free to tweak or leave it.
2. `va_start(ap, nargs);` sits just above the `if (res > nargs + 1) …` clamp; identical behaviour, just reads a touch cleaner with the clamp first. Pure taste.

Otherwise LGTM. Only thing left is the workflow run (fork PR — needs a maintainer to approve the checks) and a review; happy to run it through the BSD/macOS lanes too if that's useful.

bonomani (dismissed):
I reviewed this — **clean, no findings**:

> "The explicit argument count is consistently added to every `pickdata` call site and correctly bounds variadic destination access. The Linux interface regex change preserves matching behavior while eliminating the unintended capture group."

Three independent green lights:
- **CI** — build ×3 + manpages + shell regression all pass.
- Clean.
- **Manual** — all 30 `nargs` counts match a group-count audit of the tree; compiles warning-free; the `veth_0` crash reproduces as *fixed* (no SIGSEGV, correct `ifname`), where the same input SIGSEGVs on `main`.

Both layers in place: non-capturing group removes the trigger, explicit `nargs` is the safety net. LGTM — approving.

bonomani:
@n1tehawk — this is approved and CI-green, so it's ready to go in. Rather than merge it for you: I've sent you an invite as a repo collaborator (write access) — you found a real memory-safety bug and turned it into a clean, well-reviewed fix, which is exactly the kind of work we want more of.

Once you accept the invite (GitHub notification / email), the merge button on this PR is yours — feel free to land it yourself as your first commit with write access. From then on your branches get CI without the fork "approve workflows" step, and you can review others' PRs too.

Welcome aboard, and thanks for the fix.

n1tehawk:
> va_start(ap, nargs); sits just above the if (res > nargs + 1) … clamp; identical behaviour, just reads a touch cleaner with the clamp first. Pure taste.

I'm fine with either, I just thought this way it's directly next to the loop itself - so a reader can easily check the range. But we're talking about a 3-line distance, so it doesn't really matter. If you prefer I'll move the line up again, it may be a bit more natural to have it after `res` gets assigned.

n1tehawk:
Reworked the function comment and moved the clamp line up again (I agree it reads cleaner).

No functional changes, so the previous review/approval should still be valid.

bonomani (approved):
Re-reviewed the cosmetic update (comment reword + clamp moved above `va_start`) — no behavioral change, and re-checked clean. CI 5/5 green. LGTM 👍

PR #445, opened 2026-08-25
tests: the suite cleans up after itself, and checks that it did (#417)

### LINKED ISSUE #417: tests/lib/svcstatus-cgi.sh creates /tmp/xymon-svcstatus-cache without removing it
After building xymon on my machine, I find leftover tmp directory and files:
``` bash
$ ls -la /tmp/xymon-svcstatus-cache.1000/
total 2452
drwxr-xr-x  2 roland roland    4096 Aug 21 14:19 .
drwxrwxrwt 42 root   root    102400 Aug 21 14:25 ..
-rwxr-xr-x  1 roland roland  971736 Aug 21 14:19 _usr_local_src_Packages_xymon_build_area_xymon_4_3_31_alpha10.181218631342
-rwxr-xr-x  1 roland roland 1425008 Aug 21 14:19 _usr_local_src_Packages_xymon_build_area_xymon_4_3_31_alpha10.3574463723377
```
This seems to be created by tests/lib/svcstatus-cgi.sh:120 `svcstatus_build()`:
```bash
	bin="${TMPDIR:-/tmp}/xymon-svcstatus-cache.$(id -u)/$(printf '%s' "$ROOT" | tr -c 'A-Za-z0-9' '_').${flagkey:-plain}"
```

IMHO the test suite should remove these when it's done.

bonomani:
Fixes #417.

## What this does

The test suite cleans up after itself, and checks that it did.

**`tests/final/no-leftovers.sh`** runs after every suite run and sweeps `$TMPDIR`. If anything this run created is still there, it fails and names it. It catches *any* test that forgets to tidy up, not only the one that prompted this — so the next #417 is caught here rather than reported by somebody packaging a release.

**`tests/buildsystem/testsuite-run-disposal.sh`** starts a small throwaway suite and ends it three ways — a test fails, Ctrl-C, `kill` — and checks nothing survives each time. Those are the endings the sweep cannot watch, because on a signal it never runs at all.

So every way a run can end is covered:

| how a run ends | checked by |
|---|---|
| finishes, all passed | `no-leftovers.sh` |
| finishes, something failed | `run-disposal.sh` |
| stopped with SIGINT / SIGTERM | `run-disposal.sh` |
| killed with SIGKILL | out of scope — no trap runs |

To make the sweep precise, the runner drops a timestamp file when it starts, so *left behind* means "by this run" and not "by one last month". It is disposed of on exit, interrupt and terminate.

Also: the run order is pinned to `LC_ALL=C`, so the sequence is identical on every machine. Nothing depends on order — the suite passes 95 of 95 in reverse as well as forward.

## What it removes

`svcstatus_build()` cached the linked CGI under `$TMPDIR` and nothing ever removed it. That is #417. It now compiles each time — 510 ms more per 140-second suite run, measured.

(The first version of this PR aged the cache entries out instead. That could not work: the prune ran inside `svcstatus_build()`, so it only fired when somebody built *again* on the same machine, which a packaging build never does.)

Full suite: 95 passed, 0 failed.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -54,9 +54,25 @@ shipped-file invariants) get their own area.
 | `tests/integration/` | end-to-end scenarios spanning multiple components   |
 | `tests/lib/`      | sourced helpers (`assert.sh`, future `net.sh` etc.)    |
 | `tests/fixtures/` | shared data files (config snippets, expected outputs)  |
+| `tests/final/`    | checks about what a whole run LEAVES; held back and run last |
 
 Add a new area by PR when an existing one doesn't fit. Don't bend a
 test to fit the wrong area just to avoid creating a new directory.
+
+`tests/final/` is the one area with an order attached to it. Everything
+else runs in `LC_ALL=C` filename order and must not care; the runner
+holds `tests/final/` back and runs it after all of them, because those
+checks are about what the run as a whole left behind, which is only a
+fact once nothing else is running. Ordering it by filename would work
+until a directory sorting after "final" appeared, and would then go on
+passing while checking nothing.
+
+Each run also gets a temp directory of its own: the runner creates one
+and points `TMPDIR` at it, so `mktempdir` and anything else honouring
+TMPDIR writes inside it, and the whole directory goes when the run ends
+-- on exit, on SIGINT and on SIGTERM. That is what lets a final check
+ask "what did THIS run leave", and what keeps two suites running at the
+same time from seeing each other's files.
 
 ### Runnable vs sourced/data files
 
--- a/tests/buildsystem/testsuite-run-disposal.sh
+++ b/tests/buildsystem/testsuite-run-disposal.sh
@@ -3,9 +3,9 @@
 #
 # A run leaves nothing behind, however it ends.
 #
-# tests/final/no-leftovers.sh covers the runs that finish. It cannot cover the
-# ones that are stopped -- on a signal it never runs -- nor the runner's own
-# marker, which it excludes because the marker is still in use while it runs.
+# no-leftovers.sh covers the runs that finish, and only what is inside the run
+# directory. The endings it cannot see are here: a failing run, SIGINT, SIGTERM
+# -- and the run directory itself, which is still in use while it runs.
 # So those are here: a failing run, SIGINT, SIGTERM. SIGKILL is out of scope,
 # since no trap runs.
 #
@@ -18,7 +18,7 @@ set -eu
 . "$(dirname "$0")/../lib/assert.sh"
 ROOT=$(find_root)
 
-command -v mktemp >/dev/null 2>&1 || skip "mktemp is needed to watch the marker appear"
+command -v mktemp >/dev/null 2>&1 || skip "mktemp is needed to watch the run directory appear"
 
 work=$(mktempdir); register_cleanup "rm -rf '$work'"
 
@@ -35,7 +35,8 @@ maketree() {
 	esac
 	chmod +x "$dir/tests/dummy/t.sh"
 }
-markers() { find "$1" -maxdepth 1 -name 'xymon-runmarker.*' 2>/dev/null; }
+# One directory per run, removed however the run ends.
+runroots() { find "$1" -maxdepth 1 -name 'xymon-run.*' 2>/dev/null; }
 
 # --- a run that finishes with a failing test ---------------------------------
 d="$work/fail"; tmp="$d/tmp"; maketree "$d" failing; mkdir -p "$tmp"
@@ -46,9 +47,9 @@ set -e
 [ "$rc" -ne 0 ] || fail \
 	"the fake suite reported success with a failing test in it, so this case is
 not the one it claims to be testing"
-left=$(markers "$tmp")
+left=$(runroots "$tmp")
 [ -z "$left" ] || fail \
-	"a run that ended in failure left its marker behind:
+	"a run that ended in failure left its run directory behind:
 $left
 A failing run is the one somebody re-runs, so its leftovers are the ones that
 accumulate fastest."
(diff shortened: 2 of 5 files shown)


RolandRosenfeld:
According to the PR description this does not fix the problem I mentioned in #417.
I do not like that building a program by default leaves behind files in /tmp.
At least chkrootkit complains about them.
For building in Debian context I'll try to work around this issue by setting TMPDIR to  a directory, which I explicitly delete after the build in the clean target of debian/rules, so this is an acceptable workaround for me.
Anyway I don't think that such a cache that persists after the end of a make run is good style.

bonomani:
You were right, and the reason is worse than I had it. The prune ran inside `svcstatus_build()`, so it only fired when somebody built **again** on the same machine. You build once from a recipe and never return — so it never ran for you at all. The change was a no-op for the case you reported.

Rewritten. The suite now makes one cache directory and removes it when the run ends. The four svcstatus tests still share a single compile, which is what the cache was for, and nothing survives the run — so the `TMPDIR` workaround in `debian/rules` should not be needed. Anyone who wants the compile to persist between runs opts in with `XYMON_TEST_CACHE`, pointing at a directory of their own.

Two tests: one is #417 as you filed it — run the tests once, as a packaging build does, then look. The other runs at the end of every suite run and fails if anything at all was left in the temp directory, so the next one gets caught without anyone having to think of it in advance.

Also agreed on the general point: the old default wrote an executable to a predictable path in a world-writable directory and later ran it, checking only its mtime. That is gone.

In draft while I finish checking it. Thanks for pushing back on the first version.

bonomani:
Dropped the opt-in persistence I mentioned above — it does not earn its keep. Measured: a cache hit is 39 ms against a 333 ms compile, so keeping the cache between runs saves **one ~300 ms compile per suite run**. Not worth an environment variable, a week-long ageing rule, and a test to guard it.

So the cache is now always scoped to the run, with no way to make it outlive one. The ageing, the touch that fed it, and its test are gone with it — the change is smaller than the version you first commented on.

bonomani:
Added the test that was missing, and it was missing before this PR too: nothing anywhere asserted the cache **does its job**. It was pinned from both ends — that a build populates it, and now that nothing survives the run — and never in the middle.

That matters here specifically, because scoping the cache to the run is very nearly the same change as giving every test its own copy. Get it slightly wrong and the suite stays green while compiling four times instead of once.

`svcstatus-cache-reuse.sh` counts at the compiler, with `CC` wrapped, since a rebuild inside the same second looks exactly like a reuse if you go by mtime. Verified both ways: it passes here, and reports "compiled it 2 times" when the shared cache is bypassed.

bonomani:
One more, and it is the one I should have written first.

Nothing asserted that the cache **never serves a stale binary**. `svcstatus_build()` states it in a comment — *"any code change in the tree is newer than the cache and forces a rebuild; the cache never hides one"* — and that was the whole of the guarantee. Drop a file from the freshness list and the comment still reads exactly as true as it does now, while every svcstatus test goes green against a build of the previous source.

`svcstatus-cache-staleness.sh` touches each watched input in turn and requires a recompile. Verified by removing `svcstatus-info.h` from the list: it reports the stale binary.

It turned up a real limit too. Linux stamps file mtimes from a clock that only advances once per tick, so two files written milliseconds apart carry the same nanosecond timestamp and `find -newer` is strict — a source saved in the same tick as the cached binary is invisible to the check. Inherent to mtime comparison, `make` has the same hole, and the window is a few milliseconds during the compile. Recorded in the test rather than papered over.

bonomani:
Simplified again: there is no cache at all now.

Measured what it was worth — 1583 ms sharing one link against 2093 ms compiling four times, so **510 ms on a 140-second suite**. That does not pay for a directory in `$TMPDIR`, three traps, a hand-kept freshness list, and the risk of a stale entry letting a test pass against source no longer in the tree. `svcstatus_build()` now just compiles, and #417 becomes impossible rather than managed.

Body updated; it had been describing a design that no longer exists.

bonomani:
Restarted from scratch.

The leak came from an optimisation that brought nothing. Removed.

Added a guard so it cannot come back quietly: after every suite run a check sweeps the temp directory and fails if anything the run created is still there — any test, not just this one. A second check covers the runs that end with Ctrl-C or `kill`.

The `TMPDIR` workaround in `debian/rules` should not be needed.

RolandRosenfeld:
I really like the new simplicity of tests/lib/svcstatus-cgi.sh.

But I'm not sure, whether I like the other changes of the testsuite.
- As far as I can see, the new behavior of final dictionary isn't documented in README.md
- I didn't deeply dive into the new tests, but from my first view it looks, that two (parallel) test runs would fail here.

I'm not sure, whether adding these new tests is necessary vs. the only increase complexity of the testsuite without much effect.

bonomani:
Thanks — both of your first two points were right, and I've pushed fixes. I've marked this draft while we settle the third.

**Parallel runs did break.** Worse than a race: the sweep looked for everything under `$TMPDIR` newer than the run marker and named `xymon-*`, which is exactly what a *concurrent* run's live files are called. Measured on the code as it stood — two runs sharing a `TMPDIR`, run A exits 1 naming run B's directories while B passes. Intermittent, and it accuses a run that is behaving perfectly.

The fix makes the check simpler rather than more careful. The runner now creates one directory per run and points `TMPDIR` at it, so every test writes inside it. "What did this run leave" becomes "what is still in my own directory" — no marker, no `-newer`, no name filter, and nothing another run can be mistaken for. The directory goes when the run ends, on exit, SIGINT and SIGTERM. `testsuite-run-disposal.sh` now covers the case you raised: two runs, one `TMPDIR`, each must get its own directory and neither may leave one.

**`tests/final/` is now documented** in `tests/README.md`: the directory table, why it is held back rather than ordered by filename, and the per-run temp directory.

Two things I found while doing that, both mine:

- the disposal test called `errline()`, which is the runner's function and undefined in a test. Any exit-status mismatch was a command-not-found rather than the complaint it meant to make, and the case carried on — so that assertion had never reported anything, anywhere. #485 adds a lint rule for the class.
- the exit status a trap leaves is not portable: bash 3.2 on macOS reports 0 for both signals where bash 5 on Linux reports 130 and 143 — true of the runner before this branch as well. It is now reported rather than asserted; the cleanup is what is asserted, and that does not vary by shell.

**On complexity, I think you are pointing at something real, and I would rather split it than argue it.** This PR does three things: fix #417, add the leftovers sweep, and pin the run order to `LC_ALL=C`. Only the first is the bug. My proposal: keep here the `svcstatus-cgi.sh` fix plus `no-leftovers.sh` — 55 lines, and the thing that stops the next #417 — and move `testsuite-run-disposal.sh` (the 95-line signal harness) and the ordering pin into their own PR, so each can be judged on its own worth rather than as a package.

Say the word and I'll split it.

RolandRosenfeld (approved):
Looks good to me now.

I wondered, why the final test wasn't run on my system, that was because I applied the patch without changing the x-bit (this correct in git, but broken when applying the patch manuelly).
Maybe we should change this to ignore the x-bit and run all tests directly via bash or add another test, which checks for this (tests/client/fs-filter-common.sh currently doesn't have the x-bit set).

Anyway, this aren't problems with this PR, which I can approve now.

PR #449, opened 2026-08-25
xymonnet: drop a dead read of the certificate after releasing it

bonomani:
`setup_ssl()` reads the certificate fields, then releases its reference at the end of the certificate block:

```c
X509_free(peercert);

/* We list the available ciphers in the SSL cert data */
if (sslincludecipherlist) {          /* default on */
    ...
    certsigalg = OBJ_nid2ln(X509_get_signature_type(peercert));   /* reads a cert this function no longer owns */
```

Two things are wrong with that line, and the second is the reason to delete it.

**It is dead.** `certsigalg` was already computed earlier and emitted in the "Server certificate:" block; the value assigned here is never read again — the cipher section reports `SSL_get_cipher_name()`, not `certsigalg`. Removing the line changes no output.

**It reads through a pointer this function has released.** `SSL_get_peer_certificate()` returns the certificate with its reference count incremented, so `X509_free()` drops *our* reference and the code goes on using the pointer anyway.

## What this is not

An earlier version of this description called it a use-after-free that a hostile endpoint could exploit. That was wrong, and worth stating plainly rather than quietly editing out: the session keeps its own reference to the peer certificate, and the session is alive here — the surrounding lines use `item->ssldata`. So the object is not freed, nothing is reallocated, and there is nothing for a peer to shape.

What remains is an ownership violation: correct today only because something else happens to hold the object alive. It becomes a genuine use-after-free the moment the session is released before this block — which is exactly the kind of reordering nobody would expect to be load-bearing.

## The test

`tests/network/contest-cert-lifetime.sh` encodes an invariant stronger than "this line is gone": `X509_free(peercert)` must be the **last** mention of `peercert` in the file. That is the property worth keeping — after the release, this code has no claim on the certificate — and it forecloses the reordering above.

| check | result |
|---|---|
| `contest.c` compiles with the fix, no new warning | yes (`certsigalg` still used above, no "set but unused") |
| `tests/network/contest-cert-lifetime.sh` | PASS |
| same test against pre-fix `contest.c` | FAIL |
| `tests/buildsystem/test-suite-portability.sh` | PASS |

Found by a source pass for unfixed siblings of the milestone-2 memory-safety fixes. #488 had picked up the same one-line deletion while working nearby; it has been removed there so this PR owns the change.

### DIFF
--- a/tests/network/contest-cert-lifetime.sh
+++ b/tests/network/contest-cert-lifetime.sh
@@ -1,0 +1,54 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/network/contest-cert-lifetime.sh
+#
+# Nothing in setup_ssl() may read the peer certificate after releasing it.
+#
+# The certificate block ends with X509_free(peercert). The cipher-list section
+# that follows (entered whenever sslincludecipherlist is set, the default) then
+# recomputed the signature algorithm from that certificate -- a read through a
+# pointer this function no longer owns, and a dead one: certsigalg was already
+# emitted above and the new value was never used.
+#
+# Not a use-after-free today: SSL_get_peer_certificate() returns an
+# incremented reference, so the session still holds one and the object
+# survives. It becomes one the moment the session is released first, which is
+# why the ordering is worth pinning rather than left to hold by accident.
+#
+# setup_ssl() needs the OpenSSL headers, a real handshake and the network to
+# run, so -- like tests/server/combostatus-overflow.sh -- this binds to the
+# source: the freed certificate must not be read again. The invariant checked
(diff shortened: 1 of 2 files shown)


bonomani:
**Correction, folded into the description above.**

This was originally written up as a use-after-free that a hostile endpoint could exploit. It is not one: `SSL_get_peer_certificate()` returns the certificate with its reference count incremented — OpenSSL's contract is that it "will not be destroyed when the session containing the peer certificate is freed" — so `X509_free()` here drops only our reference and the session keeps its own.

The deletion still stands: the line is dead, and it reads a certificate this function has released. The title and body now say that instead.

I made the mistake in that direction myself — I checked the free-before-use ordering and not the refcounting.

PR #453, opened 2026-08-26
xymonnet: wait for the direction SSL_connect asked for, not for writability

### LINKED ISSUE #452: TLS handshake busy-waits: socket registered write-only while sslrunning == SSLSETUP_PENDING
## Summary

While a TLS handshake is in progress, the socket is registered with `select()` for
**writability only**. A connected socket is almost always writable, so `select()` returns
immediately and the loop spins through `setup_ssl()` -> `SSL_connect()` ->
`SSL_ERROR_WANT_READ` repeatedly until the peer's data actually arrives.

Correctness is unaffected — it makes progress, which is why SSL tests work. The cost is CPU
burned for the duration of every SSL handshake round trip.

Present in 4.3.30 and current `main`.

## Code

`xymonnet/contest.c:1154`, one or the other, never both — the comment above it says so
explicitly:

```c
if (item->readpending)
        FD_SET(item->fd, &readfds);
else
        FD_SET(item->fd, &writefds);
```

While the handshake is pending, `contest.c:1281` leaves `do_talk` false:

```c
do_talk = (item->sslrunning == 1);
```

and `contest.c:1295` therefore leaves `readpending` false:

```c
item->readpending = (do_talk && !item->silenttest && ...);
```

So the socket goes back into `writefds` on every iteration for the whole time it is
waiting on the server's ServerHello/Certificate, even though what it is actually waiting
for is readability.

## Impact

CPU only. It scales with concurrency — the default is `FD_SETSIZE / 4` (typically 256)
sockets, so a batch of SSL tests all handshaking at once will spin the main loop at full
tilt until their peers respond.

## Measured

Real `xymonnet` against a peer that accepts the TCP connection and then sends nothing, so
`SSL_connect()` stays in `WANT_READ` for the whole test:

```
$ strace -f -c -e trace=select,pselect6,read xymonnet --checkresponse --timeout=8
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 85.44    0.640534           7     84025     13997 read
 14.54    0.108989           7     14007           pselect6
```

**14,007 `pselect6` calls and 84,025 `read` calls in an 8 second window**, for a single
stalled handshake — roughly 1,750 selects/sec. A loop blocking correctly on readability
would make a handful. Wall-clock cost of that run: `user=7.96s sys=3.72s` against
`wall=42.65s`.

The `read` errors are `EAGAIN`: `SSL_connect()` polls the socket, gets nothing, returns
`WANT_READ`, and the loop immediately re-selects for *writability* — which is always
ready — instead of waiting for the readability it actually needs.

## Suggested fix

When `item->sslrunning == SSLSETUP_PENDING`, register the socket according to what
`SSL_connect()` actually asked for — `readfds` on `SSL_ERROR_WANT_READ`, `writefds` on
`SSL_ERROR_WANT_WRITE` — rather than defaulting to `writefds` because `readpending` is
false.

bonomani:
Fixes #452.

`do_tcp_tests()` registers each socket with `select()` for read **or** write, never both, choosing by `readpending`. While a TLS handshake is pending `readpending` is still 0 — `do_talk` stays false until the handshake completes — so the socket is registered for *writability* while `SSL_connect()` is blocked on a *read*. A connected socket is almost always writable, so `select()` returns immediately every iteration and the loop spins through `setup_ssl()` → `SSL_connect()` → `WANT_READ` until the peer answers.

It makes progress, which is why this has gone unnoticed. It just burns a core doing it, for one handshake round trip per test, up to `--concurrency` at a time.

The read path this makes reachable already exists and already handles a pending handshake. It was simply never selected.

**That path ended in `break`**, which leaves the whole loop over active sockets. I built a harness for this — ten healthy TLS peers behind one whose handshake is relayed a byte at a time — and it **did not** show harm. Scaled up to five slow peers interleaved through seventy-five healthy hosts at `--concurrency=256`, it still did not:

| | healthy OK | wall | user CPU |
|---|---|---|---|
| `continue` | 75/75 | 50.24s | 1.89s |
| `break` | 75/75 | 50.09s | 1.61s |

`select()` re-arms immediately and a re-scan is a pointer walk, so the recovery is cheap. `break` costs iterations, not results, and not measurably even those. Changed to `continue` anyway, as the right shape for a scan that should not abandon its remaining work, but **no user-visible harm is claimed for it**.

**A handshake can now complete in the read arm**, where the write arm has not run for that socket — `sendtxt` unsent, `readpending` unset, `silenttest` not honoured. Falling into the read there would skip the send and collect a banner for a silent test. That one *is* a real regression this change would otherwise introduce, and it returns to the loop instead.

### Measured

Peer that accepts the TCP connection and then never speaks; 8s timeout; same tree, same container, same service either way:

| | `pselect6` | `read` | read errors |
|---|---|---|---|
| before | **897** | 152880 | 887 (EAGAIN) |
| after | **20** | 488 | 1 |

### Verified

- a working TLS service still tests correctly — against a real smtps peer: `Service smtps on realtls is OK (up)`, banner `220 real.local ESMTP`
- `./tests/testsuite`: the failing set is identical before and after this change, and none of it is a network test. (Not enumerated here — which tests fail depends on the machine, and a list in a description goes stale without telling anyone.)
- the new test fails on the tree without this change

### On the test

`tests/network/tls-handshake-wait.sh` is a static guard — a behavioural run needs a stalled TLS peer plus syscall counting, which the suite has no vocabulary for and which would be timing-dependent. Same trade `tls13-support.sh` makes.

It **strips comments before asserting**: the block's own comments name the identifiers under assertion, so grepping raw text let a comment satisfy a claim about the condition beside it — deleting the `item->open` gate while leaving the comment explaining it passed until that was fixed.

Mutation-checked. Each of these is caught: dropping the `item->open` gate, dropping the pending-handshake branch, inverting the direction mapping, restoring the `break`.

### Review notes

An independent review pass found the `break`, and separately found that the test's block extraction ran to end-of-file (its terminator counted a literal `FD_SET(item->fd, &writefds)` twice, but this change makes the first one a ternary, so it never matched twice) — which is what made the `item->open` assertion a false green. Both are fixed above. It also suggested caching `SSL_get_error()` in a local rather than calling it twice; done.

Not added to `Changes` per CONTRIBUTING. Suggested wording: *xymonnet no longer busy-waits during TLS handshakes.*

## Added: a behavioural test

`tests/network/tls-handshake-cpu.sh` runs the probe against a peer that accepts the connection and then never speaks, and asserts a CPU ceiling.

The source test pins the fd registration, but whether the process then sleeps or spins is only visible by running it. Measured against a peer that accepts and never completes the handshake, `--timeout=5`: **6.8s** of CPU without this change — a saturated core for the whole stall, scaling with the wait (11.7s at `--timeout=10`) — and a few milliseconds with it. Those are the figures `tests/network/tls-handshake-cpu.sh` records and asserts against.

To be clear about what this is not: it is not a speed fix. Over 20 paired runs the wall-clock difference was +0.30s with a 95% CI of −0.34..+0.93 — nothing. The handshake takes as long as the peer takes either way. What changes is that waiting for it stops occupying a core.

---
### Where this sits

`main` → **#453 (here)** → #454 → the dialogue work.

#488 builds on #454 and needs the direction-aware handshake registration added
here, so this lands first. Merging it retargets #454 onto `main`.

### DIFF
--- a/tests/lib/tcp-silent-peer.c
+++ b/tests/lib/tcp-silent-peer.c
@@ -1,0 +1,69 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later                                  */
+/*
+ * tests/lib/tcp-silent-peer.c
+ *
+ * Accept connections on a 127.0.0.1 port and then say nothing at all, so a
+ * TLS client's handshake never completes. Prints the port on stdout so the
+ * test can name it.
+ *
+ * Holds every connection open rather than closing it: a close would let the
+ * client fail fast, and the point is to keep it waiting so the cost of
+ * waiting can be measured.
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <string.h>
+#include <time.h>
+#include <sys/socket.h>
+#include <sys/select.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+
+#define MAXHELD 64
+
+int main(int argc, char *argv[])
+{
+	int sock, held[MAXHELD], nheld = 0;
+	struct sockaddr_in addr;
+	socklen_t alen = sizeof(addr);
+	time_t deadline;
+	int seconds = (argc > 1) ? atoi(argv[1]) : 30;
+
+	sock = socket(AF_INET, SOCK_STREAM, 0);
+	if (sock < 0) { perror("socket"); return 1; }
+
+	memset(&addr, 0, sizeof(addr));
+	addr.sin_family = AF_INET;
+	addr.sin_addr.s_addr = inet_addr("127.0.0.1");
+	addr.sin_port = 0;
+	if (bind(sock, (struct sockaddr *)&addr, sizeof(addr)) < 0) { perror("bind"); return 1; }
+	if (getsockname(sock, (struct sockaddr *)&addr, &alen) < 0) { perror("getsockname"); return 1; }
+	if (listen(sock, 16) < 0) { perror("listen"); return 1; }
+
+	printf("%d\n", ntohs(addr.sin_port));
+	fflush(stdout);
+
+	deadline = time(NULL) + seconds;
+	while (time(NULL) < deadline) {
+		fd_set rfds;
+		struct timeval tv;
+		int r;
+
+		FD_ZERO(&rfds); FD_SET(sock, &rfds);
+		tv.tv_sec = 1; tv.tv_usec = 0;
+		r = select(sock + 1, &rfds, NULL, NULL, &tv);
+		if ((r > 0) && FD_ISSET(sock, &rfds)) {
+			int c = accept(sock, NULL, NULL);
+			if (c >= 0) {
+				if (nheld < MAXHELD) held[nheld++] = c;
+				else close(c);
+			}
+		}
+	}
+
+	while (nheld > 0) close(held[--nheld]);
+	close(sock);
+	return 0;
+}
--- a/tests/network/tls-handshake-cpu.sh
+++ b/tests/network/tls-handshake-cpu.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+#
+# The companion to tls-handshake-wait.sh, which reads the source. This one
+# runs the probe against a peer that accepts the connection and then says
+# nothing, so SSL_connect() never completes, and measures what waiting costs.
+#
+# The property is not speed. The handshake takes as long as the peer takes
+# either way -- over 20 paired runs the wall-clock difference was +0.30s
+# with a 95% CI of -0.34..+0.93, i.e. nothing. What changes is CPU: spinning
+# on SSL_connect() burned 6.8s on a 6.3s run, saturating a core for the whole
+# stall, and it scaled with the wait (6.8s at --timeout=5, 11.7s at 10).
+# Blocking in select() costs ~0.
+#
+# So the assertion is a CPU ceiling far below the timeout. A source test
+# cannot express that: which fd set the socket joins is visible in the code,
+# but whether the process then sleeps or spins is only visible by running it.
+
+set -eu
+. "$(dirname "$0")/../lib/assert.sh"
+root=$(find_root)
+
+require_bin XYMONNET xymonnet/xymonnet
+require_cc
+
+work=$(mktempdir); register_cleanup "rm -rf '$work'"
+mkdir -p "$work/home/etc"
+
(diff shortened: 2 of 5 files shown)


bonomani:
Housekeeping for whoever merges this: #453 and #454 touch adjacent lines and will conflict, trivially.

```
xymonnet/contest.h    int sslwantwrite;          vs   int sendagain;
xymonnet/contest.c    newtest->sslwantwrite = 0; vs   newtest->sendagain = 0;
```

Both resolve as "keep both lines" — the two fields are independent and nothing else overlaps. Verified with a trial merge; those are the only two hunks.

Merge them in either order. I'll rebase whichever lands second, just say the word.

If order is up to you, this one first: #454's remaining limitation (a write deferred with `WANT_READ` still waits on writability, so it re-attempts until the cutoff instead of blocking) is closed by the register-by-direction machinery added here, so landing this first leaves that a clean follow-up rather than something needing new plumbing.

#455 is independent of both — it only touches `protocols.cfg` and merges clean with either.

bonomani:
**Merge order.** #454 is stacked on this branch, so there is no conflict between them: merge this one first and GitHub retargets #454 to `main` on its own.

Stacking also let #454 close something it could not fix alone — a deferred write that hits `WANT_READ` needs the direction-aware registration added here to be retried safely. #454 scoped its retry to `WANT_WRITE` instead, correct on its own, and the two now read as one story rather than two that collide.

bonomani:
**For a reviewer — what to check and how.**

- **Resolves:** while a TLS handshake is in progress the socket is waited on for the wrong direction, so `select()` returns immediately every pass and **spins a core** until the peer answers (per test, up to `--concurrency`).
- **Why we need it:** colours are unchanged (not a correctness bug) — the cost is **CPU** on slow or stalled TLS peers. `tests/network/tls-handshake-cpu.sh` measures it on the machine that runs it and prints the figure — the effect is a saturated core for the length of the stall, versus tens of milliseconds.
- **How to review:** read the direction-based `select` registration (gated on `item->open`) and the two read-arm guards (`break → continue`, and the post-handshake `!readpending` hand-back to the write arm — that one is the real regression this would otherwise introduce). Then **soak it**: run on a real poller with a slow/down TLS target for a while and confirm colours are unchanged and CPU stays flat when a handshake stalls.

This is central-select-loop code, so a timing/resource behaviour that unit tests can't portably reproduce — the honest check is *run it a while and confirm nothing breaks*.

Also: this is **foundation for the dialogue work** (#488) — it builds directly on this handshake path, so this wants to land first.

PR #454, opened 2026-08-27
xymonnet: retry a write that SSL_write() did not accept, instead of dropping it

### LINKED ISSUE #451: socket_write() silently discards the payload when SSL_write() returns WANT_READ/WANT_WRITE
## Summary

For non-HTTP SSL services, when `SSL_write()` returns `SSL_ERROR_WANT_READ` or
`SSL_ERROR_WANT_WRITE`, `socket_write()` returns 0 and records nothing. The call site
treats 0 as neither an error nor a partial write, so **the payload is silently dropped**
while `readpending` stays set.

Present in 4.3.30 and current `main` (identical code).

## Code

`xymonnet/contest.c:834` — the retryable condition is discarded:

```c
res = SSL_write(item->ssldata, outbuf, outlen);
if (res < 0) {
        switch (SSL_get_error (item->ssldata, res)) {
          case SSL_ERROR_WANT_READ:
          case SSL_ERROR_WANT_WRITE:
                  res = 0;          /* <-- and nothing else */
                  break;
        }
}
```

Compare the read path at `contest.c:870`, which sets `item->sslagain = 1` so the poll loop
knows to come back (checked at `contest.c:1416`). The write path sets no such flag.

Call site, `contest.c:1315`:

```c
res = socket_write(item, outbuf, outlen);
if (res == -1) { /* error path */ }
else if (item->svcinfo->flags & TCP_HTTP) {
        item->sendtxt += res;  item->sendlen -= res;
        item->readpending = (item->sendlen == 0);
}
```

`res == 0` is neither `-1` nor `TCP_HTTP`, so nothing happens. HTTP survives by accident:
`sendlen` does not decrease, so `readpending` stays false and the write is retried on a
later pass. Non-HTTP SSL services have no such escape.

## Reproduction

Harness using the upstream `socket_write()` body verbatim, against a peer that stops
reading, with a send buffer small enough that the socket genuinely backs up:

```
C: handshake ok, TLSv1.3
C: raw SSL_write() -> -1  (SSL_ERROR_WANT_WRITE)
C: xymon socket_write() -> 0
C: call-site verdict: res==0 -> not -1, not TCP_HTTP -> PAYLOAD SILENTLY DROPPED
```

## What the consequence is, by service

The dropped payload only matters where the response depends on it:

* `bbd`, `cupsd`, `ajp13`, `rdp` — `send` with **no** `banner` option. The server never
  receives the request, so nothing comes back, `banner` stays NULL and
  `tcp_got_expected()` returns 0. The test fails with "Unexpected service response".
* `smtp`, `smtps`, `imaps`, `ftp`, `ssh`, ~20 others — `send` **plus** `banner`. The
  greeting arrives unprompted, so the probe still matches its expect string and reports
  green. The lost payload is invisible.
* `http`/`https` — exempt, `TCP_HTTP` retries the write.

## Severity: low in practice

Triggering it required a stalled peer and an 8 MB write. The current probes send a few
bytes (`QUIT\r\n`, `ehlo ...`), which will never fill a send buffer. Filing this as latent
robustness, not as a live fault.

Two things I checked that are **not** triggers, in case they seem likely: under TLS 1.3
with OpenSSL 3.0, neither post-handshake `NewSessionTicket` nor a server `KeyUpdate`
causes `SSL_write()` to return `WANT_READ` — it completes normally in both cases.

## Suggested fix

Mirror the read path so the condition survives, and give the call site a retry branch:

```c
case SSL_ERROR_WANT_READ:
case SSL_ERROR_WANT_WRITE:
        item->sslagain = 1;
        res = 0;
        break;
```

Note that simply widening the `TCP_HTTP` branch to cover all services would be wrong: the
telnet-negotiate path passes `item->telnetbuf`, not `sendtxt`, so advancing
`sendtxt`/`sendlen` there would corrupt it.

bonomani:
Fixes #451. **Stacked on #453** — base is `fix-tls-handshake-busywait`, so this diff shows only its own commit. Merge #453 first; GitHub will retarget this to `main` automatically.

`socket_write()` turned `SSL_ERROR_WANT_READ`/`WANT_WRITE` into a bare `res = 0`. At the call site 0 is neither the `-1` error case nor the `TCP_HTTP` short-write case, so for every non-HTTP service the payload was **silently discarded** — and `readpending` had already been set, so the test then waited for a reply to a command that was never transmitted.

`socket_read()` already had the vocabulary for this in `sslagain`; the write path had none.

### Only `WANT_WRITE` is retried

Clearing `readpending` puts the fd back in `writefds`, which is precisely what `WANT_WRITE` waits for. A `WANT_READ` retried there would select on a set that is always ready and busy-wait; moving the fd to `readfds` instead would land in the read branch, which would consume application data rather than retry the write. So `WANT_READ` keeps today's behaviour — the payload is not sent and the test waits out its timeout.

It is reachable only through renegotiation. TLS 1.3 session tickets and `KeyUpdate` do **not** produce it (verified against OpenSSL 3.0).

### Three things follow from clearing `readpending`

It also decides which arm the timeout takes:

1. **The `!readpending` arm never called `socket_shutdown()`** — the only place `SSL_free()`/`SSL_CTX_free()` happen — so a deferred write that never completed had its fd closed with the SSL objects still allocated. Hoisted so both arms free the session. This also closes the same leak for a **handshake** that times out, which could already reach that arm.
2. **That arm marks the test a connect failure**, which is wrong for a socket that reached TCP and TLS before stalling. It now skips that when a write is still owed.
3. **Hoisting made an older hazard reachable.** `SSLSETUP_PENDING` is `-1`, so `if (item->sslrunning)` is already true from `add_tcp_test()` while `ssldata` is still `NULL` — a connection timing out before `setup_ssl()` ran would have reached `SSL_shutdown(NULL)`. `socket_shutdown()` now guards on the objects and clears them, which also makes a second call harmless.

On (3): I could **not** get that to fault in practice — a blackholed SSL port times out cleanly before and after. It is a latent hazard closed, not a crash fixed.

### Consequences differ by service

Which is why it stayed hidden:

`socket_write()` only reaches `SSL_write()` under `if (item->sslrunning)`, so the reach is the TLS-carrying entries and nothing else:

| services | effect of a dropped payload |
|---|---|
| `smtps`, `imaps`, `pop3s`, `nntps`, `ftps`, `telnets` and any entry with `options ssl` | with `banner`, the greeting arrives unprompted, so the test **reports green** with the send quietly missing |
| the same entries without `banner` | no response at all — the test fails, loudly |
| `https` | `TCP_HTTP`, already retried, unaffected |
| every plaintext entry — `smtp`, `ftp`, `ssh`, `bbd`, `cupsd`, `ajp13`, `rdp`, … | never touches this path |

### What I verified

- a real smtps peer still tests correctly after the change
- a blackholed SSL port still times out cleanly, no crash
- `./tests/testsuite`: the failing set is identical before and after this change, and none of it is a network test. (Not enumerated here — which tests fail depends on the machine, and a list in a description goes stale without telling anyone.)
- retries are bounded by the existing per-test cutoff, checked before fd readiness

### What I could not verify

I did not reach `WANT_WRITE` through a real probe. It needs a stalled peer and a multi-megabyte payload, and the shipped probes send a few bytes — which is why this is latent rather than live. It was reproduced in a standalone harness using this function's body verbatim.

### On the test

Assertions strip comments first — this function and its call site are commented with the identifiers under assertion, so prose could otherwise satisfy a claim about the code beside it. Mutation-checked; each is caught: commenting out `sendagain = 1`, the per-call reset, the shutdown condition or the header field; deleting any of them; collapsing `WANT_READ` back into the retry arm; reverting the `ssldata` guard.

Not added to `Changes` per CONTRIBUTING. Suggested wording: *xymonnet no longer discards a service probe's payload when the SSL layer defers the write.*

## Added: a behavioural test

`tests/network/ssl-write-defer-behaviour.sh` forces `SSL_write()` to actually defer — a payload larger than the socket buffers, sent to a TLS peer that pauses before reading — and asserts every byte arrives.

This is the case the source test cannot reach: both versions "send" the data and neither reports an error, so only the peer knows how many bytes it really got.

The payload size is empirical, not arbitrary. Swept against a build without the fix: **1MB was absorbed by the buffers and never deferred at all — the test passed on the bug**. 4MB and 16MB truncated. The default is 8MB, and the reasoning is recorded in the test so the number is not magic. Verified failing without the fix and passing with it.

---
### Where this sits

`main` → #453 → **#454 (here)** → the dialogue work.

#488 builds on this. Merging #453 retargets this onto `main`.

### DIFF
--- a/tests/lib/tls-slow-reader.c
+++ b/tests/lib/tls-slow-reader.c
@@ -1,0 +1,108 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later                                  */
+/*
+ * tests/lib/tls-slow-reader.c
+ *
+ * A TLS server that completes the handshake and then does NOT read for a
+ * moment, so a client sending more than the socket buffers hold gets
+ * SSL_ERROR_WANT_WRITE part-way through. After the pause it drains
+ * everything and reports the byte count.
+ *
+ * That count is the whole point: a probe that retries the deferred write
+ * delivers all of it, one that drops the remainder delivers less, and
+ * neither reports an error -- so only the peer can tell them apart.
+ *
+ *   argv[1] certificate   argv[2] key   argv[3] output file
+ *
+ * Prints its port on stdout. Writes the byte count, or "ERR ...", to the
+ * output file.
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <sys/select.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+#include <openssl/ssl.h>
+#include <openssl/err.h>
+
+int main(int argc, char *argv[])
+{
+	int sock, c;
+	struct sockaddr_in addr;
+	socklen_t alen = sizeof(addr);
+	SSL_CTX *ctx;
+	SSL *ssl;
+	FILE *out;
+	char buf[65536];
+	long total = 0;
+	int rcvbuf = 8192;
+	fd_set rfds;
+	struct timeval tv;
+
+	if (argc < 4) { fprintf(stderr, "usage: %s CERT KEY OUTFILE\n", argv[0]); return 2; }
+
+	SSL_library_init();
+	SSL_load_error_strings();
+	ctx = SSL_CTX_new(SSLv23_server_method());
+	if (!ctx) { fprintf(stderr, "SSL_CTX_new failed\n"); return 1; }
+	if (SSL_CTX_use_certificate_file(ctx, argv[1], SSL_FILETYPE_PEM) <= 0 ||
+	    SSL_CTX_use_PrivateKey_file(ctx, argv[2], SSL_FILETYPE_PEM) <= 0) {
+		fprintf(stderr, "cannot load certificate/key\n");
+		return 1;
+	}
+
+	sock = socket(AF_INET, SOCK_STREAM, 0);
+	if (sock < 0) { perror("socket"); return 1; }
+	/* A small receive buffer makes the sender stall sooner and more reliably. */
+	setsockopt(sock, SOL_SOCKET, SO_RCVBUF, &rcvbuf, sizeof(rcvbuf));
+	memset(&addr, 0, sizeof(addr));
+	addr.sin_family = AF_INET;
+	addr.sin_addr.s_addr = inet_addr("127.0.0.1");
+	if (bind(sock, (struct sockaddr *)&addr, sizeof(addr)) < 0) { perror("bind"); return 1; }
+	if (getsockname(sock, (struct sockaddr *)&addr, &alen) < 0) { perror("getsockname"); return 1; }
+	if (listen(sock, 4) < 0) { perror("listen"); return 1; }
+	printf("%d\n", ntohs(addr.sin_port));
+	fflush(stdout);
+
+	tv.tv_sec = 25; tv.tv_usec = 0;
+	FD_ZERO(&rfds); FD_SET(sock, &rfds);
+	if (select(sock + 1, &rfds, NULL, NULL, &tv) <= 0) return 0;
+	c = accept(sock, NULL, NULL);
+	if (c < 0) return 1;
+
+	out = fopen(argv[3], "w");
+	if (!out) { perror("fopen"); return 1; }
+
+	ssl = SSL_new(ctx);
+	SSL_set_fd(ssl, c);
+	if (SSL_accept(ssl) <= 0) {
+		fprintf(out, "ERR handshake failed\n");
+		fclose(out); return 0;
+	}
+
+	/* Do NOT read yet: let the sender fill the pipe and hit WANT_WRITE. */
+	sleep(2);
+
+	for (;;) {
+		int n;
+
+		tv.tv_sec = 8; tv.tv_usec = 0;
+		FD_ZERO(&rfds); FD_SET(c, &rfds);
+		if (!SSL_pending(ssl) && (select(c + 1, &rfds, NULL, NULL, &tv) <= 0)) break;
+		n = SSL_read(ssl, buf, sizeof(buf));
+		if (n <= 0) break;
+		total += n;
+	}
+
+	SSL_write(ssl, "OK\r\n", 4);
+	fprintf(out, "%ld\n", total);
+	fclose(out);
+	SSL_shutdown(ssl);
+	SSL_free(ssl);
+	close(c); close(sock);
+	SSL_CTX_free(ctx);
+	return 0;
+}
--- a/tests/network/ssl-write-defer-behaviour.sh
+++ b/tests/network/ssl-write-defer-behaviour.sh
@@ -1,0 +1,87 @@
+#!/usr/bin/env bash
+#
+# The companion to ssl-write-retry.sh, which reads the source. This one
+# makes SSL_write() actually defer and checks whether the bytes arrive.
+#
+# Forcing the condition needs two things at once: a payload larger than the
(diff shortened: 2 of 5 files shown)


bonomani:
Housekeeping for whoever merges this: #453 and #454 touch adjacent lines and will conflict, trivially.

```
xymonnet/contest.h    int sendagain;          vs   int sslwantwrite;
xymonnet/contest.c    newtest->sendagain = 0; vs   newtest->sslwantwrite = 0;
```

Both resolve as "keep both lines" — the two fields are independent and nothing else overlaps. Verified with a trial merge; those are the only two hunks.

Merge them in either order. I'll rebase whichever lands second, just say the word.

If order is up to you, I'd take #453 before this one. The known limitation noted above — a write deferred with `WANT_READ` still waits on writability, so it re-attempts until the cutoff rather than blocking — is closed by the register-by-direction machinery #453 adds. With #453 in first, that becomes a small follow-up instead of new plumbing here, and this branch stays a single change.

#455 is independent of both — it only touches `protocols.cfg` and merges clean with either.

bonomani:
On whether this is worth merging given the bug is latent — a fair question, and the answer is not "because it might bite one day".

The code reads as a defect whether or not it is reachable. `socket_write()` returning a bare `0` gives the caller no way to tell "deferred" from "nothing to send", and the call site silently drops the buffer. That is what anyone auditing this function finds, cold — it is how this was found in the first place. Leaving it means the next reader re-derives the same state machine, reaches the same conclusion, and writes it up again.

The test is the part that pays for itself. It records the reasoning, not just the outcome:

- why the retry is scoped to `WANT_WRITE` (clearing `readpending` puts the fd in `writefds`, which is what `WANT_WRITE` waits for; `WANT_READ` there would busy-wait, and moving it to `readfds` would land in the read branch and consume application data instead),
- why **both** guards in `socket_shutdown()` are load-bearing (`sslrunning` outside, because `setup_ssl()`'s failure paths free without clearing; the pointer checks inside, because `SSLSETUP_PENDING` is `-1` and `ssldata` is NULL before `setup_ssl()` runs).

I got the second one wrong first time — a pointer-only guard, which turned a NULL-deref hazard into a double free. That is now pinned by the test rather than left as a comment somebody can tidy away.

None of which is an argument that it is urgent. It is not: nothing shipped can reach it, and #453 does not depend on it. It is an argument that closing it is the one option that costs something.

bonomani:
**For a reviewer — what to check and how.**

- **Resolves:** three things — a dropped SSL write (`socket_write()` discarded the payload on `WANT_WRITE`), an **SSL-object leak on handshake timeouts** (the old `!readpending` arm closed the fd without `socket_shutdown()`, so `SSL`/`SSL_CTX` were never freed), and `SSL_shutdown(NULL)`/double-free hardening.
- **Why we need it:** the **leak is the reachable one** — a poller that regularly times out TLS handshakes (a down/flaky SSL service) creeps in memory over time. The write-retry is latent today (probes send a few bytes) but becomes essential once multi-step SSL dialogues send real commands.
- **How to review:** read the hoisted `socket_shutdown()` (both timeout arms now free SSL), the inner `NULL` guards + pointer-clearing (load-bearing: `SSLSETUP_PENDING == -1` makes `sslrunning` truthy before `ssldata` exists), and the `sendagain` branch leaving `sendtxt`/`sendlen` intact. Then **soak it**: run with a down SSL target for a day+ and watch **RSS — it should stay flat**, with no crashes.

Central-select-loop code, not unit-testable behaviourally — the real check is *run it a while and confirm nothing breaks* (colours same, memory flat).

Also: this is **foundation for the dialogue work** (#488) — a conversation sends real commands over TLS, which exercises exactly this write path, so this wants to land first.

PR #455, opened 2026-08-27
Stop the SMTP probes speaking before the server greets them

### LINKED ISSUE #450: smtp/smtps probe pipelines EHLO+QUIT: Postfix >= 3.9 rejects every poll with 554
## Summary

The `smtp`/`smtps` tests send their `send` string as a single write containing two SMTP
commands, and `contest.c` writes **before** it reads — so the probe never waits for the
server's `220`. Postfix 3.9 and later enable `smtpd_forbid_unauth_pipelining` by default and
reject the session with `554 5.5.0 Error: SMTP protocol synchronization`.

This happens on **100% of polls** against Postfix >= 3.9. Latency only decides whether
xymonnet notices:

| timing | probe reads | xymonnet reports | server session |
|---|---|---|---|
| write beats the greeting (loopback, low RTT) | `554` only | **not ok** (false alarm) | rejected |
| greeting wins (typical WAN, and all `smtps`) | `220` then `554` | **green** | rejected |

The failing colour is **yellow** by default, not red: the expect-check path sets
`color = respcheck_color`, and `respcheck_color = COL_YELLOW` (`xymonnet.c:104`).
`--checkresponse=COLOR` overrides it; a bare `--checkresponse` leaves it yellow.

In both cases the server logs a violation, never processes the `EHLO` (`ehlo=0/1`), and
drops the connection. On most installations this is therefore silent: the check reports
green while every probe is rejected, and the mail server's log fills at the poll interval
indefinitely.

Present in 4.3.30 and current `main` (identical code).

## Root cause

`xymonnet/contest.c` always writes before it reads: `socket_write()` is called at
`contest.c:1315` in the writable branch, while the banner is only read at `contest.c:1373`
on a later `select()` pass. There is no read-then-write mode, so *any* `send` string in
`protocols.cfg` is transmitted before the server's greeting.

`xymonnet/protocols.cfg`:

```
[smtp]
   send "ehlo xymonnet\r\nquit\r\n"
   expect "220"
   options banner
   port 25
```

This violates RFC 5321 (wait for the greeting) and RFC 2920 (no pipelining before the
server announces `PIPELINING`).

## Reproduction

Postfix 3.10.13 (Debian 13), stock configuration — nothing set by hand:

```
$ postconf -h smtpd_forbid_unauth_pipelining
yes
```

**Greeting wins (WAN timing, and always the case for `smtps`)** — 10/10 on plain `smtp`,
10/10 on `smtps` (TLS 1.3, wrappermode):

```
'220 test.local ESMTP Postfix (Debian)\r\n554 5.5.0 Error: SMTP protocol synchronization\r\n'
```

```
postfix/smtpd: improper command pipelining after EHLO from ...: quit\r\n
postfix/smtpd: disconnect from ... ehlo=0/1 commands=0/1
```

xymonnet reports **green** here, because it matched the `220` it received first.

**Write beats the greeting (loopback)** — real `xymonnet` binary,
`hosts.cfg` = `127.0.0.1  smtptest  # smtp`, run with `--checkresponse`, 5/5 not ok:

```
Service smtp on smtptest is not OK : Unexpected service response
554 5.5.0 Error: SMTP protocol synchronization
```

```
postfix/smtpd: improper command pipelining after CONNECT from localhost[127.0.0.1]: ehlo xymonnet\r\nquit\r\n
```

**With the `send` line removed** — 5/5 green, no violation logged:

```
Service smtp on smtptest is OK (up)
220 test.local ESMTP Postfix (Debian)
```

## Confirmed in production

A Postfix 3.9+ server monitored by Xymon logs this at the poll interval, indefinitely
(client redacted):

```
22:32:15 mx postfix/smtpd: connect from <monitor>
22:32:15 mx postfix/smtpd: improper command pipelining after EHLO from <monitor>: quit\r\n
22:32:15 mx postfix/smtpd: disconnect from <monitor> ehlo=0/1 commands=0/1
22:37:16 mx postfix/smtpd: connect from <monitor>
22:37:16 mx postfix/smtpd: improper command pipelining after EHLO from <monitor>: quit\r\n
22:37:16 mx postfix/smtpd: disconnect from <monitor> ehlo=0/1 commands=0/1
```

Five minutes apart — the poll cadence.

The visible failure variant also occurs on the project's own instance. `www.xymon.com`, `smtp`
column:

```
Wed Aug 26 22:42:17 2026 smtp NOT ok
Service smtp on www.xymon.com is not OK : Unexpected service response
554 5.5.0 Error: SMTP protocol synchronization
Seconds: 0.030265000
```

It recovered by itself — the same column read `smtp ok` at 22:57:24 and 23:02:21.

## Why this is usually invisible to xymonnet

Three things combine so that the probe rarely notices the rejection it is causing.

**1. xymonnet reads once, then closes.** `contest.c:1416` — for a non-HTTP service
`wantmoredata` stays 0 after the first successful read, and `contest.c:1424` tears the
socket down. Whatever landed in that single read *is* the banner.

**2. The expect string is matched against the start of that buffer only** —
`contest.c:1513`:

```c
return (memcmp(test->svcinfo->exptext+test->svcinfo->expofs, test->banner, compbytes) == 0);
```

"Starts with `220`" is the entire test. Anything after it, including a `554`, is never
examined.

**3. `SSL_read()` returns at most one record.** Postfix writes the `220` and the `554` as
separate writes, so they become separate TLS records. Over TLS the single read therefore
contains only the `220` and the rejection is never seen. On plain TCP one `read()` can pick
up both at once — which is why the same exchange looks different on 25 and on 465.

So one identical rejected session produces three different verdicts, decided only by what
lands in that first read:

| first read contains | xymonnet reports |
|---|---|
| `220 ...` | **green** — the `554` is never read |
| `554 ...` (rejected before any greeting) | **not ok** — "Unexpected service response" |
| nothing — teardown beat the read | **not ok** — `banner == NULL`, `tcp_got_expected()` returns 0 at `contest.c:1502` |

The third row is a live race: Postfix disconnects immediately after the `554`. Over TLS an
abrupt close makes `SSL_read()` fail outright rather than return a short read, and
`socket_read()` swallows that into a `dbgprintf` (`contest.c:874`), so the column goes red
with no diagnostic message at all.

`imaps` and `pop3s` share this exposure — they use `options ssl` without `TCP_HTTP`, so they
too are judged on whatever the first record happened to contain. `https` escapes it because
`TCP_HTTP` sets `wantmoredata` and accumulates across reads.

Both confirmed by serving a response deliberately split across two TLS records. `imaps`,
greeting sent as `* ` then `OK IMAP4rev1 ready`:

```
Service imaps on splitimaps is not OK : Unexpected service response
```

It read the first record only — 2 bytes, short of the 4-byte `* OK` — and closed. The same
treatment applied to an `https` status line (`HTTP/1.1 2` then `00 OK\r\n...`) is
reassembled correctly:

```
HTTP/1.1 200 OK
```

So any TLS peer that splits its greeting across records, or tears the connection down
early, reads as "Unexpected service response" on `imaps`/`pop3s`/`smtps` while the same
split is harmless on `https`.

## Second symptom: false *green* under postscreen

With postscreen enabled and `postscreen_greet_action = enforce`, postscreen flags the probe:

```
postfix/postscreen: PREGREET 21 after 0 from [...]: ehlo xymonnet\r\nquit\r\n
postfix/postscreen: COMMAND PIPELINING from [...] after ehlo: quit\r\n
```

but serves an emulated session, so the probe reads a normal `220-`/`250-`/`221 Bye` dialogue
and reports **green**, 4/4. The monitor reports the service healthy while postscreen has
classified that client as a pregreeting spambot and would block it from delivering real mail.

## Conditions

The red variant only surfaces with `--checkresponse`; the expect string is otherwise never
enforced (`xymonnet.c:1597`). The protocol violation and the server-side rejection happen
regardless of that flag.

## Same pattern elsewhere

`[pop3s]` (`send "quit\r\n"`) and `[imaps]` (`send "ABC123 LOGOUT\r\n"`) also write before
the greeting. No mainstream POP/IMAP server enforces a synchronization check today, so they
are not affected in practice — but the ordering is the same.

## Workaround

Drop the `send` line so the probe only grabs the banner — the shape upstream already uses
for `[telnet]`:

```
[smtp]
   expect "220"
   options banner
   port 25

[smtps]
   expect "220"
   options ssl,banner
```

This costs nothing in coverage: against Postfix >= 3.9 the `EHLO` is already rejected
unprocessed on every poll (`ehlo=0/1`), so there is no EHLO check to lose. The server logs
`lost connection after CONNECT` instead of a pipelining violation.

## Suggested fix

`contest.c` should read the banner before writing when `options banner` is set, so `send`
strings go out after the greeting. Configuration alone cannot fix this — the
write-before-read ordering is structural.

bonomani:
Fixes #450.

`do_tcp_tests()` writes before it reads — `socket_write()` runs in the writable branch, the banner is read on a later pass — so any `send` goes out **ahead of the server's 220**. The SMTP probes sent two commands in one write, ahead of the greeting and before `PIPELINING` had been announced. That violates RFC 5321/2920.

Postfix 3.9 turned `smtpd_forbid_unauth_pipelining` on by default and answers `554 5.5.0 Error: SMTP protocol synchronization`, so **every poll of a current Postfix is rejected**.

It usually goes unnoticed: xymonnet reads one record and matches `expect` against the *start* of it, so when the 220 wins the race it reports green while the session was refused. When the write wins instead, the column goes yellow with "Unexpected service response".

### Two places define these probes

| file | entries | had |
|---|---|---|
| `xymonnet/protocols.cfg` | `smtp`, `smtps`, `submission\|msa` | `ehlo xymonnet\r\nquit\r\n` |
| `lib/netservices.c` | `smtp`, `smtps` | `mail\r\nquit\r\n` |

`lib/netservices.c` holds a compiled-in `default_svcinfo[]` used whenever `protocols.cfg` cannot be opened. Same defect, different spelling — also a pregreet, also two commands in one write. Fixing only the config would have left a violating probe one unreadable file away.

That fallback's send existed to stop sendmail logging `NOQUEUE` for a connection that greets and leaves. Removing it trades that log line for protocol correctness — the same trade the shipped config now makes, where the server logs `lost connection after CONNECT` instead. Sites with fail2ban rules counting those will want to allow their monitor.

### Nothing is lost from the verdict

`tcp_got_expected()` compares `expect` against the **start** of the banner, so the EHLO's reply was never evaluated.

Two observables do change, neither affecting pass/fail: the banner shown in the status message no longer carries EHLO/QUIT replies that happened to land in the same read, and `tcp_stats_written`/`read` shrink accordingly. Certificate collection for `smtps` is unaffected — it happens in `setup_ssl()` before any application send — and the reported `Seconds:` comes from the connection and handshake duration, not the banner exchange.

### Verified

Postfix 3.10 (Debian 13, stock config — no flags set), same build, same server either way:

| | xymonnet | Postfix log |
|---|---|---|
| before | `not OK : Unexpected service response` / `554 5.5.0 Error: SMTP protocol synchronization` | `improper command pipelining after CONNECT: ehlo xymonnet\r\nquit\r\n` |
| after | `OK (up)` / `220 test.local ESMTP Postfix (Debian)` | **no violation logged** |

`./tests/testsuite` unchanged — the two failures (`fs-sentinel-copies.sh`, `loadhosts-aliasing.sh`) are present before and after and unrelated.

### On the test

Assertions are line-anchored, so a commented-out directive cannot satisfy them. It covers both definition sites, and lists `msa` separately from `submission` even though they share one block today — if that block is ever split, the alias must not be able to reacquire a send while the test keeps passing.

Mutation-checked; each of these is caught: re-adding a send to `[smtp]`, to `[submission|msa]`, to the `smtp` fallback row, or to the `smtps` fallback row.

### Scope

This removes the violation; it does not remove the ordering that caused it. Other probes (`ftp`, `ssh`, `pop3`, `imap`, `nntp`, `clamd`, …) still write before the greeting — no mainstream server for those enforces a synchronization rule today. Making `send` safe in general means teaching `contest.c` to read the banner first: a larger behavioural change across twenty-odd services, and its own change.

Independent of #453 and #454 — different files, merges clean with either.

Not added to `Changes` per CONTRIBUTING. Suggested wording: *the smtp, smtps and submission probes no longer violate SMTP pipelining rules, which current Postfix rejects.*

## Added: a behavioural test

`tests/network/smtp-pregreet-behaviour.sh` runs the probe against a peer that answers `554 5.5.0 Error: SMTP protocol synchronization` to anything arriving before its greeting — as Postfix does with `smtpd_forbid_unauth_pipelining=yes`, its default since 3.9 — and records whether the probe spoke first.

It copies the tree's **shipped** `protocols.cfg` and redirects only the `[smtp]` port, so it covers the real entry rather than a convenient local one. It also fails a send carrying more than one command.

Verified both ways: passes on this branch, and fails with `PREGREET b'ehlo xymonnet\r\nquit\r\n'` when the old ordering is restored.

Ordering in `protocols.cfg` is only half the fix — the driver has to honour it, and that is what this checks.

### DIFF
--- a/lib/netservices.c
+++ b/lib/netservices.c
@@ -38,7 +38,7 @@ static svcinfo_t default_svcinfo[] = {
 	{ "ssh1",    NULL,                0,                  "SSH",	0, 0,	(TCP_GET_BANNER), 22 },
 	{ "ssh2",    NULL,                0,                  "SSH",	0, 0,	(TCP_GET_BANNER), 22 },
 	{ "telnet",  NULL,                0,                  NULL,	0, 0,	(TCP_GET_BANNER|TCP_TELNET), 23 },
-	{ "smtp",    "mail\r\nquit\r\n",  0,                  "220",	0, 0,	(TCP_GET_BANNER), 25 }, /* Send "MAIL" to avoid sendmail NOQUEUE logs */
+	{ "smtp",    NULL,                0,                  "220",	0, 0,	(TCP_GET_BANNER), 25 }, /* No send: speaking before the 220 is an RFC 5321/2920 violation */
 	{ "pop",     "quit\r\n",          0,                  "+OK",	0, 0,	(TCP_GET_BANNER), 110 },
 	{ "pop2",    "quit\r\n",          0,                  "+OK",	0, 0,	(TCP_GET_BANNER), 109 },
 	{ "pop-2",   "quit\r\n",          0,                  "+OK",	0, 0,	(TCP_GET_BANNER), 109 },
@@ -54,7 +54,7 @@ static svcinfo_t default_svcinfo[] = {
 	{ "bbd",     "dummy",             0,                  NULL,	0, 0,	(0), 1984 },
 	{ "ftps",    "quit\r\n",          0,                  "220",	0, 0,	(TCP_GET_BANNER|TCP_SSL), 990 },
 	{ "telnets", NULL,                0,                  NULL, 	0, 0,	(TCP_GET_BANNER|TCP_TELNET|TCP_SSL), 992 },
-	{ "smtps",   "mail\r\nquit\r\n",  0,                  "220",	0, 0,	(TCP_GET_BANNER|TCP_SSL), 0 }, /* Non-standard - IANA */
+	{ "smtps",   NULL,                0,                  "220",	0, 0,	(TCP_GET_BANNER|TCP_SSL), 0 }, /* Non-standard port - IANA. No send, as for smtp */
 	{ "pop3s",   "quit\r\n",          0,                  "+OK",	0, 0,	(TCP_GET_BANNER|TCP_SSL), 995 },
 	{ "imaps",   "ABC123 LOGOUT\r\n", 0,                  "* OK",	0, 0,	(TCP_GET_BANNER|TCP_SSL), 993 },
 	{ "nntps",   "quit\r\n",          0,                  "200",	0, 0,	(TCP_GET_BANNER|TCP_SSL), 563 },
--- a/tests/lib/smtp-greet-peer.c
+++ b/tests/lib/smtp-greet-peer.c
@@ -1,0 +1,102 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later                                  */
+/*
+ * tests/lib/smtp-greet-peer.c
+ *
+ * An SMTP server that records whether the client spoke before being
+ * greeted, and answers a pregreeter the way Postfix >= 3.9 does with
+ * smtpd_forbid_unauth_pipelining=yes (its default): 554, then hang up.
+ *
+ * Prints its port on stdout, then one word per line to the verdict file:
+ *
+ *   polite | PREGREET      -- did anything arrive before our 220?
+ *   PIPELINED              -- more than one command in a single write
+ *   commands=N
+ *
+ * Waiting half a second before greeting is what makes the pregreet
+ * observable: a client that writes on connect has already done so.
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <sys/select.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+
+static int crlfs(const char *b, int n)
+{
+	int i, c = 0;
+	for (i = 0; i + 1 < n; i++) if ((b[i] == '\r') && (b[i+1] == '\n')) c++;
+	return c;
+}
+
+int main(int argc, char *argv[])
+{
+	int sock, c, n, ncmd = 0;
+	struct sockaddr_in addr;
+	socklen_t alen = sizeof(addr);
+	char buf[2048];
+	FILE *out;
+	fd_set rfds;
+	struct timeval tv;
+
+	if (argc < 2) { fprintf(stderr, "usage: %s VERDICTFILE\n", argv[0]); return 2; }
+
+	sock = socket(AF_INET, SOCK_STREAM, 0);
+	if (sock < 0) { perror("socket"); return 1; }
+	memset(&addr, 0, sizeof(addr));
+	addr.sin_family = AF_INET;
+	addr.sin_addr.s_addr = inet_addr("127.0.0.1");
+	if (bind(sock, (struct sockaddr *)&addr, sizeof(addr)) < 0) { perror("bind"); return 1; }
+	if (getsockname(sock, (struct sockaddr *)&addr, &alen) < 0) { perror("getsockname"); return 1; }
+	if (listen(sock, 4) < 0) { perror("listen"); return 1; }
+	printf("%d\n", ntohs(addr.sin_port));
+	fflush(stdout);
+
+	tv.tv_sec = 20; tv.tv_usec = 0;
+	FD_ZERO(&rfds); FD_SET(sock, &rfds);
+	if (select(sock + 1, &rfds, NULL, NULL, &tv) <= 0) return 0;   /* nobody came */
+	c = accept(sock, NULL, NULL);
+	if (c < 0) return 1;
+
+	out = fopen(argv[1], "w");
+	if (!out) { perror("fopen"); return 1; }
+
+	/* Anything readable now arrived BEFORE our greeting. */
+	tv.tv_sec = 0; tv.tv_usec = 500000;
+	FD_ZERO(&rfds); FD_SET(c, &rfds);
+	if (select(c + 1, &rfds, NULL, NULL, &tv) > 0) {
+		n = recv(c, buf, sizeof(buf) - 1, 0);
+		if (n > 0) {
+			buf[n] = '\0';
+			fprintf(out, "PREGREET %s\n", buf);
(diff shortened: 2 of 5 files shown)


bonomani:
One thing worth flagging for anyone testing this on an existing install: **it will not take effect on upgrade.**

`make install` merges `protocols.cfg` with `build/merge-sects`, and that keeps the user's version of any section that already exists — it only adds sections that are missing:

```c
if (ewalk) {                          /* section already in the user's file */
        fprintf(curfd, "%s", ewalk->val);   /* keep THEIR version */
```

Verified rather than inferred: with an old-style `[smtp]` installed, running merge-sects against this branch's template leaves it byte-for-byte unchanged.

So on an existing installation the pregreet stays until `[smtp]`, `[smtps]` and `[submission|msa]` are edited by hand (or those sections deleted so the merge re-adds them). Fresh installs get the new file. The `lib/netservices.c` fallback rows in this PR only apply when `protocols.cfg` is absent entirely.

Not arguing for changing merge-sects here — keeping local edits is the right default — but the upgrade path should probably be called out in the release notes, otherwise this reads as fixed when it isn't yet on any existing box.

feld:
I agree with this. The old way was tolerated and it wasn't until recent that MTAs started to get strict with disobeying PIPELINING or throwing errors if the client attempts to pipeline commands before the PIPELINING capability was advertised

feld (approved):

PR #456, opened 2026-08-27
xymonnet: per-host depth, and the documentation for the dialogue stack

bonomani:
**Stack 5/5 for #457.** Base: #464. The four below it are #461 → #462 → #463 → #464.

Everything that only makes sense once the whole dialogue exists.

## The shipped SMTP entries deepen

`[smtp]`, `[smtps]` and `[submission|msa]` matched the greeting and stopped. They now EHLO, read the multi-line `250` through its terminator, and QUIT — so a server that answers `220` and then rejects every command goes red, which is the failure #457 exists to catch.

This is **the one change in the stack that alters what an existing installation reports**, and it is here rather than in #461 because `ok=` arrives in this slice: `smtp:ok=greeting` is what lets a site keep banner depth host by host while the definition deepens. Deepening before that hatch exists would leave a site no way out but to edit `protocols.cfg`.

## Per-host depth and credentials

`protocols.cfg` is one file for the whole installation: a `[pop3]` definition applies to every host carrying the test, so deepening a shipped entry changes what every site is alerted on. Two tags in hosts.cfg let a host differ from the definition without copying it:

```
   smtp:ok=greeting        check only as far as that state
   smtp:cred=acctb         bind this credentials.cfg entry
```

Neither derives a column, unlike a port — a site running mixed depths would otherwise get one column per depth and no comparable history.

## Fixes to the slices below

- the report names the alternative that matched, not the first one written
- a `timeout(N) -> fail` edge no longer ends the reachability walk, which had called every state after it unreachable

## Documentation

`xymonnet/protocols.cfg.5` is the spec: the grammar, and every decision an operator can observe — what a state consumes and leaves, the per-step budget and the idle clock, the 32 KB cap, the STARTTLS refusal over buffered data, the certificate off the upgraded session, that entering a state re-runs its action, `ok=`/`cred=`, and every complaint the parser makes.

`docs/design/protocol-dialogues.md` is for contributors: the reasoning a manual has no room for, the prior art, and the designs that were tried and rejected — a regex-matching `expect`, a separate `capture` step, deferring an ambiguous group, mermaid as the config format, and eight more. It is in the tree rather than in this body because a pull request that is closed takes its body with it, and the point of that record is to outlive the attempt.

## Where a message ends

TCP has no message boundary; the protocol states one, and there are only three ways it ever does. The stack below this had two of them -- a terminator (`until`) and the close (`eof`) -- so the protocols that send a count could be probed no further than their first bytes. This slice adds the third:

```
[dns-tcp]
   framing length(2, big)     ← every message is preceded by a 2-byte count
   port 53
   …
      expect "\x12\x34"      ← matches the start of the MESSAGE
```

- **`framing length(WIDTH, big|little)`** — an entry attribute, because framing is a property of the protocol and not of one reply. The driver assembles a whole message before any `expect` looks at it, a match consumes the message and its count together, and a `0x0A` inside a payload is data rather than the end of anything. `until` and `bytes(N)` are refused under it: the count has already answered that question.
- **`expect bytes(N)`** — a fixed-size frame for handshakes with no prefix at all, such as the RFB version string. It may not share a state with a literal, since one is decided by length and the other by content.
- **`framing terminator "SEQ"`** — a message ends at a byte sequence wherever it falls: a NUL, a blank line, a sentinel of your own. `until` cannot say this, because it compares the start of a *line*.
- **`ignore "PREFIX"`** — a message the server sends unprompted is consumed and the wait continues. IMAP's untagged lines would otherwise be read as the answer that never came, and fail-fast makes that worse rather than better: the sooner an answer is ruled out, the sooner the probe is wrong. The clocks keep their meanings — `idle` is satisfied by noise, `timeout` is not — so a server that says nothing but noise still fails.

Both make LDAP, MySQL, DNS-over-TCP and AMQP reachable past the banner, and neither costs the load-time decidability the literal-only rule was bought with: a length is decidable at every byte, exactly like a prefix.

## Verified

The suites this slice carries, all driving a real probe against the scriptable peer:

- `dialogue-depth.sh` -- the same definition **without** `ok=` must go red, and a peer that greets and then rejects everything must go red *with* it, so `ok=` cannot pass by making the test shallow enough to prove nothing. Two hosts on `cred=accta` and `cred=acctb` must each send their own account.
- `dialogue-starttls-negotiation.sh` -- with plain failures forced to red, a STARTTLS that is refused or never offered must still be **yellow**
- a length-framed message **containing newlines** must be read whole, and the same bytes under `framing line` must *not* pass — otherwise the attribute is doing nothing
- a peer promising nine bytes and sending three must **not** report OK
- two messages in one write must stay two, the second still there for the state after
- a peer chattering **before the greeting and again before the reply** must go green with `ignore`, and the same peer must go **red without it**, or the line is inert
- a peer with no noise at all must still pass, or `ignore` is eating replies.
- the certificate check in `dialogue-tls.sh` -- the upgrading peer has **its own certificate**, so the check cannot pass on the implicit handshake next door.

## The open question

Whether the fix needs branching at all — #457 states it, and `docs/design/protocol-dialogues.md` argues both sides. Three independent reviews recommended a reduced linear form, which is roughly #461 alone. The reason for the rest is that one definition is a column across the estate, and where servers differ a linear entry degrades to what they all do.

That is what the stack is for: if the linear form wins, #461 is what ships and the rest is the discussion.

## A caution

Ten bugs on this branch were found by running probes, not by reading code. Four were invisible on plain TCP and fatal with `options ssl`; one broke *legacy* services the feature does not touch; three were state that existed in one representation and did not survive into the next, including an uninitialised field that aborted the probe whenever three dialogue services ran together. Two more were reports that named the wrong line. Whatever lands here needs behavioural tests over TLS.

---

**Grammar.** [`protocols.cfg(5)`](https://github.com/xymon-monitoring/xymon/blob/feat/protocol-dialogue/xymonnet/protocols.cfg.5) is the spec — every directive, and every decision an operator can observe. It installs with xymonnet, so `man 5 protocols.cfg` on a built system says the same thing.

**Design.** [The shape](https://github.com/xymon-monitoring/xymon/blob/feat/protocol-dialogue/docs/design/protocol-dialogues.md#the-shape) · [a worked example with its state diagram](https://github.com/xymon-monitoring/xymon/blob/feat/protocol-dialogue/docs/design/protocol-dialogues.md#a-worked-example) · [branching on a value](https://github.com/xymon-monitoring/xymon/blob/feat/protocol-dialogue/docs/design/protocol-dialogues.md#branching-on-a-value) · [the reasoning](https://github.com/xymon-monitoring/xymon/blob/feat/protocol-dialogue/docs/design/protocol-dialogues.md#the-reasoning-a-manual-has-no-room-for) · [prior art](https://github.com/xymon-monitoring/xymon/blob/feat/protocol-dialogue/docs/design/protocol-dialogues.md#prior-art) · [explorations that were rejected](https://github.com/xymon-monitoring/xymon/blob/feat/protocol-dialogue/docs/design/protocol-dialogues.md#explorations-that-were-rejected) · [the linear form](https://github.com/xymon-monitoring/xymon/blob/feat/protocol-dialogue/docs/design/protocol-dialogues.md#the-linear-form) — this PR adds that file.


feld:
We also don't have any way at this time to test *explicit* TLS where you have to upgrade with a command (STARTTLS as it is sometimes called). This is probably most common with SMTP (25), SMTP Submission (587), and IMAP (143).

If we could make it possible to do a multi-step send/expect **and** be able to do a TLS upgrade **and** capture the certificate automatically -- that would be the holy grail. Right now this is a total blind spot. You can have xymon monitor SMTP on port 25, but you can't make it validate that the TLS on port 25 is working and correct.

bonomani:
Superseded by a stack of ten PRs, so each is one subject and reviewable in a sitting. Same content, verified byte-identical: the tip of the stack reproduces this branch exactly.

Reviewed bottom-up:

| PR | subject |
|---|---|
| #466 | docs: protocols.cfg(5) for multi-step dialogues |
| #467 | per-host depth and credentials (`ok=` / `cred=`) |
| #468 | four corrections to the dialogue engine |
| #469 | docs: the manual and the design note describe the code |
| #470 | the shipped SMTP entries, and one layout for every entry |
| #471 | a message boundary is whatever the protocol says it is |
| #472 | `ignore PREFIX`, and a capture that may not read what it replaces |
| #473 | a bound value is bytes |
| #474 | what a send can compute |
| #475 | framing writes as well as reads, and noise is positional |

The commits are yours, regrouped by replaying deltas in their original order — no rebase and no reordering, so messages and authorship are preserved. Where a feature spanned several commits they are folded into one, with the originals listed in the body.

Suite on the stack tip: 121 passed, 0 failed, 0 skipped.

PR #458, opened 2026-08-28
xymonnet: an https test always sends the URL's host as SNI

bonomani:
**Rescoped** from `xymonnet: send SNI for TCP-TLS services (imaps/smtps/...), on
by default` to an `https`-only change. The TCP-TLS extension is dropped, with it the `testip`
exception and the static wiring guard; that path is left to #487.

## Problem

SNI defaulted to off, so an `https` test handshook without saying which name it
wanted. A server holding a certificate per name answers that with its **default**
certificate — so `sslcert` reported on a certificate that was not the URL's,
labelled with the URL's name. The real certificate could expire unnoticed while
the page stayed green: `Host:` routes the request after the handshake, so only
the certificate belonged to someone else.

## Change

Send the URL's host, always; never an address, which RFC 6066 forbids as a
servername. An address in the URL is therefore how a test asks for no servername,
and `Host:` then carries the same address, so both halves describe that
endpoint's default.

`sni`/`nosni` are not consulted here — they are per *host*, this name is per
*URL*, and one host may carry several URLs. They and `--sni=` stay as they are
for a test that takes its servername from the host entry; none does today, which
`hosts.cfg(5)` now says, and the default is flipped to on for when one does.

## Replacing `nosni`, per hosts.cfg(5)

```
10.0.0.5  www.example.com  # https://www.example.com/ nosni
```

What that line means today, spelled out: check `www.example.com`'s content, **and
report the default vhost's certificate under `www.example.com`'s name**. Nobody
configured the second half; it is what falls out of not sending a name.

So its honest equivalent is two lines:

```
10.0.0.5  www.example.com      # https://www.example.com/ nosslcert
10.0.0.5  default.example.com  # https://default.example.com/
```

The first keeps the content check exactly as it was and drops the certificate
status rather than misattributing it. The second monitors the certificate the old
line was in fact reporting on — under the name it belongs to, where it is checked
correctly and labelled correctly.

Drop the second line if that certificate is not something you want watched. That
is a decision to stop monitoring it, not a simplification: with `nosni` you were
monitoring it whether you meant to or not.

## Test

`tests/network/sni-https-behaviour.sh` — a compiled 127.0.0.1 peer reads the
servername off the ClientHello, driven by the real `xymonnet`:

| case | URL | SNI |
| --- | --- | --- |
| name comes from the URL | `https://www.example.com/` | `www.example.com` |
| an address sends none | `https://127.0.0.1/` | *(none)* |
| `nosni` does not suppress | `https://www.example.com/ nosni` | `www.example.com` |

The host entry is named `host.example.com`, so the first row proves the name came
from the URL rather than from the entry.

All three connect to the same peer on 127.0.0.1. The second needs nothing
further — its URL names the address. The first and third name `www.example.com`,
which resolves nowhere, so the address is supplied separately: a URL's network
part is

```c
/* netloc is [username:password@]hostname[:port][=forcedIP] */     lib/url.c:362
```

and the forced IP is a field of its own, read before the port. The URL therefore
names `www.example.com` for the handshake and the `Host:` header while the socket
goes to the loopback peer. It is needed because an http test resolves the URL's
**own** host rather than column 1, so there is no other way to point a name at a
local peer.

Red without the change — *"expected SNI 'www.example.com', the peer saw ''"* —
green with it. Suite **94 passed, 0 skipped, 0 failed**; `mandoc -Tlint` clean on
the lines touched; HTML regenerated with `build/makehtml.sh`.

**Not verified:** the `snienabled` flip to on has no observable effect in this
release. Once https stops consulting it nothing reads it, so no test can pin it
— it is set so that a test later taking its servername from the host entry
inherits the decision instead of reopening it. `hosts.cfg(5)` and `xymonnet(1)`
both say it is currently without effect.

## Proposed release-note entry

> `xymonnet`: an https test now always sends the URL's host as the TLS
> servername (SNI), and never an IP address. SNI was off by default, so a server
> holding a certificate per name returned its *default* certificate and the
> `sslcert` status described a certificate that was not the URL's. The `sni` and
> `nosni` host tags and `--sni` no longer affect https tests, and the default is
> now on. A host that relied on `nosni` wants `nosslcert` on that line, plus a
> separate entry for the default virtual host if that certificate should still be
> watched — see `hosts.cfg(5)`.

#460 is the other half of making `sslcert` mean its name: this sends the name,
that checks the certificate covers it. #459 is closed: its `sni=NAME` had nothing
left to stack on here.

Refs #505.


bonomani:
Design note on the two conservative choices, so the reasoning is on record:

- **Default-on** is a revert of the 4.3.14 default-off. The only visible change is on SNI-routing servers, where the probe now validates the certificate a real client would receive, so `sslcert` values may move — a correction, and `nosni` / `--sni=off` opt out.
- **`testip` hosts send no SNI on purpose.** `testip` is frequently used with a *label* host name that is not a valid servername, and an IP-literal check cannot tell a label from a real FQDN — so auto-sending it would risk a bogus servername, which is exactly what breaks a strict server. Kept conservative here; explicit opt-in for these hosts is `sni=NAME` in #459.

bonomani:
One interaction worth recording now, while both are still in flight. It does not affect this issue as it stands — nothing here needs to change before merging — but whichever of the two lands second will otherwise ship a silent gap.

## The SNI gate is the service definition, not the connection

```c
/* xymonnet.c */
if (tt && tt->svcinfo && (tt->svcinfo->flags & TCP_SSL) && !t->host->testip) {
        ...
        if (want && !isipliteral) tt->sni = t->host->hostname;
}
```

`TCP_SSL` means "TLS from the first byte". #456 adds `starttls`, which upgrades a **plaintext** connection part-way — those services deliberately carry no `TCP_SSL` flag, because they are not TLS when the socket opens.

So with both in the tree, **submission/587, imap/143 and pop3/110 would send no SNI at all**. Those are exactly the hosted and multi-tenant ports where SNI decides which certificate you get, so the failure this issue exists to fix would persist there, and quietly: the handshake fails and the service reads as listening-but-down with an SSL error, which is the symptom described at the top of this issue.

## This is a repeat of a bug already found in the same file

`socket_read()` had the identical shape, and it broke every upgraded connection:

```c
if (item->svcinfo->flags & TCP_SSL) {   /* the definition */
```
```c
if (item->sslrunning) {                 /* the live session */
```

`socket_write()` already keyed off the live session; `socket_read()` keyed off the flag, so after an upgrade it kept calling `read()` and handed back raw TLS records as the server's reply — 510 bytes of ciphertext compared against `"250"`. Fixed in #456 by dispatching on `sslrunning`.

`TCP_SSL` is not a reliable answer to "is this connection TLS?" anywhere in `contest.c`, and this is the third place to assume it is.

## Suggested direction, for whichever merges second

Decide the servername where the handshake happens rather than from the service definition: `setup_ssl()` runs for both the implicit and the upgraded case, so a servername resolved there covers both with one gate and no coordination between the two changes.

That also removes the `xymonnet.c` → `contest.c` reach-across, which is what made the coupling invisible in the first place.

No action needed here — filing it so the second merge is deliberate rather than a surprise.

bonomani:
Reviewing this again I caught a real regression the default flip would have introduced: `snienabled` is shared, so turning it on also turns on the **HTTP** path — and that path (`httptest.c`) assigned the URL host to `tcptest->sni` with **no IP-literal guard**. So an `https://192.0.2.1/` test, which sent no SNI while the default was off, would start putting `192.0.2.1` in `server_name` — which RFC 6066 forbids and a strict server rejects.

Fixed by applying the same `inet_pton` guard to the HTTP path (per-host `sni`/`nosni` still override; the URL host is only sent when it is not an IP literal). Verified on the wire: `https://127.0.0.1/` now sends no SNI. Added a static-guard assertion that the HTTP path stays guarded, and updated the commit message. Force-pushed.

bonomani:
Conflicts with #487 in `xymonnet/xymonnet.c`.

Both send SNI on TCP-TLS probes, by different designs, so this is a decision about which to keep rather than a rebase: merging either makes the other's approach redundant rather than merely out of date. #487 additionally collides with #488 in `lib/netservices.c`.

bonomani:
## Where `main` stands

The controls already exist, and for every service except HTTP they are inert.

| | on `main` today |
|---|---|
| global | `--sni=on` / `--sni=off`, default **off** (`int snienabled = 0;`) |
| per host | `sni` / `nosni` tags in `hosts.cfg` |
| per service | nothing |
| name sent | always the URL host, no override |

The only code that ever sets the field is `httptest.c`:

```c
if (hinfo && xmh_item(hinfo, XMH_FLAG_SNI))        tcptest->sni = weburl.desturl->host;
else if (hinfo && xmh_item(hinfo, XMH_FLAG_NOSNI)) tcptest->sni = NULL;
else                                               tcptest->sni = (snienabled ? weburl.desturl->host : NULL);
```

`contest.c` then does `if (item->sni) SSL_set_tlsext_host_name(...)`. Nothing sets `sni` for a plain TCP-TLS probe, so imaps, smtps, pop3s and nntps have never sent a server name — whatever `hosts.cfg` says. So this series is not a new config surface; it wires the tags that are already documented through to the probes where they currently do nothing.

Worth deciding on its own: this changes `snienabled` from 0 to 1. `hosts.cfg(5)` records that the default was on in 4.3.13 and turned off in 4.3.14 because some SSL implementations could not complete a handshake carrying SNI. Reversing that is defensible in 2026 and is probably right, but it is a separate judgement from wiring the tags through, and neither this pull request nor the manual page argues it explicitly. If the argument is "the stacks that could not do this are long gone", it belongs in the commit message.

## Against #487

This series and #487 are the same feature on different axes: this one per host, #487 per service through `options sni` in `protocols.cfg`. Neither axis is wrong and neither has the other's, so the feature is complete only with both.

They do not compose as they stand — both assign `tt->sni` at the same point in `xymonnet.c`, so block order decides whether a host's `nosni` survives its service asking for SNI, and #460's `tt->checkname = tt->sni` would verify a name a later block replaced. The IP-literal guard would land twice, by two methods that disagree at the edges. Full analysis in https://github.com/xymon-monitoring/xymon/pull/487#issuecomment-5581409927, not repeated here.

The plan that keeps both pieces of work:

1. merge this series in order — #458, then #459, then #460
2. reduce #487 to what this series lacks: the `sni` option and the `TCP_SNI` flag in `lib/netservices.*`, a `nosni` mirror it does not yet have, its `protocols.cfg.5` text and its test
3. combine the two axes in the decision this series already makes, rather than ranking them:

```c
int svcsni   = (tt->svcinfo->flags & TCP_SNI)   ? 1 : 0;
int svcnosni = (tt->svcinfo->flags & TCP_NOSNI) ? 1 : 0;
int said_no  = (t->host->sni < 0) || svcnosni;
int said_yes = (t->host->sni > 0) || svcsni;
int want     = !said_no && (said_yes || snienabled);
```

A "no" from either axis wins, a "yes" from either raises the default, and neither outranks the other — the axes are orthogonal, so a host flag covering every service on one host and a service option covering every host for one service have no containment between them to rank by.

Nothing else here changes: the host flags, `sni=NAME` and the certificate check stay as they are. Only the line computing `want` moves, and only once #487 is reduced.

*(Corrected: an earlier version of this comment proposed a `want` line in which the host flag won outright.)*


---

A different direction, which would make the name part of the test rather than a host tag, is written up in #505. It does not change the case for merging this series.

PR #459, opened 2026-08-28
xymonnet: add "sni=NAME" to send an explicit TLS servername

bonomani:
Stacked on #458 (base `feat/sni-tcp-tls`) — review/merge that first.

## What & why

#458 sends the **host name** as the SNI servername. That is wrong when the host name can't be the certificate's FQDN:

- a host monitored by **IP** (or `testip`) has no usable name to send (#458 deliberately sends no SNI for it);
- a host whose **identity is a friendly/business name** while the certificate uses a **technical FQDN** — you don't want to rename the host (all its history, alerts, acks are keyed on the name) just to fix SNI.

`NAME:` can't solve this: it changes the *display* name, not the *servername* (which comes from the host name). So this adds **`sni=NAME`** to send an explicit servername while the host name stays the identity.

## Behaviour

- `sni` → SNI on, servername = host name (as in #458)
- `sni=mail.example.com` → SNI on, servername = `mail.example.com` (used even on a `testip`/IP-only host — the operator vouched for it)
- `nosni` → off

`do_tcp_tests()` picks the name (explicit `sni=NAME` if set, else the hostname for a non-`testip` host) and feeds it through **one IP-literal guard**: RFC 6066 forbids an address in SNI, so `sni=1.2.3.4` sends nothing, exactly as an IP-literal hostname does. `NAME` must be a host name.

## Why it's safe

- `sni=NAME` registers as `XMH_SNINAME` with key `"sni="`, which shares the `"sni"` prefix and so also trips the on-flag. Bare `sni` does **not** match the `"sni="` lookup; `nosni` matches neither — existing behaviour is untouched.
- The only two `xmh_item_idx()` callers test solely for `== -1` (known vs unknown tag), so the shared prefix can't misroute anything.
- New enum value appended before `XMH_LAST`, so no existing item index shifts.

## Verification

Built in-tree and run against an SNI-reporting TLS peer:

| hosts.cfg | servername sent |
|-----------|-----------------|
| `… imaps sni=mail.example.com` | `mail.example.com` |
| `… testip imaps sni=mx.example.org` | `mx.example.org` |
| `… imaps sni=203.0.113.9` (IPv4) | *(none)* |
| `… imaps sni=2001:db8::1` (IPv6) | *(none)* |
| `… imaps` (name = FQDN) | the FQDN |
| `… imaps nosni` | *(none)* |

`tests/network/sni-tcp-tls.sh` (static guard) gains block-scoped checks for the tag registration, the `sniname` field, the read, and its use through the guarded assignment. `tests/network/sni-tcp-tls-behaviour.sh` (behavioural, drives the built xymonnet against a ClientHello-reading peer) gains cases that `sni=NAME` is sent — including on a `testip` host — and `sni=<ip>` is not. Both mutation-checked.

## Docs

`common/hosts.cfg.5` documents the new form (kept short) and states `NAME` must be a host name; its HTML is regenerated from source per `CONTRIBUTING.md`.

---

**Draft, and likely to be dropped.** The case this exists for is narrow: a
`testip` host whose configured name is a label rather than the certificate's
FQDN, where the hostname is the wrong servername to send and #458 therefore
sends none. A new `hosts.cfg` key is a lot of permanent surface for that, and
#458 stands on its own without it — `nosni` already covers the "do not send"
direction, which is the one operators actually reach for.

Kept open rather than closed because the need may yet show up in practice, and
because two things currently lean on it: #460 reads `host->sniname` in nine
places, and #458's description points here for the `testip` opt-in. Both want a
small adjustment if this does go.

### DIFF
--- a/common/hosts.cfg.5
+++ b/common/hosts.cfg.5
@@ -722,6 +722,7 @@ test will go red. E.g. to check that your server only uses
 strong encryption (128 bits or better), use "sslbits=128".
 
 .IP "\fBsni\fR"
+.IP "\fBsni=\fRNAME"
 .IP "\fBnosni\fR"
 Enable or disable SNI (Server Name Indication) for SSL tests. SNI is
 on by default and sends the host name as the servername; most TLS
@@ -730,6 +731,13 @@ servers now expect it, and many pick the certificate by it. Use
 name, or set the default globally with the "--sni" option to
 .I xymonnet(1).
 A host tested by its IP address sends no SNI.
+
+Use "sni=NAME" to send a specific servername instead of the host name
+- for a host tested by IP, or whose host name is a display label
+rather than the certificate's FQDN. It also enables SNI. NAME must be
+a host name, not an IP address (RFC 6066 does not allow an address in
+SNI); an IP is logged and no SNI is sent. HTTP tests take the SNI name
+from the URL instead.
 
 
 .IP "\fBDOWNTIME\fR=\fIday\fR:\fIstarttime\fR:\fIendtime\fR[,\fIday\fR:\fIstarttime\fR:\fIendtime\fR]"
--- a/docs/manpages/man5/hosts.cfg.5.html
+++ b/docs/manpages/man5/hosts.cfg.5.html
@@ -791,6 +791,8 @@ Section: File Formats (5)<br/>Updated: Version 4.3.30: 4 Sep 2019<br/><a href="#
   </dd>
   <dt id="sni"><a class="permalink" href="#sni"><b>sni</b></a></dt>
   <dd></dd>
+  <dt id="sni~2"><a class="permalink" href="#sni~2"><b>sni=</b>NAME</a></dt>
+  <dd></dd>
   <dt id="nosni"><a class="permalink" href="#nosni"><b>nosni</b></a></dt>
   <dd>Enable or disable SNI (Server Name Indication) for SSL tests. SNI is on by
       default and sends the host name as the servername; most TLS servers now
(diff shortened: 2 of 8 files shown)


bonomani:
How the name fields relate, for weighing whether this tag is needed:

- **host name** (column 2) — identity + connection target + SNI servername
- **`NAME:`** — display label only; does *not* change the servername
- **`sni=NAME`** (this PR) — servername only, for when the identity must stay a friendly/business name while the certificate uses a technical FQDN
- **several names on one IP** — several host entries, each sending its own SNI (verified: xymonnet does no connection dedup, so each line is its own handshake)

So `NAME:` and `sni=` solve opposite halves — display vs servername. The friendly-identity / technical-FQDN case genuinely needs `sni=`, because `NAME:` only moves the display and the servername would otherwise stay the (wrong) host name.

bonomani:
Reviewing this I noticed the explicit `sni=NAME` bypassed the IP-literal guard that #458 applies to the hostname, so `sni=1.2.3.4` would have put an address in the servername — which RFC 6066 forbids and no server routes on. Reworked the block so both sources (explicit name and hostname) feed **one** IP-literal guard: `sni=<ip>` now sends nothing, while `sni=<name>` (even on a `testip` host) still works. Verified against an SNI-reporting peer (IPv4 and IPv6 literals both suppressed); manpage now states `NAME` must be a host name.

bonomani:
Added a **behavioural** test alongside the static guard, matching the compiled-peer pattern #453/#454 introduced: `tests/lib/tls-sni-peer.c` reads the ClientHello and reports the SNI, and `tests/network/sni-tcp-tls-behaviour.sh` runs the built xymonnet against it to check the servername actually on the wire — default (hostname), `nosni`, `testip`, an IP-literal name, `sni=NAME` (incl. on a testip host), and `sni=<ip>` (suppressed). This exercises the loadhosts prefix-parsing and the IP-literal guard for real, not just at the source level. Mutation-checked (flipping the default off, dropping the send, ignoring `sniname`, or neutering the IP-literal guard each fail it).

bonomani:
Re-reviewed after the IP-literal rework: no correctness issues. The `sni=NAME` override is consistently registered (`XMH_SNINAME` / `sni=`), loaded onto the host, validated against IP literals through the shared guard, and applied to the TCP-TLS tests, with the HTTP path unchanged. Static + behavioural tests cover it.

bonomani:
Closing. The case this existed for -- a host whose configured name is not the name to send -- is a new `hosts.cfg` key for one narrow situation, and #458 no longer gives it anywhere to stand: that PR is now https-only, where the servername comes from the URL and a per-host key is the wrong granularity anyway.

If an explicit servername is wanted for a non-http test later, #505's grammar expresses it without a new key -- `host` is the name, `#connect=` the address -- which is the same separation `sni=NAME` was reaching for.

Nothing depends on this being merged: #460 reads `host->sniname` and would drop those references with it.

Issue #457, opened 2026-08-27
RFC: xymonnet cannot test a protocol conversation, only its first line


bonomani:
Filing this as the problem statement behind #450 and the implementation stacked in the pull requests listed below, because those read as unrelated bugs and they are one.

**Scope.** This issue is the problem only. The shape proposed for it, the decisions taken, the prior art and the explorations that were rejected live in [`docs/design/protocol-dialogues.md`](https://github.com/xymon-monitoring/xymon/blob/dialogue-10-startfeature/docs/design/protocol-dialogues.md) on the tip of the stack; what is implemented is the stack in the table below, #461 being the linear form on its own. Each thing is stated once and amended in one place.

## The problem

A probe can express one string out and one prefix match back:

```
[smtp]
   send "ehlo xymonnet\r\nquit\r\n"
   expect "220"
```

**It speaks first.** SMTP, IMAP, POP3, FTP and NNTP all transmit before the client may say anything. Postfix 3.9 answers a client that speaks first with `554 5.5.0 Error: SMTP protocol synchronization`, so the shipped entries provoke an error and then match the greeting they were going to match anyway.

**It cannot see past the greeting.** A server that answers `220` and then rejects every command is indistinguishable from a healthy one, which is the failure operators actually get.

**It cannot use a value the server sent.** APOP needs `md5(challenge + password)` from the greeting; SMTP/IMAP AUTH needs base64 after a `334`. Neither is expressible, so authenticated probes -- the ones that show the service *works* -- are out of reach.

Underneath, there is no state machine. The driver has to hold a position across `select()` wakeups, choose which fd set the socket joins, accumulate partial replies, and tell "does not match" from "not enough yet". `contest.c` read a banner into a buffer and compared it once. The ordering fix on main corrects the SMTP entries inside that single-shot model; it cannot reach the second or third problem.

## One definition, several depths

`protocols.cfg` is one file for the whole installation: a `[pop3]` definition applies to every host carrying the test. Deepening a shipped entry therefore changes what every site is alerted on -- so the unit being fixed is not a probe against one server, it is a column across the estate. Two consequences follow, and both are constraints on any fix rather than features of one: a site must be able to keep the old, shallow depth host by host while the definition deepens, and a fleet whose servers are configured differently -- some offering APOP, some STARTTLS -- cannot be described by whatever they all happen to do. How each is answered: the first by the per-host depth in #456, the second by branching in #462.

## The implementation

[...]

PR #461, opened 2026-08-29
xymonnet: a protocols.cfg entry is a sequence, not one line each way

bonomani:
A `protocols.cfg` entry could express one string out and one prefix match back, and it sent that string before the server had greeted. SMTP, IMAP, POP3, FTP and NNTP all transmit before the client may speak, and Postfix ≥ 3.9 answers a client that speaks first with `554 5.5.0 Error: SMTP protocol synchronization`.

This slice makes an entry a sequence: `send` and `expect` run in the order they are written, `expect … until …` consumes a multi-line reply through its terminator, and a conversation that buffers 32 KB with nothing matching fails naming the service rather than growing for as long as the server cares to talk. An entry that is a single `send` followed by a single `expect` keeps the old write-on-connect behaviour, so nothing shipped changes.

It fixes the first of the three problems in #457 on its own, and is the part none of the reviews disputed.

**`xymonnet/protocols.cfg` is untouched here** — the shipped entries are deepened in #480, and #477 adds `smtp:ok=greeting` as the per-host way to decline that. But five of them do change code path: the classifier sets `TCP_DIALOGUE` when `first_is_expect || nsend > 1 || nexp > 1`, and `[smtp]`, `[smtps]`, `[submission|msa]`, `[rsync]` and `[svn]` are expect-only, so they run the driver from this slice on. What they send and match is identical; what executes is not. A service that is one `send` followed by one `expect` does keep the old path.

**Verified.** `tests/lib/dialogue-peer.c` is a scriptable peer the suites drive a real probe against.

- `dialogue-verdict.sh` carries the headline: a server that greets `220` and then fails every command was **green before, red after**. Its third row pins the other direction -- a plain send-and-match service must behave exactly as it did, since the driver is opt-in and a regression there reaches services this does not touch.
- `protocol-dialogue.sh` and `dialogue-multiline.sh` cover ordering and `until`; `smtp-no-pregreet.sh` covers the shipped SMTP entries.
- `dialogue-buffer-cap.sh` checks that `MAX_DIALOGUE_BYTES` is real: a peer sending ~50 KB that never terminates its reply must fail **by name** rather than by running out of time, with a ~20 KB reply that does terminate as the control.

---

**Design.** [The shape](https://github.com/xymon-monitoring/xymon/blob/dialogue-09-framedsend/docs/design/protocol-dialogues.md#the-shape) · [the reasoning behind it](https://github.com/xymon-monitoring/xymon/blob/dialogue-09-framedsend/docs/design/protocol-dialogues.md#the-reasoning-a-manual-has-no-room-for) · [what was tried and rejected](https://github.com/xymon-monitoring/xymon/blob/dialogue-09-framedsend/docs/design/protocol-dialogues.md#explorations-that-were-rejected) — in the tree, added by #479, further up the stack.

---
### Stack

**3 of 15** in the dialogue stack: base #454, next #462. The full chain is listed in #457; read bottom-up.

### DIFF
--- a/lib/netservices.c
+++ b/lib/netservices.c
@@ -66,6 +66,68 @@ static svcinfo_t default_svcinfo[] = {
 
 static svcinfo_t *svcinfo = default_svcinfo;
 
+/*
+ * Append one step to a service's dialogue, preserving file order. Every alias
+ * in an [a|b|c] header gets its own copy: the records are freed independently,
+ * so they cannot share a list.
+ */
+static svcstep_t *add_svcstep(svcinfo_t *rec, int type, unsigned char *text, int len)
+{
+	svcstep_t *step, *walk;
+
+	step = (svcstep_t *)calloc(1, sizeof(svcstep_t));
+	step->type = type;
+	step->len  = len;
+	step->text = (unsigned char *)malloc(len + 1);
+	memcpy(step->text, text, len);
+	step->text[len] = '\0';
+
+	if (rec->steps == NULL) rec->steps = step;
+	else {
+		for (walk = rec->steps; (walk->next); walk = walk->next) ;
+		walk->next = step;
+	}
+
+	return step;
+}
+
+
+/*
+ * Where a quoted string ends. Deliberately the same rule getescapestring
+ * uses -- scan to the next '"' -- so the two agree on where the value stops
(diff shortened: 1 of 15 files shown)


bonomani:

PR #462, opened 2026-08-29
xymonnet: named states, bound values and named credentials

bonomani:
A linear dialogue cannot see past the greeting in a useful way, and cannot use a value the server sent. This slice adds named states as edge targets, so a state's alternatives each decide where the dialogue goes next; values bound out of a reply and read by a later step; and `credentials NAME`, which names an entry in `etc/credentials.cfg` instead of writing a secret into a world-readable file.

Together these are what make APOP and AUTH expressible — a probe that shows the service *works*, not that it answers.

**Verified.** `dialogue-values.sh` and `dialogue-validation.sh` drive the peer, which recomputes the digest and the base64 itself, so a probe sending a wrong value fails there rather than passing on a string comparison.

---

**Design.** [A worked example with its state diagram](https://github.com/xymon-monitoring/xymon/blob/dialogue-09-framedsend/docs/design/protocol-dialogues.md#a-worked-example) · [branching on a value](https://github.com/xymon-monitoring/xymon/blob/dialogue-09-framedsend/docs/design/protocol-dialogues.md#branching-on-a-value) · [the shape](https://github.com/xymon-monitoring/xymon/blob/dialogue-09-framedsend/docs/design/protocol-dialogues.md#the-shape) · [why matching is literal, and what was rejected](https://github.com/xymon-monitoring/xymon/blob/dialogue-09-framedsend/docs/design/protocol-dialogues.md#explorations-that-were-rejected) — in the tree, added by #479, further up the stack.

---
### Stack

**4 of 15** in the dialogue stack: base #461, next #463. The full chain is listed in #457; read bottom-up.

### DIFF
--- a/lib/netservices.c
+++ b/lib/netservices.c
@@ -198,6 +198,8 @@ static void free_svcsteps(svcinfo_t *rec)
 		if (st->target)  xfree(st->target);
 		if (st->label)   xfree(st->label);
 		if (st->varname) xfree(st->varname);
+		if (st->user) { memset(st->user, 0, strlen(st->user)); xfree(st->user); }
+		if (st->pass) { memset(st->pass, 0, strlen(st->pass)); xfree(st->pass); }
 		if (st->re)      freeregex((pcre2_code *)st->re);
 		xfree(st);
 		st = next;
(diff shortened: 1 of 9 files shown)


bonomani:

PR #463, opened 2026-08-29
xymonnet: starttls as a dialogue step, and a per-step time budget

bonomani:
`options ssl` means TLS from the first byte, which is not what submission (587), IMAP (143) or SMTP (25) do. `starttls` upgrades in place: everything before the step is plaintext, everything after is encrypted, on the same socket. Reading the certificate off the upgraded session — so 587/143/25 reach the `sslcert` column — comes with the assertions in #478; this slice does the handshake.

If the server has sent anything still unread when the upgrade begins, the test fails rather than proceeding — the handshake has to start on the first byte after the go-ahead, and data arriving before it is the STARTTLS plaintext-injection flaw.

Also here: `timeout N` as a per-step budget, armed against the step the dialogue is on and re-armed when it advances, rather than one deadline for the whole test. It becomes `timeout(N) -> TARGET` in #464, where a budget gains an edge like every other condition.

**Verified.** `dialogue-tls.sh`. Four bugs on this branch were invisible on plain TCP and fatal with `options ssl`, so the TLS suites are the ones that matter.

---

**Design.** [A worked example with its state diagram](https://github.com/xymon-monitoring/xymon/blob/dialogue-09-framedsend/docs/design/protocol-dialogues.md#a-worked-example), which upgrades on 587 · [the reasoning](https://github.com/xymon-monitoring/xymon/blob/dialogue-09-framedsend/docs/design/protocol-dialogues.md#the-reasoning-a-manual-has-no-room-for) — in the tree, added by #479, further up the stack.

---
### Stack

**5 of 15** in the dialogue stack: base #462, next #464. The full chain is listed in #457; read bottom-up.

### DIFF
--- a/lib/netservices.c
+++ b/lib/netservices.c
@@ -404,7 +404,8 @@ static void emit_step(svclist_t *first, svcstep_t *tmpl)
 	for (walk = first; (walk); walk = walk->next) {
 		svcstep_t *st = add_svcstep(walk->rec, tmpl->type, tmpl->text, tmpl->len);
 
-		st->action = tmpl->action;
+		st->action  = tmpl->action;
+		st->seconds = tmpl->seconds;
 		if (tmpl->target)  st->target  = strdup(tmpl->target);
 		if (tmpl->label)   st->label   = strdup(tmpl->label);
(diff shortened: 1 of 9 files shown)


bonomani:

PR #464, opened 2026-08-29
xymonnet: verdicts, refused ambiguity, and the checks a graph can make

bonomani:
What a dialogue reports, and what the parser refuses to accept.

- **Three verdicts.** `-> success|warning|fail`. A failure took `respcheck_color` — one colour for every way of failing — so a *busy* server and a *broken* one reported identically. `warning` is the point: a server refusing an optional capability is not down, and calling it down teaches operators to ignore the column.
- **Overlapping alternatives are refused when the file is read.** Matching is prefix-anchored, so two patterns collide iff one is a prefix of the other — the check is complete rather than heuristic, which is what lets it refuse instead of warn.
- **Three checks only a graph can make**: a dead end that reaches no verdict, a state nothing jumps to, and a wait no clock bounds. `docs/design/protocol-dialogues.md` says why they are worth having and where they mislead.
- `eof`, `start`, `transport`, `always`, and `idle(N)` — the clock that restarts when data arrives, as distinct from the total budget.

**Two changes the title does not cover.** `timeout N` from #463 becomes `timeout(N) -> TARGET`: a budget is a condition like any other and now says where expiry leads, so an entry written against #463 needs updating. And `tests/buildsystem/test-suite-portability.sh` gains a rule that a test sourcing `assert.sh` must declare `#!/usr/bin/env bash` rather than `#!/bin/sh` — unrelated to verdicts, and here only because the suites this slice adds are what tripped it.

**Verified.** `dialogue-terminals.sh`, `dialogue-ambiguity.sh`, `dialogue-eof.sh`, `dialogue-grammar.sh`, `dialogue-idle.sh`, `dialogue-timeout.sh`. The busy and broken peers must come out of one run with different colours; a peer answering one byte per second, never pausing long enough to be idle, must stay green.

`dialogue-buffer-cap.sh` is rewritten here against the grammar this slice adds — a named state, a per-step budget, `-> success`/`-> fail` and `eof` — so the bound is checked the way an entry will actually be written. It arrived with the driver in #461, in the only grammar that existed then.

---

**Design.** [Why matching is literal, and how the graph checks mislead](https://github.com/xymon-monitoring/xymon/blob/dialogue-09-framedsend/docs/design/protocol-dialogues.md#the-reasoning-a-manual-has-no-room-for) · [deferring an ambiguous group, and other rejected designs](https://github.com/xymon-monitoring/xymon/blob/dialogue-09-framedsend/docs/design/protocol-dialogues.md#explorations-that-were-rejected) · [prior art](https://github.com/xymon-monitoring/xymon/blob/dialogue-09-framedsend/docs/design/protocol-dialogues.md#prior-art) — in the tree, added by #479, further up the stack.

---
### Stack

**6 of 15** in the dialogue stack: base #463, next #476. The full chain is listed in #457; read bottom-up.

### DIFF
--- a/lib/netservices.c
+++ b/lib/netservices.c
@@ -899,29 +899,46 @@ char *init_tcp_services(void)
 				if (txt) xfree(txt);
 			}
 		}
-		else if (strncmp(l, "timeout ", 8) == 0) {
+		else if (strncmp(l, "timeout(", 8) == 0) {
 			/*
-			 * How long the wait that follows may take. Without one, a
-			 * step that never matches is bounded only by --timeout, which
-			 * ends the whole test and cannot say which step stalled.
+			 * "timeout(N) -> TARGET": how long this state may take, and
+			 * where to go when it does not. Without one, a state that never
+			 * matches is bounded only by --timeout, which ends the whole
+			 * test and cannot say which state stalled.
 			 *
-			 * The budget covers the write as well as the read: a peer that
-			 * will not accept the send is as stuck as one that will not
-			 * answer it, and the global ceiling still wins either way.
+			 * The budget covers writing the action as well as reading the
+			 * reply: a peer that will not accept the send is as stuck as one
+			 * that will not answer it. --timeout still wins, so per-state
+			 * budgets can never sum past it.
 			 */
-			char *arg = skipwhitespace(l + 8);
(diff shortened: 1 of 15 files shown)


bonomani:

PR #465, opened 2026-08-30
xymonnet: the value layer reaches a real authentication

bonomani:
Stacked on #456. Adds what a `send` can compute, makes a framed connection framed in both directions, and decides `ignore` by position.

## Why

Separately, `framing length(W)` taught the driver to *read* a counted message and left writing one to the file, which cannot do it: the count depends on how long an expanded `${...}` turns out to be. Length-framed protocols could be read from and never spoken to.

## What

**Expansions.** `${sha1:}`…`${sha512:}`, `${rmd160:}` — `lib/digest.c` already implemented all of them. `${hmac-md5:KEY,MESSAGE}`…`${hmac-sha512:}` (RFC 2104). `${hex:}`, `${len:}`, `${unbase64:}`.

The HMAC is checked against the RFC 2202 and RFC 4231 vectors, including the key-longer-than-block branch. The argument split is made on the comma the *file* writes, at brace depth zero, before either half is expanded — so a password containing a comma is not mistaken for the separator.

**`${unbase64:}` is not a convenience.** RFC 2195 sends the CRAM-MD5 challenge in base64 and hashes it *decoded*. It needed a decoder that returns a length and honours `=` padding — `base64decode()` did neither, so a decoded nonce stopped at its first NUL and carried one or two rubbish bytes on the end. It is now a wrapper over the counted one, identical for the text its callers pass it.

**Write-side framing.** Under `length` the count is written in the declared width and byte order; under `terminator` the sequence is appended. `line` is untouched — its entries have always written their own `\r\n`, and adding one would change every send that exists. A constant send too long for its own count is refused when the file is read; an expanding one at the step, which is the one thing reading the file cannot settle.

**`ignore` is positional.** Refusing a prefix that also starts an `expect`, by analogy with two overlapping alternatives, was the wrong analogy — it made the one protocol `ignore` exists for the one protocol it could not express. IMAP's greeting *is* an untagged `* OK` line; the untagged lines after it are noise; they share seven characters. The alternatives are tried first and a message is skipped only when none could be it. That also fixes an ignored record split across reads, which broke the scan without marking anything skipped and failed a healthy server — same block, cannot be fixed apart.

**One list.** `dlg_expansion()` is read both by the check that refuses a mistyped name and by the driver that runs it. They nearly drifted already: `check_expansions()` used `${sha1:...}` as its example of a name that "parses cleanly and expands to nothing".

## Tests

| test | proves |
|---|---|
| `dialogue-cram.sh` | CRAM-MD5 end to end — the peer recomputes the answer with OpenSSL's `HMAC()`; control is the same exchange with the wrong secret |
| `dialogue-expansions.sh` | every function against `openssl` reference values; control is the same HMAC under a second key |
| `dialogue-framed-send.sh` | count in both byte orders and the terminator, asserted as hex on the wire; **control is a line-framed send, which must go out with nothing added**; both refusal paths |
| `dialogue-async.sh` | a real IMAP exchange — the same `* ` prefix taken as the greeting and skipped three lines later |

`dialogue-conventions.sh` did not recognise `framing` or `ignore` as directives, so any entry using either ended there and the rest went unjudged — including everything added here. Widening it exposed a flaw in its own roff extractor (literal blocks ran together). Both fixed, and the widened rule verified against a planted violation it could not previously see.

Two existing tests changed because the work made their assumptions false: `dialogue-validation.sh` used `${sha1:...}` as its *unknown* function, and `protocol-dialogue.sh` guarded against freeing `md5hash()`'s static buffer — that hazard is gone, so the assertion inverts to "every function result must be freed", which guards a real leak.

## Still out of reach

MySQL native auth needs `XOR`; SCRAM needs PBKDF2 with a server-supplied iteration count; `transport udp` is deferred. Different class of addition, scoped separately.

Suite: 121 passed, 0 failed, 0 skipped.


bonomani:
Superseded by a stack of ten PRs, so each is one subject and reviewable in a sitting. Same content, verified byte-identical: the tip of the stack reproduces this branch exactly.

Reviewed bottom-up:

| PR | subject |
|---|---|
| #466 | docs: protocols.cfg(5) for multi-step dialogues |
| #467 | per-host depth and credentials (`ok=` / `cred=`) |
| #468 | four corrections to the dialogue engine |
| #469 | docs: the manual and the design note describe the code |
| #470 | the shipped SMTP entries, and one layout for every entry |
| #471 | a message boundary is whatever the protocol says it is |
| #472 | `ignore PREFIX`, and a capture that may not read what it replaces |
| #473 | a bound value is bytes |
| #474 | what a send can compute |
| #475 | framing writes as well as reads, and noise is positional |

The commits are yours, regrouped by replaying deltas in their original order — no rebase and no reordering, so messages and authorship are preserved. Where a feature spanned several commits they are folded into one, with the originals listed in the body.

Suite on the stack tip: 121 passed, 0 failed, 0 skipped.

PR #467, opened 2026-08-30
xymonnet: choose depth and credentials per host, not per definition

bonomani:
`svc:ok=STATE` ends a dialogue green at a named state, and `svc:cred=NAME` picks the credentials entry, both as hosts.cfg tags — so one definition serves hosts checked to different depths and under different accounts, instead of forcing a copy of the entry per depth.
---
Part of the dialogue stack, sliced out of #456 and #465 so each PR is one subject. Reviewed bottom-up: this one is based on `dialogue-05-manpage`.


bonomani:

PR #468, opened 2026-08-30
xymonnet: four corrections to the dialogue engine

bonomani:
Four fixes and the tests that catch them, none of them user-visible syntax:

- the certificate must come off the **upgraded** session, not the plaintext one
- a failure names the alternative that **matched**, not the first one written
- a `timeout` edge does not end the reachability walk — it briefly made every state after one look unreachable
- `expect ... as NAME` binds the reply on the line that **produces** it, so it cannot name an earlier state's reply
---
Part of the dialogue stack, sliced out of #456 and #465 so each PR is one subject. Reviewed bottom-up: this one is based on `dialogue-06-perhost`.

### DIFF
--- a/lib/netservices.c
+++ b/lib/netservices.c
@@ -456,6 +456,44 @@ static void check_graph(svcinfo_t *rec)
 			 * ->next regardless would make every state reachable and
 			 * the check useless.
 			 */
+			/*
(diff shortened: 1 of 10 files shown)


bonomani:

PR #469, opened 2026-08-30
docs: the manual and the design note describe the code

bonomani:
Nine documentation commits: what protocols.cfg.5 got wrong about the code, DIAGNOSTICS, the certificate, `-> fail` being red regardless of `--checkresponse`, and the design note kept as a description of the code rather than of a pull request. Documentation only.
---
Part of the dialogue stack, sliced out of #456 and #465 so each PR is one subject. Reviewed bottom-up: this one is based on `dialogue-10-binding`.


bonomani:

PR #470, opened 2026-08-30
xymonnet: the shipped SMTP entries hold the conversation, and one layout for every entry

bonomani:
The shipped `[smtp]`/`[smtps]`/`[submission|msa]` entries carry the conversation through EHLO rather than stopping at the greeting.

Then the CONVENTIONS the manual states — attributes before states, one action and one wait per state, every expect naming where it goes, clocks above the waits they bound — and `tests/buildsystem/dialogue-conventions.sh`, which checks every entry the project publishes or tests against them. The parser enforces none of it, so prose that nothing checks drifts.
---
Part of the dialogue stack, sliced out of #456 and #465 so each PR is one subject. Reviewed bottom-up: this one is based on `dialogue-11-manual`.

### DIFF
--- a/docs/design/protocol-dialogues.md
+++ b/docs/design/protocol-dialogues.md
@@ -108,7 +108,8 @@ Submission on 587, upgrading to TLS and authenticating:
    state farewell
       send   "quit\r\n"
       expect "221"                -> success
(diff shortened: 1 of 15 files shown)


bonomani:

PR #471, opened 2026-08-30
xymonnet: a message boundary is whatever the protocol says it is

bonomani:
TCP says nothing about where a message ends; a protocol states it in one of three ways, and until now only one of them was expressible.

- `expect bytes(N)` — a frame decided by length, waiting for exactly N bytes
- `framing length(W, big|little)` — the peer sends a count, then that many bytes: DNS-over-TCP, MySQL, LDAP, AMQP
- `framing terminator "SEQ"` — a sequence ends the message wherever it falls, for a protocol of your own
- and the seven misshapen state layouts the manual only *asked* for are now refused when the file is read

Under length and terminator framing the driver assembles a whole message before anything looks at it, so an `expect` matches the start of the **message** rather than of whatever has arrived.
---
Part of the dialogue stack, sliced out of #456 and #465 so each PR is one subject. Reviewed bottom-up: this one is based on `dialogue-14-conventions`.


bonomani:

PR #472, opened 2026-08-30
xymonnet: ignore PREFIX, for servers that speak unprompted

bonomani:
IMAP sends untagged `*` lines whenever it likes — an EXISTS, an EXPUNGE, a BYE warning — and NNTP and XMPP have their own. A state waiting for a tagged reply read one, failed to match it, and reported a server behaving exactly as its RFC says as down. Fail-fast made it worse: the faster an answer is ruled out, the sooner the probe is wrong.

`ignore PREFIX` says such a message is not an answer — consumed, deciding nothing, ending no state. `idle` is satisfied by it, `timeout` is not, so a server that says nothing but noise still runs out of time.

Also: a capture may not read the value it is about to replace. `banner ~ ... as banner` freed the bytes the match was still reading, and the ovector offsets then indexed freed memory.
---
Part of the dialogue stack, sliced out of #456 and #465 so each PR is one subject. Reviewed bottom-up: this one is based on `dialogue-18-terminator`.


bonomani:

PR #473, opened 2026-08-30
xymonnet: a bound value is bytes, so it keeps every one of them

bonomani:
`expect bytes(N)` and length framing exist for the services whose replies are binary — DNS-over-TCP, LDAP, MySQL, AMQP — and those carry NULs in the first few bytes. `dlg_set()` stored a bound value with `strdup()`, so it was cut at the first one: `${name}` expanded to the stump, and a regex tested against the name saw only what preceded the NUL.

Nothing reported an error. The probe compared almost nothing and reported **green** — the feature added for framed protocols did not work on them.

The length is carried with the value now, through storage, expansion, regex matching, capture, and the wipe that clears a password.

`dialogue-nul.sh` drives four peers with the same twelve bytes `AB\0CDEFGHIJK`, each green entry spending the value somewhere different. Reverted, the peer records `shortbytes 2 4142` and all three go red.
---
Part of the dialogue stack, sliced out of #456 and #465 so each PR is one subject. Reviewed bottom-up: this one is based on `dialogue-20-ignore`.

### DIFF
--- a/lib/digest.c
+++ b/lib/digest.c
@@ -19,7 +19,12 @@ static char rcsid[] = "$Id$";
 
 #include "libxymon.h"
 
-char *md5hash(char *input)
+/*
+ * Hash a counted buffer. A dialogue may hash a value it read from the wire,
+ * and a framed reply is bytes rather than a string -- strlen() would digest
+ * only the part of it that precedes the first NUL.
+ */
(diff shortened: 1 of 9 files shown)


bonomani:

PR #474, opened 2026-08-30
xymonnet: what a send can compute

bonomani:

`lib/digest.c` already implemented every digest they need, so what was missing was the reach:

- `${sha1:}` … `${sha512:}`, `${rmd160:}`
- `${hmac-md5:KEY,MESSAGE}` … `${hmac-sha512:}` — RFC 2104, checked against the RFC 2202 and RFC 4231 vectors including the key-longer-than-block branch
- `${hex:}`, `${len:}`, `${unbase64:}`

**CRAM-MD5 says which of those are load-bearing.** RFC 2195 sends the challenge base64 and hashes it *decoded*, so `${unbase64:}` is the difference between the right digest and a plausible wrong one — and it needed a decoder that returns a length and honours `=` padding, neither of which `base64decode()` did. `dialogue-cram.sh` drives the real exchange; the peer recomputes the answer with OpenSSL's `HMAC()`, and the control is the same exchange with the wrong secret.

The function list lives in `dlg_expansion()`, read both by the check that refuses a mistyped name and by the driver that runs it. They nearly drifted already: `check_expansions()` used `${sha1:...}` as its example of a name that "parses cleanly and expands to nothing".
---
Part of the dialogue stack, sliced out of #456 and #465 so each PR is one subject. Reviewed bottom-up: this one is based on `dialogue-21-bytesvalue`.


bonomani:

PR #475, opened 2026-08-30
xymonnet: framing writes as well as reads, and noise is positional

bonomani:
**A framed connection is written framed.** `framing length(W)` taught the driver to *read* a counted message and left writing one to the file, which cannot do it — the count depends on how long an expanded `${...}` turns out to be. A length-framed *request* carrying any value was unwritable, and the framed protocols could be read from and never spoken to. The count is now written in the declared width and byte order, and a terminator appended. `line` is untouched, and the test's control is a line-framed send that must go out with **nothing** added.

A message too long for its count is refused on both of the only two occasions it can be: a constant send when the file is read, an expanding one at the step. That boundary is the one thing reading the file cannot settle.

**Noise is positional.** Refusing an ignored prefix that also starts an `expect`, by analogy with two overlapping alternatives, was the wrong analogy — it made the one protocol `ignore` exists for the one protocol it could not express. IMAP's greeting *is* an untagged `* OK` line; the untagged lines after it are noise; they share seven characters. The alternatives are tried first, and a message is skipped only when none could be it.

That also fixes an ignored record split across reads, which broke the scan without marking anything skipped and failed a healthy server — same block, cannot be fixed apart.
---
Part of the dialogue stack, sliced out of #456 and #465 so each PR is one subject. Reviewed bottom-up: this one is based on `dialogue-22-expansions`.


bonomani:

PR #476, opened 2026-08-30
docs: protocols.cfg(5) for multi-step dialogues

bonomani:
The manual for the dialogue grammar the earlier slices built. Documentation only — no code changes.

Every `.IP` keyword in the page is a directive `lib/netservices.c` accepts at this commit, and every directive it accepts has an entry: `send`, `expect` with `until` and the `-> TARGET` edges, `state`, `start`, `always`, `eof`, `timeout`, `idle`, the `~` and `else` branches, `capture`/`capture-regex`, `credentials`, `starttls`, `transport`, `options`, `port`. Checked against the parser rather than written beside it — an earlier draft of this page documented `goto`, a bare `fail` and a `when … end` block, none of which the parser has ever taken.

---
### Stack

**7 of 15** in the dialogue stack: base #464, next #477. The full chain is listed in #457; read bottom-up.


bonomani:

PR #477, opened 2026-08-30
xymonnet: choose depth and credentials per host, not per definition

bonomani:
`svc:ok=STATE` ends a dialogue green at a named state and `svc:cred=NAME` picks the credentials entry, both as hosts.cfg tags — so one definition serves hosts checked to different depths and under different accounts, instead of a copy of the entry per depth.

---
### Stack

**8 of 15** in the dialogue stack: base #476, next #478. The full chain is listed in #457; read bottom-up.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,7 @@
 .*.swp
 .DS_store
 # Simulated Subversion default ignores end here
+
+# Written by build/genconfig.sh from this machine probes; committing it
+# pins one host answers on everybody else build.
+include/config.h
--- a/lib/netservices.c
+++ b/lib/netservices.c
@@ -611,7 +611,7 @@ static void emit_step(svclist_t *first, svcstep_t *tmpl)
  *
  *     NAME	username	password
  */
-static int lookup_credentials(char *name, char **user, char **pass)
+int lookup_credentials(char *name, char **user, char **pass)
 {
 	char fn[PATH_MAX];
 	FILE *fd;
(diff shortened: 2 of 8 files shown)


bonomani:
Fixed the `ok=` bug found in review, in this PR rather than on top of it — `ok=` is this PR's feature, so it should ship correct rather than be repaired three PRs later. The commit is amended, and the seven PRs above have been rebased.

**The bug.** The verdict was taken on *entering* the named state, before that state's own `expect` had run. So `ok=<start state>` reported green the moment the TCP connect succeeded — a server that accepted the connection and then sent garbage, or sent nothing at all, was reported up. A depth that checks nothing is worse than no depth, because it looks like a check.

**The fix.** A state ends when the next one begins. Every edge resolves to its target's `STEP_LABEL` and every transition runs through `dlg_run_instant()`, so that label is the one place a state ends — the verdict is taken there, against the state just left. A state that *fails* never reaches it: the failing verdict is already set and the dialogue has stopped, so a bad greeting is still red.

**The old test passed for the wrong reason.** Its `shallow` host would have passed with a greeting of `599 broken`. Two controls added, both tagged `ok=greeting`:

| host | server | must be |
|---|---|---|
| `badgreet` | answers `599 broken` | red |
| `silent` | accepts the connection, says nothing | red |

Reverted, `badgreet` reports **green** — that is the bug, reproduced.

Suite: 121 passed, 0 failed.

bonomani:

PR #478, opened 2026-08-30
xymonnet: four corrections to the dialogue engine

bonomani:
Four fixes to code that landed in the earlier slices, and the tests that catch them.

**It does replace one piece of syntax.** `capture as NAME` and `capture-regex "…" as N1;N2` are removed, and the two forms that supersede them are added: `expect "…" as NAME`, which binds on the line that produced the reply, and `NAME ~ "…" as N1;N2`, which reads a bound value. The old spelling named its operand by position — a `capture` written one state too late bound the earlier state's reply and said nothing — which is why it is gone rather than kept alongside. `protocols.cfg(5)` is updated in the same commit.

- the certificate must come off the **upgraded** session, not the plaintext one — an assertion added here, over code that #463 already had
- a failure names the alternative that **matched**, not the first one written
- a `timeout` edge does not end the reachability walk — it briefly made every state after one look unreachable
- `expect ... as NAME` binds the reply on the line that **produces** it, so it cannot name an earlier state's reply

---
### Stack

**9 of 15** in the dialogue stack: base #477, next #479. The full chain is listed in #457; read bottom-up.

### DIFF
--- a/docs/manpages/man5/protocols.cfg.5.html
+++ b/docs/manpages/man5/protocols.cfg.5.html
@@ -124,10 +124,10 @@ Section: File Formats (5)<br/>Updated: Version 4.3.30: 4 Sep 2019<br/><a href="#
       that end the test where it stands:
     <p class="Pp"></p>
     <pre>   -&gt; <i>LABEL</i>        continue in that state
-   -&gt; success      the service is up (green)
+   -&gt; success      the service is up
    -&gt; warning      answering, but not as it should (yellow)
-   -&gt; fail         the test failed (red)</pre>
-    <p class="Pp"><b>warning</b> is the one worth reaching for: a server that
+   -&gt; fail         the test failed (the <b>--checkresponse</b> colour)</pre>
+    <p class="Pp"><b>warning</b> is the one worth reaching for. A server that
         answers correctly but refuses an optional capability is not down, and
         reporting it as down teaches operators to ignore the column.</p>
     <p class="Pp"><b>state</b> names the state that follows it -- the expects up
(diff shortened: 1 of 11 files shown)


bonomani:
**Heads-up from verifying the stack: this PR does not pass its own tests.**

Building and running the dialogue suite against each PR in the chain independently:

| PR | result |
|---|---|
| #476 | green |
| #477 | green |
| #478 | **red** — `dialogue-depth`, `dialogue-terminals`, `dialogue-starttls-negotiation` |
| #479 | **red** — same three |
| #480 | **red** — same three |
| #481 … #484 | green |

**This is inherited, not caused by the slicing.** I checked out the original commit `25d64698c` from the old #456 branch and it fails the same three, so PR #456 was red in its middle too — it just was not visible, because only the tip of a single PR ever gets tested.

**Where it starts:** bisected to `35bf73467` *"report the alternative that matched, not the first written"*. Before it `dialogue-terminals` passes; from it onward the tree is red until the fixture rewrite in `e764e2aa9` *"refuse the seven shapes"*, which lands in #481.

**What breaks:** verdict colours. `-> fail` yields yellow instead of red, and `else -> warning` yields red instead of yellow — consistent with blaming a different step than before and taking the colour from it.

**To make the chain green at every step**, the fixture adjustments in `e764e2aa9` (and the parts of `070e3af23` / `7ef5a6e95` that touch these three tests) would need to move down to sit with `35bf73467`. That is a reorder of your commits, so I have not done it — say the word and I will.

bonomani:
**Retracting the previous comment — it was wrong.**

The three failures I reported were an artefact of my own testing, not a property of this PR. I was checking out each slice and running `make`, which is timestamp-driven: checking out an older commit leaves object files newer than the sources, so make skipped rebuilding and I was testing a binary that did not correspond to the tree.

Rebuilt from clean (`*.o` and archives deleted) at this PR's head, all three pass:

```
dialogue-depth=PASS  dialogue-terminals=PASS  dialogue-starttls-negotiation=PASS
```

CI had it right all along — the `build (server)` job on this exact SHA reports 112 passed, 0 failed. I should have trusted that over my own local run and checked the discrepancy before posting.

The bisect to `35bf73467` in that comment is void for the same reason. There is no red middle in this chain.

Sorry for the noise. The one genuine CI failure on the chain is `manual pages match their sources`, which is a separate matter and I am looking at it now.

bonomani:

PR #480, opened 2026-08-30
xymonnet: the shipped SMTP entries hold the conversation, and one layout for every entry

bonomani:
The shipped `[smtp]`/`[smtps]`/`[submission|msa]` entries carry the conversation through EHLO rather than stopping at the greeting.

The three are written as named states — `greeting`, `ehlo`, `farewell` — and not as a flat sequence, because that is what makes the depth tag work. `ok=` takes its verdict when a **named** state ends, so a stateless entry offers a host no way to decline: protocols.cfg(5) documents `smtp:ok=greeting` for exactly this case, and against a flat entry it would have been silently inert — no error, no effect, the whole exchange sent anyway. With the states it means banner depth, host by host, which is the old behaviour for anyone who wants to keep it.

Each waiting state carries `timeout(10)`, which is what the global default already was (`xymonnet.c:79`), and the per-test cutoff still bounds the test as a whole (`contest.c:2076`). The clocks attribute a timeout to the state it happened in; they do not lengthen anything.

Two commits, so the only change that reaches the wire is revertable on its own: the layout and its test, then the shipped entries.

**What it reports.** The `ehlo` state declares its verdicts instead of inheriting the generic "unexpected response" colour: `4` → warning, `5` → fail, `eof` → fail. Measured against a real ESMTP server over implicit TLS, driving this file as it ships:

| | old entry | this entry |
|---|---|---|
| healthy | green | green |
| `500` to EHLO | green | **red** |
| `421` to EHLO | green | **yellow** |
| closes after `220` | green | **red** |
| no reply to EHLO | green | **red**, via `timeout(10)` |
| never greets | yellow | red |
| `500` + `smtps:ok=greeting` | — | green |

The failure names its step: `Unexpected service response: state ehlo got "5"`.

**One parser fix rides with it,** because the entries are unwritable without it. `always -> success` was read as a jump to a state nobody declared, so the edge was dropped at link time and `farewell` had no way out — the `always` arm resolved only state names, where the six other edge arms map the three verdicts and protocols.cfg(5) defines a target once for all of them. The runtime already honoured a verdict on a jump (`contest.c`, `case STEP_JUMP`); only the parser never emitted one. `dialogue-validation.sh` called the shipped file quiet while grepping seven remembered phrases that didn't include this one; it now counts every `Service NAME:` line, and fails against the unfixed parser naming all four services.

**Not addressed here:** the `greeting` state has no verdicts, so a 5xx greeting stays the generic yellow, exactly as today.

Then the CONVENTIONS the manual states — attributes before states, one action and one wait per state, every expect naming where it goes, clocks above the waits they bound — and `tests/buildsystem/dialogue-conventions.sh`, which checks every entry the project publishes or tests against them. The parser enforces none of it, so prose that nothing checks drifts.

---
### Stack

**11 of 15** in the dialogue stack: base #479, next #481. The full chain is listed in #457; read bottom-up.

### DIFF
--- a/docs/design/protocol-dialogues.md
+++ b/docs/design/protocol-dialogues.md
@@ -40,7 +40,13 @@ condition ::= expect "…" [ until "…" ] [ as NAME ] | NAME ~ "…"
 
 **Targets** are a state, or one of `success`, `warning`, `fail` -- green, yellow, red. `warning` is the point: a server answering correctly but refusing an optional capability is not down, and calling it down teaches operators to ignore the column. An edge saying `fail` is red whatever `--checkresponse` is set to; a reply matching no alternative takes that option's colour, yellow by default.
 
-**Order is file order.** A state's lines run as written and the first edge that fires leaves, so a `~` above an `expect` decides before the socket is read at all. An earlier draft specified a fixed precedence regardless of where lines were written; the implementation does not do that and should not, because a precedence the reader cannot see is what turns a declaration back into a program.
+**Layout.** Entry attributes -- `transport`, `port`, `options`, `start` -- are written first, in any order among themselves, because they describe the definition rather than any step; then each `state` with its lines indented under it.
+
+**One action, one wait.** A state does one thing -- a `send`, a `starttls`, a `credentials` -- and waits for one answer: the clock that bounds the wait, and the `expect` lines that end the state, each naming where its answer leads. A state that acts without waiting leaves on `always`; a state that decides on a value already bound needs no action at all. Because a state holds one action and one wait, the order of its lines carries no meaning of its own: the entry is declarative down to the state, and sequence lives in the edges between states rather than in the lines within one.
+
+The parser enforces none of it: attributes are accepted anywhere, indentation is ignored, several actions and several waits in one state are accepted, and an `expect` with no `->` falls through to the line below. One ordering rule is real and unenforced: a `timeout` or `idle` arms the wait that follows it, so a clock below a state's `expect` lines bounds nothing. `protocols.cfg(5)` lists the five rules under CONVENTIONS, and `tests/buildsystem/dialogue-conventions.sh` checks every entry in the manual, in the shipped `protocols.cfg` and in the fixtures against them -- an entry that breaks one on purpose says `# conventions: permissive` and says why. Prose nothing checks is how the manual came to document a `when ... end` grammar no shipped code ever had.
+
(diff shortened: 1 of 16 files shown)


bonomani:
**Heads-up from verifying the stack: this PR does not pass its own tests.**

Building and running the dialogue suite against each PR in the chain independently:

| PR | result |
|---|---|
| #476 | green |
| #477 | green |
| #478 | **red** — `dialogue-depth`, `dialogue-terminals`, `dialogue-starttls-negotiation` |
| #479 | **red** — same three |
| #480 | **red** — same three |
| #481 … #484 | green |

**This is inherited, not caused by the slicing.** I checked out the original commit `25d64698c` from the old #456 branch and it fails the same three, so PR #456 was red in its middle too — it just was not visible, because only the tip of a single PR ever gets tested.

**Where it starts:** bisected to `35bf73467` *"report the alternative that matched, not the first written"*. Before it `dialogue-terminals` passes; from it onward the tree is red until the fixture rewrite in `e764e2aa9` *"refuse the seven shapes"*, which lands in #481.

**What breaks:** verdict colours. `-> fail` yields yellow instead of red, and `else -> warning` yields red instead of yellow — consistent with blaming a different step than before and taking the colour from it.

**To make the chain green at every step**, the fixture adjustments in `e764e2aa9` (and the parts of `070e3af23` / `7ef5a6e95` that touch these three tests) would need to move down to sit with `35bf73467`. That is a reorder of your commits, so I have not done it — say the word and I will.

bonomani:
**Retracting the previous comment — it was wrong.**

The three failures I reported were an artefact of my own testing, not a property of this PR. I was checking out each slice and running `make`, which is timestamp-driven: checking out an older commit leaves object files newer than the sources, so make skipped rebuilding and I was testing a binary that did not correspond to the tree.

Rebuilt from clean (`*.o` and archives deleted) at this PR's head, all three pass:

```
dialogue-depth=PASS  dialogue-terminals=PASS  dialogue-starttls-negotiation=PASS
```

CI had it right all along — the `build (server)` job on this exact SHA reports 112 passed, 0 failed. I should have trusted that over my own local run and checked the discrepancy before posting.

The bisect to `35bf73467` in that comment is void for the same reason. There is no red middle in this chain.

Sorry for the noise. The one genuine CI failure on the chain is `manual pages match their sources`, which is a separate matter and I am looking at it now.

bonomani:

PR #481, opened 2026-08-30
xymonnet: how a message ends, what a value keeps, what a state may be

bonomani:
TCP says nothing about where a message ends; a protocol states it in one of three ways, and until now only one of them was expressible.

- `expect bytes(N)` — a frame decided by length, waiting for exactly N bytes
- `framing length(W, big|little)` — the peer sends a count, then that many bytes: DNS-over-TCP, MySQL, LDAP, AMQP
- `framing terminator "SEQ"` — a sequence ends the message wherever it falls, for a protocol of your own
- and the seven misshapen state layouts the manual only *asked* for are refused when the file is read

**And the two fixes that make those work on the data they exist for.** A bound value was stored with `strdup()`, so a reply carrying a NUL — which is every message these protocols send — was cut at the first one: `${name}` expanded to the stump, a regex saw only what preceded it, and nothing said so. The probe compared almost nothing and reported **green**. The length is now carried with the value through storage, expansion, matching, capture and the wipe that clears a password.

That fix rewrites the subject copy in `dlg_capture()`, which is the copy the use-after-free fix introduced, so the two travel together rather than the fix trailing the feature it repairs.

`dialogue-nul.sh` drives four peers with the same twelve bytes `AB\0CDEFGHIJK`. Reverted, the peer records `shortbytes 2 4142` and all three green entries go red.

---
### Stack

**12 of 15** in the dialogue stack: base #480, next #482. The full chain is listed in #457; read bottom-up.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,6 @@
 .DS_store
 # Simulated Subversion default ignores end here
 
-# Written by build/genconfig.sh from this machine probes; committing it
-# pins one host answers on everybody else build.
+# Written by build/genconfig.sh from this machine's probes; committing it
+# pins one host's answers on everybody else's build.
 include/config.h
--- a/docs/design/protocol-dialogues.md
+++ b/docs/design/protocol-dialogues.md
@@ -17,34 +17,34 @@ entry rather than deleting it.
 
 ```
 [service|alias]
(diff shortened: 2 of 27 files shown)


bonomani:

PR #482, opened 2026-08-30
xymonnet: ignore PREFIX, for servers that speak unprompted

bonomani:
IMAP sends untagged `*` lines whenever it likes — an EXISTS, an EXPUNGE, a BYE warning — and NNTP and XMPP have their own. A state waiting for a tagged reply read one, failed to match it, and reported a server behaving exactly as its RFC says as down. Fail-fast made it worse: the faster an answer is ruled out, the sooner the probe is wrong about it.

`ignore PREFIX` says such a message is not an answer — consumed, deciding nothing, ending no state, binding nothing. `idle` is satisfied by it, `timeout` is not, so a server that says nothing but noise still runs out of time.

Up to eight prefixes, written on the entry beside `framing`, because noise is a property of the protocol rather than of one reply.

---
### Stack

**13 of 15** in the dialogue stack: base #481, next #483. The full chain is listed in #457; read bottom-up.

### DIFF
--- a/docs/design/protocol-dialogues.md
+++ b/docs/design/protocol-dialogues.md
@@ -17,10 +17,11 @@ entry rather than deleting it.
 
 ```
 [service|alias]
-   transport tcp          entry attributes: transport, port, options, framing and start
+   transport tcp          entry attributes: transport, port, options, framing, ignore and start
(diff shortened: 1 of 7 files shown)


bonomani:

PR #483, opened 2026-08-30
xymonnet: what a send can compute

bonomani:

`lib/digest.c` already implemented every digest they need, so what was missing was the reach:

- `${sha1:}` … `${sha512:}`, `${rmd160:}`
- `${hmac-md5:KEY,MESSAGE}` … `${hmac-sha512:}` — RFC 2104. Checked against a peer that computes the same HMAC with OpenSSL and rejects a wrong one, so a mistake in the padding or the key handling fails there rather than passing on a string comparison. The published RFC 2202/4231 vectors are not run: both suites use a short key, so the key-longer-than-block branch is exercised by no test here.
- `${hex:}`, `${len:}`, `${unbase64:}`

**CRAM-MD5 says which of those are load-bearing.** RFC 2195 sends the challenge base64 and hashes it *decoded*, so `${unbase64:}` is the difference between the right digest and a plausible wrong one — and it needed a decoder that returns a length and honours `=` padding, neither of which `base64decode()` did. `dialogue-cram.sh` drives the real exchange; the peer recomputes the answer with OpenSSL's `HMAC()`, and the control is the same exchange with the wrong secret.

The function list lives in `dlg_expansion()`, read both by the check that refuses a mistyped name and by the driver that runs it. They nearly drifted already: `check_expansions()` used `${sha1:...}` as its example of a name that "parses cleanly and expands to nothing".

---
### Stack

**14 of 15** in the dialogue stack: base #482, next #484. The full chain is listed in #457; read bottom-up.


bonomani:

PR #486, opened 2026-08-31
xymonnet: start FEATURE, and the outcome it binds

bonomani:
**14 of 14** in the dialogue stack: base #484. The full chain is listed in #457; read bottom-up.

`starttls` was one protocol's verb for something the client does — FTP calls it `AUTH TLS`, POP3 `STLS`. `start FEATURE` names it instead: the word after `start` selects code here rather than a state, and is reserved. `tls` is the only feature; naming another is refused when the file is read. The entry attribute `start NAME` becomes `begin NAME` to free the word, and reads better for what it was already doing.

## The outcome is a bound value

```
   state secure
      start tls
      tls ~ "^ok"          -> ehlo-tls
      tls ~ "no shared"    -> warning
      else                 -> fail
```

`${tls}` holds `ok`, or the reason OpenSSL gave up — `certificate verify failed`, `no shared cipher`, `wrong version number`.

Before this, a server that **agreed** to STARTTLS and then failed the handshake ended the test red whatever the entry said, while a server that **refused** could be routed to `warning` by an ordinary `expect "454"` edge. Those are the same operational fact — TLS not usable, service up — and only one of them could be reported as such.

Binding the result needs no new condition keyword: the state branches on `${tls}` with the same `~` edges every other bound value uses. An entry that tests nothing still fails, because no alternative matches and fail-fast ends it there.

## Refused when the file is read

- `start FEATURE` for any feature but `tls` — a typo cannot leave a probe quietly doing less than it says
- `expect … as tls` — a capture would shadow the outcome

## Verified

`dialogue-tls-outcome.sh` drives a peer that agrees to STARTTLS and then answers the ClientHello in plaintext: the entry routes it to **yellow**. Its control is the same entry against a peer that completes the handshake, which must stay **green** — otherwise the routing would be reporting the outcome it was handed rather than the one that happened.


bonomani:

PR #487, opened 2026-08-31
xymonnet: send the tested hostname as the TLS server name

bonomani:
## The problem

A TLS server asked for no server name answers with its **default** certificate.

`xymonnet` reads the certificate from the session to fill the `sslcert` column, so wherever one address serves several names, the expiry Xymon reports belongs to whichever certificate that address answers with — not to the name the test is for. The column looks authoritative and is wrong, which is the failure mode a monitoring system can least afford.

Nothing in `protocols.cfg` ever parsed an SNI name: `item->sni` was set only by the HTTP tests (`httptest.c`), so **no protocol probe has ever sent one** — neither `options ssl` nor a STARTTLS upgrade.

## The change

`options sni` hands the tested hostname to the test:

```
[imaps]
   options ssl,banner,sni
   port 993
```

| layer | change |
|---|---|
| `lib/netservices.h` | `TCP_SNI` flag |
| `lib/netservices.c` | parse the bare `sni` option, beside `alpn=` |
| `xymonnet/xymonnet.c` | set `tt->sni` from the tested hostname when the flag is set |

`setup_ssl()` is **untouched** — it already passes `item->sni` to `SSL_set_tlsext_host_name()`, so this only supplies the value the TLS code was already looking for.

## The address guard

The name is sent, never the address. RFC 6066 does not allow an address literal in a `server_name` extension, and a server that enforces it **aborts the handshake** rather than ignoring the field — so an unguarded version would turn a quiet reporting gap into a red test for anyone whose `hosts.cfg` entry is an IP. A host entered by address sends no server name even with the option set.

Tested by shape rather than `inet_pton()` so no new header is needed on the platforms this builds for: a legal hostname cannot be all digits and dots, because the last label cannot be numeric, and a colon appears only in an IPv6 literal.

## Blast radius

**No shipped entry enables it.** `smtps`, `imaps` and `pop3s` keep todays

**Overlaps with #488, and duplicates #458.** Two different problems:

- **#458 #459** in `xymonnet/xymonnet.c` — the same feature by a different design, opened three days before this one. That is a decision about which to keep, not a rebase: merging either makes the other's approach redundant.
- **#488** in `lib/netservices.c`, `lib/netservices.h`, `xymonnet/protocols.cfg.5` and the generated page. No line is claimed by both — the edits abut. This adds a single option-parsing line to `netservices.c`; #488 rewrites that file (+465/-44 on 371 lines), so the cheap resolution is to carry the one line into #488 rather than rebase this around it.

### DIFF
--- a/docs/manpages/man5/protocols.cfg.5.html
+++ b/docs/manpages/man5/protocols.cfg.5.html
@@ -102,7 +102,16 @@ Section: File Formats (5)<br/>Updated: Version 4.3.30: 4 Sep 2019<br/><a href="#
     <pre>   banner - include received data in the status message
    ssl - service uses SSL so perform an SSL handshake
    telnet - service is telnet, so exchange telnet options
-   alpn=protocol1,protocol2 - specify ALPN protocols to negotiate during SSL handshake</pre>
+   alpn=protocol1,protocol2 - specify ALPN protocols to negotiate during SSL handshake
+   sni - send the tested hostname as the TLS server name</pre>
+    <p class="Pp"><b>sni</b> matters wherever one address serves several names.
+        A TLS server that is asked for no server name answers with its DEFAULT
+        certificate, so the expiry Xymon reports in the <b>sslcert</b> column is
+        that of whichever certificate the address happens to answer with, and
+        not the one the test is named for -- wrong, and indistinguishable from
+        right. The name tested is sent, never the address: RFC 6066 does not
+        allow an address literal in the server_name extension, so a host entered
+        in hosts.cfg by IP sends nothing even with the option set.</p>
     <p class="Pp"></p>
     <p class="Pp"></p>
   </dd>
--- a/lib/netservices.c
+++ b/lib/netservices.c
@@ -247,6 +247,7 @@ char *init_tcp_services(void)
 					else if (strncmp(opt, "alpn=", 5) == 0) {
 						first->rec->alpns = strdup(opt+5);
 					}
+					else if (strcmp(opt, "sni") == 0)    first->rec->flags |= TCP_SNI;
 					else errprintf("Unknown option: %s\n", opt);
 
 					opt = strtok(NULL, ",");
--- a/lib/netservices.h
+++ b/lib/netservices.h
@@ -19,6 +19,7 @@
 #define TCP_SSL        0x0004
 #define TCP_HTTP       0x0008
 #define TCP_ALPN       0x0010
+#define TCP_SNI        0x0020	/* send the tested hostname as the SNI server name */
 
 typedef struct svcinfo_t {
 	char *svcname;
--- a/tests/network/sni-support.sh
+++ b/tests/network/sni-support.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# tests/network/sni-support.sh
+#
+# Guard for the xymonnet SNI support: `options sni` on a protocols.cfg entry
+# sends the tested hostname as the TLS server name.
+#
+# Without it, a host serving several names on one address answers a handshake
+# that carries no server name with its DEFAULT certificate -- so the expiry
+# reported in the sslcert column belongs to whichever certificate that address
+# answers with, not to the name the test is for. It is wrong quietly, which is
+# the worst way for a monitoring system to be wrong.
+#
+# The wiring spans three layers:
+#   - lib/netservices.h : the TCP_SNI flag,
+#   - lib/netservices.c : parsing the `sni` option into the service flags,
+#   - xymonnet/xymonnet.c: handing the tested hostname to the test, which
+#     setup_ssl() then passes to SSL_set_tlsext_host_name().
+#
+# Exercising it for real needs a TLS server holding two certificates and
+# selecting on the server name; the build CI already compiles these files, so
+# this is a static guard that the wiring across the layers (and the manpage)
+# survives future edits -- the same tradeoff tests/network/alpn-support.sh
+# makes for ALPN. Skips only when the source files are absent (e.g. an
+# autopkgtest run against an installed package with no source tree); a present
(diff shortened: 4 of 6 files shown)


bonomani:
Conflicts with #488 in `docs/manpages/man5/protocols.cfg.5.html`, `lib/netservices.c`, `lib/netservices.h` and `xymonnet/protocols.cfg.5`.

The `lib/netservices.c` collision is one line of option parsing against a file #488 rewrites (+465/-44 on a 371-line file), so it is cheapest resolved by carrying that line into #488 rather than by rebasing this onto it.

See also #458 — that is a design question rather than an ordering one.

bonomani:
## Where `main` stands

The controls already exist, and for every service except HTTP they are inert.

| | on `main` today |
|---|---|
| global | `--sni=on` / `--sni=off`, default **off** (`int snienabled = 0;`) |
| per host | `sni` / `nosni` tags in `hosts.cfg`, documented in `hosts.cfg(5)` |
| per service | nothing |
| name sent | always the URL host, no override |

The only code that ever sets the field is `httptest.c`, which reads the host tags and falls back to the global default. `contest.c` then does `if (item->sni) SSL_set_tlsext_host_name(...)`. Nothing sets `sni` for a plain TCP-TLS probe, so imaps, smtps, pop3s and nntps have never sent a server name — whatever `hosts.cfg` says.

That matters for this pull request on its own, before any question of composing it with #458:

```c
if (tt && tt->svcinfo && (tt->svcinfo->flags & TCP_SNI) &&
    t->host->hostname && !sni_name_is_address(t->host->hostname)) {
        tt->sni = t->host->hostname;
}
```

The condition reads neither `t->host->sni` nor `snienabled`. So a host carrying `nosni` — a tag `hosts.cfg(5)` documents today, and which `httptest.c` honours — would silently receive SNI on its imaps test as soon as the service definition carries `options sni`, with no way to except it. The tag exists, the manual page promises it, and this path ignores it. Whatever happens with #458, that wants fixing here.

## Against #458 → #459 → #460

This and that series are the same feature configured on different axes: this one per service, that one per host. Neither axis is wrong, and neither has the other's.

They do not compose as they stand. Both assign `tt->sni` at the same point in `xymonnet.c`:

```c
/* #460 */
if (tt && tt->svcinfo && (tt->svcinfo->flags & TCP_SSL)) {
        int want = (t->host->sni != 0) ? (t->host->sni > 0) : snienabled;
        char *name = t->host->sniname ? t->host->sniname
                                      : (t->host->testip ? NULL : t->host->hostname);
        if (want && name) { /* inet_pton guard */ if (!isipliteral) tt->sni = name; }
        tt->checkname = tt->sni;
}
```

Stacked, whichever block runs second wins, and #460 sets `tt->checkname = tt->sni` inside its own block, so the certificate check would verify a name a later block replaced. The IP-literal guard also lands twice, by two methods that disagree at the edges: `inet_pton()` against the shape test here, where `sni_name_is_address("")` returns 1 and `123.456` reads as an address.

## The config surface the two add up to

#458 also flips `snienabled` from 0 to 1, so with the series merged SNI is on by default. That matters for what is still missing:

| | enable | disable | name |
|---|---|---|---|
| global | on (after #458), `--sni=on` | `--sni=off` | — |
| per host | `sni` | `nosni` | `sni=NAME` (#459) |
| per service | `options sni` | **missing** | — |

With no per-service `nosni`, a service can raise the default but never lower it — and once the default is on, `options sni` asks for what it already has. A protocol that breaks under SNI cannot be excluded except by turning it off for the whole host, which loses it for that host's other TLS services.

## What to do

Merge the series, then reduce this pull request to the part the series does not have:

- keep `lib/netservices.c` and `lib/netservices.h` — the `sni` option and the `TCP_SNI` flag
- add the mirror: a `nosni` option and a `TCP_NOSNI` flag, two more lines
- keep `xymonnet/protocols.cfg.5` and `tests/network/sni-support.sh`
- drop `sni_name_is_address()` and the `xymonnet.c` block; the series already guards the address case, and reads the host tags this block ignores
- combine the two axes in the decision the series already makes:

```c
int svcsni   = (tt->svcinfo->flags & TCP_SNI)   ? 1 : 0;
int svcnosni = (tt->svcinfo->flags & TCP_NOSNI) ? 1 : 0;
int said_no  = (t->host->sni < 0) || svcnosni;
int said_yes = (t->host->sni > 0) || svcsni;
int want     = !said_no && (said_yes || snienabled);
```

A "no" from either axis wins; a "yes" from either raises the default. The two axes are orthogonal — a host flag covers every service on one host, a service option covers every host for one service — so neither contains the other and ranking them is arbitrary. Both disables report that something breaks, a peer that rejects an unknown name or a probe that cannot carry the extension, and ranking them makes one broken case silently override another.

That leaves `sni` on a host with `nosni` on its service sending nothing, deliberately: a service-level no says the probe cannot carry it, and a host flag cannot grant a capability. Worth stating in both manual pages rather than leaving it to be read out of the expression.

Not proposed: a per-host-per-service axis. The two cases that occur — this server breaks, this protocol cannot — are both covered once `nosni` exists, and syntax for the third is config surface nobody has asked for.

*(Corrected: an earlier version of this comment had the host flag winning outright over the service option. The axes are orthogonal, so that ranking had no basis.)*

bonomani:
## A different answer: let the test carry the name

Everything above argues about where to attach a server name to a test that has none. Xymon's own `http` and `ldap` tests don't have that problem, because they are URLs — the name is in the test. Generalised to the other services, `imaps://mail.example.com=10.0.0.5/` sends `mail.example.com` and connects to `10.0.0.5`, per test rather than per host, which is finer than anything the four open pull requests reach.

That direction is written up in #505, with the grammar, the full legacy-to-new table, and the implementation constraints, and is not repeated here.

It does not argue against merging this work. Eight TLS services send no server name at all today, so `sslcert` reports whichever certificate a shared address answers with, and the `sni`/`nosni` tags `hosts.cfg(5)` documents are inert for every one of them. #505 is scoped after the dialogue work in #457, not in front of it.

PR #488, opened 2026-08-31
xymonnet: a probe holds its conversation, and can upgrade to TLS

bonomani:
Stacked on #454. An entry in `protocols.cfg` becomes an ordered conversation instead of one send and one match.

**The first three commits are not part of the feature.** They fix bugs already in `main` — an alias's `send`/`expect` copied with `strdup()` while its length is carried separately, the banner buffer grown through an unchecked `realloc()`, telnet's option buffer unchecked and its banner leaked. Take them on their own merits or cherry-pick them; the dialogue work starts at the fourth commit. The `X509_get_signature_type(peercert)` line is **not** here — #449 owns it.

## Problem

`do_tcp_tests()` writes before it reads, so every greet-first probe sent its command **ahead of the greeting**. For SMTP that violates RFC 5321, and Postfix ≥ 3.9 answers `554 Error: SMTP protocol synchronization` — #450, where the probe matched the `220` it raced and reported the service **up** while every session was rejected. It was fixed there by deleting the send, which left the probe with only a banner to look at. The same shape was live in `ftp`, `pop`, `imap`, `nntp` and `snpp`.

Those entries also never checked the answer to their own command: `expect` matched the first read, which is the greeting. A server that greets and then fails everything read as healthy. And a plaintext port could not reach a certificate at all — `options ssl` is TLS from the first byte, a different port and service, while **STARTTLS on 25 is how most mail is actually encrypted**.

## Change

```
[smtp]
   expect "220" until "220 "
   send   "ehlo xymonnet\r\n"
   expect "250" until "250 "
```

- **Order is the file's order.** `expect` first waits; `send` first asks.
- **`until`** says where a multi-line reply ends, and accepts a bare final code (`250` with no text), which RFC 5321 4.2 permits.
- **`start tls`** upgrades where the line sits, through the same `setup_ssl()` as `options ssl`, so a plaintext port reaches `sslcert`. As the **first** step it is TLS from the first byte — implicit TLS, said in order. Two words because the request is protocol-specific and the step is not: FTP asks `AUTH TLS`, POP3 `STLS`, SMTP and IMAP `STARTTLS`.
- **`start iac`** negotiates telnet options where the line sits, and answers them. IAC belongs to the telnet protocol, not to port 23 — MUDs and BBSes negotiate on their own ports.
- **`expect ""`** matches any reply and rejects none: it asks that the server answered, not what it said. `[oratns]` and `[ircd]` use it, their replies being unmatchable — one behind a length-prefixed header, one where the IRC numeric is the second token on a line starting with the server's own name.
- **An entry with no steps is a connect check**, which is all some protocols allow here: `ldap`, `ldaps`, `qmtp`, `qmqp`, `netbios-ssn` and `lpd` each want a length-prefixed binary request, and `bbd` answers nothing until the client half-closes. No keyword marks them; the set is pinned in `protocols-stepless.sh`, so adding one means changing a test on purpose.
- **A line that cannot be resolved refuses the entry** — unknown directive or option, `start` with anything but `tls`/`iac`, a second `start tls`, a mistyped `until`, `options ssl` with `start tls`. The entry then sends nothing at all, not even a ClientHello, and reports yellow with the reason, with or without `--checkresponse`.

`ssh` and `vnc` lose their send: their version exchange is followed by binary, so it only bought a protocol violation. Before this, every entry stopped at its greeting.

**No shipped entry uses an option any more.** `ssl`, `banner` and `telnet` are still accepted — they are in files people already have, and an unknown option refuses an entry — but they are the older spellings of a leading `start tls`, of an `expect`, and of `start iac`. `options telnet` is *parsed into* a leading `start iac`, so the two cannot behave differently. `alpn=` is unchanged and unused by shipped entries.

## Behaviour changes an operator will see

1. **A reply split across TCP segments no longer reports down.** Every entry is now run by the dialogue driver. The older send-and-match arm judged the **first read alone**, so a reply whose first segment was shorter than the `expect` string read as short and a healthy server reported down. Eight entries move onto the driver (`clamd`, `spamd`, `oratns`, `amqp`, `cupsd`, `ajp13`, `rdp`, `ircd`); a five-scenario differential — matching, wrong, silent, split, reply-then-close — agrees on four and differs only on the split one, where the old arm is wrong. That arm remains for the compiled-in `http`/`https` tests, which never come from this file.
2. **`[smtp]`/`[submission]` require ESMTP** — `250` to `EHLO`, not just a banner. A HELO-only server now reports down. RFC 5321 3.2 describes that fallback and this grammar cannot express it (choosing between two replies needs branching), so the note above the entry says to copy it with `helo`. A false "down" is visible and fixable in one line; a false "up" on a server refusing every session is invisible.
3. **`[imap*]` read the tagged result** of `LOGOUT`; `* BYE` is sent whether the command worked or not.
4. **`[nntp*]` accept `201`** as well as `200`, per RFC 3977, and check the `205` to `QUIT`.
5. **A service that accepts the connection and never answers now reports down.** `oratns`, `ircd`, `telnet` and `telnets` used to go green on an open port alone — a wedged listener was indistinguishable from a healthy one, with only an empty banner column to show for it.
6. **`protocols.cfg` is now required.** If it cannot be read, every TCP service reports as a refused definition naming the file, instead of falling back to probes compiled into the binary — those were older and weaker, so columns kept reporting while the checks behind them had quietly changed. The compiled table is now **empty**. It had shrunk to names and ports that made a missing file yellow rather than red, but covered 28 of the 49 services the file defines: the same fault, two colours, decided by which names were hardcoded years ago. `find_tcp_service()` now refuses every service instead, which also fixes a latent crash — `contest.c` dereferences the returned definition unconditionally, so `svc:port` with no `protocols.cfg` hit a NULL that those 28 names were hiding.
7. **`/etc/services` is no longer consulted for ports.** A locally-added entry with no `port` line **stops resolving** — loudly (`Trying to scan port 0 …`), but it is a config that used to work. All 37 shipped entries name their port; `[amqps]` had none in any source, and `[smtps]` gains **465**, left out when IANA reassigned it and given back by RFC 8314.
8. **Telnet option refusals are written.** They never were, on either path. `start iac` answers `WILL` with `DONT` and `DO` with `WONT`, and — per RFC 854 — answers `WONT`/`DONT` with nothing: replying to those starts an exchange that never ends, which `do_telnet_options()` did, inherited from netcat. A 5250 or 3270 host still cannot be checked beyond its port: it sends no text at all, waiting for a terminal to negotiate itself.

## Supersedes #461–#486

That stack went further — fourteen directives to this one's five, with branching, named states, per-host credentials and framing. Scope decided it, not correctness: one reviewable change on #453/#454 rather than fifteen drafts read bottom-up. The branches are untouched on the remote if that machinery is wanted later.

## Milestone

**Not milestoned deliberately.** #453, #454 and #449 are in 4.3.31; this is not, and should not be until someone other than its author has read it. It does not match that milestone's shape — the rest of 4.3.31 is bounded fixes and build changes, this is ~4,000 lines redefining how every TCP service is checked — and the changes above want a release note rather than a changelog line. 4.3.32 looks like the right home once reviewed. It does not block #453 or #454, which stand on their own.

## Depends on #453 and #454

A conversation writes several times. #454 fixes `socket_write()` discarding a payload on `SSL_ERROR_WANT_WRITE`; #453 makes a handshake wait on the event it asked for, which matters more once `start tls` handshakes **mid-session**. For a plaintext dialogue neither is needed: `sendagain` is only set inside `if (item->sslrunning)`.

## Tests

32 network tests (21 added here), 3 C harnesses, 35 fixtures. Every assertion was checked against an unfixed build — several review findings were tests that passed while verifying nothing, so "it passes" is not evidence on its own.

- **Scripted peers**: ordering, multi-line and coalesced replies, partial lines, the STARTTLS upgrade (EHLO either side, certificate reached, failed handshake not green, plaintext not carried across), telnet options spread across reads, a server that waits to be answered, `WONT` not answered, each refusal, verdicts, bounds.
- **Recorded transcripts**: `protocol-realworld.sh` copies `xymonnet/protocols.cfg` **unmodified** and replays sessions from Gmail, Office 365, GNU, Rebex and gmane, reading the peer's own log so a wrong command or an early stop fails even when the colour is green.
- **Live servers**, while preparing this: `smtps`/`imaps`/`pop3s` on Gmail, `ldaps` on freeipa and db.debian.org, `nntps` on eternal-september and individual.net, `amqps` on three CloudAMQP hosts, `telnet` on telehack and a public BBS — all green, with `sslcert` where TLS is involved. `telnets` has no reachable non-5250 server and is the one entry verified only by the pattern its siblings confirm.
- **ASan**: the parser runs for real and every `send`, `expect` and `until` is read back by the length recorded with it.
- **Invariants**: every entry declares a port; no port comes from `/etc/services`; the compiled table carries no definitions; every entry is driver-run; the stepless set is exactly as expected; no shipped entry uses an option a step replaces; and parser, manual and file header list the same options — that last had drifted twice, and an unknown option now refuses an entry, so stale documentation would take a service down.

All 32 pass locally and CI is green. A full local `tests/testsuite` run is not a reliable number on a developer machine: the suite leaks SysV semaphores and shared memory, so after enough runs the box sits at its IPC limit and unrelated server tests cannot create a status channel. Pre-existing, and CI in a clean environment is the reading to trust.

## Documentation

`protocols.cfg(5)` described one send and one match, and **its opening example was the pregreet itself**. It now shows a conversation, documents `until`, `start tls`, `start iac` and the empty expect, says what happens when an entry or the whole file cannot be read, and adds *WHAT CANNOT BE EXPRESSED* — anchored matching, binary dialogues beyond a single exchange, and the absence of branching. `protocols.cfg`'s own header still described the single-exchange model and the `/etc/services` lookup; it was rewritten too.


bonomani:
### Note for the dialogue stack (#461 → #486)

This PR adds `ignore`, and so does the stack (#482). They collide, and **git will not tell you**: merging leaves four conflicts, and resolving them naively still produces a tree that does not build.

Recorded here so whoever rebases the stack onto a `main` containing this does not rediscover it.

**What overlaps**

| | this PR | the stack |
|---|---|---|
| `svcinfo_t` fields | `ignoretext[8]`, `ignorelen[8]`, `ignorecount` | the same three, identical names and types |
| parser | `strncmp(l, "ignore ", 7)` | the same branch |
| applied | after the send -- the reply match, and the keep-reading test | in the driver, on a dialogue wait (`dlg_skip_ignored()`) |

Because the fields sit at different places in the header, a merge keeps **both** declarations and both parse branches: a duplicate-member compile error, and a second `ignore` branch that is unreachable.

**What the stack should absorb**

1. drop its own `ignoretext[8]` / `ignorelen[8]` / `ignorecount` and use the ones already in `svcinfo_t`
2. drop its own `ignore` parse branch
3. gate application on `TCP_DIALOGUE`, which the stack already decides once at load time: `dlg_skip_ignored()` when set, the classic-path `skip_ignored()` otherwise
4. refuse an entry that carries **both** `greeting` and states -- both claim the first read, which is ambiguous and decidable when the file is read

**Why the gating lands there and not here**

`main` has no `TCP_DIALOGUE`; there is only one path, so this PR has nothing to distinguish. The stack is what introduces the second mode, so it is what has to say which mode applies.

Worth noting that a `greeting` entry classifies as *classic* under the stack's own rule -- `greeting` sets a field rather than a step, so `nsend=1`, `nexp=1`, and the first step is the `send`. The two shapes are therefore separable by the flag that already exists, and the semantics are not really in conflict: both say "a message spoken out of turn is not an answer", applied at whatever counts as a wait in each mode. This PR additionally excludes the greeting from it, because IMAP greets with an untagged `* OK` and would otherwise skip its own greeting.

PR #492, opened 2026-09-01
xymonnet: check what allocations return, and copy escaped text by length

bonomani:
Three independent bugs in code that is on `main` today. One commit each; none depends on the others.

## 1. An alias's `send`/`expect` is copied with `strdup()`, its length carried separately

`[a|b]` gives every service after the first its own copy:

```c
walk->rec->sendtxt = strdup(first->rec->sendtxt);
walk->rec->sendlen = first->rec->sendlen;
```

`getescapestring()` resolves `\xNN`, so that text may contain a NUL. `strdup()` stops there; `sendlen` does not. `contest.c` then writes `sendlen` bytes out of the shorter allocation, or matches `explen` bytes against it — a heap over-read past the end of the object.

Needs a `protocols.cfg` using `\x00` in a `send` or `expect` on an aliased entry, so nothing shipped triggers it. The file is operator-editable.

## 2. The banner buffer is grown through an unchecked `realloc()`

```c
item->banner = (unsigned char *)realloc(item->banner, item->bannerbytes+len+1);
...
memcpy(item->banner+item->bannerbytes, buf, len);
```

Assigning `realloc()` straight back loses the old pointer on failure, and the `memcpy()` on the next line writes through NULL. `len` comes from the peer. Reachable only under memory pressure; the recovery is to keep what was collected and stop adding.

## 3. Telnet: the option buffer is unchecked, and the banner leaks

`do_telnet_options()` uses `malloc()` for the option buffer without checking it, on a length the peer controls.

It also replaces the banner by `strdup()`ing over it, leaking whatever was there. The obvious repair — free the old one first — is **wrong**: the read arm sets `item->telnetbuf = item->banner`, so `inp` walks the very buffer being replaced, and freeing first leaves it dangling for the copy to read. Copy first, free after, and clear `telnetbuf`, which otherwise still points at the freed buffer.

## Provenance

Found while working on #488, which is stacked three deep behind #453 and #454. These fix `main` and have nothing to do with that feature, so they are here instead of waiting behind it.

## Checks

- each commit builds on `main` and produces a binary
- `tests/network/` and `tests/buildsystem/` clean, except `rrd-api-compat.sh`, which fails identically on unmodified `main`
- the dialogue stack (#461–#486) does **not** fix any of these: it rewrites the parser and reproduces 1 verbatim (`strdup((char *)txt)` with the length set separately), and leaves 2 and 3 untouched

### DIFF
--- a/lib/netservices.c
+++ b/lib/netservices.c
@@ -218,7 +218,16 @@ char *init_tcp_services(void)
 			if (first) {
 				getescapestring(skipwhitespace(l+4), &first->rec->sendtxt, &first->rec->sendlen);
 				for (walk = first->next; (walk); walk = walk->next) {
-					walk->rec->sendtxt = strdup(first->rec->sendtxt);
+					/*
+					 * By length, not strdup(): getescapestring()
+					 * resolves \xNN, so the text may contain a NUL.
+					 * strdup() stops there while sendlen still says how
+					 * long it was, and contest.c writes sendlen bytes
+					 * out of the shorter copy.
+					 */
+					walk->rec->sendtxt = (unsigned char *)malloc(first->rec->sendlen + 1);
+					memcpy(walk->rec->sendtxt, first->rec->sendtxt, first->rec->sendlen);
+					walk->rec->sendtxt[first->rec->sendlen] = '\0';
 					walk->rec->sendlen = first->rec->sendlen;
 				}
 			}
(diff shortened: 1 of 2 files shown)


bonomani:
Closing: opened on my own initiative and not wanted. The three fixes it carries stay at the front of #488, which is where they were.

Nothing here is lost — #488's first three commits are the same changes (alias `send`/`expect` copied by length, the banner `realloc()` checked, telnet's option buffer checked and its banner no longer leaked), still separable and cherry-pickable from there if they are ever wanted on `main` ahead of the dialogue work.

PR #499, opened 2026-09-03
ci: refuse AI attribution instead of only forbidding it

bonomani:
`AGENTS.md` forbids crediting an AI tool in anything that lands here, and several
tools append these trailers by default. Stating the rule leaves the failure
silent: six commits on `main` carry one already, out of 6762 — a session URL, a
`Co-Authored-By` naming a model, an `Assisted-By` naming Copilot. Nobody typed
any of them.

One workflow, checking the two places a tool writes on its own: the commits a
pull request adds, and the pull request's own title and body. It runs on a push
to `main` as well, which is the one thing a pull request check cannot see — both
permitted merge methods write a new commit whose message the person merging types
after the last check has run, so the only moment it exists is after the merge.
That run reports rather than blocks; the commit is already there.

Two things it does not do. It does not look at history: those six stay where they
are, and rewriting merged history to strip metadata would cost far more than it
buys. And it does not read the lines a pull request adds to files — a trailer
arrives because a tool appended it to a message, while a credit inside a file's
contents is something a person put there, and a person who does that has read
`AGENTS.md` and gone around it. That case cost an awk parser for git's diff
format and a temp-repo harness to prove it, and it still could not see a binary
blob or a file marked `-diff`, so what it refused was the easy half of a case
nobody was making. Both are `AGENTS.md`'s to answer, not a pattern's.

### The pattern

A **prefix** — a credit trailer key, or `Generated with` — **AND a tool name**,
plus the session forms that carry no prefix. Both halves are load-bearing: the
prefix alone would refuse a human co-author and a page generated with mandoc; the
tool name alone would refuse the prose `AGENTS.md` and `CONTRIBUTING.md` have to
contain. The prefix is anchored to the start of a line, because an emitted trailer
opens its line while prose cites the same strings mid-sentence. The tool name is
read in the display name only, up to the `<`: a contributor is not the tool they
share an address with, and reading the whole line refused
`Co-Authored-By: Jane Doe <jane@openai.com>` for her employer's domain.

What that does not separate is a contributor from a tool they share a *name*
with, and `Co-Authored-By: Claude Dupont <claude.dupont@example.fr>` is refused.
That is a cost, not an impossibility: an allowlist of display names does separate
them, built and measured rather than assumed, and it passed five of nine emitted
shapes — `Claude 5 <noreply@anthropic.com>` and `Cursor Composer <noreply@cursor.sh>`
among them — to spare three human names, because a list of model names goes stale
between releases. A refusal a maintainer reads costs less than a miss nobody sees,
in a check that leaves the history it would land in alone. The vendor list is a
floor; `AGENTS.md` governs whatever it misses.

### Verified

Every row run against this head, not read:

| check | result |
| --- | --- |
| 21 refuse / 8 pass fixtures | green |
| 22 pattern mutants: every trailer key, every vendor, both prefixes, the robot marker, the three session forms, the leading whitespace class, the display-name narrowing | red, 22 of 22 |
| the commit range switched from two dots to three | red — a trailer the base branch added blames a pull request that is clean |
| a clean pull request whose base branch moved on carrying a trailer | passes |
| `01e0ce587` on `main`, a real trailer commit, as the head side | refused, naming the commit and quoting the line |
| an empty base or head sha | exits 1 where it would otherwise report `commits clean` |
| this pull request's own commits, and its real title and body | pass |

The history the range test scans is built inside the step, by the same function
that scans the real one.


bonomani:
Merging without a second review, using the maintainers bypass, and saying so as CONTRIBUTING asks.

The row that allows it is *build, CI, test framework* — "yes, when tests cover it or the pull request carries the evidence". This carries both: the `no AI attribution` check runs on this pull request itself and is green, and the pattern has a non-vacuity probe over 22 alternatives, each with a line that fails if that alternative is deleted.

What prompted merging it now rather than waiting longer: the first outside contribution to the packaging arrived with `Co-authored-by: Cursor <cursoragent@cursor.com>` on four commits, which is exactly the shape AGENTS.md forbids and nothing here was catching. The pattern refuses that string and leaves a human `Signed-off-by` alone — checked against both before merging.

Consequence worth naming: open pull requests carrying such a trailer will go red on their next push and need their commits amended.

PR #503, opened 2026-09-05
maint: add an audit script for the upstreaming trackers (#29, #106)

bonomani:
Both tracking issues — #29 (Terabithia patches) and #106 (`devel` commits) — carry an identical **Rule** block. Checking the trackers against it has meant re-deriving the checks by hand each time, which is slow and unreliable: several audits this cycle reported defects that turned out to be artefacts of the throwaway regexes rather than problems in the issues.

`.github/scripts/audit-trackers.py` makes those checks reproducible. It groups them the way the failures actually cluster:

| group | what it checks |
|---|---|
| **structural** | one line on its own — verdict count, checkbox/icon/verdict coherence, delegated-line form, group member counts |
| **relational** | what one line asserts about another — ownership terminates (no delegation loop), a pointer names every claimant, a carrier contains the change, a counterpart names you back |
| **measured** | tracker against code — citations resolve and are reachable from `main` or `devel`, one abbreviation per commit, in-main claims carry evidence, no closed-unmerged PR named as carrying |

The **relational** group is the reason this exists. Every clause in the Rule block constrains a single line, so two lines could each be valid while holding no decision between them — a delegation pointing at a line that points back. Those invariants are now written into the Rule block on both issues as well.

```
.github/scripts/audit-trackers.py --repo .            # live, via gh
.github/scripts/audit-trackers.py --offline 29.md 106.md
```

Exits non-zero on failure, so a workflow can run it later. No shipped file is touched — it sits in `.github/scripts/`, not `build/` or `tools/`.

Both issues pass at the current head.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QHTauwKUURyLoiL1zLo62h

### DIFF
--- a/.github/scripts/audit-trackers.py
+++ b/.github/scripts/audit-trackers.py
@@ -1,0 +1,185 @@
+#!/usr/bin/env python3
+"""Audit the Xymon upstreaming trackers (#29 Terabithia patches, #106 devel commits)
+against the Rule block both issues carry.
+
+    ./audit-trackers.py [--repo DIR] [--tera DIR] [--offline A.md B.md]
+
+Exit 0 when every invariant holds, 1 otherwise.  Checks are grouped as:
+  structural  - one line in isolation
+  relational  - claims one line makes about another   <- where the real defects live
+  measured    - the tracker against the code (needs --repo, and --tera for #29)
+
+Filter notes, learned the hard way; changing them causes false positives:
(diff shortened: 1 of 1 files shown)


bonomani:
Closing — this is maintainer tooling for the #29/#106 trackers rather than something the project needs to carry. It lives on `bonomani/xymon`, branch `maint/audit-trackers`.

Issue #519, opened 2026-09-14
licensing: no holder on new files, six licence statements giving three answers, and advertising clauses in shipped code


bonomani:
**Tracking issue.** The audit is below; each problem moves into its own
issue as it is opened, listed under [Tracked issues](#tracked-issues).
The first is #524.

Every file added since the GitHub migration carries a licence marking with no
copyright holder; the project states its own licence in six places, giving
three different answers; and code carrying advertising clauses is compiled into both the server
and the client.

## Status

Audit of all 928 tracked files at 464af920b. Method, so the numbers can be
reproduced: "added since 2026-01-01" uses `git log --follow --diff-filter=A`,
so a file renamed this year counts from when its content first appeared;
markings are looked for anywhere in the file, not only in a header; and a
holder counts whether it is written `Copyright (C) Name Year` or bare
`(C) Name Year`.

| | Files |
|---|---|
| Added since 2026-01-01 | 141 |
| **of those, containing the word "copyright" anywhere** | **0** |
| — carrying an `SPDX-License-Identifier: GPL-2.0-or-later` header | 120 |
| — carrying no licence marking at all | 21 |
| Older files (787) carrying a copyright notice | 302 |
| — naming Henrik Størner and nobody else | 269 |
| — naming someone or something else, as well or instead | 32 |
| — `COPYING`, which names the Free Software Foundation | 1 |
| Older files with no copyright notice at all | 485 |

## Problems, by urgency

Three things are wrong, and none of them is complicated.

1. **Our licence label does not match our contents.** `COPYING`, GitHub and
   xymon.com all say "GPL version 2", full stop. That is not what is in the
   tree.
2. **Some of the code belongs to other people, and we never wrote it down
   properly.** Two files try to list it. They disagree with each other and
   each gets something wrong.
3. **Nothing written in the last twelve years says who wrote it.** No file
   added since 2014 names a copyright holder.

Seventeen distinct problems follow from those three. The classes below are
about consequence, not effort.

**Urgent — and this does not mean a release is blocked.** Nothing in this
issue blocks a Debian or FreeBSD release. Debian has shipped Xymon for some
fifteen years with these files in it and its licence reviewers looked at them;
FreeBSD ships it too. Nobody is waiting on us.

These three are urgent in a narrower and more practical sense: **doing the
other work first makes them worse or harder.** U1 gets harder the moment a
file-marking pass runs over the tree. U2 can be destroyed by an innocent
[...]

PR #520, opened 2026-09-14
tree: name a copyright holder in every file added since the migration

bonomani:
Roland pointed out on the mailing list that `tests/lib/fake-xymond.c` carries
`SPDX-License-Identifier: GPL-2.0-or-later` and no copyright line, and asked
whether a file can be GPL-2+ with no copyright owner. The audit in #519 shows
it is not one file: **all 138 files added since 2026-01-01 lack a holder**, 119
with an SPDX line and 19 with no licence marking at all.

A GPL notice with no holder records the licence but not who granted it, and not
who answers for the code. This fixes the first of the four items in #519.

## What this does

Each file gains `Copyright (C) 2026 <name>`, where the name is the author of
the commit that added the file, from `git log --diff-filter=A`:

| Author | Files |
|---|---|
| Bruno Manzoni | 125 |
| Markus Häll | 10 |
| Roland Rosenfeld | 2 |
| Scot Kreienkamp | 1 |

The 19 files that had no marking at all also gain the `SPDX-License-Identifier:
GPL-2.0-or-later` line the other 119 already carried.

## What a reviewer cannot infer from the diff

- **Names, no e-mail addresses.** The older headers carry `Henrik Storner
  <address>`; these do not. Publishing four people's addresses is not mine to
  decide. Say the word and they go in.
- **Three of you are named without being asked.** Markus, Roland, Scot — this
  asserts that you hold copyright on files you wrote. If you would rather not
  be named, or want a different name or an employer, say so and I will change
  it before merge.
- **`tree:` is not one of the components `CONTRIBUTING.md` lists.** 122 of the
  138 files are under `tests/`, but 16 are not, so `tests:` would be wrong.
  There is no component for a tree-wide metadata change; if we want one, that
  is a rule to add rather than a label to fudge here.
- **This does not settle version 2 against version 2-or-later.** Every line
  added here says or-later, which is what these files already said — the
  inconsistency with `COPYING` is item 2 in #519 and is for the packagers to
  decide, not for this pull request.
- The project forbids crediting an AI tool in anything that lands here
  (`AGENTS.md`), so there is no such trailer on the commit.

## Verified

- Full CI green on the current head: both server and client builds, the shell
  regression suite, and the manual-page check.
- `./tests/testsuite` locally on an unbuilt tree: 43 passed, 52 skipped, 0 failed.
- `bash -n` on all 106 modified shell scripts and on `tests/testsuite`;
  `python3 -m py_compile build/manpage-html.py`; all four workflow files parse
  as YAML.

The first revision of this pull request failed CI, and the reason is worth
recording. I argued that a comment-only change could not break a build and did
not compile it. Three harnesses open a block comment on the SPDX line and
continue it:

```c
/* SPDX-License-Identifier: GPL-2.0-or-later
 *
 * Driver for the real generate_wml_statuscard(), ...
```

Inserting a self-contained `/* Copyright ... */` under that line closed the
block early and left the rest of the header as code. The three tests that
compile a harness against source extracted from the tree —
`tests/web/wmlgen-path-bounds.sh`, `tests/rrd/rrddef-empty-list.sh`,
`tests/server/xymond-gap-validity.sh` — caught it. Those three files now carry
the copyright as a `" * "` continuation instead.

Refs #519


bonomani:
Moved to draft. Not abandoned — it is ahead of a question it depends on.

The change names a copyright holder in the 138 files added since the
migration. That is mechanical and the diff is easy to read, but which name
belongs on a file is a question about the project, not about the patch:

- every copyright line in the tree today names one person, with year ranges
  ending in 2011. Fifteen years of contributions carry no holder at all, so
  "who holds copyright here" has not been answered for a long time and this
  pull request would answer it for a third of the tree by default.
- the tree states two different things about its own licence. `COPYING` is
  the GPL v2 text and the old headers say "version 2", while the 119 marked
  new files say `GPL-2.0-or-later`. Distribution is unaffected, but adding
  holders without settling that writes the ambiguity into more files.
- 19 of the new files carry no licence marking at all, mostly the
  infrastructure ones. Whether they should be marked, and as what, is part
  of the same question.

So it waits on the licensing thread rather than on review. It will come out
of draft when there is an answer to apply, and the diff will be regenerated
then rather than rebased -- by that point the file list will have moved.

No review needed meanwhile.

bonomani:
Heads-up on a file collision, not a review comment: #547 changes `tests/buildsystem/testsuite-run-disposal.sh` substantially (it is the #539 runtime fix). This PR adds one copyright header line to the same file.

Whichever of the two lands second will need a rebase. It is a trivial one — a header line against the body of the file — but worth knowing rather than discovering at merge time.

PR #521, opened 2026-09-14
rpm: ship /etc/xymon-client/clientlaunch.d in both packages

skreienk:
The client is missing the creation of directory  /etc/xymon-client/clientlaunch.d, which is already created in the debian distribution, so this brings it to parity as well as creating the missing directory.


skreienk:
Found a problem, will redo.

PR #522, opened 2026-09-14
RPM: Create missing clientlaunch.d

skreienk:
The Debian distro already has this directory.  The RPM misses it, so this brings it into parity.


skreienk:
I will reopen on the correct spec file.

PR #523, opened 2026-09-14
RPM: reload a running httpd after installing the Apache config for Xymon

skreienk:
Xymon already places the httpd config, but it does not reload it.  This will cause the httpd to load the config.


skreienk:
I will reopen on the correct spec file.

PR #527, opened 2026-09-20
loadhosts: name the top-level page /, so PAGE= selects it instead of everything

### LINKED ISSUE #526: xymond_client: PAGE= rules in analysis.cfg apply to every host on the top-level page
## The problem

A `PAGE=`-qualified section of `analysis.cfg` applies to every host on the Xymon
front page, whatever page it names. Hosts on named pages filter correctly, so the
report reads as "the main page ignores `PAGE=`".

Reported from a live server: a `PAGE=DeltaChat` section carrying eight `PROC`
checks was applied to a host that is not on that page, which then went red on
every one of them while its own processes passed.

```
⚠ chatmail-turn (found 0, req. between 1 and 1)
⚠ chatmail-metadata (found 0, req. between 1 and 1)
⚠ doveauth (found 0, req. between 1 and 1)
⚠ opendkim (found 0, req. 1 or more)
⚠ nginx (found 0, req. 1 or more)
⚠ master (found 0, req. 1 or more)
✓ ntpd (found 1, req. 1 or more)
✓ sshd (found 1, req. 1 or more)
✓ cf-execd (found 1, req. between 1 and 2)
```

## Resolved on the reporting install

With the fix applied, xymond restarted and the affected clients forced to
re-submit, the unwanted `PROC` checks cleared from the main-page host, and the
hosts actually on the `DeltaChat` page kept theirs. Both halves matter — a fix
that simply stopped the rules applying anywhere would clear the first.

Also seen there before patching, and worth recording because it is the same root
cause seen from the other side: a host on the front page *and* on `DeltaChat`
reported its pagepath list as `dcpx.feld.me|DeltaChat,` — the trailing comma is
the front-page element failing to survive the list. Depending on ordering it
either dangles like that or vanishes outright, taking the host's front-page
membership with it.

## Confirmed on the reporting install

The report says the host is "100% not on the DeltaChat page", which is equally
consistent with it being on the top-level page (this defect) or on some *other*
named page (a different one, since named pages filter correctly). The reporter
was asked to settle it:

```
xymon <server>:1984 'xymondboard page=^$ fields=hostname' | sort -u
```

The contaminated host is in that list — it is on the unnamed top page. So the
mechanism below is the one producing the symptom, not merely a defect that
happens to exist.

Worth noting for anyone hitting this: *every* host in that list was collecting
the same rules. It is usually only noticed on the one where the missing
processes are conspicuous.

## Mechanism

`ruleset()` in `xymond/client_config.c` tokenises the host's pagepath list and
rejects a rule when a token fails to match:

```c
pgmatchres = pgexclres = -1;
pgtok = strtok(pagenamecopy, ",");
while (pgtok) { ... }
if (pgexclres == 1) continue;
if (pgmatchres == 0) continue;   /* -1 is not 0 -> the rule is kept */
```

The top-level page's pagepath is the empty string (`lib/loadhosts.c:301`,
`pghead->pagepath = strdup("")`). `strtok("")` returns `NULL`, the loop body never
runs, `pgmatchres` stays at its `-1` sentinel, and `-1 != 0`, so the rule is kept.

`analysis.cfg.5:529` documents the front page as having the fixed name `/`, and
`lib/loadalerts.c:915` implements exactly that for `alerts.cfg`
(`strdup((*alert->location == '\0') ? "/" : alert->location)`, backported in
`4abd6f3e0`, 2016). `client_config.c` never got the same treatment, so `PAGE=/`
appears to work only because `PAGE=`*anything* matches the front page.

## Second defect in the same block

`pagenamecopy` is `strdup`'ed once outside the rule loop and walked with
`strtok()`, which writes NULs over the separators. After the first rule tokenises
it the commas are gone, so every later rule sees only the **first** pagepath of a
host listed on several pages.

Measured with a harness against the real `client_config.c`
(`PAGE=deltachat` / `LOAD 99.0 99.9`; 5.0 is the default that means "no rule matched"):

| host | pagepaths | before | after | expected |
|---|---|---|---|---|
| fronthost | *(top page)* | 99.0 | 5.0 | 5.0 |
| homehost | `home` | 5.0 | 5.0 | 5.0 |
| dchost | `deltachat` | 99.0 | 99.0 | 99.0 |
| multihost, rule alone | `home,deltachat` | 99.0 | 99.0 | 99.0 |
| multihost, one `DISK * 90 95` line above it | `home,deltachat` | **5.0** | 99.0 | 99.0 |

Adding an unrelated `DISK` line at the top of the file changes whether a `PAGE=`
rule matches.

Both date from `2564fb6b6` (2007), which introduced the multi-page list.

## Proposal

Name the front page `/` at the point of comparison, the way `loadalerts.c`
already does; require a `PAGE=` rule to match rather than merely not fail, so a
comparison that never happened cannot pass for a successful one; and give each
rule its own copy to tokenise.

## Workaround

Named pages filter correctly. Give front-page hosts a real `page` in `hosts.cfg`,
or scope the section with `HOST=` / `CLASS=` / `DISPLAYGROUP=`, which are all
evaluated before the page logic. `EXPAGE=` does not help: it hits the same
sentinel and never excludes a front-page host.

## Out of scope

`XMH_ALLPAGEPATHS` (`lib/loadhosts.c:555-563`) appends each pagepath with a comma
separator only when the buffer is already non-empty, so an empty top-page path
contributes nothing and emits no separator: a host on both the front page and
`home` reports `home`, losing its front-page membership. Same root cause, but
fixing it means changing the format every consumer reads — `webaccess.c` ACLs,
`xymond` board matching, `xymond_alert` — so it wants its own change.

bonomani:
> **Merge order: #531 → #527 → #532 → #533.** All four merge clean in any order (verified: merged, built, suite green), so this is a preference, not a constraint. Out of order, the only effect is that `alerts.cfg.5` (corrected here) and `xymon.1` (corrected in #533) disagree on the top page's name until both land.

The top-level page has no pagepath of its own — it is the empty string, from the
pagelist head in `lib/loadhosts.c:301` — and an empty string is not a name that
can be matched. Two places worked around that and one did not, so a page filter
on the front page did the opposite of what it said.

`ruleset()` (`xymond/client_config.c`) tokenised the host's pagepath list and
rejected a rule when a token failed to match. With no pagepath there was no
token, nothing was compared, `pgmatchres` kept its `-1` sentinel, and `-1 != 0`
let the rule through: **every `PAGE=`-qualified rule in `analysis.cfg` applied to
every host on the front page.** Such a host collected a foreign page's `PROC`
rules and went red on processes it does not run, while hosts on named pages
filtered correctly — which is what made it read as "the main page ignores
`PAGE=`". Reported from a live server, with the evidence, in #526.

`XMH_ALLPAGEPATHS` (`lib/loadhosts.c:577` after this change) had the mirror-image defect: it
appended each pagepath raw, and an empty element cannot be told from no element
— it either vanished (the separator is only written for a non-empty buffer) or
left a trailing comma a tokeniser drops. A host on the front page *and* a named
page reported only the named one, so no `PAGE=/` rule could reach it, in
`analysis.cfg` or in `alerts.cfg` — whose `criteriamatch()` has named that page
`/` since `4abd6f3e0` (2016) without being able to repair the list it is handed.

This names the page `/` where it is produced, which is what `analysis.cfg(5)` and
`alerts.cfg(5)` document, and requires a `PAGE=` rule to *match* rather than
merely not fail. `alerts.cfg.5` still told the reader to write `PAGE=%^$` for the
front page, which `criteriamatch()` has been unable to match for a decade; it now
says `/` as `analysis.cfg.5` does.

The same block shared one `strdup`'ed buffer across the whole rule list and
walked it with `strtok()`, which writes NULs over the separators: after the first
page-filtered rule the commas were gone and every rule behind it saw only the
first pagepath. Whether a `PAGE=` rule matched a multi-page host depended on how
many such rules preceded it in the file. Each rule now tokenises its own copy,
and only when it has a page filter to apply.

Both `analysis.cfg` defects go back to `2564fb6b6` (2007), which introduced the list.

### Evidence

`analysis.cfg`, via a harness against the real `client_config.c`
(`PAGE=deltachat` / `LOAD 99.0 99.9`; 5.0 is the default meaning "no rule matched"):

| host | pagepaths | before | after |
|---|---|---|---|
| fronthost | *(front page)* | **99.0** | 5.0 |
| bothhost | front page + `home` | **99.0** | 5.0 |
| homehost | `home` | 5.0 | 5.0 |
| dchost | `deltachat` | 99.0 | 99.0 |
| multihost, one page-filtered rule above it | `home,deltachat` | **5.0** | 99.0 |

`alerts.cfg`, via the real `xymond_alert --test` (`PAGE=/` → `MAIL`):

| host | before | after |
|---|---|---|
| fronthost | 1 recipient | 1 — unchanged |
| bothhost (front page + `home`) | **0** | **1** |
| homehost | 0 | 0 — unchanged |

Each of the three defects was reintroduced separately against the new tests, and
each is caught by a named assertion:

| reintroduced | caught by |
|---|---|
| `ruleset()` stops naming the front page | `PAGE=/ did not reach the top-level page through a XMH_PAGEPATH caller` |
| `ALLPAGEPATHS` swallows the empty element | `the top-level page lost its name: XMH_ALLPAGEPATHS must report it as /` |
| one `strtok` buffer shared across rules | `PAGE=deltachat missed a host on home,deltachat once another page-filtered rule preceded it` |

Suite: **95 passed, 0 skipped, 2 failed** — `tests/network/smtp-pregreet-behaviour.sh`
and `tests/server/disk-unavailable-honest-message.sh`, both failing identically on
unmodified `main` in the same environment (checked by rebuilding without the change
and rerunning the two). `tests/buildsystem/test-suite-portability.sh` passes, the
regenerated `alerts.cfg.5.html` leaves the `.TH` line alone, and the changed C
compiles warning-free under the tree's `-Wall`. Built and run on Linux only.

Both tests fail correctly when the binary under test is broken: `alerts-page-filter.sh`
originally wrapped `xymond_alert --test` in a command substitution and called
`fail` inside it, which exits only the subshell — the script carried on, reported
the missing output as an assertion mismatch, and exited 0, so a crashed binary
could have been recorded as a pass. It now assigns first and fails in the main
shell; verified by pointing it at a broken binary (`rc=1`, message naming the
fixture).

**Verified on the reporting deployment.** Everything else here is a synthetic
fixture on one Linux box, so the reporter was asked for the two things only a
live install can answer.

*First, before patching* — is the contaminated host actually on the unnamed top
page, or on some *other* named page, which filters correctly here and would mean
a second defect this PR does not address?

```
xymon <server>:1984 'xymondboard page=^$ fields=hostname' | sort -u
```

It is on the top page. The alternative reading is ruled out.

*Then, after patching* — he applied the batch, restarted xymond and forced the
affected clients to re-submit, so nothing was stale from the previous
five-minute interval:

- the unwanted `PROC` checks **cleared** from the main-page host
- the hosts actually on the `DeltaChat` page **still have theirs**

The second is the one that matters: a fix that merely stopped the rules applying
anywhere would satisfy the first on its own.

Caveat on attribution: all four PRs in this batch were applied together, so this
verifies the batch. The `analysis.cfg` behaviour is this PR's alone — #533 does
not touch `xymond/client_config.c` — and an accidentally half-patched state
settles it. With #533 applied and this PR absent, the reporter still saw the
leak, and `XMH_ALLPAGEPATHS` still read empty.

**The second defect, sighted in production.** Before patching, a host on the
front page *and* on `DeltaChat` reported:

```
dcpx.feld.me|DeltaChat,
```

That trailing comma is the empty top-page element failing to survive the list —
the defect described above, which until now rested entirely on fixtures. Which
variant appears depends on ordering: the comma dangles, or the element vanishes
and the host silently loses its front-page membership. With this PR the same
host reads `/,DeltaChat`.

The reporter, independently and before seeing this PR, proposed the same fix:
*"perhaps we should somehow try to make it not be an empty string but instead be
a root path like '/' for the main page?"*

### What a reviewer cannot infer from the diff

- **Front-page hosts lose rules they were silently collecting.** That is the
  point, but it changes what such a host reports on upgrade.
- **`xymond_client --local` changes in both directions, and one of them removes
  checks.** `localhostinfo()` pins all hosts to the top page
  (`lib/loadhosts.c:486`), which local mode now reads as `/` rather than as no
  page at all. `PAGE=<name>` rules stop applying, where before they all did
  (verified: 99.0 → 5.0) — that is the intended direction. But `EXPAGE=/` now
  excludes *every* host there, where before it excluded none, so a local-mode
  threshold written behind one stops being applied and that check stops alerting
  (verified: `EXPAGE=/ / LOAD 66.0 66.9` gives 66.0 on `main`, 5.0 here). Losing a
  rule you were relying on is the more disruptive half, so it is worth calling out
  above the other. Release-note wording: *"`analysis.cfg` rules qualified with
  `PAGE=` no longer apply to hosts on the Xymon front page, and under
  `xymond_client --local` the whole host is now treated as being on the front
  page: `PAGE=<name>` rules no longer apply there and `EXPAGE=/` now excludes
  everything. Scope such rules with `HOST=`, `CLASS=` or `DISPLAYGROUP=`."*
  **This reaches client machines, not the server.** `xymond_client` is installed
  by `client/Makefile:88 install-localclient`, and `client/xymonclient.sh:101`
  runs it when the client was configured with `LOCALCLIENT=yes` — so the change
  lands on a *client* upgrade, in a different maintenance window from the server
  one, and bites hardest where `localclient.cfg` was copied from the server's
  `analysis.cfg`.
- **The `pagepath` field in `@@status` channel messages** now carries `/` rather
  than the empty string for front-page hosts (`xymond/xymond.c:780`, `893`,
  `2279`). In-tree consumers are unaffected — `xymond_alert` feeds it to
  `criteriamatch()`, which mapped `""` to `/` already — but a third-party worker
  module testing that field for emptiness would see the change.
- **`webaccess.c:80` is unaffected, by inspection rather than by test:** it
  truncates each element at the first `/`, so `"/"` becomes `""` — the same ACL
  key the old empty string produced. Not exercised by the suite.
- **`xymondboard`'s `page=` filter is untouched:** it matches on
  `xmh_item_multi(XMH_PAGEPATH)` (`xymond/xymond.c:3468`), not `ALLPAGEPATHS`.
- **Two lines are deliberately unreachable and untestable.** Once the page always
  has a name, `pgmatchres` can only be 0 or 1, so requiring `!= 1` is the same
  test as `== 0`; it is kept as the half that does not depend on every caller
  remembering to name the page, and the code comment says so. The `pagename &&`
  NULL guard is likewise unreachable — both callers check `hinfo` first — and
  replaces a `strdup(NULL)`.
- **`xymond/xymond.c:2575` has a pre-existing unbounded write this change
  nudges.** `handle_notify()` does `malloc(1024 + strlen(msgtext))` then
  `sprintf("%s|%s|%s\n%s", hostname, testname, XMH_ALLPAGEPATHS, msgtext)` — the
  1024 bytes of slack must cover the hostname, the testname and the whole
  comma-separated pagepath list. This adds one element plus a separator for every
  host on the front page as well as named pages, so it moves an already-unbounded
  write in the wrong direction. Not fixed here; filed as #528.
- **Left alone, and now more exposed:** `ruleset()` caches per hostname while its
  callers disagree about which page item to pass — `get_cpu_thresholds` and the
  PROC checks use `XMH_ALLPAGEPATHS`, `get_inode_thresholds` and
  `get_paging_thresholds` use `XMH_PAGEPATH`. For a multi-page host the first
  getter called freezes the ruleset for the rest, so a `PAGE=` rule is accepted or
  rejected depending on call order. Verified unchanged by this PR (identical
  before and after); it wants its own change, since the fix is a decision — the
  pagename joins the cache key, or the callers agree on one item.
- **Follow-ups filed rather than folded in:** #528 (unbounded `sprintf` above),
  #529 (`criteriamatch()` still treats an undecided `PAGE=` filter as a match —
  the same sentinel this removes from `analysis.cfg`), #530 (`/` is honoured by
  `analysis.cfg` and `alerts.cfg` and ignored by `xymondboard` and the event,
  notification, acknowledgement and perfdata page filters).

Fixes #526


bonomani:
Pushed `956c9a985`: `tests/server/alerts-page-filter.sh` wrapped `xymond_alert --test` in a command substitution and called `fail` inside it. `fail` exits, but an exit inside `$(...)` ends only the subshell — the script carried on, reported the missing output as an assertion mismatch, and exited 0, so a crashed binary could have been recorded as a pass. It now assigns first and fails in the main shell, with the message naming which fixture was in play.

Found while reviewing #532, which had the same shape. No change to `lib/loadhosts.c` or `xymond/client_config.c`.

bonomani:
Cross-reference for whoever merges these: #533 changes what `page=` filters accept on five more surfaces and documents the `/` name in `common/xymon.1`. It deliberately does **not** touch `xymond/alerts.cfg.5`, because this PR already rewrites that paragraph — so there is no conflict, but after both land it is worth checking that `alerts.cfg.5`, `analysis.cfg.5` and `xymon.1` all say `/` and none still recommends `%^$`.

Also relevant to this PR: on the regex-based filters `matchregex()` refuses only NULL, not an empty string, so `page=^$` used to reach top-level hosts and no longer does. #533 carries the release-note wording for that.

feld (approved):
Tested and discussed in chat with Bruno

PR #533, opened 2026-09-21
matching: give the top-level page one name the page filters match it by

### LINKED ISSUE #530: The top-level page's name / is honoured by analysis.cfg and alerts.cfg, and ignored by every other page filter
`analysis.cfg(5)` and `alerts.cfg(5)` both document the top-level page as having
the fixed name `/`, and after #527 both honour it. Every other page filter in the
tree does not, because they read `XMH_PAGEPATH`, which still returns the empty
string for that page:

| surface | call site | can select the front page? |
|---|---|---|
| `analysis.cfg` `PAGE=` | `xymond/client_config.c:373` | yes, after #527 |
| `alerts.cfg` `PAGE=` | `lib/loadalerts.c:916` | yes, since 4abd6f3e0 |
| `xymondboard page=` | `xymond/xymond.c:3470` | **no** |
| event log page filter | `lib/eventlog.c:196`, `:214` | **no** |
| notification log page filter | `lib/notifylog.c:223`, `:238` | **no** |
| acknowledgement log page filter | `lib/acknowledgementslog.c:256`, `:271` | **no** |
| `perfdata.cgi` page filter | `web/perfdata.c:407` | **no** |

All five of the "no" rows call `xmh_item_multi(hinfo, XMH_PAGEPATH)` and match the
result directly, and `XMH_PAGEPATH` returns `host->page->pagepath` unchanged
(`lib/loadhosts.c:540`) — which for the top-level page is `""`, set at
`lib/loadhosts.c:301`. `namematch()` and `matchregex()` both refuse an empty
needle, so no pattern selects those hosts, and `/` in particular does not.

## The consequence

An admin who reads the manual page writes `PAGE=/`, sees it work in
`analysis.cfg` and `alerts.cfg`, and then writes the same thing in
`eventlog.cgi`, `notifications.cgi` or a `xymondboard page=/` query, where it
silently selects nothing. The name is authoritative in two places and unknown in
five.

## Why it is a separate change

#527 named the page at two of the three producers — `XMH_ALLPAGEPATHS`, and
`ruleset()` for the callers that pass `XMH_PAGEPATH`. Naming it at
`XMH_PAGEPATH` itself would reach the rest, but that accessor also feeds page
*rendering* and URL construction — `web/findhost.c:201-239`,
`web/datepage.c:128`, `web/svcstatus-info.c:1005-1022` build hyperlinks out of
it, where an empty string is the correct relative path and `/` is not. So this
cannot be a one-line change to the accessor; it needs either a separate
"name for matching" accessor that the seven filter sites call, or a shared
helper they all pass their token through.

That choice is the reason this is its own issue rather than more of #527: the
same `"" -> "/"` mapping is now written out at `lib/loadhosts.c:577`,
`xymond/client_config.c:373` and `lib/loadalerts.c:916`, and the fix worth
making is the one that leaves exactly one copy — not a fourth.

bonomani:
> **Merge order: #531 → #527 → #532 → #533.** All four merge clean in any order (verified: merged, built, suite green), so this is a preference, not a constraint. Merged before #527, `alerts.cfg.5` still recommends `%^$` while `xymon.1` here says `/` — the two manual pages disagree until #527 lands.

The top-level page has no pagepath of its own: it is the empty string
(`lib/loadhosts.c:301`). What that costs depends on how the filter compares.

`namematch()` refuses an empty needle outright, so `alerts.cfg`'s `PAGE=` could
not select that page at all — which is why `criteriamatch()` has mapped `""` to
`/` by hand since `4abd6f3e0`. `matchregex()` and the raw `pcre2_match()` calls
in the log filters refuse only NULL, so an empty subject *is* matchable:

| pattern | subject `""` | subject `"/"` |
|---|---|---|
| `.*` | 1 | 1 |
| `^$` | **1** | 0 |
| `/` | 0 | 1 |
| `^/` | 0 | 1 |

So `xymondboard page=^$` reached those hosts and `page=/` did not — the opposite
of what `analysis.cfg(5)` tells an admin to write, and undocumented besides.

**An earlier revision of this description said no pattern could reach the top
page. That was wrong**, and it was the justification for the whole change. The
defect is narrower: the *documented* name did not work on five of six surfaces,
and the page was reachable only by an undocumented spelling.

| surface | `page=/`-style before | after |
|---|---|---|
| `alerts.cfg` `PAGE=` | ✅ (hand-written mapping) | ✅ (same mapping, shared) |
| `xymondboard page=` | ❌ | ✅ |
| event / notification / acknowledgement log, **both** include and exclude | ❌ | ✅ |
| `perfdata.cgi` | ❌ | ✅ |

### The break this takes

`^/$` starts selecting the top page; **`^$` stops**. Verified live against a
two-level fixture:

```
page=/     -> [deephost fronthost]    unanchored: also every nested page
page=^/$   -> [fronthost]             the spelling to use
page=^$    -> []                      used to select fronthost
```

`alerts.cfg.5` on `main` actively tells admins to write `%^$`, so this breaks a
filter written from the shipped documentation. Taken deliberately: one name, the
documented one, on every surface. `common/xymon.1` documents it here, including
that these filters are unanchored — `/` alone matches every pagepath containing a
separator, exactly as `sub` matches `subpage`. `alerts.cfg.5` is corrected in
**#527**, which rewrites the same paragraph; neither PR touches the other's man
page, so they do not conflict, and whoever merges second should check that both
now say `/`.

*Release note:* `page=` filters name the Xymon frontpage `/` — write `page=^/$`
on `xymondboard`, `eventlog.cgi`, `notifications.cgi`, `acknowledgements.cgi` and
`perfdata.cgi`. `page=^$`, which worked before, no longer selects it.

### Why a helper and not a second accessor

The issue proposed either. A `XMH_PAGEPATH_MATCH` accessor would carry the
mapping automatically through `xmh_item_multi()`'s NULL terminator, which is a
real advantage — but it costs an enum value, an `xmh_item_name[]` entry, a
`switch` case, and permanently two ways to ask one question, so the next person
writing a page filter can pick the wrong one and nothing complains.

`XMH_PAGEPATHTITLE` (`lib/loadhosts.c:547-549`) already maps an empty pagetitle
to `"Top Page"` inside `xmh_item()` — the same shape, so the cost I first called
prohibitive is already paid once in that `switch`. What the accessor would *not*
have saved is edits: each call site still has to name the new item, so the count
is identical. Its one real advantage is that the mapping rides through
`xmh_item_multi()`'s NULL terminator automatically.

The deciding evidence is that **all six sites use the pagepath only for
matching** — none of them displays it or builds a URL from it. That is what made
the one-line-per-site cost the cheaper half of the trade. The sites that *do*
build hyperlinks from `XMH_PAGEPATH` (`web/findhost.c:201-239`,
`web/datepage.c:128`, `web/svcstatus-info.c:1005-1022`) are untouched, and `""`
stays correct for them — which is also why changing the accessor itself was never
an option.

### Where it lives, and why not beside the accessor

`lib/matching.c`, next to `namematch()` and `matchregex()` — their refusal of an
empty needle is the reason it exists.

`lib/loadhosts.c` was the first choice and is wrong: `loadhosts.o` is in
`libxymoncomm.a` only, while `eventlog.o`, `notifylog.o`,
`acknowledgementslog.o` and `loadalerts.o` are in **both** archives. Putting it
there left four objects in `libxymon.a` depending on a symbol not in that
archive. It links today only because every in-tree consumer of `libxymon.a`
(`web/`, `xymongen/`, `xymonnet/`) happens to link `libxymoncomm.a` as well.
`matching.o` is in both, so the question does not arise.

### Evidence

`tests/server/board-page-filter.sh` drives a real `xymond` over a **two-level**
page fixture — without the nested page it cannot see that these filters are
unanchored. It pins `^/$` selecting the top page and neither the named nor the
nested one, unanchored `/` still reaching the nested page, and `^$` selecting
nothing. Mutation-tested:

| mutation | caught |
|---|---|
| board filter back to the raw pagepath | ✅ |
| helper stops naming the top page | ✅ |

Suite: **94 passed, 0 skipped, 2 failed** — `tests/network/smtp-pregreet-behaviour.sh`
and `tests/server/disk-unavailable-honest-message.sh`, both failing identically on
unmodified `main`. `tests/buildsystem/test-suite-portability.sh` passes,
`git diff --check main..HEAD` is quiet, the regenerated `xymon.1.html` leaves the
`.TH` line alone. Linux only.

**Verified on a live install.** Everything above is a fixture on one Linux box.
The reporter of #526 ran the four user-facing cases on his own server with the
patch applied:

| check | result |
|---|---|
| `xymondboard page=^/$` | lists the same hosts `page=^$` used to |
| `xymondboard page=^DeltaChat$` | unchanged — named pages were never affected |
| `eventlog PAGEMATCH=^/$` | only main-page hosts |
| `eventlog EXPAGEMATCH=^/$` | *"Only events on other pages are shown."* |

The last row is the one worth having. The exclude halves of the three log
filters are the part of this change with no automated coverage — the test drives
`xymondboard` only — and until now the claim that they were converted correctly
rested on reading the diff. They are now driven end to end.

(On that install the CGI is reached as `eventlog.sh` rather than `eventlog.cgi`,
which is a packaging detail of the deployment, not of this change.)

### What a reviewer cannot infer from the diff

- **`criteriamatch()` loses an unguarded dereference.** `*alert->location` was
  read with no NULL check; the helper takes NULL. Behaviour is otherwise
  identical for every input.
- **The one rule a new page filter must follow** is to apply the helper to the
  value being compared, not to the result of `xmh_item_multi()` — that returns
  NULL to end its iteration and has to keep doing so. It is stated at the
  function definition, and it is the one thing the accessor form would have made
  automatic.
- **The exclude halves are converted too.** An earlier revision mapped only the
  include filter in the three log modules, so `PAGEMATCH=.*&EXPAGEMATCH=/` would
  have shown every top-level host the admin asked to exclude. Both halves now go
  through the helper.
- **Still unconverted, deliberately:** `lib/webaccess.c:75` — no group entry can
  name the top page, so page-based access can never be granted for front-page
  hosts. Same defect class on an access-control surface, filed as **#535** and
  blocked on #527: the two-line fix is *inert* without it, because
  `XMH_ALLPAGEPATHS` is still the empty string and the loop body never runs
  (prototyped and confirmed inert). Also unconverted: `xymond`'s
  `XMH_PAGEPATH=` filter spelling, and the sixteen `XMH_PAGEPATH` /
  `XMH_ALLPAGEPATHS` sites in `xymond/client_config.c`, which #527 handles.
- **Found while reviewing this, unrelated to it:** `ruleset()` caches per
  hostname while six callers pass `XMH_PAGEPATH` and eleven pass
  `XMH_ALLPAGEPATHS`, so for a multi-page host the first getter called decides
  the ruleset for all the rest. Filed as **#534**; unaffected by this change.
- **The title says "the page filters", not "every page filter"**, because of
  `webaccess.c` above.
- **Overlaps #527 by subject, not by code.** #527 names the top page in
  `XMH_ALLPAGEPATHS` and in `ruleset()`; this names it at the six filter sites
  that read `XMH_PAGEPATH`. Cut from `main`, different files apart from
  `lib/loadalerts.c`, where #527 does not touch line 916. Once both land, the
  two hand-written copies in #527 could fold into this helper — a follow-up, not
  a prerequisite.

Fixes #530


bonomani:
Retitled: "every page filter" → "the page filters". `lib/webaccess.c:75` is a page filter this does not reach, and the body already carves it out, so the title should not claim otherwise. Filed as #535, blocked on #527 — verified that the two-line fix is inert without it, because `XMH_ALLPAGEPATHS` is still the empty string and the loop body never runs.

Also filed #534 from this review: `ruleset()` caches per hostname while six callers pass `XMH_PAGEPATH` and eleven pass `XMH_ALLPAGEPATHS`, so for a multi-page host the first getter called decides the ruleset for all the rest. Unrelated to this change and unaffected by it.

feld (approved):
Tested and discussed in chat with Bruno

PR #538, opened 2026-09-21
Move DEFAULT section in analysis.cfg to the bottom again.

### LINKED ISSUE #537: analysis.cfg: analysis.d should be placed before DEFAULT
analysis.cfg.DIST says, that the DEFAULT section should be placed at the end of the file.

But 5513733dc54d5ac6cbb9207077b9094a9222c73d places
```
# Local additions: any file dropped into the directory below is included
optional directory /usr/lib/xymon/server/etc/analysis.d
```
below the DEFAULT section.

I tried to move my local changes from analysis.cfg (placed between CLASS=linux and DEFAULT before) into a new file analysis.d/local.cfg, which resulted in
```
HOST=luv
    DISK /home 96 98
```
not working any more, xymon generates the following warning for host luv:
```
/home (92% used) has reached the WARNING level (90%)
```

Moving the `optional directory analysis.d` block up (between CLASS=linux and DEFAULT section), everything works as expected.

RolandRosenfeld:
PR #222 added the "optional directory @XYMONHOME@/etc/analysis.d" section below the "DEFAULT" section. This doesn't work as expected, because the "DEFAULT" section always needs to be at the bottom.

This change moves the "DEFAULT" section to the bottom again.

It fixes #537.


bonomani (approved):
Confirmed, and the placement is right for a reason worth recording, since the
diff on its own looks like a cosmetic move.

**Why between `CLASS=linux` and `DEFAULT`, and not at the top.** Rules go into
one list in file order (`client_config.c:471`) and `getrule()` returns on the
first match, so a rule is only reached if everything before it failed to match.
`DEFAULT` creates no rule of its own — it clears the section qualifiers
(`:748`, then `:1636`), which is what makes everything under it unqualified and
therefore matching every host. That is why `DISK * 90 95` was eating
`HOST=luv / DISK /home 96 98`.

Moving the include *above* `CLASS=linux` would fix that too, and would be
wrong: the class section is all `IGNORE` rules for `/dev`, `/run`,
`/sys/fs/cgroup` and `/run/user/N`. A local fragment containing a catch-all
such as `DISK * 95 98` would then intercept those pseudo-filesystems, which sit
at 100% on every Linux host, and turn them red everywhere. Worse, an admin with
such a fragment would never receive a new `IGNORE` added upstream later —
exactly the regression the drop-in directory exists to prevent.

So: after the exclusions, before the catch-all. That is the placement here.

**Cause, for the record.** #222 is ours, and it applied one uniform rule —
"each shipped config ends with `optional directory …`". That is right for
`graphs.cfg` (last definition wins), right for the env files (`setenv(…, 1)`,
last assignment wins), harmless for the configs that ship nothing active, and
wrong here. Upstream got it right in 2015: `bff5b5ef5` put this same line
immediately above the `DEFAULT` section, and the analysis.cfg of that era had
no `CLASS=` section, so the question you had to answer did not exist for them.

**Two more files have the same root cause**, both found while checking this
one, and neither is yours to carry: `rrddefinitions.cfg`, where `xtreeAdd`
keeps the first definition of a key so a fragment cannot override `[disk]`; and
`client-local.cfg`, where the fix is a decision about `--merge-clientlocal`
rather than a moved line. They are tracked separately.

Thanks for diagnosing it properly rather than just working around it locally —
the report named the mechanism, which is what made the rest findable.

PR #542, opened 2026-09-21
etcfiles: read rrddefinitions.d before the shipped sections

### LINKED ISSUE #540: rrddefinitions.cfg: a .d fragment cannot override a shipped definition
`rrddefinitions.cfg` ships `optional directory @XYMONHOME@/etc/rrddefinitions.d`
as its **last** line, and a fragment dropped there cannot override a definition
the shipped file already makes. Only brand-new keys take effect.

Same root cause as #537: #222 applied one uniform rule — each shipped config
ends with its include — to files whose resolution order differs.

## Why the position matters here

Definitions are not an ordered list. `xymond_rrd.c:171` loads them into a tree
keyed by section name and `get_rrd_definition()` does an exact `xtreeFind`,
falling back to `""` only when the key is absent. Position would therefore be
irrelevant — except for duplicates: `xtreeAdd` returns
`XTREE_STATUS_DUPLICATE_KEY` (`lib/tree.c:98`) **without replacing**, and the
loader ignores the return value. **The first definition of a key wins.**

So with the include last, `[disk]` in a fragment loses to the shipped `[disk]`,
silently. With the include first, the fragment is loaded first and wins.

Unlike #537 there is no arbitration to make: the shipped file has no exclusions
that a fragment could trample, so "first" is simply correct.

## What upstream did

`bff5b5ef5` (J.C. Cleaver, 2015) put this line at line 7, near the top, before
any section. Our port moved it to the end.

## Fix

Move the include above the first `[...]` section. One line.

## Also worth deciding, separately

Whether a fragment should be able to *override* a shipped definition at all, or
only add new ones. Moving the line grants override; leaving it grants addition
only. This issue assumes override is wanted, on the grounds that it is the
point of a drop-in directory — but it is a choice, and the shipped comment
should say which one was made.

bonomani:
The include was the last line of `rrddefinitions.cfg`, so a fragment dropped
into `rrddefinitions.d` could only add new keys — never override one the
shipped file already defines. This moves it above the sections.

## Why position matters, given it is not an ordered list

`xymond_rrd.c:171` loads the definitions into a tree keyed by section name, and
`get_rrd_definition()` does an exact `xtreeFind` with a fallback to `""`. So
order ought to be irrelevant — except for duplicate keys: `xtreeAdd` returns
`XTREE_STATUS_DUPLICATE_KEY` (`lib/tree.c:98`) **without replacing**, and the
loader ignores the return value. **The first definition of a key wins.**

Last in the file therefore meant last to be considered, and `[disk]` in a
fragment lost to the shipped `[disk]`, with no warning.

## The choice this makes

There are two positions, and they grant different things. #540 records this as
a decision rather than an obvious fix, so it belongs in the PR too.

**Include last — a fragment can only add.** The shipped definitions stay
authoritative and keep receiving upstream changes. A site can define
`[mytest]`, but cannot touch `[disk]` or the default `[]` without editing the
shipped file — which is the thing a drop-in directory exists to avoid.

**Include first — a fragment can override.** What this PR does. A site can
change the retention of any test, including the global default.

**Why the second.** The common local need here is longer retention, and
retention lives in `[]`, the one section option 1 makes unreachable. A
drop-in mechanism that cannot express the most common local change is not
worth much.

The risk that made the equivalent choice hard in #537 does not arise. There,
the danger was a *pattern* — a local `DISK *` silently swallowing the shipped
`/dev` and `/run` exclusions. Here every section is an explicit key: overriding
`[]` means writing `[]`, which is a deliberate act, not an accident.

And the blast radius is bounded by the file's own first paragraph: *"Changes
here only apply to newly created RRD files."* A bad override degrades new
graphs gradually rather than breaking existing ones, and is undone by fixing
the fragment.

The comment above the line now records which choice was made and why, so it
does not get tidied to the end again.

## Origin

Upstream `bff5b5ef5` put this line near the top in 2015. #222 moved it to the
end when it generalised the mechanism to every config, applying one uniform
position to files whose resolution order differs — the same cause as #537.

Closes #540.


RolandRosenfeld (approved):
Looks good to me.

